PR #397, opened 2017-01-25
Change time updates to current time for duplicate sync out records

edmofro:


Chris-Petty:
@edmofro fine to delete related branch yea?

edmofro:
Yeah think so!

src/sync/SyncQueue.js
```diff
@@ -71,7 +71,9 @@ export class SyncQueue {
                                 recordType,
                                 record.id)
                               .length > 0;
-          if (!duplicate) {
+          if (duplicate) {
+            duplicate.changeTime = new Date().getTime();
```
edmofro: Just looking at this again, it's a good example of why variables should provide an indication of booleanness by being called things like isDuplicate

PR #415, opened 2017-04-27
#406 usage aggregation logic fix

### LINKED ISSUE #406: Item usage not aggregated across batches correctly
Build Number: 30

Description: It looks like item.dailyUsage() is summing the dailyUsage for each batch, which would over-estimate usage when there are multiple batches. If daily usage for batch 1 is 15, and daily usage for batch 2 is also 15, it looks like item.dailyUsage() would return 30. The spreadsheet example linked in issue #405 also includes something about this. It seems like the correct approach to get the total daily usage for an item across batches would be to sum up the total usage for all batches and the number of days each batch was used, then do the division in Item.

Comments: Happy to take a crack at this ticket or #405, but I don't have a server I could test on at the moment..

Chris-Petty:
Refactors dailyUsage as described in #406.

fixes #406

### DIFF
--- a/src/database/DataTypes/Item.js
+++ b/src/database/DataTypes/Item.js
@@ -13,8 +13,8 @@ export class Item extends Realm.Object {
   }
 
   get dailyUsage() {
-    const startDate = new Date(Date.now() - USAGE_PERIOD_MILLISECONDS); // 90 Days ago
     const endDate = new Date();
+    const startDate = new Date(endDate - USAGE_PERIOD_MILLISECONDS); // 90 Days ago
     return this.dailyUsageForPeriod(startDate, endDate);
   }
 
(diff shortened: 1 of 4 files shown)


Chris-Petty:
@gkaplan You might like to have a pass at this 💃

src/database/DataTypes/ItemBatch.js
```diff
@@ -16,37 +12,18 @@ export class ItemBatch extends Realm.Object {
...
-    const currentDate = new Date();
-    let usagePeriod = millisecondsToDays(USAGE_PERIOD_MILLISECONDS);
-    if (transactionBatches.length === this.transactionBatches.length) {
-      // This item batch has no transaction batches older than the usage period constant,
-      // use the actual amount of time it has been around for as the usage period
-      usagePeriod = millisecondsToDays(currentDate.getTime() - this.addedDate.getTime());
-    }
-    const dailyUsage = usagePeriod ? totalUsage / usagePeriod : 0;
-    return dailyUsage;
+    throw new Error('ItemBatch.dailyUsage() is a deprecated function');
```
Chris-Petty: I should probably remove this method before PR.

src/database/DataTypes/Item.js
```diff
@@ -11,19 +13,35 @@ export class Item extends Realm.Object {
...
+      const batchAddedDate = batch.addedDate;
+      itemAddedDate = batchAddedDate < itemAddedDate ? batchAddedDate : itemAddedDate;
+    });
+    return itemAddedDate;
   }
 
   get earliestExpiringBatch() {
     if (this.batches.length === 0) return null;
-    let earliestBatch = this.batches.find((batch) => batch.totalQuantity > 0);
+    let earliestBatch = this.batches.find(batch => batch.totalQuantity > 0);
```
Chris-Petty: I've run prettier locally which culled the parentheses. We don't have an eslint rule against not parenthesising single params in a arrow function and I'm becoming indifferent towards enforcing this coding style.

andreievg: looks tidier to me

PR #416, opened 2017-04-27
#413 sync version compatibility

### LINKED ISSUE #413: App sends version on sync for version compatibility check
Build Number: 30

On sync mobile should send the app version to the server so that the server can check if the app version is compatible with the server version. Server sends back yes or no. If no response, mobile should assume incompatibility (old server version). Should warn mobile user.

Initialisation page can show compatibility error on fail. In normal use normal "sync error" at top should be shown.

Chris-Petty:
If we are using semantic versioning, this should be in a major (2.0.0) release as it it will not work with servers without the following implemented:

Servers need REST API updated to handle
`/sync/v2/compatibility?app=msupply_mobile&app_version=${appVersion}`

Fixes #413


src/sync/Synchroniser.js
```diff
@@ -100,12 +107,49 @@ export class Synchroniser {
...
+   * - reponse.compatible === false should reject the promise sad :(
+   * -- In the above case, response.error should specify the highest compatible version number of
+   *    mobile
+   * - else reject the promise, server isn't handling this check at all and needs updating
+   */
+  async checkServerCompatibility() {
+    const appVersion = await getAppVersion();
+    const serverURL = this.settings.get(SYNC_URL);
+    const response = await fetch(
+      `${serverURL}/sync/v2/msupply_mobile_compatibility/?app_version=${appVersion}`,
```
Chris-Petty: @craigdrown @nmadruga @andreievg Requested y'all to review particularly for agreement on the REST api structure as depicted here

Chris-Petty: as far as mSupply server is concerned, `/?` will be replaced with`?`. So there is an opportunity to drop that forward `/`.

I've not done so as all the other mobile server fetches use the `/?` pattern. If it is more standard REST to not have the `/` we could refactor this here.

Chris-Petty: @edmofro has 2cents: `/` doesn't matter, but suggests:
`v2/compatibility?app=msupply_mobile&app_version=...`
Which I think is far nicer. BRB.

andreievg:
if mobile version does not request compatibility check (current mobile and older) should the msupply server not allow sync or auth ? 
should this comment be in mantis issue ?

src/sync/Synchroniser.js
```diff
@@ -100,12 +107,49 @@ export class Synchroniser {
...
+      // Likely server hasn't extended it's API to handle this method and needs updating
+      throw new Error('Unexpected version verification response from server');
+    }
+    if (!responseJson.compatible && responseJson.error && responseJson.error.length > 0) {
+      // Server should provide compatible version number in this error
+      throw new Error(responseJson.error);
+    } else if (!responseJson.compatible) {
+      // If error wasn't sent but compatible was false
+      throw new Error('App version not compatible with this server');
+    }
```
edmofro: This could be simpler than two conditional blocks:
```
if (!responseJson.compatible) {
  // Not compatible, send the error provided by the server indicating compatible version if possible
  throw new Error(responseJson.error || 'App version not compatible with this server');
}

File: src/sync/Synchroniser.js [hunk 3]
```diff
@@ -133,14 +177,15 @@ export class Synchroniser {
...
-      `${serverURL}/sync/v2/queued_records/?from_site=${thisSiteId}&to_site=${serverId}`,
+      `${serverURL}/sync/v2/queued_records?from_site=${thisSiteId}&to_site=${serverId}`,
       {
         method: 'POST',
         headers: {
           Authorization: this.authenticator.getAuthHeader(),
         },
         body: JSON.stringify(records),
-      });
+      }
```
edmofro: I know it's 'prettier' doing this, but doesn't it break the eslint rule about dangling commas?

Chris-Petty:
I'm leaning towards your solution @edmofro, simply smarter/cleaner. The work _isn't_ done on the other end.

The thought process here was that this would mean that mobile wouldn't have to be familiar with other servers, respective servers would handle it; your suggestion works great provided mobile is only used with mSupply (which frankly is a pretty safe assumption save some kind of divine open source intervention), but then mobile must store info about servers x, y and z if it comes to that. Or different servers would have to all have the same API versioning and features haha.

edmofro:
@Chris-Petty you laugh, but I actually think your final comment is fair. We provide the code open source, but if some other party wants to integrate with it, they either have to fork and change mSupply Mobile, or provide exactly matching APIs.

Chris-Petty:
Divine open source intervention is forking mobile, got it! I realise more sense in your point; servers are going to have to mimic mSupply anyway save a major refactoring of mobile sync. Mimicking the sync versioning is the least of concerns there!

I'm going to close this, do a PR on mSupply implementing your suggestion. Make the changes on this branch here then reopen/new PR.

Chris-Petty:
New commits!?

Chris-Petty:
@edmofro Done on server end too. Review!

I added max version, then removed it, as it'd probably cause more hassle than benefit given mSupply not using any kind of semantic versioning.

Minimum version should be changed to `383` from `380` with the August release

Chris-Petty:
Killing, can really just rely on API versioning and end points to give back a useful error if no implemented.

Issue #401, opened 2017-03-09
Allow creating new supplier invoices (not from central supplier)


edmofro:
Motivation:
Stores sometimes get medicines from someone other than their main supplier. For example, they might run out of paracetamol, and buy some new stocks from the local private pharmacy. Or a donor might turn up with a whole lot of asthma inhalers. We need to allow them to enter this stock into inventory in a more elegant way that just increasing stock through a stocktake.

Solution:
On the Supplier Invoices page, have a 'New Invoice' button. This matches the functionality of the analogous button on the Customer Invoices page: clicking it brings up a list of Suppliers (defined on the central server) which can be filtered and selected. Selecting a supplier creates a new supplier invoice with that supplier connected, with the ability to add item lines.

Gotchas:
- We still don't want to confuse mobile by adding batch and expiry information, so just add the batches with none (much like we do for a stocktake that has no existing lines to increase)
- What prices to use?

andreievg:
Feature implemented in -> https://github.com/sussol/mobile/releases/tag/v2.0.0

PR #430, opened 2017-05-21
#401 mobile supplier invoices

andreievg:


src/pages/ExternalSupplierInvoicePage.js
```diff
@@ -0,0 +1,441 @@
...
+
+  addNewSIitem(item) {
+    const { database, transaction } = this.props;
+    database.write(() => {
+      const transactionItem = createRecord(database, 'TransactionItem', transaction, item);
+      createRecord(
+        database,
+        'TransactionBatch',
+        transactionItem,
+        createRecord(database, 'ItemBatch', item, '')
```
Chris-Petty: So this creates a new ItemBatch then the user edits and saves the `batch`(code) field; It's possible to have multiple entries for the same batch (given there may be a preexisting batch with the same `batch` string). @craigdrown or someone might want to confirm, but that is probably no deal.

I think it be better if there was a way to not create the corresponding itemBatches until finalise (where it can use the batch code entered by user. Finalising stock takes may have a clue on how to do this (and there is intent to make stock takes per batch, so maybe stock takes will reflect what we do here!).

jtrein:
#### !spam 
This PR review deserves a Macho Man

download

src/database/DataTypes/Transaction.js
```diff
@@ -157,25 +197,26 @@ export class Transaction extends Realm.Object {
...
-          itemBatch.numberOfPacks - transactionBatch.numberOfPacks;
-        itemBatch.packSize = transactionBatch.packSize;
-        itemBatch.numberOfPacks = newNumberOfPacks;
-        itemBatch.expiryDate = transactionBatch.expiryDate;
-        itemBatch.batch = transactionBatch.batch;
-        itemBatch.costPrice = transactionBatch.costPrice;
-        itemBatch.sellPrice = transactionBatch.sellPrice;
-        database.save('ItemBatch', itemBatch);
-      });
+    const externalInvoice = this.isExternalSI;
```
andreievg: renamed to `isExternalInvoice` and `isIncomingInvoice` I think it goes without saying and references that repetitive string validations (in both functions) shouldn't be happening on every iteration when they can happen once and rammed.  Good practice re booleans though, I missed that one cheers.

src/database/DataTypes/TransactionBatch.js
```diff
@@ -10,6 +11,10 @@ export class TransactionBatch extends Realm.Object {
     return this.numberOfPacks * this.packSize;
   }
 
+  get itemCode() {
```
andreievg: Used in `ExternalSupplierInvoice.js`, since table rows are transactionBatches and not transactionItems like most other pages.. and no reference to item.code from TransactionBatch realm schema..have to improvise (itemCode should really be added to TransactionBatch and db migrated.. should I ?)

src/database/DataTypes/TransactionBatch.js
```diff
@@ -55,7 +72,13 @@ export class TransactionBatch extends Realm.Object {
     if (!this.costPrice) return 0;
     return this.costPrice * this.numberOfPacks;
   }
-
+  /**
+   * removes associated item batch
+   * @param  {Realm} database   App wide database
+   */
+  removeItemBatch(database) {
+    database.delete('ItemBatch', this.itemBatch);
```
andreievg: Agree, although it's only used in ExternalSupplierInvoices (for the time being) it may cause problems if someone careless uses this function in other transactions:
Solitions:
* rename to associate with external supplier invoices item batch and check that parent transaction isExternalSI (since itemBatch is created by externalSI, it's very safe)
* check of totalQuantity 0 ? is it ok to remove empty batches ?
* (best solution) when creating new line, do not create itemBatch, and in transaction.confirm(), check if itemBatch exists, if not create. (this will require, item code to be in transactionBatch, thus migration and big rethink of createRecord because of back and forward relationships between itemBatch, transactionItem, item and transactionBatch.. by the way why so many parent <-> child relationships in mobile ? *note 1.. and also transaction.confirm will start to become a bit of a monster, especially with your comments re pack size 1 translation being in transaction.confirm)

*note 1- instead of child parent back and forth, can we somehow store or access database object within Realm.Object ?, then we can eventually move away from 'circular' relationships, store reference by id and still be able to use getters of parents|children via database.objects(type).filtered(by-id)

src/database/utilities/utilities.js
```diff
@@ -1,3 +1,4 @@
+import util from 'util';
```
andreievg: it did indeed, should enforce ESLint errors in sussol/mobile before commits

src/database/utilities/utilities.js
```diff
@@ -20,3 +21,44 @@ export function addBatchToParent(batch, parent, createItem) {
   if (item.batches && item.batches.find(currentBatch => currentBatch.id === batch.id)) return;
   item.addBatch(batch);
 }
+/**
+ * Removes TransactionBatch...
```
andreievg: * 1 - Remove ItemBatch from TransactionBatch
 * 2 - Remove TransactionBatch from TransactionItem
 * 3 - if TransactionItem has no batches, remove TransactionItem from Transaction

Above as per full function comment.  Had to ... the end of first line since explanation is not a one liner

src/database/utilities/utilities.js
```diff
@@ -20,3 +21,44 @@ export function addBatchToParent(batch, parent, createItem) {
...
+ * 1 - Remove ItemBatch from TransactionBatch
+ * 2 - Remove TransactionBatch from TransactionItem
+ * 3 - if TransactionItem has no batches, remoe TransactionItem from Transaction
+ * @param {Realm}   database   App wide database
+ * @param {Transaction}   transaction  context transaction
+ * @param {TransactionBatch} transactionBatch transaction batch to remove
+ */
+export function removeTransactionBatchUtil(database, transaction, transactionBatch) {
```
andreievg: I though it was more like a static function.. similar to addBatchToParent.. I will refactor and relocate to Transaction, can see a bit of improvements having it in Transaction.. especially in onDeleteConfirm `SupplierInvoicesPage`

src/database/utilities/utilities.js
```diff
@@ -20,3 +21,44 @@ export function addBatchToParent(batch, parent, createItem) {
...
+  // find transaction item
```
andreievg: yeah .. although it won't be called if transaction is finalised, could be good incase someone decided to use carelessly. will amend when refactoring into Transaction

src/database/utilities/utilities.js
```diff
@@ -20,3 +21,44 @@ export function addBatchToParent(batch, parent, createItem) {
...
+}
+/**
+ * Returns class as JSON, warning: since the schema has child <-> parent (circular))
+ * printing via console.log or json Stringify will be infinite..
+ * use util.format('[Circular]', returnJson) to log
+ * @param {Realm}   database   App wide database
+ * @param {Transaction}   transaction  context transaction
+ * @param {TransactionBatch} transactionBatch transaction batch to remove
+ */
+export function classAsJson(classInstance) {
```
andreievg: so to reply to other comment relating to Transaction.confirm, classAsJson and transactionBatch.batchAsSinglePackJson()

batchAsSinglePack was created as a non mutable function that minimises changes to .confirm .. the name kind of says it all, no real reason to go deep into it's implementation when looking at .confirm, and classAsJson might be useful in the future.. or perhaps not.  I can migrate changed to .confirm(), just thinking it might become a bit messy..
I read about option c and inlines, batchAsSinglePackJson is quite pure function, no state changes, it compartmentalises functionality and makes for easier digestion but .. good question is it worth the overhead ? even though it runs on finalise only.. 

need to talk about this in person, I am keen to change as per suggested, just want to drive my point of view a little :), I'll try to remember talking about sortDataBy

src/pages/index.js
```diff
@@ -19,6 +19,9 @@ import { StocktakesPage } from './StocktakesPage';
 import { SupplierInvoicePage } from './SupplierInvoicePage';
 import { SupplierInvoicesPage } from './SupplierInvoicesPage';
 import { RealmExplorer } from './RealmExplorer';
+import { ExternalSupplierInvoicePage,
+         checkForFinaliseError as checkForExternalSupplierInvoiceError
```
andreievg: yes, just noticed, it complained.

src/sync/syncTranslators.js
```diff
@@ -153,4 +153,5 @@ export const SEQUENCE_KEYS = new SequenceKeyTranslator({
   requisition_serial_number: 'requisition_serial_number_for_store_',
   requisition_requester_reference: 'requisition_requester_reference_for_store_',
   stocktake_serial_number: 'stock_take_number_for_store_',
+  supplier_invoice_serial_number: 'supplier_invoice_number_for_store_'
```
andreievg: store id get's added to this values.. and mSupply desktop does some parsing 'magic'... sync api should already know which store id POSTs are coming from though...

src/pages/ExternalSupplierInvoicePage.js
andreievg: from what I've seen in mSupply desktop it is possible to have multiple batch names (this is where 'merge' comes in).. I do agree a better solution will be to create itemBatch on finalise as per my previous comment.. I don't think it will solve the multiple batch problem though.. the juice of stocktake batch creating is in Stocktake.adjustInventory .. ill check it out see what is common

andreievg:
So, TransactionBatch should always have ItemBatch associated with it, unless we are changing quite a bit of sync logic, small change to Realm Scheme and another case: function in createRecord.  @Chris your comment: `I think it be better if there was a way to not create the corresponding itemBatches until finalise` -> will require major changes, I can elaborate further if you need, let me know.. and ` Finalising stock takes may have a clue on how to do this` -> stock take transaction/batch are both a different kind of beast (from normal transactions) in both mSupply desktop and mobile, and differ in Scheme and Sync.  I did spend quite a bit of time trying to make ItemBatchless, TransactionBatch to work, only to realise that sync breaks it. So reverted back to the way I did it originally.

src/pages/ExternalSupplierInvoicePage.js
```diff
@@ -0,0 +1,444 @@
...
+        title: tableStrings.batch_cost_price,
+        alignText: 'center',
+      },
+      {
+        key: 'remove',
+        width: 1,
+        title: tableStrings.remove,
+        alignText: 'center',
+      },
+    ];
```
edmofro: That's a lot of columns! Do they all fit?

src/pages/ExternalSupplierInvoicePage.js
```diff
@@ -0,0 +1,444 @@
...
+        break;
+      case 'numberOfPacks':
+      case 'packSize':
+        sortDataType = 'number';
+        break;
+      default:
+        sortDataType = 'realm';
+    }
+    // calculate and set total price
+    const tPrice = transactionBatches.reduce(
```
edmofro: No abbreviations

src/pages/ExternalSupplierInvoicePage.js
```diff
@@ -0,0 +1,444 @@
...
+  }
+  /**
+   * Returns updated data according to searchTerm, sortBy and isAscending.
+   * Dealing with TransactionBatch as oppose to TransactionItems
+   * to be able to add miltiple batches for item
+   * also total price is calculated here
+   */
+  getFilteredSortedData(searchTerm, sortBy, isAscending) {
+    const { database, transaction } = this.props;
+    console.log(transaction.items);
```
edmofro: Left a console.log in

src/pages/ExternalSupplierInvoicePage.js
```diff
@@ -0,0 +1,444 @@
...
+      case 'batch':
+        sortDataType = 'string';
+        break;
+      case 'numberOfPacks':
+      case 'packSize':
+        sortDataType = 'number';
+        break;
+      default:
+        sortDataType = 'realm';
+    }
```
edmofro: My preference in situations where you need to get a single value based on a key is to use a dictionary (js object) rather than a switch case.
```
const sortDataType = {
  itemName: 'string',
  itemCode: 'string',
  ...
}[sortBy] || 'realm'; // Use realm as the default sortDataType
```

src/pages/ExternalSupplierInvoicePage.js
```diff
@@ -0,0 +1,444 @@
...
+      selection.forEach(transactionBatchID => {
+        const transactionBatch = transactionBatches.find(tBatch =>
+                                    tBatch.id === transactionBatchID);
+
+        database.delete('ItemBatch', transactionBatch.itemBatch);
+        database.delete('TransactionBatch', transactionBatch);
+      });
+
+      transaction.pruneBatchlessItems(database);
+    });
```
edmofro: This logic should also move into the database type

src/pages/ExternalSupplierInvoicePage.js
```diff
@@ -0,0 +1,444 @@
...
+  // transaction batches
+  onDeleteConfirm() {
+    const { selection } = this.state;
+    const { transaction, database } = this.props;
+    const transactionBatches = transaction.transactionBatches(database);
+    database.write(() => {
+      // Find each selected trasnaction batch by id
+      selection.forEach(transactionBatchID => {
+        const transactionBatch = transactionBatches.find(tBatch =>
+                                    tBatch.id === transactionBatchID);
```
edmofro: More abbreviations...

src/database/utilities/createRecord.js
```diff
@@ -206,6 +210,24 @@ function createStocktakeBatch(database, stocktakeItem, itemBatch) {
   return stocktakeBatch;
 }
 
+// Creates a supplier invoice (Transaction) and adds it to the customer (Name)
+function createSupplierInvoice(database, supplier, user) {
+  const currentDate = new Date();
+  const invoice = database.create('Transaction', {
+    id: generateUUID(),
+    serialNumber: getNextNumber(database, SUPPLIER_INVOICE_NUMBER),
```
edmofro: **Important** Have you talked with others about how locally assigned supplier invoice numbers avoid conflicts with those that are centrally assigned?

src/database/DataTypes/Transaction.js
```diff
@@ -157,21 +210,62 @@ export class Transaction extends Realm.Object {
...
+   * Confirm this transaction, generating the associated item batches, linking them
+   * to their items, and setting the status to confirmed.
+   * @param  {Realm}  database The app wide local database
+   * @return {none}
+   */
+  confirmExternalSI(database) {
+    if (this.isConfirmed) throw new Error('Cannot confirm as transaction is already confirmed');
+    if (this.isFinalised) throw new Error('Cannot confirm as transaction is already finalised');
+
+    this.pruneRedundantBatches(database);
```
andreievg: I though there will be too many conditionals if we combine these two together.  Let's see how it looks though, will commit as advised.

src/database/DataTypes/Transaction.js
```diff
@@ -157,21 +210,62 @@ export class Transaction extends Realm.Object {
...
+    this.transactionBatches(database).forEach(transactionBatch => {
+      const { itemBatch,
+              batch,
+              packSize,
+              numberOfPacks,
+              expiryDate,
+              costPrice,
+             } = transactionBatch;
+      // Assuming createRecords will set packSize to 1
```
andreievg: incoming sync always set's pack size to one, createRecord set packSize to one (and in this case ItemBatch is unchanged until this stage), there are numerous comments re mobile pack size === 1.
// Every item batch in mobile should be pack-to-one
// Pack to one all mobile data

ItemBatch remains at 1, but TransactionBatch for external supplier invoice, will keep existing pack size and quantity, so on finalisation or re-sync it will be reflective of the paper trail (and easier to 'reconcile'), thus the extra in 'TransactionBatch' sync changes

src/database/DataTypes/TransactionItem.js
```diff
@@ -87,6 +86,47 @@ export class TransactionItem extends Realm.Object {
...
     this.pruneEmptyBatches(database);
   }
+  /**
+   * Sets Quantity for external supplier invoice, having one Batch per
+   * TransactionLine
+   * @param {int}quantity
+   */
+  setActualBatchQuantity(database, quantity) {
+    const transBatch = this.batches[0];
+    transBatch.packSize = quantity;
```
andreievg: I won't explain, I'll delete, thanks for finding useless function.. it must have been used somewhere, but retired.

src/sync/incomingSyncUtils.js
```diff
@@ -372,13 +372,18 @@ export function createOrUpdateRecord(database, settings, recordType, record) {
       const item = getObject(database, 'Item', record.item_ID);
       itemBatch.item = item;
       item.addBatchIfUnique(itemBatch);
-      const packSize = parseNumber(record.pack_size);
+      let incomingPackSize = parseNumber(record.pack_size);
+      let packSize = 1;
+      if (transaction.isExternalSI) {
```
andreievg: As per above packSize === 1 comment, it's for external supplier invoices to keep transaction batches as is. Since they can have pack size other then 1, and we want to keep the entered pack size

src/pages/ExternalSupplierInvoicePage.js
```diff
@@ -251,53 +263,14 @@ export class ExternalSupplierInvoicePage extends GenericPage {
...
-      case 'remove':
-        return {
-          type: 'checkable',
-          icon: 'md-remove-circle',
-          isDisabled: this.props.transaction.isFinalised,
-        };
-    }
+
+    const renderCellGenerator =
+      renderCellMapping[key] || defaultRenderCellWrapper;
```
andreievg: what do you think about this kind of mapping ? 
Ideally defaults would go somewhere common.
And remove (in next commit) can go into renderCellMapping, and instead of being disabled, can change type to text or ideally .. hide the column all together on finalised

src/pages/SupplierInvoicesPage.js
```diff
@@ -100,8 +153,55 @@ export class SupplierInvoicesPage extends GenericPage {
...
+            placeholderText={modalStrings.start_typing_to_select_supplier}
+            queryString={'name BEGINSWITH[c] $0'}
+            sortByString={'name'}
+            onSelect={name => {
+              this.onNewSupplierInvoice(name);
+              this.setState({ isCreatingInvoice: false });
+            }}
+            onClose={() => this.setState({ isCreatingInvoice: false })}
+            title={modalStrings.search_for_the_customer}
+          />
```
andreievg: I do agree, should this be done in this pr ? I think it's getting too big, I can branch of it and make requested changes.

edmofro: Definitely leave for now - it will only need to be refactored when we merge in master anyway

src/database/DataTypes/Transaction.js
```diff
@@ -142,13 +150,58 @@ export class Transaction extends Realm.Object {
...
-    this.items.forEach((transactionItem) => {
-      if (transactionItem.totalQuantity === 0) itemsToPrune.push(transactionItem);
+    const transactionBatchesToPrune = [];
+    this.items.forEach(transactionItem => {
+      if (transactionItem.totalQuantity === 0) {
+        itemsToPrune.push(transactionItem);
+        transactionBatchesToPrune.concat(transactionItem.batches);
+      }
     });
+    database.delete('TransactionBatch', transactionBatchesToPrune);
```
edmofro: Actually this is unnecessary - the ```TransactionItem``` takes care of deleting any batches attached to it

src/database/DataTypes/Transaction.js
```diff
@@ -123,6 +121,16 @@ export class Transaction extends Realm.Object {
   }
 
   /**
+   * Remove the transaction item passed as param
+   * @param  {Realm}  database        App wide local database
+   * @param  {TransactionItem}  TransactionItem to remove
+   * @return {none}
+   */
+  removeTransactionItem(database, transactionItem) {
+    database.delete('TransactionItem', transactionItem);
```
edmofro: No, as the ```TransactionItem``` deletes any batches using its destructor

src/database/DataTypes/Transaction.js
```diff
@@ -121,6 +120,49 @@ export class Transaction extends Realm.Object {
...
+   */
+  removeSelf(database) {
+    if (this.isFinalised) throw new Error('Cannot delete finalised transaction');
+    if (!this.isValid()) return;
+    const transactionBatchesIds =
+      this.getTransactionBatches(database).map(transactionBatch => transactionBatch.id);
+
+    this.removeTransactionBatchesById(database, transactionBatchesIds);
+
+    database.delete('Transaction', this);
```
edmofro: This should move into the destructor, so you can always just call `database.delete` and it will clean up all of the batches etc.

PR #432, opened 2017-07-17
Get off react native fork

edmofro:
- Upgrade dependencies to latest
- Move from NavigationExperimental to React Navigation
- Incorporate a little redux


edmofro:
Fixes #436 
Fixes #435

index.ios.js
```diff
@@ -1,8 +1,6 @@
 /**
- * Sample React Native App
```
Chris-Petty: hahahah

edmofro: heh

src/localization/tableStrings.js
```diff
@@ -76,7 +76,7 @@ export const tableStrings = new LocalizedStrings({
     name: 'Arana',
     no_change: 'Akea te bitaki',
     number_of_batches: 'Mwatin taian batch',
-    number_recieved: 'Mwaitin ae reke raoi',
+    number_received: 'Mwaitin ae reke raoi',
```
Chris-Petty: Received always catches me

src/navigation/Navigator.js
```diff
@@ -0,0 +1,18 @@
...
+* Tupaia MediTrak
+* Copyright (c) 2017 Beyond Essential Systems Pty Ltd
+**/
+
+import { StackNavigator as createStackNavigator } from 'react-navigation';
+
+import { PAGES } from '../pages';
+const routes = {};
+// Add all pages to the navigation routes
+Object.entries(PAGES).forEach(([routeName, page]) => { routes[routeName] = { screen: page }; });
```
Chris-Petty: Cool

src/pages/StocktakesPage.js
```diff
@@ -79,10 +80,10 @@ export class StocktakesPage extends GenericPage {
     const { stocktakes, selection } = this.state;
     const { database } = this.props;
     database.write(() => {
-      const stocktakesToDelete = [];
+      let stocktakesToDelete = [];
       for (let i = 0; i < selection.length; i++) {
         const stocktake = stocktakes.find(s => s.id === selection[i]);
-        if (stocktake.isValid() && !stocktake.isFinalised) stocktakesToDelete.push(stocktake);
+        if (stocktake.isValid() && !stocktake.isFinalised) stocktakesToDelete = stocktake;
```
Chris-Petty: Wont this just set stocktakesToDelete to be the last record in the selection? I think the master implementation is correct.

src/pages/index.js
```diff
@@ -1,11 +1,19 @@
...
+import React from 'react';
+
 import { CustomerPage } from './CustomerPage';
 import { CustomersPage } from './CustomersPage';
 import { CustomerInvoicePage,
          checkForFinaliseError as checkForCustomerInvoiceFinaliseError,
        } from './CustomerInvoicePage';
 import { CustomerInvoicesPage } from './CustomerInvoicesPage';
-import { FirstUsePage } from './FirstUsePage';
+export { FirstUsePage } from './FirstUsePage';
```
Chris-Petty: exports should be after all imports.

src/widgets/NavigationBar.js
```diff
@@ -0,0 +1,83 @@
...
+    height: HEADER_HEIGHT,
+    flexDirection: 'column',
+    justifyContent: 'center',
+  },
+  backIcon: {
+    marginRight: 30,
+    ...textStyles,
+  },
+  leftSection: {
+    ...sectionStyle,
```
Chris-Petty: Ohhhh I really like this rather than doing it in the JSX.

src/widgets/PageInfo.js
```diff
@@ -104,7 +110,7 @@ function renderInfoComponent(isEditingDisabled, columnIndex, rowData, rowIndex)
       break;
   }
   // If null or empty string, use single space to avoid squishing row
-  let infoString = rowData.info && String(rowData.info);
+  let infoString = (rowData.info || rowData.info === 0) && String(rowData.info);
```
Chris-Petty: if rowData.info is falsy then it can't `=== 0`. If rowData is truthy then the second half is not evaluated. So the second half of this condition is redundant.

edmofro: Pretty sure 0 is falsey actually

src/pages/index.js
edmofro: Sweet

src/pages/StocktakesPage.js
edmofro: Ummm...yes. This is clearly wrong. I don't remember doing it!

edmofro: Oh right I would have done this when I was testing realm. They changed the toString method of objects, which broke our re-implementation of database.delete. Will revert this!

Chris-Petty: If only we had time to write testing suites :'(

src/widgets/PageInfo.js
Chris-Petty: !0 is true, so yes falsy.

I understand now, because 0 is falsy this becomes a weird edge case.

PR #433, opened 2017-07-17
Fix requisition usage calculation bug

edmofro:


src/database/DataTypes/Item.js
```diff
@@ -89,7 +89,7 @@ export class Item extends Realm.Object {
     const fromDate = addedDate > startDate ? addedDate : startDate;
     const periodInDays = millisecondsToDays(endDate - fromDate);
     const usage = this.totalUsageForPeriod(fromDate, endDate);
-    return usage / periodInDays;
+    return usage / (periodInDays === 0 ? 1 : periodInDays); // Avoid divide by zero
```
andreievg: periodInDays || 1 ?

edmofro: Yeah that's much better. I was nervous of -1 also being falsey (is that true?), but thinking about it, we never want a negative periodInDays anyway!

PR #440, opened 2017-07-21
Remove all flow weak

edmofro:


DhanyaHerath:
**Crash Instances**
- when deleting a customer invoice
- when searching item list to add an item to a new customer invoice

**Can't create a supplier invoice from mobile**
- Supplier names are not returned when searching, so a new invoice can't be created. 
-Reproduction Steps: Create new supplier invoice, type in search box to search for supplier - no suppliers returned. 

**Quantities in Supplier Invoice allows for fractional values**
- When adding new items and editing existing items, the quantity box allows you to use fractional values (eg. 0.999), and doesn't round the value. When the same thing is done for Customer invoices, it rounds the quantity value. 
- Reproduction Steps: Add new item or edit existing item line on a supplier invoice - quantity can be changed to 0.999 or any other decimal value. 

**Heading for Supplier Invoice Creation says “Search for the customer”**
- When creating a new supplier invoice and searching for new supplier, the headings says “Search for the customer”
- Reproduction Steps: Create new supplier invoice - this will take you to the search screen for suppliers.

edmofro:
> **Can't create a supplier invoice from mobile**
> - Supplier names are not returned when searching, so a new invoice can't be created. 
> -Reproduction Steps: Create new supplier invoice, type in search box to search for supplier - no suppliers returned. 

No suppliers set to visible in the store you had on the tablets when I left, I suspect this is still the same?

edmofro:
> Quantities in Supplier Invoice allows for fractional values
> 
> When adding new items and editing existing items, the quantity box allows you to use fractional values (eg. 0.999), and doesn't round the value. When the same thing is done for Customer invoices, it rounds the quantity value.
> Reproduction Steps: Add new item or edit existing item line on a supplier invoice - quantity can be changed to 0.999 or any other decimal value.

@craigdrown This was to support partial packs being sent, which I'm pretty sure mSupply desktop allows...valid?

src/pages/CustomerInvoicePage.js
```diff
@@ -35,56 +34,32 @@ const MODAL_KEYS = {
...
+      searchTerm: '',
+      sortBy: 'itemName',
+      isAscending: true,
+    };
+    autobind(this);
+  }
+
+  updateDataFilters(newSearchTerm, newSortBy, newIsAscending) {
+    // We use != null, which checks for both null or undefined (undefined coerces to null)
+    if (newSearchTerm != null) this.dataFilters.searchTerm = newSearchTerm;
```
andreievg: One things that is bothers me with refactor to composition.. duplicate updateDataFilters and very similar this.dataFilters declaration.

andreievg: Also sortDataBy function.
I see the case for manual sorting, it's when the field is not in a realm schema. So let's say we are trying to use realm to sort otherPartyName (which is getter) we will get -> property doesn't exist (I assume realm does `Object.getOwnPropertyDescriptor` check before sorting, which will return undefined for getters, unlike `typeof` which will return `string` for otherPartyName).. I would prefer us using realm as much as possible, so in sortDataBy, it would be good to:

```
// check !data or data.length < 1, return []
// check data.length < 1 return []
// check !Object.getOwnPropertyDescriptor(data[0], sortBy), return manualSort()
```
it would also be nice if manualSort can figure out sortDatType from schema, but I see some problems with currentScheme, i.e serialNumber is a string.

Not to be hypocritical and to enforce consistency (using realm as much as possible), Transaction.removeTransactionBatchesById should be refactored to use filtered().

I think this can be done in another PR, once this one is merged I will create

src/pages/SupplierInvoicePage.js
```diff
@@ -107,56 +208,234 @@ export class SupplierInvoicePage extends GenericPage {
...
+      <PageInfo columns={infoColumns} isEditingDisabled={this.props.transaction.isFinalised} />
+    );
   }
 
-  renderCell(key, transactionItem) {
+  renderCell(key, transactionBatch) {
+    const isEditable = !this.props.transaction.isFinalised;
+    const type = isEditable ? 'editable' : 'text';
+    const editableCell = {
+      type: type,
```
andreievg: Should only be able to edit Quantity field if !isExternalSupplierInvoice

edmofro: Errmm...why?

src/database/DataTypes/Transaction.js
```diff
@@ -163,19 +163,17 @@ export class Transaction extends Realm.Object {
...
-      if (transactionItem.totalQuantity === 0) {
-        itemsToPrune.push(transactionItem);
-      }
-    });
-    database.delete('TransactionItem', itemsToPrune);
+  pruneRedundantBatches(database) {
+    const batchesToRemove = this.getTransactionBatches(database)
+                              .filtered('numberOfPacks = 0');
+
+    database.delete('TransactionBatch', batchesToRemove);
```
andreievg: Never mind, TransactionBatch destructor does this. nice!

src/pages/CustomerInvoicePage.js
edmofro: First comment: Yes, it bothers me too. I thought enough thinking/refactor for one PR though, can work out a way to simplify in another smaller refactor.

Agree about using realm as much as possible, but if you look inside the sorting utility function, you'll see we are using realm as much as possible already. Doing something clever to work out the sortDataType would also be nice.

src/database/DataTypes/Transaction.js
edmofro: Exactly! We want as much of the clean up logic to be handle inside the destructors, that's why I designed the database to have destructors in the first place. Saves having to think about what you need to clean up every time, and then forgetting :-)

edmofro:
Still to handle:

**Heading for Supplier Invoice Creation says “Search for the customer”**
- When creating a new supplier invoice and searching for new supplier, the headings says “Search for the customer”
- Reproduction Steps: Create new supplier invoice - this will take you to the search screen for suppliers.

**Crash Instances**
- when deleting a customer invoice
- when searching item list to add an item to a new customer invoice

I've gotta finish off a bit of tupaia stuff today, then will get straight onto this.

src/pages/SupplierInvoicePage.js
andreievg: To minimise mistake (foul play), on receiving end -> I don't think batch name, batch expiry, pack size, cost price, should be editable (since normal supplier invoice-> inter-store transfer and stock transfer from central store should set this accurately, it's good to see this fields though). I can start an e-mail chain to see what everyone else thinks?

edmofro: Already discussed with @craigdrown, he thought ok and maybe even good to have editable. After all, mistakes (foul play) can happen on both ends ;-)

andreievg:
Still to handle:

**Heading for Supplier Invoice Creation says “Search for the customer”**
( I thought I changed this in External Supplier Invoices PR.. must have either forgot or lost in transit )

**Crash Instances**
- when deleting a customer invoice
- when searching item list to add an item to a new customer invoice

( can't replicate, works fine )

**Can't create a supplier invoice from mobile**
as per your comment:
* inSupply Desktop, log into subject mobile store
* Under Special->Stores, click on subject mobile store
* At the bottom, Use Query ->  Query Name

screen shot 2017-08-01 at 11 39 45 am

* after query, click make found names visible in stores, resync mobile 

(this should be documented, will follow up )

**Quantities in Supplier Invoice allows for fractional values**
( this is good as is, I will write an e-mail re pack size / quantity best practices, and should be documented )

src/pages/SupplierInvoicePage.js
```diff
@@ -1,100 +1,201 @@
...
+  // transaction batches
+  onDeleteConfirm() {
+    const { selection } = this.state;
+    const { transaction, database } = this.props;
+    database.write(() => {
+      transaction.removeTransactionBatchesById(database, selection);
+    });
+    this.setState({ selection: [] }, this.refreshData);
+  }
+
```
andreievg: somewhere here: 
function onSelectionChange

src/pages/SupplierInvoicePage.js
```diff
@@ -107,56 +208,234 @@ export class SupplierInvoicePage extends GenericPage {
...
-            <View style={globalStyles.pageTopRightSectionContainer} />
-          </View>
-          {this.renderDataTable()}
-        </View>
-      </View>
+      <GenericPage
+        data={this.state.data}
+        refreshData={this.refreshData}
+        renderCell={this.renderCell}
+        renderTopLeftComponent={this.renderPageInfo}
```
andreievg: onSelectionChange={..

src/pages/index.js
```diff
@@ -57,7 +55,7 @@ export const FINALISABLE_PAGES = {
     finaliseText: 'finalise_supplier_invoice',
   },
   externalSupplierInvoice: {
-    checkForError: checkForExternalSupplierInvoiceError,
+    checkForError: checkForSupplierInvoiceError,
```
andreievg: * externalSupplierInvoice key not needed
* supplierInvoice.checkForError = checkForSupplierInvoiceError

src/pages/index.js
```diff
@@ -24,11 +24,10 @@ import { StocktakeEditPage,
 import { StocktakeManagePage } from './StocktakeManagePage';
 import { StocktakesPage } from './StocktakesPage';
 import { SupplierInvoicePage } from './SupplierInvoicePage';
-import { SupplierInvoicesPage } from './SupplierInvoicesPage';
+import { SupplierInvoicesPage,
+         checkForFinaliseError as checkForSupplierInvoiceError,
```
andreievg: checkForFinaliseError from `./SupplierInvoicePage`

src/pages/SupplierInvoicePage.js
```diff
@@ -1,100 +1,201 @@
...
+    this.state = {
+      totalPrice: 0,
+      modalKey: null,
+      modalIsOpen: false,
+      selection: [],
+    };
+    autobind(this);
+  }
+
+  updateDataFilters(newSearchTerm, newSortBy, newIsAscending) {
```
andreievg: my eslint complains here on updateDataFilters line: `updateDataFilters should be placed after getModalTitle (react/sort-comp)`

src/pages/SupplierInvoicePage.js
Chris-Petty: What's the point of making an external supplier invoice if you can say how much you received? Where else would the quantity value come for an external invoice?

In both cases the quantities in the unfinalised invoices needs to be editable, so that the receiving store can process it, make adjustments and finalise. This leaves a trail of any discrepancies that have occurred.

In terms of the other fields, being able to edit and correct and discrepancies is what is needed - it might be a mistake, it might be foul play, but without the ability for the receiver to make changes there isn't a discernible audit trail.

Quick non-foul example: Supplier claimed to send batch A, customer received batch B. Batch B has a recall because bad. Good thing receiving store corrected the received batch according to label/inspection before finalising supplier invoice, now they aren't dispensing a bad batch.

andreievg: @Chris-Petty , this is a shared SupplierInvoicePage now, I was talking about SupplierInvoice not created but received.  So when Supplier Invoice comes from central store:

screen shot 2017-08-03 at 2 18 11 pm

Editable fields: Their Ref, Comment, pack size, quantity, batch, expiry, cost price, delete.. and can add a new line.. It looks to me like it's adding room for error. @craigdrown  ?

edmofro: @andreievg Chris understood, and his example relates to your concern exactly. Editable everything is the way @craigdrown and I talked about.

andreievg: ok if you are both happy that you can edit pack size, co$t price, batch, expiry, Their Ref, Comment and can add new lines, all for Supplier Invoice coming from central store, then discard these comments.

Chris-Petty:
Crash on deleting supplierInvoice:

image

andreievg:
re undefined.removeSelf.. in original external supplier invoices pr, I made a commit https://github.com/sussol/mobile/pull/430/commits/f320951fe844617c76d7dcfa664b3bc7e952737f,
it doesn't show for some reason in the file changes for that pr.. but it was working in that branch, may need to add to Transaction datatype as per that commit:

```JS
  removeSelf(database) {
    if (this.isFinalised) throw new Error('Cannot delete finalised transaction');
    const transactionBatchesIds =
      this.getTransactionBatches(database).map(transactionBatch => transactionBatch.id);

    this.removeTransactionBatchesById(transactionBatchesIds);

    database.delete(this);
  }
```

src/database/DataTypes/Transaction.js
andreievg: I found a bug, it wasn't evident when ItemBatches were deleted from within ExternalSupplierInvoices page.  As per issue #449, origin isExternalSupplierInvoice check in TransactionBatch destructor
Please change:
```
transaction.isExternalSupplierInvoice
return this.otherParty.isExternalSupplier
```
```
Name.isExternalSupplier
return this.type === 'facility' && this.isSupplier;
```

edmofro: Done

src/pages/SupplierInvoicePage.js
edmofro: Done

src/pages/SupplierInvoicePage.js
edmofro: Done

src/pages/index.js
edmofro: Done

src/pages/index.js
edmofro: Done

edmofro:
Fixes #449

edmofro:
Think all issues are now addressed in this PR

@DhanyaHerath Could you build a new apk and give the app another test run? Particularly focusing on the bugs you found first time round

PR #446, opened 2017-08-02
#332 fix lines si between mobiles

### LINKED ISSUE #332: Handle stock transfers
Build Number: N/A

Description: If a customer invoice for a customer which happens to be another store is finalised in mobile, it should generate a supplier invoice in the customer store (which then gets synced back to the server)

Reproducible: Haven't checked yet - may not be an issue?

Reproduction Steps:

Comments:

nmadruga:
Branched from #440 
fixes #332

### DIFF
--- a/src/dataMigration.js
+++ b/src/dataMigration.js
@@ -59,7 +59,7 @@ const dataMigrations = [
     migrate: database => {
       // Changed SyncQueue to expect no more than one SyncOut record for every record in database.
       // Assume that last SyncOut record is correct.
-      const allRecords = database.objects('SyncOut').sorted('changeTime');
+      const allRecords = database.objects('SyncOut').sorted('changeTime').snapshot();
       database.write(() => {
         allRecords.forEach(record => {
           const hasDuplicates = allRecords.filtered('recordId == $0', record.id).length > 1;
--- a/src/dataMigration.js
+++ b/src/dataMigration.js
@@ -59,7 +59,7 @@ const dataMigrations = [
     migrate: database => {
       // Changed SyncQueue to expect no more than one SyncOut record for every record in database.
       // Assume that last SyncOut record is correct.
-      const allRecords = database.objects('SyncOut').sorted('changeTime').snapshot();
+      const allRecords = database.objects('SyncOut').sorted('changeTime');
       database.write(() => {
         allRecords.forEach(record => {
           const hasDuplicates = allRecords.filtered('recordId == $0', record.id).length > 1;
--- a/src/dataMigration.js
+++ b/src/dataMigration.js
@@ -69,7 +69,7 @@ const dataMigrations = [
             // Transactions and Requisitions need to be synced after all their children records
             // for 2.0.0 interstore features
             record.changeTime = new Date().getTime();
-            this.database.update('SyncOut', record);
+            database.update('SyncOut', record);
           }
         });
       });
--- a/src/dataMigration.js
+++ b/src/dataMigration.js
@@ -59,11 +59,18 @@ const dataMigrations = [
     migrate: database => {
       // Changed SyncQueue to expect no more than one SyncOut record for every record in database.
       // Assume that last SyncOut record is correct.
-      const allRecords = database.objects('SyncOut').sorted('changeTime');
+      const allRecords = database.objects('SyncOut').sorted('changeTime').snapshot();
       database.write(() => {
         allRecords.forEach(record => {
-          const hasDuplicates = allRecords.filtered('recordId == $0', record.id).length > 0;
-          if (hasDuplicates) database.delete('SyncOut', record);
+          const hasDuplicates = allRecords.filtered('recordId == $0', record.id).length > 1;
+          if (hasDuplicates) {
+            database.delete('SyncOut', record);
+          } else if (record.recordType === 'Transaction' || record.recordType === 'Requisition') {
+            // Transactions and Requisitions need to be synced after all their children records
+            // for 2.0.0 interstore features
+            record.changeTime = new Date().getTime();
+            this.database.update('SyncOut', record);
+          }
         });
       });
     },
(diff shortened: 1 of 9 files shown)


src/sync/SyncQueue.js
```diff
@@ -55,36 +55,52 @@ export class SyncQueue {
...
+        const recordsToDelete = this.database
+          .objects('SyncOut')
+          .filtered('recordId == $0', record.id);
         this.database.delete('SyncOut', recordsToDelete);
       }
       switch (changeType) {
         case CREATE:
         case UPDATE:
         case DELETE: {
+          let duplicate = false;
```
edmofro: Change to `isDuplicate` to indicate it is a boolean

src/sync/SyncQueue.js
```diff
@@ -55,36 +55,52 @@ export class SyncQueue {
...
-                              .length > 0;
+
+          if (recordType === 'Transaction' && record.isFinalised) {
+            // Remove old snapshots from the same transaction to be sync with lines
+            const result = this.database
+              .objects('SyncOut')
+              .filtered('recordType == $0 && recordId == $1', recordType, record.id);
+
+            Object.entries(result).forEach(([key, value]) => {
+              this.database.delete('SyncOut', value);
```
edmofro: IMO this is overkill. I would say just update the `changeTime` for any record with `isDuplicate === true`. If you want to go further, any record with `isDuplicate && record.isFinalised`, but I think the former is fine and makes the logic simpler.

nmadruga: Cool! Chris explained here to me, I see what you mean know. If that way the record will go to the end of the line like that is fine. We just also have to search for `recordType == CREATE` records also when we update the `changeTime`.

src/sync/SyncQueue.js
nmadruga: Indeed

src/sync/SyncQueue.js
edmofro: I would say you'll be ok to change the duplicate search to ignore `changeType`, and if it's a duplicate update both `changeTime` and `changeType`...after all we only really need to send through the last thing that has happened. So long as mSupply doesn't balk if we send through a delete for a record that doesn't yet exist on the server.

Chris-Petty:
A thought: Might have to migrate sync out table so there is only one record.

edmofro:
@Chris-Petty You mean so that any finalised transactions sync after their lines? That's kind of a good point. Yuck.

edmofro:
Looks great!

src/dataMigration.js
```diff
@@ -52,4 +54,18 @@ const dataMigrations = [
...
+      // Assume that last SyncOut record is correct.
+      const allRecords = database.objects('SyncOut').sorted('changeTime');
+      database.write(() => {
+        allRecords.forEach(record => {
+          const hasDuplicates = allRecords.filtered('recordId == $0', record.id).length > 0;
+          if (hasDuplicates) database.delete('SyncOut', record);
+        });
+      });
+    },
+  },
```
edmofro: This is nice, but it isn't enough. If there's a transaction that's been updated once, then had lines added and then finalised, the sync queue will still be out of order after this migration code is run. Soz.

Chris-Petty: Well it'll currently just deletes the whole queue because `> 0` haha. Given changing to `> 1`. I got what you mean. It's a bit dirty but I could just add all record type 'Transaction' (and requisitions?) SyncOuts to the end of the queue right?

src/dataMigration.js
```diff
@@ -52,4 +54,25 @@ const dataMigrations = [
...
   },
+  {
+    version: '2.0.0',
+    migrate: database => {
+      // Changed SyncQueue to expect no more than one SyncOut record for every record in database.
+      // Assume that last SyncOut record is correct.
+      const allRecords = database.objects('SyncOut').sorted('changeTime');
+      database.write(() => {
+        allRecords.forEach(record => {
+          const hasDuplicates = allRecords.filtered('recordId == $0', record.id).length > 1;
```
andreievg: I think you need snapshot or slice when deleting within forEach of sorted realm array

Chris-Petty: Yup it's in look at commits :D

Chris-Petty:
Should test mSupply can handle receiving transaction batches before the transaction.

This would happen if someone made an invoice, added some items and finalised within the 10min sync period. @adrianwb might know off the top of his head, otherwise I'll check server side tomorrow.

adrianwb:
@Chris-Petty Off the top of my head, the server won't care - even if this happens (and I would expect that the server would always receive an insert record when the transaction is created), the line would temporarily be orphaned until the transaction arrives, which isn't a problem, unless of course the transaction never arrives. Will do no harm to test though, just to be sure!

andreievg:
What about delete record, no create or update record previously.

adrianwb:
@andreievg If the server gets a delete sync record and the referenced record doesn't exist, then it will just ignore it but it will generate a log message.

src/pages/SupplierInvoicePage.js
```diff
@@ -431,11 +432,9 @@ export class SupplierInvoicePage extends React.Component {
 }
 
 export function checkForFinaliseError(transaction) {
-  if (transaction.items.length === 0) {
-    return modalStrings.add_at_least_one_item_before_finalising;
-  } else if (transaction.totalQuantity === 0) {
-    return modalStrings.stock_quantity_greater_then_zero;
-  }
+  if (!transaction.isExternalSupplierInvoice) return null;
```
edmofro: I would hope no internal supplier invoices are being sent without any items/quantity...

Chris-Petty: You'd hope, but it can be done. Just did it. mSupply doesn't validate what it sends out.

Chris-Petty: Maybe a mantis issue in around that is appropriate...

PR #450, opened 2017-08-08
#448 V3 sequence numbers

### LINKED ISSUE #448: V3 Sequence numbers
##### Mobile
* [x] Incoming Requisitions, Supplier Invoice to be assigned next sequence number if sequence number is -1, sync changes (sync will be disables while processing incoming records)

##### mSupply
* [ ] When duplicating 'transfer' record, the SI and Response requisition should have sequence number -1
* [ ] When creating CI from central store to mobile, on finalised set sequence number -1 for SI

Chris-Petty:
Listens to database changes with 'sync' param and builds a queue of records to run post sync operations on. These are obviously run after sync.

Initial sync ignores the queue and just crawls necessary tables (`checkTables()`) for records.

If a sync is interrupted, it'll run a `checkTables()` also. 

fixes #448 
branched off https://github.com/sussol/mobile/pull/446


src/mSupplyMobileApp.js
```diff
@@ -57,6 +56,7 @@ class MSupplyMobileAppContainer extends React.Component {
     this.userAuthenticator = new UserAuthenticator(this.database, this.settings);
     const syncAuthenticator = new SyncAuthenticator(this.settings);
     this.synchroniser = new Synchroniser(database, syncAuthenticator, this.settings);
+    this.postSyncProcessor = new PostSyncProcessor(this.database);
     this.scheduler = new Scheduler();
     const initialised = this.synchroniser.isInitialised();
```
edmofro: Can we change `initialised` to `isInitialised`? Been caught out on a boolean recently, good to have the indicator

src/mSupplyMobileApp.js
```diff
@@ -83,21 +83,26 @@ class MSupplyMobileAppContainer extends React.Component {
...
+        // Only reauthenticate if currently logged in
         this.userAuthenticator.reauthenticate(this.onAuthentication);
       }
     }, AUTHENTICATION_INTERVAL);
   }
 
   componentDidMount() {
     BackHandler.addEventListener('hardwareBackPress', this.handleBackEvent);
+    if (this.state.initialised && this.synchroniser.isSyncing()) {
+      // Sync was interrupted
```
edmofro:
```
// Sync was interrupted, so ensure any recently synced records have been properly post sync processed
```

src/mSupplyMobileApp.js
```diff
@@ -281,6 +288,4 @@ function mapStateToProps({ navigation: navigationState }) {
   return { currentTitle, finaliseItem, navigationState };
 }
 
-export const MSupplyMobileApp = connect(
-  mapStateToProps,
-)(MSupplyMobileAppContainer);
+export const MSupplyMobileApp = connect(mapStateToProps)(MSupplyMobileAppContainer);
```
edmofro: So much noise from prettier!

src/pages/RequisitionsPage.js
```diff
@@ -29,7 +29,8 @@ export class RequisitionsPage extends React.Component {
     this.state = {
       selection: [],
     };
-    this.requisitions = props.database.objects('Requisition');
+    // Hide Requisitions that have not been updated with a serialNumber in postSync
+    this.requisitions = props.database.objects('Requisition').filtered('serialNumber != "-1"');
```
edmofro: Move this logic into UIDatabase (see Customer for e.g.)

src/sync/SyncQueue.js
```diff
@@ -55,36 +55,52 @@ export class SyncQueue {
...
+        const recordsToDelete = this.database
+          .objects('SyncOut')
+          .filtered('recordId == $0', record.id);
         this.database.delete('SyncOut', recordsToDelete);
       }
       switch (changeType) {
         case CREATE:
         case UPDATE:
         case DELETE: {
+          let duplicate = false;
```
edmofro: Can we change to ```isDuplicate```?

Chris-Petty:
Testing points:

If sync pull occurs but doesn't acknowledge records with serialNumber: '-1', either records shouldn't be allocated serialNumbers or if they are, then the next sync of that record cannot cause a second allocation.

src/sync/PostSyncProcessor.js
```diff
@@ -0,0 +1,164 @@
...
+    }
+    // Add new entry at end of Map
+    this.recordQueue.set(record.id, recordType);
+  }
+
+  /**
+   * Runs post checks across tables, ensuring data is correct.
+   * Tables manually added as to not iterate over tables that don't have any post processing.
+   */
+  checkTables() {
```
edmofro: I think this method name could be a bit more descriptive. Like `processAnyFoos` where `Foo` is a word that means unprocessed record.

src/sync/PostSyncProcessor.js
```diff
@@ -0,0 +1,164 @@
...
+    }
+  }
+
+  /**
+   * Builds an array of post process functions based on conditions, runs a database write call
+   * to apply all the post process functions if there are any.
+   * @param  {object} record     The record changed
+   * @return {array}  An array of functions to be called for the given record
+   */
+  postProcessRequisition(record) {
```
edmofro: `generatePostProcessesForRequisition`

src/sync/PostSyncProcessor.js
```diff
@@ -0,0 +1,164 @@
...
+    return processes;
+  }
+
+  /**
+   * Builds an array of post process functions based on conditions, runs a database write call
+   * to apply all the post process functions if there are any.
+   * @param  {object} record     The record changed
+   * @return {array}  An array of functions to be called for the given record
+   */
+  postProcessTransaction(record) {
```
edmofro: `generatePostProcessesForTransaction`

src/sync/PostSyncProcessor.js
```diff
@@ -0,0 +1,164 @@
...
+    this.actionQueue = [];
+  }
+
+  /**
+   * Delegates records to the correct utility function
+   * @param  {string} recordType  The type of record changed (from database schema)
+   * @param  {object} record      The record changed
+   * @return {none}
+   */
+  delegateByRecordType(recordType, record) {
```
edmofro: `enqueuePostProcesses[ForRecordType]` (latter segment optional)

src/sync/Synchroniser.js
```diff
@@ -84,24 +94,41 @@ export class Synchroniser {
 
   /**
    * Return whether the synchroniser has been completely or partially initialised.
-   * @return {string} Either 'complete', 'partial', 'uninitialised'
+   * @return {string} Either 'complete', 'partial', 'uninitialised' TODO: make sure this is corrected
```
Chris-Petty: Part of me was hoping this would slip through haha. I'm pretty sure it's safe to change to `either 'true' or 'false'`, though I considered using this tweaked instead of  creating `SYNC_IS_SYNCED`.

src/sync/incomingSyncUtils.js
```diff
@@ -551,25 +563,33 @@ export function sanityCheckIncomingRecord(recordType, record) {
...
   if (!hasAllNonBlankFields) return false; // Return early if record already not valid
   const hasRequiredFields = requiredFields[recordType].canBeBlank.reduce(
     (containsAllFieldsSoFar, fieldName) =>
       containsAllFieldsSoFar &&
-      record[fieldName] !== null &&    // Key must exist
+      record[fieldName] !== null && // Key must exist
       record[fieldName] !== undefined, // May be blank, i.e. just ''
-      hasAllNonBlankFields); // Start containsAllFieldsSoFar as result from hasAllNonBlankFields
+    hasAllNonBlankFields
+  ); // Start containsAllFieldsSoFar as result from hasAllNonBlankFields
```
Chris-Petty: Nothing changed, I originally tried something here and didn't revert the pretty-ness.

src/sync/PostSyncProcessor.js
```diff
@@ -113,7 +108,7 @@ export class PostSyncProcessor {
    * @param  {object} record     The record changed
    * @return {array}  An array of functions to be called for the given record
    */
-  postProcessRequisition(record) {
+  generatePostProcessesForRequisition(record) {
```
edmofro: One more change now I see it... `generateProcessorsForRequisition` 
- It generates functions that have an effect on something, so they're 'processors' more than they are 'processes'
- We know it's post sync, that's the name of the class, so we can probably drop the 'post' bit.)

src/sync/PostSyncProcessor.js
```diff
@@ -113,7 +108,7 @@ export class PostSyncProcessor {
...
     const processes = [];
```
edmofro: `const processors = [];`

src/sync/PostSyncProcessor.js
```diff
@@ -1,25 +1,20 @@
...
 
 const { REQUISITION_SERIAL_NUMBER, SUPPLIER_INVOICE_NUMBER } = NUMBER_SEQUENCE_KEYS;
 
+import autobind from 'react-autobind';
+
 export class PostSyncProcessor {
   constructor(database) {
     this.database = database;
     this.recordQueue = new Map(); // Map of [recordId, recordType]
     this.actionQueue = [];
```
edmofro: `this.processorQueue = [];`

src/sync/PostSyncProcessor.js
```diff
@@ -55,11 +50,11 @@ export class PostSyncProcessor {
...
       .objects('Requisition')
-      .forEach(record => this.delegateByRecordType('Requisition', record));
+      .forEach(record => this.enqueuePostProcessesForRecordType('Requisition', record));
 
     this.database
       .objects('Transaction')
-      .forEach(record => this.delegateByRecordType('Transaction', record));
+      .forEach(record => this.enqueuePostProcessesForRecordType('Transaction', record));
 
     this.runActionQueue();
```
edmofro: `this.runQueuedProcessors();`

src/sync/PostSyncProcessor.js
```diff
@@ -0,0 +1,157 @@
...
+ */
+
+import { getNextNumber, NUMBER_SEQUENCE_KEYS } from '../database/utilities';
+
+const { REQUISITION_SERIAL_NUMBER, SUPPLIER_INVOICE_NUMBER } = NUMBER_SEQUENCE_KEYS;
+
+export class PostSyncProcessor {
+  constructor(database) {
+    this.database = database;
+    this.recordQueue = new Map(); // Map of [recordId, recordType]
```
andreievg: Interesting, the first time Map() is used in mobile, I guess it's easier to iterate

src/sync/PostSyncProcessor.js
```diff
@@ -0,0 +1,157 @@
...
+    this.actionQueue = [];
+  }
+
+  /**
+   * Delegates records to the correct utility function
+   * @param  {string} recordType  The type of record changed (from database schema)
+   * @param  {object} record      The record changed
+   * @return {none}
+   */
+  delegateByRecordType(recordType, record) {
```
andreievg: Umm. This looks workable, but not very extendable:
Can this class be made in a way where one can add to a queue of postProcess functions, i.e.:
```JS
this.postSyncProcessor = new PostSyncProcessor(this.database);
this.postSyncProcessor.addProcess('Requisition', (record, database) => {
  if (record.serialNumber === '-1') {
    record.serialNumber =  getNextNumber(database, REQUISITION_SERIAL_NUMBER)
    return true; // to tell postSyncProcessor to do this.database.update(recordType, record));
 } 
  return false;
});
```
In PostSyncProcessor class
```JS
addProcess(recordType, callback) {
  if(!this.processList[recordType]) this.processList[recordType] = [];
  this.processList[recordType].push(callback);
}

onDatabaseEvent(changeType, recordType, record, causedBy) {
...
if (causedBy !== 'sync' || recordType === 'SyncOut' || !Object.keys(this.processList).includes(recordType)) return;
...

  processRecordQueue() {
   this.database.write(() => {
    this.recordQueue.forEach((recordType, recordId) => { // ahh this is confusing for Map(), usually it's (key, value), here it's (value, key).
     
      const internalRecord = this.database.objects(recordType).filtered('id == $0', recordId)[0];
      
      this.dealWithRecordType(recordType, internalRecord);
      this.database.update('Requisition', record));
    });
  });
    this.recordQueue = []; 
  }

dealWithRecordType(recordType, record) {
  if (this.processList[recordType]) {
   this.processList[recordType].forEach(callback => {
   if (callback(record, this.database))
    this.database.update(recordType, record));
   });
  }   
}
```
Of course, code can look tidier then my markdown.

And similar logic for checkTables.

This way PostSyncProcessor becomes dynamic. It's easier to add multiple postSyncProcesses without touching PostSyncProcess logic.

src/sync/incomingSyncUtils.js
```diff
@@ -230,8 +233,11 @@ export function createOrUpdateRecord(database, settings, recordType, record) {
       const thisStoreId = settings.get(THIS_STORE_ID);
       const sequenceKey = SEQUENCE_KEYS.translate(record.name, EXTERNAL_TO_INTERNAL, thisStoreId);
       // Don't accept updates to number sequences
-      if (database.objects('NumberSequence')
-                  .filtered('sequenceKey == $0', sequenceKey).length > 0) break;
+      if (
+        database.objects('NumberSequence').filtered('sequenceKey == $0', sequenceKey).length > 0
```
andreievg: this time prettier, didn't really make things prettier..

src/sync/Synchroniser.js
```diff
@@ -56,52 +57,77 @@ export class Synchroniser {
...
     const syncIsInitialised = this.settings.get(SYNC_IS_INITIALISED);
     return syncIsInitialised && syncIsInitialised === 'true';
   }
 
   /**
+   * Return whether the whether or not the app is syncing. Useful for checking if sync was
+   * interrupted on app close or crash
+   * @return {boolean} 'true' if the app was (in case of app crash or close) or currently is syncing
+   */
+  isSyncing() {
```
andreievg: not a fan of having isSyncing in different places with similar meaning.  MSupplyMobileAppContainer, this.state.isSyncing, after site login in and fully synced initial dump. And here, if site is syncing, during or after initial dump  Maybe we can also have two callbacks to set MSupplyMobileAppContainer.state.isSyncing, at the same time this.setting.set(SYNC_IS_SYNCING.. or leave for now but we all take not for refactor

src/sync/incomingSyncUtils.js
Chris-Petty: Subjective haha. The bonus with prettier is that if we don't debate it, we don't spend time debating/implementing about how exactly to format something best.

src/sync/Synchroniser.js
```diff
@@ -56,52 +57,77 @@ export class Synchroniser {
     // initialisation, in which case we are continuing a failed partial initialisation. If the
     // serverURL is different, it is either completely fresh, or the URL has been changed so we
     // should start afresh
-    const isFresh = serverURL !== this.serverURL;
-    if (isFresh) this.database.write(() => { this.database.deleteAll(); });
+    const oldUrl = this.settings.get(SYNC_URL);
+    const isFresh = !oldUrl || serverURL !== oldUrl;
```
Chris-Petty: They get to enjoy multiple sites records obviously!

src/mSupplyMobileApp.js
```diff
@@ -83,21 +83,27 @@ class MSupplyMobileAppContainer extends React.Component {
...
+      if (this.state.currentUser !== null) {
+        // Only reauthenticate if currently logged in
         this.userAuthenticator.reauthenticate(this.onAuthentication);
       }
     }, AUTHENTICATION_INTERVAL);
   }
 
   componentDidMount() {
     BackHandler.addEventListener('hardwareBackPress', this.handleBackEvent);
+    if (this.state.isInitialised && this.synchroniser.isSyncing()) {
```
Chris-Petty: Both Synchroniser and PostSyncProcessor create listeners in their constructors so it should be fine (and both a instantiated on the mSupplyMobileApp constructor). 

I added the remove listeners to componentWillUnmount for good measure, but it's actually probably redundant.

Chris-Petty: Both Synchroniser and PostSyncProcessor create listeners in their constructors so it should be fine (and both a instantiated on the mSupplyMobileApp constructor). 

I added the remove listeners to componentWillUnmount for good measure, but it's actually probably redundant.

src/sync/PostSyncProcessor.js
Chris-Petty: Yea insertion order is guaranteed (key advantage here) and the object methods are pretty semantic.

`myMap.has(thisKey)` vs `!myObject[thiskey]`
`myMap.remove(thisKey)` vs ~~I guess making a new object with all the same properties except the one to remove~~ [Wow I never knew!](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete) `delete myObject[thiskey]`

src/mSupplyMobileApp.js
andreievg: So, when app is in the background, wouldn't component unmount ? Then when it's opened up again, PostSyncProcessor stops getting events from db ?

src/sync/incomingSyncUtils.js
andreievg: true. agree.

src/sync/PostSyncProcessor.js
```diff
@@ -0,0 +1,157 @@
...
+
+  /**
+   * Iterates through records added through listening to sync, adding needed actions
+   * to actionQueue. Runs the actionQueue, making the changes.
+   */
+  processRecordQueue() {
+    this.recordQueue.forEach((recordType, recordId) => {
+      // Use local database record, not what comes in sync. Ensures that records are
+      // integrated information (definitely after sync is done)
+      const internalRecord = this.database.objects(recordType).filtered('id == $0', recordId)[0];
```
edmofro: My belief is that in a case where you should always expect a condition to be met, you shouldn't safety check for that condition. If you do, there may be something going wrong that is silenced by your safety check. If you don't, the underlying problem is very brutally and obviously exposed by a crash (and a bugsnag report! heh)

src/sync/Synchroniser.js
edmofro: Tasty treat

src/sync/Synchroniser.js
edmofro: Agree

src/mSupplyMobileApp.js
edmofro: That's a good question @andreievg. If you're right, we need to change it in the page specific listeners too

Chris-Petty:
https://github.com/sussol/mobile/pull/450

Chris-Petty:
Wrong button wrong comment wrong place

src/sync/PostSyncProcessor.js
```diff
@@ -0,0 +1,157 @@
...
+ */
+  onDatabaseEvent(changeType, recordType, record, causedBy) {
+    // Exit if not a change caused by incoming sync
+    if (causedBy !== 'sync' || recordType === 'SyncOut') return;
+    if (this.recordQueue.has(record.id)) {
+      // Check if already in queue, remove old
+      this.recordQueue.delete(record.id);
+    }
+    // Add new entry at end of Map
+    this.recordQueue.set(record.id, recordType);
```
Chris-Petty: We could, but it'd only save a few milliseconds of processing the record queue (by moving recordType checking to during sync instead).

src/mSupplyMobileApp.js
Chris-Petty: Just did a quick test with a console log in `MSupplyMobileAppContainer` In `constructor`, `componentWillMount`, `componentDidMount` and `componentWillUnmount`.

First 3 only called on app start (putting in background doesn't seem to trigger any). Closing app triggers unmount. Apparently at that point this.database.removeAllListeners() is undefined anyway!

Plot gets deeper a little: Closing the app with back button triggers unmount, but going to the android open app manager view and swiping the app away _doesn't_ trigger the unmount code.

andreievg: App shouldn't close with back button. Should still be 'idle' in the background and should call mount when it's opened back up. I will make sure our testers today will try the following:
Press back button, go back in app, try to sync a record with -1 sequence number and see if updates.

src/sync/PostSyncProcessor.js
andreievg: Ability to add more functions/processors to PostSyncProcess is required in Inter-Store requisition, where Supply(Response) Requisition is synced, CustomerInvoice should be created and linked to that Requisition.  TransactionItems should be added mimicking RequisitionItems, this should be done after full successful sync.  I can refactor PostSyncProcessor in Requisition branch if it's not done here.

src/mSupplyMobileApp.js
Chris-Petty: Listeners and sync do not appear to operate when the app is in the background. Perhaps applies to all app code, hence componentWillUnmount doesn't get called when the app is closed while in background.

edmofro: We knew this, or at least @craigdrown did

src/sync/Synchroniser.js
Chris-Petty: #455 issue created for this, so we can get this PR going along while I resolve issue that before release.

PR #456, opened 2017-08-11
Inter store requisitions current

andreievg:


src/pages/RequisitionsPage.js
```diff
@@ -202,4 +262,16 @@ RequisitionsPage.propTypes = {
   genericTablePageStyles: PropTypes.object,
   topRoute: PropTypes.bool,
   navigateTo: PropTypes.func.isRequired,
+  settings: PropTypes.object.isRequired,
```
Chris-Petty: You've added this but I don't see it used?

src/pages/SupplyRequisitionPage.js
```diff
@@ -0,0 +1,362 @@
...
+  }
+}
+
+SupplyRequisitionPage.propTypes = {
+  database: PropTypes.object.isRequired,
+  genericTablePageStyles: PropTypes.object,
+  topRoute: PropTypes.bool,
+  runWithLoadingIndicator: PropTypes.func.isRequired,
+  requisition: PropTypes.object.isRequired,
+  settings: PropTypes.object.isRequired,
```
Chris-Petty: Added but unused?

src/pages/SupplyRequisitionPage.js
```diff
@@ -0,0 +1,362 @@
...
+      searchTerm: '',
+      sortBy: 'itemName',
+      isAscending: true,
+    };
+    autobind(this);
+  }
+
+  /**
+   * Respond to the user editing the number in the required quantity column
+   * @param  {string} key             Should always be 'requiredQuantity'
```
Chris-Petty: First check in method is for 'suppliedQuantity.'

src/pages/SupplyRequisitionPage.js
```diff
@@ -0,0 +1,362 @@
...
+    this.dataFilters = {
+      searchTerm: '',
+      sortBy: 'itemName',
+      isAscending: true,
+    };
+    autobind(this);
+  }
+
+  /**
+   * Respond to the user editing the number in the required quantity column
```
Chris-Petty: Might be worth noting here that it is updating the related customer invoice too.

src/pages/SupplyRequisitionPage.js
```diff
@@ -0,0 +1,362 @@
...
+        },
+      ],
+      [
+        {
+          title: `${pageInfoStrings.supplying_store_name}`,
+          info: requisition.otherStoreName.name,
+        },
+        {
+          title: `${pageInfoStrings.months_stock_required}:`,
+          info: Math.round(requisition.monthsToSupply),
```
Chris-Petty: I thought the month picker was integer based anyway?

src/pages/SupplyRequisitionPage.js
```diff
@@ -0,0 +1,362 @@
...
+  }
+
+  renderCell(key, requisitionItem) {
+    const { requisition } = this.props;
+    switch (key) {
+      case 'monthlyUsage':
+      case 'suggestedQuantity':
+      case 'requiredQuantity':
+      case 'ourStockOnHand':
+        return Math.round(requisitionItem[key]);
```
Chris-Petty: I know this is done else where in the code base, but I think we should show the user if there is a decimal, because hopefully they'll raise it as an issue. If we are internally setting quantities to decimal values or receiving decimal quantities from mSupply, we really need to know about that.

src/pages/SupplyRequisitionPage.js
```diff
@@ -0,0 +1,362 @@
...
+          <AutocompleteSelector
+            options={this.props.database.objects('Item')}
+            queryString={'name BEGINSWITH[c] $0 OR code BEGINSWITH[c] $0'}
+            queryStringSecondary={'name CONTAINS[c] $0'}
+            sortByString={'name'}
+            onSelect={(item) => {
+              const { database, requisition } = this.props;
+              database.write(() => {
+                if (!requisition.hasItemWithId(item.id)) {
+                  createRecord(database, 'RequisitionItem', requisition, item);
```
Chris-Petty: Does this make TransactionItem in the related customer invoice?

src/pages/SupplyRequisitionPage.js
```diff
@@ -0,0 +1,362 @@
...
+              if (newComment !== this.props.requisition.comment) {
+                this.props.database.write(() => {
+                  this.props.requisition.comment = newComment;
+                  this.props.database.save('Requisition', this.props.requisition);
+                });
+              }
+              this.closeModal();
+            }}
+          />
+          );
```
Chris-Petty: Indentation is out.

src/localization/buttonStrings.js
```diff
@@ -21,8 +21,11 @@ export const buttonStrings = new LocalizedStrings({
     new_supplier_invoice: 'New Supplier Invoice',
     past: 'Past',
     use_suggested_quantities: 'Use Suggested Quantities',
+    request_requisition: 'REQUEST STOCK',
+    response_requisition: 'SUPPLY STOCK',
+    use_requested_quantity: 'Use Requested Quantities',
```
Chris-Petty: Should really be 'use_requested_quantities'.

src/pages/SupplyRequisitionPage.js
```diff
@@ -0,0 +1,362 @@
...
+            sortable: true,
+          },
+          {
+            key: 'itemName',
+            width: 4,
+            title: tableStrings.item_name,
+            sortable: true,
+          },
+          {
+            key: 'ourStockOnHand', // TODO: should this column name be consistent with other pages
```
Chris-Petty: Should it? Unless we have excess column width then it'll be good to keep this as is, I wouldn't bother changing the other pages to conform.

src/pages/SupplyRequisitionPage.js
```diff
@@ -0,0 +1,362 @@
...
+            width: 1.5,
+            title: tableStrings.our_stock,
+            sortable: true,
+            alignText: 'right',
+          },
+          {
+            key: 'stockOnHand',
```
Chris-Petty: theirStockOnHand

src/pages/index.js
```diff
@@ -17,6 +17,9 @@ import { RequisitionsPage } from './RequisitionsPage';
 import { RequisitionPage,
          checkForFinaliseError as checkForRequisitionFinaliseError,
        } from './RequisitionPage';
+import { SupplyRequisitionPage,
+         checkForFinaliseError as checkForSupplierRequisitionError
+       } from './SupplyRequisitionPage'; // TODO finalise checking
```
Chris-Petty: Indeed TODO.

src/sync/incomingSyncUtils.js
```diff
@@ -274,8 +277,9 @@ export function createOrUpdateRecord(database, settings, recordType, record) {
         serialNumber: record.serial_number,
         requesterReference: record.requester_reference,
         comment: record.comment,
-        enteredBy: getObject(database, 'User', record.user_ID),
+        enteredBy: getObject(database, 'User', record.user_ID), // TODO: should be..
```
Chris-Petty: Should be what?

src/sync/outgoingSyncUtils.js
```diff
@@ -118,10 +116,10 @@ function generateSyncData(settings, recordType, record) {
         ID: record.id,
         date_entered: getDateString(record.entryDate),
         user_ID: record.enteredById,
-        name_ID: settings.get(THIS_STORE_NAME_ID),
+        name_ID: record.otherStoreName.id,  // TODO: check if there is a case when outgoing req might not have otherStoreName
```
Chris-Petty: Was checked?

package.json
```diff
@@ -4,7 +4,7 @@
     "email": "info@sussol.net"
   },
   "name": "mSupplyMobile",
-  "version": "1.1.6",
+  "version": "2.0.0",
   "private": true,
   "license": "proprietary",
```
andreievg: updated

src/dataMigration.js
```diff
@@ -73,6 +73,27 @@ const dataMigrations = [
...
+                                settings.get(SETTINGS_KEYS.SUPPLYING_STORE_NAME_ID));
+      if (nameResults.length < 1) throw new Error('Supplying Store Name ID missing from settings');
+      const mainSupplyingStoreName = nameResults[0];
+
+      const requisitions = database.objects('Requisition')
+                                   .filtered('otherStoreName = Null').snapshot();
+
+      database.write(() => {
+        requisitions.forEach(requisition => {
+          if (!requisition.isRequest) return;
```
andreievg: I am pretty sure realm doesn't store getters + functions from datatype in app data. It will construct Requisition data type as it load realm records. So yeah it will have the getter without migration.

In V3 mSupply api four types of requisition, finalised(fn) and suggested(sg) for both isReponse and isRequest, so either isResponse or isRequest have same 'safeness' level

src/dataMigration.js
```diff
@@ -73,6 +73,27 @@ const dataMigrations = [
...
+          // The reason for database.update is we don't want to update finalized transactions
+          // and incase database.save is replaced by automatic way of invoking db event listeners
+          // simply assigning property may cause changes to finalized transaction to be synced.
```
andreievg: Should not re sync finalised requisitions as it will cause v2, v3 to replicate requisition. And as for otherStoreName (i.e name_ID), mSupply has migration code that already assigns correct name_ID.

src/database/DataTypes/Requisition.js
```diff
@@ -47,6 +51,10 @@ export class Requisition extends Realm.Object {
     this.daysToSupply = months * 30;
   }
 
+  get isLinkedToRequisition() {
+    return typeof this.linkedRequisition === 'object';
```
andreievg: Good point, I am pretty sure realm will store 'undefined' in object key that has no value. I will change this check to keep with consistency.

src/dataMigration.js
Chris-Petty: I'm happy with first answer, but second answer is a bit hazy. They types are `// imprest, forecast, request or response`, that doesn't have anything to do with statuses. If a store somehow has a requisition of type 'imprest' or 'forecast', it'll run this migration code.

src/dataMigration.js
Chris-Petty: Should add note of why we don't want them being synced, as you just described.

src/database/DataTypes/Requisition.js
```diff
@@ -61,6 +69,26 @@ export class Requisition extends Realm.Object {
...
+            filtered('linkedRequisition.id = $0', this.id).length > 0) return;
+    const transaction = createRecord(database, 'CustomerInvoice',
+                                  this.otherStoreName, /* TODO: USER */);
+    this.items.forEach(requisitionItem => {
+      const transactionItem = createRecord(database, 'TransactionItem',
+                                  transaction, requisitionItem.item);
+      database.save('TransactionItem', transactionItem);
+    });
+    transaction.linkedRequisition = this;
+    this.linkedTransaction = transaction;
```
Chris-Petty: This occurs on initial sync with an error, I assume from the post processing call.

`null is not an object (evaluating 'linkedRequisition.linkedTransaction = transaction')`

Chris-Petty: @sinclairsystems and I looking into it

Chris-Petty: Actually, I've obviously pointed out the wrong bit of code.

src/database/DataTypes/Requisition.js
```diff
@@ -61,6 +69,26 @@ export class Requisition extends Realm.Object {
...
   }
 
+  createCustomerInvoice(database) {
+    if (this.isRequest || this.isFinalised) {
+      throw new Error('Cannot create invoice from Finalised or Request Requistion ');
+    }
+    if (database.objects('Transaction').
+            filtered('linkedRequisition.id = $0', this.id).length > 0) return;
+    const transaction = createRecord(database, 'CustomerInvoice',
+                                  this.otherStoreName, /* TODO: USER */);
```
andreievg: Updated, set enteredBy for current customerInvoice to be current user

src/database/DataTypes/TransactionItem.js
```diff
@@ -48,7 +48,7 @@ export class TransactionItem extends Realm.Object {
         this.transaction.isFinalised)) {
       return this.item.totalQuantity + this.totalQuantity;
     }
-    return this.item.totalQuantity;
+    return this.item.totalQuantity;  // TODO think about this logic, will never hit this return, createRecord, new invoice is always confirmed
```
andreievg: Ahh that's true.  It's not used in SupplierInvoice currently, I can see it could be.

src/database/DataTypes/TransactionItem.js
```diff
@@ -72,7 +72,10 @@ export class TransactionItem extends Realm.Object {
    * are finalised, whereas customer invoices immediately influence stock levels.
    * @param {double} quantity The total quantity to set across all batches
    */
-  setTotalQuantity(database, quantity) {
+   // TODO: document function param
```
andreievg: just did it

src/database/DataTypes/TransactionItem.js
```diff
@@ -72,7 +72,10 @@ export class TransactionItem extends Realm.Object {
...
+  setTotalQuantity(database, requestedQuantity, referenceQuantity = -1) {
+    let quantity = requestedQuantity;
+    if (referenceQuantity > 0) quantity = Math.min(requestedQuantity, referenceQuantity);
```
andreievg: As per Edwin's comment, `My belief is that in a case where you should always expect a condition to be met, you shouldn't safety check for that condition. If you do, there may be something going wrong that is silenced by your safety check. If you don't, the underlying problem is very brutally and obviously exposed by a crash (and a bugsnag report! heh)` in https://github.com/sussol/mobile/pull/450, I don't see a case where setTotalQuanity is called on finalised Transaction. If we were to check for finalised here, we might as well check it for every data type function... and that's a bit of refactoring

src/database/DataTypes/RequisitionItem.js
```diff
@@ -25,6 +25,18 @@ export class RequisitionItem extends Realm.Object {
     const daysToSupply = this.requisition ? this.requisition.daysToSupply : 0;
     return Math.ceil(Math.max((this.dailyUsage * daysToSupply) - this.stockOnHand, 0));
   }
+
+  get linkedTransactionItem() {
+    if (this.isRequest || !this.requisition.linkedTransaction) return null;
+    const resultList = this.requisition.linkedTransaction.items.filtered('item.id = $0', this.item.id);
```
andreievg: Yes indeed that would also work, and it's a little bit less code, so I changed.

src/database/DataTypes/TransactionItem.js
edmofro: My comment doesn't apply here @andreievg, because if it is called when the transaction is finalised, it won't cause any obvious error. It's a very different situation to the one I commented on.

src/pages/CustomerInvoicesPage.js
```diff
@@ -92,9 +92,15 @@ export class CustomerInvoicesPage extends React.Component {
       });
     }
     this.setState({ selection: [] }, this.refreshData); // Clear any invoices selected for delete
-    this.props.navigateTo('customerInvoice',
-                          `${navStrings.invoice} ${invoice.serialNumber}`,
-                          { transaction: invoice });
+    // Navigate to requisition if one is linked to selected customer invoice
+    const { linkedRequisition } = invoice;
+    const navigateToParams = linkedRequisition ?
```
andreievg: wow, that's sad, I've put it under Requisition.. changed

src/sync/outgoingSyncUtils.js
andreievg: I can't think of a case, but will leave this to adjust last (in case there is a case)

src/database/UIDatabase.js
```diff
@@ -13,10 +13,14 @@ export class UIDatabase {
       case 'Supplier':
         results = results.filtered('isVisible == true AND isSupplier == true');
         break;
+      case 'mSupplyStore':
```
andreievg: Yes indeed kudos. I think all cases should be returns, don't see how break-ing every case is helpful here

edmofro: There is a principle in the react community of 'single return'. We try to follow where it is easy

src/pages/RequisitionsPage.js
```diff
@@ -112,26 +126,51 @@ export class RequisitionsPage extends React.Component {
...
+      case 'otherStoreName':
+        return requisition.otherStoreName.name;
       case 'status':
         return formatStatus(requisition.status);
       case 'delete':
         return {
           type: 'checkable',
           icon: 'md-remove-circle',
-          isDisabled: requisition.isFinalised,
+          isDisabled: requisition.isFinalised || !requisition.isRequest,
```
andreievg: isRequest used 4 times, 2 directly and two with not, can make isResponse getter, if it's more readable

src/pages/RequisitionsPage.js
```diff
@@ -112,26 +126,51 @@ export class RequisitionsPage extends React.Component {
...
     }
   }
 
+  renderRequisitionTypeToggle() {
+    return (
+      <View style={globalStyles.horizontalContainer}>
+        <PageButton
+          style={[localStyle.toggleButton, this.state.isRequestView ?
+                            localStyle.selectedButton : {}]}
+          text={buttonStrings.request_requisition} /* TODO: change to agreed name  */
```
andreievg: yes request stock and supply stock

src/pages/RequisitionsPage.js
andreievg: thanks, good catch, I was using it earlier but forgot to remove.

src/database/UIDatabase.js
edmofro: Just reading around...it's less than unanimously agreed upon. Up to your preference

src/pages/SupplyRequisitionPage.js
andreievg: Yes, this pr lacks comments, hard!

src/pages/SupplyRequisitionPage.js
andreievg: Umm I copy pasted from RequisitionPage, ( I don't see any use for Math.round.. ).. changed in both.

src/pages/SupplyRequisitionPage.js
andreievg: Umm how many decimals should we allow ? 1, 2 or 3 ?

src/pages/SupplyRequisitionPage.js
andreievg: oops this modal show not exists here

src/pages/SupplyRequisitionPage.js
andreievg: DataType: RequisitionItem.stockOnHand.

src/pages/SupplyRequisitionPage.js
andreievg: yes removed thanks

src/localization/buttonStrings.js
andreievg: yes indeed, replaced

src/sync/incomingSyncUtils.js
andreievg: Wasn't sure if it there is a need to change this, look like there is no need.

src/pages/index.js
andreievg: I don't see any reason to do before finalise checking for SupplyRequisition.

src/database/DataTypes/Requisition.js
andreievg: yes that's a bug, fixed

src/dataMigration.js
andreievg: I think it should be fine to add a otherStoreName as mainSupplyingStoreName for any requisition prior to v3. As they will only be requisitions for main supplying store

src/dataMigration.js
andreievg:
```JS
// The reason for database.update is we don't want to update finalized transactions
// and incase database.save is replaced by automatic way of invoking db event listeners
// simply assigning property may cause changes to finalized transaction to be synced.
```
That should be descriptive enough

edmofro: What's going on here?
> Should not re sync finalised requisitions as it will cause v2, v3 to replicate requisition.
Surely not?

src/database/DataTypes/TransactionItem.js
andreievg: Very good, changed to have finalised check first.

src/dataMigration.js
andreievg: I should adjust my comment and say, possible will cause v2, v3 to replicate requisition. But I do think that finalised requisition should not be updated

src/pages/SupplyRequisitionPage.js
```diff
@@ -0,0 +1,334 @@
...
+  database: PropTypes.object.isRequired,
+  genericTablePageStyles: PropTypes.object,
+  topRoute: PropTypes.bool,
+  runWithLoadingIndicator: PropTypes.func.isRequired,
+  requisition: PropTypes.object.isRequired,
+};
+
+export function checkForFinaliseError() {
+  return null;
+}
```
edmofro: ...this seems unnecessary. Or are you going to update with an actual check?

src/database/DataTypes/TransactionItem.js
```diff
@@ -70,12 +70,16 @@ export class TransactionItem extends Realm.Object {
    * N.B. For customer invoices, will create and delete transaction batches as appropriate.
    * N.B. Supplier invoices do not take effect on the rest of the stock until they
    * are finalised, whereas customer invoices immediately influence stock levels.
-   * @param {double} quantity The total quantity to set across all batches
+   * @param {Database} Realm DB wrapper
+   * @param {double} requestedQuantity try to set this quanity against all batches
+   * @param {double} referenceQuantity max available quantity or -1 to use requestedQuantity
    */
-  setTotalQuantity(database, quantity) {
+  setTotalQuantity(database, requestedQuantity, referenceQuantity = -1) {
```
edmofro: Actually `referenceQuantity` is always being passed in as `transactionItem.availableQuantity`, and come to think of it we can probably always just cap `requestedQuantity` at `availableQuantity` if the transaction `isOutgoing`

src/pages/CustomerInvoicesPage.js
```diff
@@ -144,7 +150,7 @@ export class CustomerInvoicesPage extends React.Component {
         return {
           type: 'checkable',
           icon: 'md-remove-circle',
-          isDisabled: invoice.isFinalised,
+          isDisabled: invoice.isFinalised || invoice.isLinkedToRequisition,
```
andreievg: The invoice should have a title, from requisition {serialNumber}. And when you press it the invoice, will navigate to requisition. Yes more verbose to the user would be nice!

src/pages/SupplyRequisitionPage.js
andreievg: Maybe? 
```js
get therStockOnHand() {
  // check requisition is response 
  // return 0 or even throw and error
}
```
This also bugged me, stockOnHand not very representative

src/sync/outgoingSyncUtils.js
```diff
@@ -192,6 +191,8 @@ function generateSyncData(settings, recordType, record) {
         category_ID: record.category && record.category.id,
         confirm_time: getTimeString(record.confirmDate),
         store_ID: settings.get(THIS_STORE_ID),
+        requisition_ID: record.isLinkedToRequisition ?
```
andreievg: agree

src/database/DataTypes/Requisition.js
```diff
@@ -61,6 +65,26 @@ export class Requisition extends Realm.Object {
     this.addItem(requisitionItem);
   }
 
+  createCustomerInvoice(database, user) {
+    if (this.isRequest || this.isFinalised) {
+      throw new Error('Cannot create invoice from Finalised or Request Requistion ');
+    }
+    if (database.objects('Transaction').
+            filtered('linkedRequisition.id = $0', this.id).length > 0) return;
```
andreievg: This is for sync, in case of resync and CustomerInvoice arrived before Requisition. And if CustomerInvoice arrived after Requisition.. this is why this should only be done in postSyncProcess

src/database/DataTypes/Requisition.js
```diff
@@ -61,6 +65,26 @@ export class Requisition extends Realm.Object {
...
+    const transaction = createRecord(database, 'CustomerInvoice',
+                                  this.otherStoreName, user);
+    this.items.forEach(requisitionItem => {
+      const transactionItem = createRecord(database, 'TransactionItem',
+                                  transaction, requisitionItem.item);
+      database.save('TransactionItem', transactionItem);
```
andreievg: Sounds good, createRecord handles it all :)

src/database/DataTypes/RequisitionItem.js
```diff
@@ -25,6 +25,17 @@ export class RequisitionItem extends Realm.Object {
     const daysToSupply = this.requisition ? this.requisition.daysToSupply : 0;
     return Math.ceil(Math.max((this.dailyUsage * daysToSupply) - this.stockOnHand, 0));
   }
+
+  get linkedTransactionItem() {
+    if (this.isRequest || !this.requisition.linkedTransaction) return null;
+    return this.requisition.linkedTransaction.items.filtered('item.id = $0', this.item.id)[0];
```
andreievg: done

src/database/UIDatabase.js
```diff
@@ -47,6 +51,7 @@ function translateToCoreDatabaseType(type) {
   switch (type) {
     case 'Customer':
     case 'Supplier':
+    case 'mSupplyStore':
```
andreievg: sounds better. updated

src/localization/buttonStrings.js
```diff
@@ -21,8 +21,11 @@ export const buttonStrings = new LocalizedStrings({
     new_supplier_invoice: 'New Supplier Invoice',
     past: 'Past',
     use_suggested_quantities: 'Use Suggested Quantities',
+    request_requisition: 'REQUEST STOCK',
+    response_requisition: 'SUPPLY STOCK',
+    use_requested_quantities: 'Use Requested Quantities',
   },
-  gil: {
+  gil: {  // TODO: request_requisition, response_requisition, use_requested_quantities
```
andreievg: updated

src/mSupplyMobileApp.js
```diff
@@ -91,6 +91,7 @@ class MSupplyMobileAppContainer extends React.Component {
 
   onAuthentication(user) {
     this.setState({ currentUser: user });
+    this.postSyncProcessor.setUser(user);
```
andreievg: reduxify mobile .. september/october..

src/pages/CustomerInvoicePage.js
```diff
@@ -97,10 +97,11 @@ export class CustomerInvoicePage extends GenericPage {
    */
   onEndEditing(key, transactionItem, newValue) {
     if (key !== 'totalQuantity') return;
+    const { database } = this.props;
     this.props.database.write(() => {
```
andreievg: thanks, I missed that

src/pages/CustomerInvoicePage.js
```diff
@@ -97,10 +97,11 @@ export class CustomerInvoicePage extends GenericPage {
...
-      const quantity = Math.min(parsePositiveInteger(newValue), transactionItem.availableQuantity);
-      transactionItem.setTotalQuantity(this.props.database, quantity);
-      this.props.database.save('TransactionItem', transactionItem);
+      transactionItem.setTotalQuantity(database, parsePositiveInteger(newValue),
+                                                 transactionItem.availableQuantity);
+      database.save('TransactionItem', transactionItem);
```
andreievg: I saw this in original CustomerInvoicePage, so thought to keep:
```JS
transactionItem.setTotalQuantity(this.props.database, quantity);
      this.props.database.save('TransactionItem', transactionItem);
```

Will look into .dot notation assignments in the depth to setTotalQuantity, tomorrow

src/dataMigration.js
```diff
@@ -73,6 +73,27 @@ const dataMigrations = [
...
         });
       });
+
+      const nameResults = database.objects('Name').filtered('id = $0',
+                                settings.get(SETTINGS_KEYS.SUPPLYING_STORE_NAME_ID));
+      if (nameResults.length < 1) throw new Error('Supplying Store Name ID missing from settings');
+      const mainSupplyingStoreName = nameResults[0];
+
+      const requisitions = database.objects('Requisition')
+                                   .filtered('otherStoreName = Null').snapshot();
```
andreievg: Thanks, nice catch, updated.
As per capital Null, I didn't want to go deep into research, just found the following (since latest realm js docs didn't include undefined/null value search)
https://github.com/realm/realm-js/issues/162 , then pr file changes https://github.com/realm/realm-js/pull/386/files there an enum .. I didn't know if case sensitive
 `enum class Type { None, Number, String, KeyPath, Argument, True, False, Null } type; `, will check with = null and == null tomorrow

src/pages/RequisitionsPage.js
```diff
@@ -112,26 +126,51 @@ export class RequisitionsPage extends React.Component {
...
+        />
+        <PageButton
+          style={[localStyle.toggleButton, !this.state.isRequestView ?
+                            localStyle.selectedButton : {}]}
+          text={buttonStrings.response_requisition}
+          onPress={this.toggleRequisitionViewType}
+        />
+      </View>
+    );
+  }
```
andreievg: Will look into that tomorrow or refactor after the trip, I missed the ToggleBar component..

src/pages/index.js
```diff
@@ -17,6 +17,9 @@ import { RequisitionsPage } from './RequisitionsPage';
 import { RequisitionPage,
          checkForFinaliseError as checkForRequisitionFinaliseError,
        } from './RequisitionPage';
+import { SupplyRequisitionPage,
+         checkForFinaliseError as checkForSupplierRequisitionError,
```
andreievg: Can `checkForError:` be undefined in FINALISABLE_PAGES ? I updated assuming so, will check tomorrow

src/sync/incomingSyncUtils.js
```diff
@@ -264,6 +264,9 @@ export function createOrUpdateRecord(database, settings, recordType, record) {
     }
     case 'Requisition': {
       let status = REQUISITION_STATUSES.translate(record.status, EXTERNAL_TO_INTERNAL);
+      let otherStoreName;
+      const nameResults = database.objects('Name').filtered('id = $0', record.name_ID);
+      if (nameResults && nameResults.length > 0) otherStoreName = nameResults[0];
```
andreievg: Awesome, thanks. much better

src/sync/incomingSyncUtils.js
```diff
@@ -353,8 +360,10 @@ export function createOrUpdateRecord(database, settings, recordType, record) {
         status: STATUSES.translate(record.status, EXTERNAL_TO_INTERNAL),
         confirmDate: parseDate(record.confirm_date),
         theirRef: record.their_ref,
+        linkedRequisition,
       };
       const transaction = database.update(recordType, internalRecord);
+      if (linkedRequisition) linkedRequisition.linkedTransaction = transaction;
```
andreievg: Hmm, I will, in case the future requisition will have a persistent link to transaction.. this field is not present in mSupply so can't be synced.. but if updated, we could miss a .save

src/pages/SupplyRequisitionPage.js
```diff
@@ -0,0 +1,334 @@
...
+    if (key !== 'suppliedQuantity') return;
+    const { database } = this.props;
+    database.write(() => {
+      const transactionItem = requisitionItem.linkedTransactionItem;
+      if (!transactionItem) return;
+      transactionItem.setTotalQuantity(database, parsePositiveInteger(newValue),
+                                                 transactionItem.availableQuantity);
+      requisitionItem.suppliedQuantity = transactionItem.totalQuantity;
+      database.save('TransactionItem', transactionItem);
+      database.save('RequisitionItem', requisitionItem);
```
andreievg: but suppliedQuantity has changed.. suppliedQuantity -> record.actualQuan, and should be synced

src/pages/SupplyRequisitionPage.js
```diff
@@ -0,0 +1,334 @@
...
+          editableType: 'text',
+        },
+      ],
+    ];
+    return (
+      <PageInfo
+        columns={infoColumns}
+        isEditingDisabled={requisition.isFinalised}
+      />
+    );
```
andreievg: Ahh refactor septemeber/october

src/pages/SupplyRequisitionPage.js
andreievg: it's gone.. will check

src/pages/RequisitionsPage.js
andreievg: Implemented, looks so much nicer.

src/sync/outgoingSyncUtils.js
Chris-Petty: @edmofro if `record.linkedRequisition` is _false_, then it'll assign _false_ to `requisition_ID` rather than null/undefinted

edmofro: And...when would `record.linkedRequisition` be false?

src/pages/SupplyRequisitionPage.js
Chris-Petty: Don't bother on this PR, it's just an over all app idea/discussion

src/pages/index.js
edmofro: Yeah the FinaliseModal will run a checkForError function if it exists, otherwise it will just allow finalising without any check

src/pages/SupplyRequisitionPage.js
Chris-Petty: I'm wondering if it is a good practice to throw errors rather than return null with such checks else where. Would make bugsnag logs be more informative!

src/sync/incomingSyncUtils.js
edmofro: Can we please stop thinking about .save and other notifiers as related to sync. They call listeners, and any part of the app may want to know about the change of a given data type, so we just have to consistently make sure we call .save any time we use dot notation. If that's too hard, we will have to rule out dot notation all together, and always just use database.update.

To talk specifically to your point about it not being synced here, well at the point you are in incomingSyncUtils you are using SyncDatabase anyway, so any changes will not be added to the SyncQueue (as they will have `causedBy === 'sync'`)

src/pages/SupplyRequisitionPage.js
edmofro: Basically the rule is:
At the point you use dot notation to change a data object, you must call .save
If you just call a method that manipulates the data object, you don't call .save. You can feel save in the knowledge that if the method uses dot notation to change it, it is the methods responsibility to call .save - so long as writer of the method followed the rules, you are fine (perhaps a good idea to check inside the method if you want to be really sure)

src/pages/SupplyRequisitionPage.js
Chris-Petty: As for your idea with the get, that along with a get ourStockOnHand could be a nice tweak and minor refactor. I'm happy not bothering though given current time constraints.

src/pages/RequisitionsPage.js
```diff
@@ -31,7 +31,8 @@ export class RequisitionsPage extends React.Component {
       isRequestView: true,
       isCreatingRequisition: false,
     };
-    this.requisitions = props.database.objects('Requisition');
+    this.requisitions = props.database.objects('Requisition')
+                                      .filtered('serialNumber != "-1"');
```
Chris-Petty: @edmofro If you're wondering why we're back to doing this in the Requisition page:

SyncPostProcessor uses the UIDatabase, so when trying to find records with serialNumber '-1' it'd never find any because they were already filtered out.

I suppose a different solution would be make another switch option in UIDatabase for 'RawRequisitions' or something that doesn't do any filtering.

Or another function objectsDirect(type) that just passes straight to the realm objects without any filters.

edmofro: Yeah I worked that out after a bit...sorry for sending you astray. I don't think you should at this stage, but if you were going to add to the UIDatabase, I'd go in the other direction...keep `Requisition` as the 'raw' type, and add a new type for the filtered requisitions.

P.S. listening carefully to this, you might hear a hint that the postSyncProcessor shouldn't be using the UIDatabase ;-)

Chris-Petty: Yea, that would be another quick fix (Your P.S.)

edmofro: Quick and correct IMO. Code smells include:
- A bug happened (start sniffing...)
- The 'post sync' processor, which has nothing to do with UI, is using something called a 'UIDatabase' (there it is!)

src/sync/incomingSyncUtils.js
andreievg: ok gotcha.. I would prefer using update..

src/pages/SupplyRequisitionPage.js
andreievg: As per comment above where setTotalQuantity is used in CustomerInvoicePage -> before I refactored.

My original comment wasn't very good ( I misread your comment requisition item -> transaction item ). And yes, will leave 'sync' out of save and update talk and use them all the time dot notation assignment is used.

src/pages/SupplyRequisitionPage.js
andreievg: I am glad you mentioned this actually. Request requisition.stockOnHand should be 'dynamic' until finalised and update (alongside suggestedQuantity), not set in createRecord. I have other thought on this subject, will try to update by mid Tomorrow otherwise will make an issue.

src/pages/RequisitionsPage.js
andreievg: yes I agree, but, this doesn't help in deciphering differences in db wrappers:
```JS
    const database = new Database(schema);
    this.database = new UIDatabase(database);
    this.settings = new MobileAppSettings(this.database);
    migrateDataToVersion(this.database, this.settings);
    this.userAuthenticator = new UserAuthenticator(this.database, this.settings);
    const syncAuthenticator = new SyncAuthenticator(this.settings);
    this.synchroniser = new Synchroniser(database, syncAuthenticator, this.settings);
    this.postSyncProcessor = new PostSyncProcessor(this.database, this.settings);
```
A bit too messy in my opinion, this.database, database

PR #459, opened 2017-08-13
#455 initial sync site safety

### LINKED ISSUE #455: Can change initial sync site part way through initialising
Build Number: 2.0.0 alpha

Description: A initial sync is stopped part way through pulling the records for store A (closing the app for example). You open the app and reenter the details, but decide enter store B. Sync will resume where it left off with store A and sync any records queued for store B. You end up with a mess. Sync for store B if live is also silently broken (store B would not pull records such as supplier invoices, because store A took them from server queue).

Reproducible: Yes

Reproduction Steps: Follow description

Comments: We check the URL matches when restarting a sync, but nothing else. Should remember with settings the store details entered and compare (we can also use this to auto fill when reopening the app).

Chris-Petty:
fixes #455


src/authentication/SyncAuthenticator.js
```diff
@@ -65,8 +62,8 @@ export class SyncAuthenticator {
    * @return {string} Authorization header value
    */
   getAuthHeader() {
-    const username = this.settings.get('SyncSiteName');
-    const password = this.settings.get('SyncSitePasswordHash');
+    const username = this.settings.get(SYNC_SITE_NAME);
+    const password = this.settings.get(SYNC_SITE_PASSWORD_HASH);
```
Chris-Petty: Only real change, rest is prettier

PR #464, opened 2017-08-15
Add sync dashboard

edmofro:


src/sync/SyncModal.js
```diff
@@ -0,0 +1,167 @@
...
+    alignSelf: 'center',
+  },
+  progressDescription: {
+    fontSize: 18,
+    color: 'white',
+    textAlign: 'center',
+    marginBottom: 10,
+  },
+  row: {
+    width: Dimensions.get('window').width / 3,
```
edmofro: Hmmm...yeah maybe. I'll try it.

edmofro: Nope it does need to be on the row. Ready for merge!

package.json
```diff
@@ -4,7 +4,7 @@
     "email": "info@sussol.net"
   },
   "name": "mSupplyMobile",
-  "version": "2.0.0",
+  "version": "2.0.1",
```
nmadruga: Is this the right version? Aren't we going to release as 2.0.0?

edmofro: Yes you're right...will revert this

edmofro:
Those are good points @nmadruga! I think they do improve the UX, but for now they are not critical to Timor rollout, so I will leave them for you or the guys to chase up when Timor is finished. I am on loan, and probably overdue to return to Tupaia, so will focus on the critical.

package.json
Chris-Petty: We will release as 2.0.0, if the version number gets whack all we have to do is commit back to 2.0.0 and publish.

That said, I'd prefer something like 2.0.0-rc1 until release we increment the rc number.

edmofro: As I said when you made this comment earlier @Chris-Petty, I don't think you can do this with our smart android versioning atm. Can try to make it happen after looking at a couple of higher priorities, but have made v2.0.0 for now

nmadruga:
@edmofro Yes, indeed. I'll do that now

PR #466, opened 2017-08-16
Fixes #460

### LINKED ISSUE #460: Crash when clicking in "New Item" while creating requisitions
Build Number: Current master (last commit on 11/08/17 fd402ef6a296fa75f771dfe4b12640dc8c4a9c71)

Description: Mobile crashing after selecting first item on new requisition. It happening periodically. 

Reproducible: Not always

Reproduction Steps: This time was, after creating 2 other requisitions with items, editing one item quantity and finalising. I had sync them also before creating the new one that crashed when I tried to create new item.

Comments: I was running new mSupply changes branch #8293 on background. The mobile was running on Genymotion on the same machine.

edmofro:


package.json
```diff
@@ -4,7 +4,7 @@
     "email": "info@sussol.net"
   },
   "name": "mSupplyMobile",
-  "version": "2.0.0",
+  "version": "2.0.1",
```
Chris-Petty: @edmofro @nmadruga @andreievg It's a bit late now but in development it'd have been better to use 2.0.0-rc(what ever number) or beta tag. I think it'd be tidier to keep the next release as 2.0.0 (which we can still do as we haven't released these testing versions).

edmofro: Pretty sure our clever android build number logic doesn't like that, I've tried with Tupaia App before. Probably not hard to fix that though.

PR #469, opened 2017-08-17
Fix bug qith sync error to push new records

nmadruga:


Chris-Petty:
Good find, though might have been better to use [Map.prototype.clear()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/clear):
`this.recordQueue.clear()`

Issue #473, opened 2017-08-17
Customer invoices created for requisitions are confusing


nmadruga:
Build Number: current master (last commit 17/08/17 34a449a27baaab071b4817085c9bedb0282e3a1b)

Description: If you click on a Customer Invoice created by one request made to your store, instead of being sent to a Customer Invoice page, you get in the Requisiton Page. So the number of the invoice you clicked is not showed, at the top you see "Requisition 55" which is the number of the response requisition and in the comment you can see "From request requisition 99".

Here is the list with response requisitions made from Customer Store: Baucau Refferal Hospital
![screenshot_20170817-152326](https://user-images.githubusercontent.com/16461988/29397060-fdaeb210-836f-11e7-8208-b3cbb43b1818.png)

If you open it, you can see that it was creatd from request requisition 99 (serial_number on Baucau) and has it's own serial_number 55
![screenshot_20170817-152346](https://user-images.githubusercontent.com/16461988/29397051-eee8a768-836f-11e7-82f1-5f08fa22f005.png)

Here is the list of Customer Invoices and the 1st item is the CI created for that requisition from Customer: Baucau Referral Hospital, with invoice number 181.
![screenshot_20170817-152403](https://user-images.githubusercontent.com/16461988/29397045-df4979e0-836f-11e7-9474-79bcec79b6d0.png)

If you open it, takes you to the Requisitions page (that's why is showing the number 55 on the top). But look the comments, it tells the number of the request, and where is the invoice number?
![screenshot_20170817-152417](https://user-images.githubusercontent.com/16461988/29397038-d7bc4478-836f-11e7-9598-1d56ed24ab45.png)

Besides that, as we had seen before it says "Supplying Store" where should be "Customer Store" and everything gets very confusing.

Reproducible: Always

Reproduction Steps:

Comments:

Issue #472, opened 2017-08-17
Weird value "placeholderUser" as value in response requisitions


nmadruga:
Build Number: current master (last commit 17/08/17 34a449a27baaab071b4817085c9bedb0282e3a1b)

Description: "Entered By": placeholderUser. We think that would be better just to hide that field in case of response requisition. @edmofro said the format of the Date could be improved also.
![screenshot_20170817-152346](https://user-images.githubusercontent.com/16461988/29396922-19118fa6-836f-11e7-819e-c2556056db7d.png)

Reproducible: Always

Reproduction Steps:

Comments:

edmofro:
This is because the user_ID that comes in with the requisition record is just duplicated by mSupply server from the request requisition that generated this response requisition. So it's the user that entered the original requisition on the requesting site. That kind of does make sense, but seeing as we can't show it here because those users don't sync (is that true @adrianwb?) I have removed entered by from the info section.

adrianwb:
@edmofro Yes, that's right - users don't sync (unless we manually force it) so the receiving site most likely won't have a user record for the person who entered the original request. For response requisitions entered via desktop from the SAMES store, they will have a user_ID from someone at SAMES.

edmofro:
@adrianwb That's fine, we don't need to display the 'Entered By'...no one really cares in mobile, particularly not on a requisition. 

When you say 'For response requisitions entered via desktop from the SAMES store, they will have a user_ID from someone at SAMES.', what does that mean? When a response requisition is interacted with, you attach that user_ID? Or when a response requisition is generated from a request requisition that was created at SAMES, it gets the request creator's user_ID?

adrianwb:
Neither actually. I mean that it is possible for SAMES to enter response requisitions directly for any store without the store having had to generate an original request requisition (Customer -> Requisitions when logged into the SAMES store) - in that case, the user ID will be from the person who entered it.

edmofro:
Ah, right! So what is the desktop site behaviour when a response
requisition syncs in as a result of a request requisition in another store
and site? We should replicate that in mobile. The first user to interact
with it gets attached in user_ID? Or does it always stay as the original
user attached to the initiating request requisition?

Issue #471, opened 2017-08-17
Response requisition is showing "Supplier Invoice" that should be customer/requester


nmadruga:
Build Number: current master (last commit 17/08/17 34a449a27baaab071b4817085c9bedb0282e3a1b)

Description: If you are on tab "Supply Stock" you see the response requisitions made from another store (customer) to your store (supplier). When you open the requisition the Requisition Page (that is used also for request requisition) has one field "Supplier Store" in that case should be "Customer Store" or "Requester Store". But, is very important that the field stay like it is if you opened an request requisition (on tab "Request Stock").
![screenshot_20170817-152326](https://user-images.githubusercontent.com/16461988/29396541-97f28576-836c-11e7-8bbc-b649066bb8f9.png)
![screenshot_20170817-152346](https://user-images.githubusercontent.com/16461988/29396601-d00c6d82-836c-11e7-98b2-fa777a9b0694.png)

Reproducible: Always

Reproduction Steps: With 2 mobiles running, make store A send request to store B. Then sync both. On Store A mobile you should see the requisition on tab "Request Stock" and when you open it should say "Supplier Store": Store B correctly. But, on Store B mobile see requisitions on tab Supply Stock. Check if the response to the one you created is there, when you open it is saying "Supplier Store": Store A which is incorrect.

Comments:

PR #475, opened 2017-08-17
Change inter store requisition interface design

edmofro:
Major changes:
- Separated customer and supplier requisitions into pages with buttons
in appropriate sections of the menu
- Customer invoice no longer created automatically on syncing in a
response requisition
- Customer invoice created (and finalised) on finalising response
requisition

And fix a few things along the way:
- Add customer invoice number to customer requisition info section
- Change ‘Supplying Store’ to customer in customer requisition info
section
- Fix date formatting in ‘Entered Date’ everywhere
- Remove unhelpful ‘Entered By’ on both type of requisition
- Changed finalise message to be different for customer and supplier
requisitions


edmofro:
I still need to:
- [ ] Add a sensible enteredBy
- [ ] Work out why names are unreliably syncing
- [ ] Test further downstream: do the customer invoices generated from requisitions create a supplier invoice in the other store?

nmadruga:
Wow, that's looking way more user friendly. The Customer Invoice be created just after Customer Requisition is finalised very good. The Labels now make sense: fix #472 and fix #471. I don't know if I like the formatting on the dates yet. I didn't see the difference and it's cutting the time always. Also the customer name is cut, maybe we cold give more space there?
screenshot_20170818-104906
screenshot_20170817-151955
But besides those minor issues for me that change is wonderful. It solves the problem fix #473

edmofro:
Thanks @nmadruga! I agree about the customer name, and will work on that. However, for the entry date the fix was actually in another library (see [sussol-utilities](https://github.com/sussol/sussol-utilities/commit/a0b5151ccf7a7ffd1a36c150a5e8029316745b38)), so you just need to run `yarn` and it will be fixed :-)

Chris-Petty:
@Nicole just FYI fix #someIssueNumber doesn't work unless it is in the top comment or in the comment/commit description box when trying to merge.

src/localization/modalStrings.js
```diff
@@ -20,7 +20,8 @@ export const modalStrings = new LocalizedStrings({
     edit_the_requisition_comment: 'Edit the requisition comment',
     edit_their_reference: 'Edit their reference',
     finalise_customer_invoice: 'Finalise will lock this invoice permanently.',
-    finalise_requisition: 'Finalise will send this requisition and lock it permanently.',
+    finalise_supplier_requisition: 'Finalise will send this requisition and lock it permanently.',
+    finalise_customer_requisition: 'Finalise will generate a finalised customer invoice, adjust inventory, and lock this requisition permanently.',
```
Chris-Petty: Is this up to date with this implementation? The customer invoice already exists yes? To the user it doesn't exist sure.

Speaking of which, is it possible to just hold off assigning a serial number until the related requisition is finalised?

src/database/UIDatabase.js
```diff
@@ -7,18 +7,27 @@ export class UIDatabase {
...
-        return results.filtered(
-          "isVisible == true AND isSupplier == true AND type == 'facility'"
-        );
+        return results.filtered('isVisible == true AND isSupplier == true AND type == "facility"');
       case 'Item':
         return results.filtered('isVisible == true');
+      case 'RequestRequisition':
+        return results.filtered('type == "request"');
+      case 'ResponseRequisition':
+        return results.filtered('serialNumber != "-1" AND type == "response"');
```
Chris-Petty: Breaks post sync processor again I believe. This is the right way to do this. I'll commit my suggested fix in a moment.

edmofro: Doesn't break post sync, uses different term. Is the right way to do it.

src/localization/modalStrings.js
edmofro: Yes up to date with implementation. Doesn't exist to user, wording is user facing. When training, I think you should never mention that it exists under the hood...that's complicated. To them, we create it when they finalise the requisition.

edmofro: Not sure what your serial number point is, but you can try anything you like in the weekend ;-)

PR #494, opened 2017-08-21
Minor localisation and visual tweaks

andreievg:


src/pages/CustomerRequisitionPage.js
```diff
@@ -156,8 +156,8 @@ export class CustomerRequisitionPage extends React.Component {
       ],
       [
         {
-          title: `${pageInfoStrings.months_stock_required}:`,
-          info: requisition.monthsToSupply,
+          title: `${pageInfoStrings.customer}:`,
+          info: requisition.otherStoreName ? requisition.otherStoreName.name : '',
```
edmofro: I'm guessing this is to give more room for the Customer (and below, Supplier) name. Good call probably, though there was some niceness to having the two 'editable' fields in the same column.

PR #501, opened 2017-08-22
PostSyncProcessor clearing record queue more

Chris-Petty:


Chris-Petty:
I haven't tested this, maybe hold off. Should be safe though. **Famous last words**

PR #509, opened 2017-08-23
#xxx-onLogin-connection-should-timeout

andreievg:


src/authentication/UserAuthenticator.js
```diff
@@ -13,6 +13,7 @@ const {
  } = AUTH_ERROR_CODES;
 
 const AUTH_ENDPOINT = '/sync/v3/user';
+const CONNECTION_TIMEOUT_PERIOD = 10000; // 10 second timeout for authenticating connection
```
edmofro: I like to use 10 * 1000 to make it obvious that it is milliseconds, clean to change and avoid a potential mistake

src/authentication/UserAuthenticator.js
```diff
@@ -102,3 +106,13 @@ export class UserAuthenticator {
...
+
+// Promise will be called parallel to authenticating connection
+function createConnectionTimeoutPromise() {
+  return new Promise((resolve, reject) => {
+    const id = setTimeout(() => {
+      clearTimeout(id);
+      reject(new Error(CONNECTION_FAILURE));
+    }, CONNECTION_TIMEOUT_PERIOD);
+  });
+}
```
edmofro: All looks good @andreievg. I would recommend pulling this out into a more globally used utility that can be used instead of `fetch` everywhere. That's my setup in Tupaia App - I've got a `TupaiaApi` class that you can call 'get' or 'post' on, and it will handle the fetching and timeout.

andreievg: This should be globally available I agree, but I don't think we should wrap fetch because:
* UserAuthenticate uses -> authenticateAsync which is imported from 'sussol-utils', which imports from 'authentication-utilities' -> and only there fetch is used, it seems to deep to change.
* Have to change the createConnectionTimeoutPromise to take timeout period as an argument, and of course estimate how long each type of fetch should take and deduce timeout period, it's easy for Login but what about fetching records ? and what if we want to change how many records to fetch ?  

We should definitely be setting timeout for each fetch operation, but for now this fixes the urgent problem of Login never timing out (you can say the same goes for timeout when syncing, but that can be reset by login in and out manually.. for now)

PR #512, opened 2017-08-24
Fix Chris' crash

edmofro:


andreievg:
```JS
handleBackEvent() {
     const { navigation } = this.navigator.props;
```

you want to check here @edmofro , this is where it crashes

edmofro:
Right you are @andreievg, was moving too fast. Should be fixed now. Again, untested.

PR #514, opened 2017-08-24
Fix pages with complex nav state becoming unresponsive due to redux-persist

### LINKED ISSUE #513: RC9 - Pages UNRESPONSIVE
I have only tested this in emulator, don't have access to tablet until tomorrow.

Every page is unresponsive, no verbose or error. (emulated with Genymotion, android 6 or 4)

Reproduction:
Go to SupplierInvoicePage, RequisitionPage, CustomerInvoicePage, etc.  Inputs seems to freeze, clicking on anything doesn't work,including hardware back button.

I really really hope it's my emulator playing up (or I've done something wrong). If it's not, then -> Team, no more merging or submiting PR's unless tested !!!

edmofro:
Fixes #513

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -25,7 +25,7 @@ const store = createStore(
   applyMiddleware(thunk),
 );
 
-const persistedStore = persistStore(store, { storage: AsyncStorage });
+const persistedStore = persistStore(store, { blacklist: ['navigation'], storage: AsyncStorage });
 
 function App() {
   return (
--- a/src/main.js
+++ b/src/main.js
@@ -25,6 +25,9 @@ const store = createStore(
   applyMiddleware(thunk),
 );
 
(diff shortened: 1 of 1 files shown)


edmofro:
Looked further into it. It's because in those pages we have an enormous prop passed through, which is taking forever to serialise and persist to AsyncStorage. So there are two possible fixes:
- This one (blacklisting navigation from redux-persist)
- Refactor to not have the massive nested realm objects in the redux state

In the interests of simplicity and time, let's just go with the blacklisting option.

Chris-Petty:
@DhanyaHerath once there is a APK courtesy of @edmofro can you test this please?

edmofro:
On the way. Would be a good time to do another sweep through the app, my testing is always isolated to the thing I'm doing, and results in bugs like this awful one slipping through. That's an apology.

Issue #408, opened 2017-03-29
Stock Item usage lookup


andreievg:
Ability to check average usage per month for an item (or other specified period). Currently can only see when creating a requisition, should be able to look it up (probably best in stock item page)

Chris-Petty:
related #441

andreievg:
https://github.com/sussol/mobile/pull/520

Chris-Petty:
It's better that in #520 top comment to write "fixes #520". This issue will get an automatic reference shown here and when #520 is merged it'll automatically resolve this issue.

Chris-Petty:
Fixed in #520

PR #520, opened 2017-09-10
#408 Stock Item usage lookup

andreievg:
* Monthly Usage in stock page
* Batch information in stock page

### DIFF
--- a/src/localization/tableStrings.js
+++ b/src/localization/tableStrings.js
@@ -10,6 +10,7 @@ export const tableStrings = new LocalizedStrings({
     actual_quantity: 'ACTUAL\nQUANTITY',
     available_stock: 'AVAILABLE STOCK',
     batch_expiry: 'EXPIRY',
+    batch: 'Batch',
     category: 'Category',
     code: 'CODE',
     comment: 'COMMENT',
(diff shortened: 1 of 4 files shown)


andreievg:
screen shot 2017-09-11 at 3 57 25 pm

andreievg:
it seems like inconsistencies in font's character length
screen shot 2017-09-12 at 5 08 40 pm

andreievg:
So going with a comma and two spaces as per:
screen shot 2017-09-12 at 5 18 55 pm

GaryWilletts:
I realise I'm late to the party (didn't see stuff happening on Balsamiq etc) but the batch stuff seems really wordy. Why not have Batch: _batch name_, Expiry: _expiry date_, Quantity: _quantity_ i.e. ditch the name part. Just confusing and takes up extra screen space I think.

andreievg:
@GaryWilletts, I updated as per your suggestion, yes it does make it easier on the eye, thanks. And there was nothing on Balsamiq for this (there was an issue to include monthly usage in the expansion summary, and while implementing that we also implemented showing batch info, was quite easy).

Below is the new view (three batches, one with full info, next one with no name, and third one with no name or expiry):

screen shot 2017-09-28 at 2 14 31 pm

GaryWilletts:
Nice. Would be awesomer if you treated the 3 values as columns (fixed width divs?) so they all lined up (much quicker to read at a glance). Then you could put a "-" when there's no value. Gotta be a sinch for a chap of your talents ;-)

andreievg:
Thank you Gary.

In the StockPage Expansion component is used with PageInfo component (both written prior to may with mobile), there is a limit to what can be displayed with PageInfo (atm only text, editable text and 'selectable'). Since this PR focused on a low priority issue, and actually the batch info display wasn't even a part of that issue, I didn't really want to spend excessive time modifying (or replacing) PageInfo component to handle 'custom' columns and create another custom table <- as we like to create everything generic/reusable (which will be a sinch, but +time +complexity +100+ extra lines..).  I did try to align batch info columns with via text (tab, spaces) but in the font we use, character widths are inconsistent.

Given the importance and priority of the feature (low, low), simplicity of implementation (currently quite simple) and again simplicity of interface (only show what needs to be shown).. taking into account those three -> We should keep this change and this PR minimalistic (as is :)

PR #523, opened 2017-09-14
Filtering in realm explorer

andreievg:
* Can now filter selected table


andreievg:
screen shot 2017-09-14 at 4 10 00 pm

PR #528, opened 2017-09-18
Sync now accepts all transaction types

edmofro:


src/sync/syncTranslators.js
```diff
@@ -68,7 +68,11 @@ export const TRANSACTION_TYPES = new SyncTranslator({
   supplier_invoice: 'si',
   supplier_credit: 'sc',
   inventory_adjustment: 'in',
-  prescription: 'pi',
+  prescription: 'pi', // From here down just provided for sync purposes, not actually used in mobile
```
andreievg: Looks good.  A thought though:
* Should also have a way to handle all other types that are not found in SyncTranslators (currently if type translation does not exist, it will be set as 'undefined'). Maybe we should change SyncTranslator -> if translation is not found, set it to 'key' value (untranslated).  This way if the type is used in the future, and it's not in translator, we can just migrate the changes. And + we don't need to add all the types that are in the system.

edmofro:
Yeah that's probably the best idea.

PR #536, opened 2017-10-09
Update to latest generic pages and avoid deep equals

andreievg:


andreievg:
https://github.com/sussol/react-native-generic-table-page/pull/6

src/pages/GenericPage.js
```diff
@@ -64,4 +64,9 @@ GenericPage.propTypes = {
   finalisableDataType: PropTypes.string,
   refreshData: PropTypes.func.isRequired,
   topRoute: PropTypes.bool,
+  isDataCircular: PropTypes.bool,
+};
+
+GenericPage.defaultProps = {
+  isDataCircular: true,
```
Chris-Petty: Thats a pretty specific name that doesn't convey the difference very well. I'd just go straight `useDeepEquals` or `useShallowEquals`, depending on what you made the default in `react-native-generic-table-page`

Chris-Petty: Talked about it. We'll just deal with it for now (i.e. never going to be changed).

Issue #519, opened 2017-09-08
Stocktake by batch


andreievg:
Myanmar feature requirements, for version 2.1.0 -> Should be able do stock take by batch

Things to consider:

* Minimal or no changes to existing interface for current mobile users
  * Toggle for by [By Item][By Batch] -> this will work if batches are not deleted until finalisation (i.e. say there are two batches for the same item with quantity 50, when reducing quantity to 1 in [By Item], then switching to [By Batch], the batch that was emptied, remains with zero quantity
* Should be able to add new batch line
* On finalisation 0 stock batches to be removed

Chris-Petty:
fixed in #542

PR #542, opened 2017-10-12
#519 stocktake by batch

andreievg:
Tested the following:
* Upgrading from previous version with one finalised and one unfinalised stocktake
* Creating new Stocktake in this version and afterward, resyncing changes
* Finalising Stocktake and resyncing

Have not tested unfinalised Stocktake created in previous version, then clearing data, installing new version and resyncing (this will cause all StocktakeBatches in unfinalised stocktake to have totalCountedQuantity = 0).. Nothing can be done apart from server side migration, which is excessive in this case because ideally stocktake should be created/adjusted/finalised in one sitting.

This PR requires https://github.com/sussol/react-native-generic-table-page/pull/9 to be merged and published


src/database/DataTypes/StocktakeItem.js
```diff
@@ -55,48 +73,75 @@ export class StocktakeItem extends Realm.Object {
...
-   * @param  {Realm}   database        App wide database
-   * @param  {object}  transactionItem The transaction item
-   * @return {none}
+   * Function is used when adjusting inventory for the StocktakeItem
+   * rather then individual StocktakeBatches. Logic:
+   * Increasing -> From latest expiry to earliest expiry StocktakeBatch, increase
+   * to the maximum of snapshot quantity, after that if still have quantity
+   * to increase, apply it to earlies expiry batch.
+   * Reducing -> From earliest expiry to latest expiry StocktakeBatch, decrease
+   * quantity by maximum === current counted quantity
```
nmadruga: From what I understood here: "decrease quantity by maximum === current counted quantity" means that will not decrease below the current counted quantity for the batch? What is the current counted quantity? That's probably what got me here.

src/pages/expansions/StocktakeEditExpansion.js
```diff
@@ -0,0 +1,205 @@
...
+  }
+
+  renderAddBatchButton() {
+    const addNewBatch = () => {
+      this.props.database.write(() => {
+        this.item.createNewBatch(this.props.database);
+      });
+      this.refreshData();
+    };
+    return ( // TODO: localise addNewBatch
```
nmadruga: Left TODO?

src/database/DataTypes/StocktakeItem.js
Chris-Petty: I think latest/earliest can get mixed up in what they mean. I think it's because 'latest' often refers to dates in the past. Earliest/soonest/nearest and latest/furthest ? Not sure, I could be off base here.

Chris-Petty: What the argument for doing additions from latest to earliest?

src/database/DataTypes/StocktakeItem.js
```diff
@@ -55,48 +73,75 @@ export class StocktakeItem extends Realm.Object {
...
    */
-  applyBatchAdjustments(database, transactionItem) {
-    this.batches.forEach((stocktakeBatch) => {
-      const transactionBatch = transactionItem.getBatch(stocktakeBatch.itemBatchId);
-      let difference = 0;
-      if (transactionBatch) { // If a matching transaction batch, work out the difference
-        difference = transactionItem.transaction.isIncoming ?
-                            transactionBatch.totalQuantity :
-                            -transactionBatch.totalQuantity;
+  setCountedTotalQuantity(database, quantity) {
```
Chris-Petty: I've read through this a lot and thought about test cases. Seems solid.

src/database/utilities/createRecord.js
```diff
@@ -191,6 +191,7 @@ function createStocktakeBatch(database, stocktakeItem, itemBatch) {
     stocktake: stocktakeItem.stocktake,
     itemBatch: itemBatch,
     snapshotNumberOfPacks: numberOfPacks,
+    countedNumberOfPacks: numberOfPacks, // countedNumberOfPacks === snapshotNumberOfPacks
```
Chris-Petty: I'd rather an explanation somewhere why the 2 fields are equal.

src/pages/GenericPage.js
```diff
@@ -37,12 +37,12 @@ export class GenericPage extends React.Component {
...
 
-  // Refetch data and render the list any time sync changes data displayed, or the
-  // record is finalised
+  // Refetch data and render the list any time a listener is triggered for a data type
+  // that a page subscribes to listen to or if record is finalised
   onDatabaseEvent(changeType, recordType, record, causedBy) {
-    if ((causedBy === 'sync' &&
-        this.props.dataTypesSynchronised &&
-        this.props.dataTypesSynchronised.indexOf(recordType) >= 0) ||
+    const dataTypesArray = causedBy === 'sync' ? this.props.dataTypesSynchronised
```
Chris-Petty: This review stopped here. Will do another for the rest in morning.

src/pages/expansions/StocktakeEditExpansion.js
Chris-Petty: I think it's worthwhile "Request changes" a review even if only typos and feedback on code comments.

src/database/DataTypes/StocktakeItem.js
andreievg: If StocktakeItem quantity is changed to say zero, then increase (but below snapshot quantity), we should increase the last expiry batch first to follow FEFO rule.

src/pages/expansions/StocktakeEditExpansion.js
andreievg: Forgot to remove the TODO:

src/dataMigration.js
```diff
@@ -126,4 +126,24 @@ const dataMigrations = [
...
+      // so without below migration, unfinalised Stocktake countedTotalQuantity will be 0 for
+      // all StocktakeBatches and thus for all StocktakeItems, so we 'reset' unfinalised Stocktakes
+      // and set StocktakeBatches countedTotalQuantity === snapshotTotalQuantity
+      const unfinalisedStocktakeBatches = database.objects('StocktakeBatch')
+                                            .filtered('stocktake.status != "finalised"');
+      database.write(() => {
+        unfinalisedStocktakeBatches.forEach(StocktakeBatch => {
+          database.update('StocktakeBatch', {
+            id: StocktakeBatch.id,
+            countedNumberOfPacks: StocktakeBatch.snapshotNumberOfPacks,
```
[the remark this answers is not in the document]
andreievg: Yes indeed. Stocktake should ideally be created and finalised in one sitting anyways. I.e. if Customer Invoices are created or Supplier Invoices are finalised while there is a stocktake in progress, stocktake logic will be affected.

src/database/DataTypes/Stocktake.js
```diff
@@ -81,83 +73,65 @@ export class Stocktake extends Realm.Object {
...
   /**
-   * Finalises this stocktake, creating transactions to apply the stock changes to inventory
+   * Get or create reducing invoice for this stocktake
+   * @param {Realm}         database
+   * @param {Realm.Object}  user     The current user logged in
+   * @return {Realm.Object} Transaction for reduction
+   */
+  getReductions(database, user) {
+    if (!this.reductions) {
+      this.reductions = createRecord(database, 'InventoryAdjustment', user, new Date(), false);
```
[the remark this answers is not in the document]
andreievg: Yeah, I was trying to minimise passing too many parameters.. figured out a better way.. as per commit: https://github.com/sussol/mobile/pull/542/commits/6e50e41397d9ccdbac1962ca66cdf337ee9ed354

src/database/DataTypes/Stocktake.js
```diff
@@ -81,83 +73,65 @@ export class Stocktake extends Realm.Object {
...
-    // Prune off any items that did not get counted
-    database.delete('StocktakeItem', uncountedItems);
+  adjustInventory(database, user) {
+    // Get list of all StocktakeBatches associated with this stocktake
+    const stocktakeBatches = database.objects('StocktakeBatch')
+                             .filtered('stocktake.id = $0', this.id);
+     // Delete all StocktakeBatches that have been created by stocktake
+     // but have not been changed
+    database.delete('StocktakeBatch', stocktakeBatches.filter(stocktakeBatch =>
+      stocktakeBatch.snapshotTotalQuantity === 0 && stocktakeBatch.difference === 0));
```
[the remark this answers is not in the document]
andreievg: Only deleting StocktakeBatches that have been created as a result of adjustment in this stocktake.  Should we delete all stocktake batches + stocktake items that have not been changed ? What about historic data for Items with no stock changes ? And if no stock adjustments, having an empty stocktake ?

src/database/DataTypes/StocktakeItem.js
```diff
@@ -55,48 +73,75 @@ export class StocktakeItem extends Realm.Object {
...
-    return this.countedTotalQuantity !== undefined &&
-           this.countedTotalQuantity !== null &&
-           this.countedTotalQuantity < this.minimumTotalQuantity;
-  }
-
-  addBatch(stocktakeBatch) {
-    this.batches.push(stocktakeBatch);
-    if (!this.countedTotalQuantity) this.countedTotalQuantity = stocktakeBatch.countedTotalQuantity;
-    else this.countedTotalQuantity += stocktakeBatch.countedTotalQuantity;
+    const countedTotalQuantity = this.countedTotalQuantity;
```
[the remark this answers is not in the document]
andreievg: Since it's a getter now, saving a little bit of processing. I.e. only doing `getTotal(this.batches, 'countedTotalQuantity');` once , not three times.

src/database/utilities/createRecord.js
andreievg: I think they should be equal by default, otherwise need to check for null everywhere, and if null assume countedNumberOfPacks === snapshotNumberOfPacks.. And also need to check difference between null and 0.  I am not sure why we originally didn't assume countedNumberOfPacks === snapshotNumberOfPacks when stocktakeBatch is created.

src/pages/StocktakeEditPage.js
```diff
@@ -83,24 +83,23 @@ export class StocktakeEditPage extends React.Component {
   }
 
   renderCell(key, item) {
+    const isEditable = !this.props.stocktake.isFinalised;
     switch (key) {
       default:
         return item[key];
-      case 'countedTotalQuantity':
+      case 'countedTotalQuantity': {
+        const emptyCellContents = isEditable ? '' : tableStrings.no_change;
```
Chris-Petty: If the stocktake is finalised and there was no change, shouldn't this item be pruned anyway? Or does this cover there being no net change, but a redistribution amongst batches?

src/pages/StocktakeEditPage.js
```diff
@@ -83,24 +83,23 @@ export class StocktakeEditPage extends React.Component {
   }
 
   renderCell(key, item) {
+    const isEditable = !this.props.stocktake.isFinalised;
```
Chris-Petty: I feel like this variable could be better as isFinalised (just through object destructuring), kind of an extra layer of abstraction but not in a useful way to me. But I assume that you did isEditable as that made more sense to you so I'm not fussed.

src/sync/incomingSyncUtils.js
```diff
@@ -329,7 +329,7 @@ export function createOrUpdateRecord(database, settings, recordType, record) {
         itemBatch: itemBatch,
         snapshotNumberOfPacks: parseNumber(record.snapshot_qty) * packSize,
         packSize: 1, // Pack to one all mobile data
-        expiry: parseDate(record.expiry),
+        expiryDate: parseDate(record.expiry),
```
Chris-Petty: Was this a bug? OutcomingSyncUtils has it correct 🤔

src/pages/expansions/StocktakeEditExpansion.js
```diff
@@ -0,0 +1,207 @@
...
+  renderFooter() {
+    return null;
+  }
+
+  renderPageInfo() {
+    const infoColumns = [
+      [
+        {
+          title: 'By Batch:',
+          info: this.item.item.name,
```
Chris-Petty: I think `this.item` would be better named as `this.stocktakeItem` in order to avoid this quirk.

Chris-Petty: Is there a limitation stopping props.data just being props.stocktakeItem and using that directly? Or is a benefit to binding it to `this` in the constructor? I'd think that either case is pointing at a object in a live Realm.Results object right?

src/pages/expansions/StocktakeEditExpansion.js
```diff
@@ -0,0 +1,207 @@
...
+import { dataTableStyles, expansionPageStyles } from '../../globalStyles';
+import { PageButton, ExpiryTextInput, PageInfo } from '../../widgets';
+import { GenericPage } from '../GenericPage';
+import { parsePositiveInteger } from '../../utilities';
+import { tableStrings, buttonStrings } from '../../localization';
+
+/**
+* Renders page to be displayed in StocktakeEditPage -> expansion.
+* @prop   {Realm}               database      App wide database.
+* @state  {Realm.Results}       items         the stocktakeItems of props.stocktake.
```
Chris-Petty: This isn't actually in state. Shouldn't it be just one stocktakeItem? Capitalise 'the'.

Would be good to explain props.data and props.refreshParent.

src/pages/expansions/StocktakeEditExpansion.js
```diff
@@ -0,0 +1,207 @@
...
+    this.props.database.write(() => {
+      switch (key) {
+        case 'countedTotalQuantity': {
+          const newCountedQuantity = parsePositiveInteger(newValue);
+          if (newCountedQuantity === null) return;
+
+          stocktakeBatch.countedTotalQuantity = parsePositiveInteger(newValue);
+          // Should refresh StocktakeEditPage, because no changes are made to
+          // StocktakeItem object, but countedTotalQuantity getter result will be different
+          this.props.refreshParent();
```
Chris-Petty: Wouldn't a "dataTypesLinked" on the parent genericPage make this redundant?

src/pages/GenericPage.js
```diff
@@ -37,12 +37,12 @@ export class GenericPage extends React.Component {
...
+                                              : this.props.dataTypesLinked;
```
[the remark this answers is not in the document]
andreievg: This is what my prettier does to this function:
from:
```js
const dataTypesArray = causedBy === 'sync' ? this.props.dataTypesSynchronised
                                               : this.props.dataTypesLinked;
    if ((dataTypesArray && dataTypesArray.indexOf(recordType) >= 0) ||
        (recordType === this.props.finalisableDataType && record.isFinalised)) {
      this.props.refreshData();
    }
```
to:
```js
    const dataTypesArray = causedBy === 'sync'
      ? this.props.dataTypesSynchronised
      : this.props.dataTypesLinked;
    if (
      (dataTypesArray && dataTypesArray.indexOf(recordType) >= 0) ||
      (recordType === this.props.finalisableDataType && record.isFinalised)
    ) {
      this.props.refreshData();
    }
```

I am not at all convinced that it's more readable or prettier..

src/pages/StocktakeEditPage.js
andreievg: I think it's easier to understand `isEditable` rather then deduce what is : `isFinalised` and how it applies to the state of cell render.

src/pages/expansions/StocktakeEditExpansion.js
andreievg: yes indeed! good point.

src/pages/expansions/StocktakeEditExpansion.js
andreievg: no benefit you are right.

src/sync/incomingSyncUtils.js
andreievg: yeah it was a bug

src/pages/StocktakeEditPage.js
andreievg: https://github.com/sussol/mobile/pull/542/commits/598c8a26a5c648bc090e0d78445ce77175efffbc 
Also should only display No change for StocktakeBatches not for StocktakeItem (as it would be pruned now)

src/localization/tableStrings.js
```diff
@@ -33,7 +33,7 @@ export const tableStrings = new LocalizedStrings({
     monthly_usage: 'MONTHLY\nUSE',
     name: 'NAME',
     no_batch_name: 'no batch name',
```
Chris-Petty: 'No batch name'?

src/pages/GenericPage.js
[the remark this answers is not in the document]
Chris-Petty: I think that's more readable and far prettier just from consistent indentation (though I think the original copy pasted bad). Obviously this is pretty subjective/opinionated, but that's a big advantage of prettier; it has the final say much like eslint rules.

src/pages/StocktakeEditPage.js
Chris-Petty: Fair enough.

src/database/DataTypes/Stocktake.js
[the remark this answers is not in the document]
Chris-Petty: What historic data?
I thought finalising a stocktake already pruned stocktakeItems with no change? 
I thought stocktake wouldn't let you finalise without making any edits?

To those last 2 I just looked. The old finalise method did delete stocktakeItems with no count. The finalise check didn't check if there were any edits.

I think that it should delete only new/'fresh' stocktakeBatches. Then should check if the remaining batches have a `countedNumberOfPacks` set, implying it's been counted. If none of them are set, then the stocktakeItem wasn't counted what so ever so should be removed from stocktake. I believe the confirmation of exisiting batch counts is what you mean by historic data?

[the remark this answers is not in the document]
Chris-Petty: If what you mean by no stock adjustments you mean say, a stocktake with only 5 items that were all counted but there were no differences, that is still a valid stocktake. Just the ideal result (which rarely happens). Given my above way of pruning they shouldn't be removed.

src/database/DataTypes/Stocktake.js
```diff
@@ -76,88 +68,80 @@ export class Stocktake extends Realm.Object {
...
+   * Finalises this stocktake, creating transactions to apply the stock changes to inventory
+   * @param {Realm}         database App wide local database
+   * @param {Realm.Object}  user     The current user logged in
+   */
+  finalise(database, user) {
     // Set the stocktake finalise details
     this.finalisedBy = user;
-    this.stocktakeDate = date;
+    this.stocktakeDate = new Date();
+    // Ajust stocktake inventory
```
Chris-Petty: Adjust

src/database/DataTypes/StocktakeBatch.js
```diff
@@ -18,10 +27,40 @@ export class StocktakeBatch extends Realm.Object {
...
+    return this.countedTotalQuantity - this.snapshotTotalQuantity;
+  }
+
   set countedTotalQuantity(quantity) {
     this.countedNumberOfPacks = this.packSize ? quantity / this.packSize : 0;
   }
 
+  /**
+   * Finalising StocktakeBatch will adjust inventory appropriately and will add
+   * new TransactionBatch in reducing or increasing Transactino for this Stocktake
```
Chris-Petty: Transactino -> Transaction

src/database/DataTypes/StocktakeBatch.js
```diff
@@ -1,12 +1,21 @@
...
+    }
+  }
+
   get snapshotTotalQuantity() {
     return this.snapshotNumberOfPacks * this.packSize;
   }
 
   get countedTotalQuantity() {
-    return this.countedNumberOfPacks ? this.countedNumberOfPacks * this.packSize : null;
+    return this.countedNumberOfPacks * this.packSize;
```
Chris-Petty: What if countedNumberOfPacks is null (being that it is optional in schema)?

src/database/DataTypes/StocktakeItem.js
```diff
@@ -26,6 +34,16 @@ export class StocktakeItem extends Realm.Object {
...
   }
 
+  /**
+   * Check to see if any StocktakeBatches have been adjusted for this StocktakeItem
+   * Will return true if any StocktakeBatches were changed, even if net quantity change
+   * for StocktakeItem is 0.
+   * @return  {boolean} True if StocktakeBatches have adjustments
+   */
+  get hasBatchWithQuantityChange() {
+    return this.batches.some(stocktakeBatch => stocktakeBatch.difference !== 0);
```
Chris-Petty: I've alluded to it elsewhere, but would it be better to check for `!stocktakeBatch.countedNumberOfPacks`. Check that for null means that where you use this getter, it'll correctly show a diff value of `0` where batches were counted but with no net change.

src/database/utilities/createRecord.js
```diff
@@ -191,6 +191,8 @@ function createStocktakeBatch(database, stocktakeItem, itemBatch) {
     stocktake: stocktakeItem.stocktake,
     itemBatch: itemBatch,
     snapshotNumberOfPacks: numberOfPacks,
+    // Assume counted quantity === snapshot quantity, when StocktakeBatch is created
+    countedNumberOfPacks: numberOfPacks,
```
Chris-Petty: Argh this assumption breaks my previous comment's line of thinking :/

Chris-Petty: They really haven't "counted" anything at this point, so it's a bit of a misappropriation to automatically set it it `numberOfPacks`. That leaves you with no point of which to distinguish what has and hasn't been counted by the stocktaker.

Chris-Petty: I suppose you would need to use the snapshot quantity for the totalQuantity of the stocktakeItem if the numberOfPacks for a batch was null, but that isn't particularly hard.

src/database/DataTypes/StocktakeBatch.js
andreievg: For all newly created StocktakeBatches and StocktakeBatches in finalised Stocktake, countedNumberOfPacks will not be null. And migration deals with StocktakeBatches that are not finalised.  This leave one more scenario, if there is an unfinalised Stocktake, and instead of migrating, the app is uninstalled and then new version is installed (will add a case for this in Incoming sync and will make StocktakeBatch.numberOfPacks required with default value = 0)

src/database/DataTypes/StocktakeItem.js
andreievg: This adds a bit more complexity.. in calculating difference and in calculating countedTotalQuantity.
Take this example: Two StocktakeBatches for one StocktakeItem
* No adjustment in StocktakeBatches, so StocktakeItem.difference === 0 and StocktakeItem.countedTotalQuantity === null (that's easy)
* Adjustment to one StocktakeBatch... StocktakeItem.countedTotalQuantity === (countedTotalQuantity of the changed StocktakeBatch and snapshotQuantity for the second StocktakeBatch, even though it's countedNumberOfPacks is null)

It's much easier to assume that countedNumberOfPacks === snapshotNumberOfPacks when a StocktakeBatch is created.

src/database/utilities/createRecord.js
andreievg: Pre this PR the placeholder for actual quantity is 'No change', this implies there is no change to quantity for a given item, so for the user it does not suggest 'Not counted'.

I see your point thought.. how does a user distinguish between counted item and an item that is counted but the quantity has not changed..

src/database/DataTypes/StocktakeBatch.js
Chris-Petty: I think you're changing away from this as discussed but if not:

Wouldn't it be more consistent to assume StocktakeBatch.countedNumberOfPacks is StocktakeBatch.snapshotNumberOfPacks, keeping the assumption made in `createRecord`?

src/database/utilities/createRecord.js
andreievg: As per our conversation, will change `No Change` to `Not Counted` and keep exisiting (pre this PR) logic

andreievg:
so the change back to null countedNumberOfPacks wasn't as complex as I thought.

When ready to merge please let me know I will do a final full test

src/pages/StocktakeEditPage.js
```diff
@@ -90,8 +90,8 @@ export class StocktakeEditPage extends React.Component {
       case 'countedTotalQuantity':
         return {
           type: isEditable ? 'editable' : 'text',
-          cellContents: item.hasBatchWithQuantityChange ? item.countedTotalQuantity : '',
-          placeholder: tableStrings.no_change,
+          cellContents: item.hasCountedBatches ? item.countedTotalQuantity : '',
```
Chris-Petty: Minor suggestion: stocktakeItem rather than item, just to be consistent with the other methods.

andreievg: good point

andreievg:
Did the final test, as per the first comment in this pr.  All tests OK.

Also:
_Have not tested unfinalised Stocktake created in previous version, then clearing data, installing new version and resyncing (this will cause all StocktakeBatches in unfinalised stocktake to have totalCountedQuantity = 0).. Nothing can be done apart from server side migration, which is excessive in this case because ideally stocktake should be created/adjusted/finalised in one sitting._

Does not apply anymore as countedNumberOfPacks remains null by default

PR #554, opened 2017-10-31
#540 Parse as int, use set method, right align value

### LINKED ISSUE #540: Bugsnag: TransactionBatch.numberOfPacks must be of type: number
## Error in mSupply Mobile

**Error** in **index.android.bundle:733**
TransactionBatch.numberOfPacks must be of type: number

[View on Bugsnag](https://app.bugsnag.com/sustainable-solutions-nz-ltd/msupply-mobile/errors/59dd87b392bd570017522854?event_id=59dd87b392bd570017522853&i=gh&m=ci)

## Stacktrace

    index.android.bundle:733 - 

[View full stacktrace](https://app.bugsnag.com/sustainable-solutions-nz-ltd/msupply-mobile/errors/59dd87b392bd570017522854?event_id=59dd87b392bd570017522853&i=gh&m=ci)

Chris-Petty:
~~Was meant to fix issue #540, but it doesn't really make sense. Doesn't seem to consistently throw the bug anymore, even if I switch back to master branch...~~

Fixes #540, I was testing with spacebar rather than backspace. Has to actually be an empty string for the bug to occur.

Changes are consistent with implementation elsewhere in app.

### DIFF
--- a/src/pages/SupplierInvoicePage.js
+++ b/src/pages/SupplierInvoicePage.js
@@ -8,7 +8,7 @@ import PropTypes from 'prop-types';
 import autobind from 'react-autobind';
 import {
   formatDate,
-  parsePositiveFloat,
+  parsePositiveInteger,
   sortDataBy,
 } from '../utilities';
 import { createRecord } from '../database';
@@ -88,7 +88,7 @@ export class SupplierInvoicePage extends React.Component {
     database.write(() => {
       switch (key) {
         case 'numberOfPacks':
-          transactionBatch.numberOfPacks = parsePositiveFloat(newValue);
+          transactionBatch.setTotalQuantity(database, parsePositiveInteger(newValue));
           break;
         case 'expiryDate':
           transactionBatch.expiryDate = newValue;
@@ -338,7 +338,7 @@ export class SupplierInvoicePage extends React.Component {
             key: 'numberOfPacks',
             width: 1,
             title: tableStrings.quantity,
-            alignText: 'center',
+            alignText: 'right',
           },
           {
             key: 'expiryDate',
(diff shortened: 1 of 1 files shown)


src/pages/SupplierInvoicePage.js
```diff
@@ -88,7 +88,7 @@ export class SupplierInvoicePage extends React.Component {
     database.write(() => {
       switch (key) {
         case 'numberOfPacks':
-          transactionBatch.numberOfPacks = parsePositiveFloat(newValue);
+          transactionBatch.setTotalQuantity(database, parsePositiveInteger(newValue));
```
andreievg: If a the pack size field is re-introduced, this will be a gotcha (because of `his.numberOfPacks = this.packSize ? quantity / this.packSize : 0;` in `setTotalQuantity`. Also if there is a value in this field and invalid value is entered, the field will be set to 0... 
Would rather a simple 
```
const numberOfPacks = parsePositiveInteger(newValue);
if (numberOfPacks !== null) transactionBatch.numberOfPacks = numerOfPacks;
```

Chris-Petty: parsePositiveInteger will return 0 for all invalid inputs as of the next release of that module, so  that check will be redundant :). I can see that packsize may well sneak it's way back as a interface field, so sure.

andreievg: why would parsePositiveInteger return 0 ? This will override previous value of all inputs if invalid value is entered

andreievg: As per discussion in sussol-utils PR, sounds like parsePositiveInteger will be returning 0 even if value entered in invalid.. I am fine with this for now.

But can you make a comment in onEndEditing re the first part of my comment (i.e. if the pack size is re introduced we cannot use setTotalQuantity

Chris-Petty: Just realised to change the switch to 'quantity' rather than 'numberOfPacks'. The column is called Quantity, so I think this is clearer as it is now. If that the column actually represents number of packs, the header should say that too to be really clear in my opinion. Adding comment as suggested too.

Issue #516, opened 2017-09-07
Sync modal and hardware back button closes app


Chris-Petty:
Build Number: 2.0.0

Description:
If you press the hardware back button with the sync modal up, it'll close the app if you are on the main menu. I would expect it to take me back from the sync modal back to the main menu.

Reproducible: Yes

Reproduction Steps:
1. From main menu open sync dashboard/modal
2. Press hardware back button

1. From supplier invoice page open sync dashboard/modal
2. Press Hardware back button twice

Comments:
This is because it's a sync *modal*, simply overlaying the current navigation card. The back button affects navigation hidden under the sync modal. If you are deeper in navigation, say supplier invoices, the behaviour is the same but you have to press the hardware back button more times. Suppler invoices -> Menu page -> exit app

I like that it is a modal and the current aesthetic, but the straight forward way to work with hardware back button it'd have to change implementation to actually being a page of sorts. Benefit of the modal is that it is very light weight performance wise and you don't lose your place in a table.

Chris-Petty:
Duplicate of #355

Issue #355, opened 2016-09-22
Hardware back button should dismiss modals


edmofro:
Build Number: 23

Description: If there is a modal open on screen, pressing the back button on the device will navigate the app in the background, rather than closing the modal as expected

Reproducible: Every time

Reproduction Steps: 
1. Open a modal by tapping 'Finalise' on any record
2. Tap the hardware back button
3. Look closely at the app behind the semi-transparent modal background 

Comments:

Issue #346, opened 2016-09-14
Pending sync record statistic would be great 


ujwalSussol:
Build Number: 20

Description: Like mSupply we need a message that says  XX sync record to send ......   this will make the user aware.

Reproducible:

Reproduction Steps:

Comments:

ujwalSussol:
@Chris-Petty   While testing mSupply mobile, I still think we need a text on the main navigator stressing the number of data waiting to be synced....    On our desktop app the information on the navigator actually helps the Nigerian user to switch on their internet....  Some places use the internet in a very stingy manner.

Chris-Petty:
This should be really easy to do at the moment, just adding it to the manual sync modal. Would that be enough or would you push for being in the top navigation bar? Picture of sync modal for context:
![image](https://user-images.githubusercontent.com/7684221/30527909-ede750fa-9c81-11e7-9bab-e3280b96248a.png)

We could add some indicator of number of sync records here.

Chris-Petty:
@ujwalSussol Would the below be sufficient?

![image](https://user-images.githubusercontent.com/7684221/32259859-73e92ec6-bf28-11e7-8794-a3212fc41182.png)

PR #557, opened 2017-11-01
#355 modal fixes

Chris-Petty:
Fixes hardware back button navigating/closing app rather than dismissing modal #355 #516 
Shows waiting sync count #346 

image

### DIFF
--- a/src/sync/SyncModal.js
+++ b/src/sync/SyncModal.js
@@ -16,33 +16,34 @@ import {
     PROGRESS_LOADING,
 } from './constants';
 
-const getStatusMessage = (progress, total, isSyncing, errorMessage, progressMessage) => {
-  let message = '';
+export function SyncModal({ database, isOpen, onClose, onPressManualSync, state }) {
+  const getStatusMessage = (progress, total, isSyncing, errorMessage, progressMessage) => {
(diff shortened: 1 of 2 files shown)


src/sync/SyncModal.js
```diff
@@ -16,33 +16,34 @@ import {
...
-    message = 'Loading change count...';
-  } else {
-    message = progressMessage ? `${progressMessage}\n` : '';
-    message += `${progress} of ${formatPlural('@count record', '@count records', total)} updated`;
-  }
+    if (errorMessage !== '') {
+      message = errorMessage;
+    } else if (!isSyncing) {
+      const recordsToSyncCount = database.objects('SyncOut').length;
+      message = recordsToSyncCount > 0 ? `${recordsToSyncCount} Records Waiting` : 'Sync Complete';
```
Chris-Petty: This is main change, most of this diff is from moving the function inside the SyncModal function to get scope of the database prop (added in this commit)

Chris-Petty: @edmofro might like to do something similar in meditrak. So you are tagged.

andreievg:
Nice changes

Issue #579, opened 2017-12-11
'Not counted' remains in stocktake column when same amount is entered


michaelnunan:
In the latest version, when you're doing a stocktake and you enter an actual amount that is the same as the snapshot amount, the cell automatically reverts to say 'Not Counted'. 
Whilst I realise why it's doing this (because it will make no change, so it's literally not doing a stock adjustment with it), it's misleading, particularly when multiple people are working on a stocktake and ESPECIALLY when you're jumping around items (not ideal practice but happens a lot).
It would be better to keep the entered amount appearing in the cell or revert it to 'No Change' rather than jumping back to 'Not Counted'. In our most recent stocktake, people repeatedly got confused, returned to lines and counted them again.

andreievg:
Fixed in https://github.com/sussol/mobile/pull/580

This should have been the default behaviour

Chris-Petty:
Duplicate of #575

Issue #573, opened 2017-12-04
Current stock shows no number for 0 stock


Chris-Petty:
Build Number: 2.1.0

Description: 
Current stock shows no number for 0 stock

Reproducible: Yes

Reproduction Steps: Look at current stock
![image](https://user-images.githubusercontent.com/7684221/33578251-d30f22ba-d9aa-11e7-93e5-c6cf84a738fd.png)

Comments: Something passive aggressive about testing

andreievg:
Fixed in https://github.com/sussol/mobile/pull/580

PR #580, opened 2017-12-11
#579-and-#573

andreievg:
Fixes https://github.com/sussol/mobile/issues/579
Fixes https://github.com/sussol/mobile/issues/573

### DIFF
--- a/src/database/DataTypes/StocktakeItem.js
+++ b/src/database/DataTypes/StocktakeItem.js
@@ -109,40 +109,40 @@ export class StocktakeItem extends Realm.Object {
           stocktakeBatch.countedTotalQuantity = stocktakeBatch.snapshotTotalQuantity;
           database.save('StocktakeBatch', stocktakeBatch);
         });
-      } else {
-        // Actual Quantity is not snaphot quantity
-        const isIncreasingQuantity = difference > 0;
-        const sortedBatches = this.batches.sorted('expiryDate', isIncreasingQuantity);
+        return;
+      }
+      // Actual Quantity is not snaphot quantity
+      const isIncreasingQuantity = difference > 0;
+      const sortedBatches = this.batches.sorted('expiryDate', isIncreasingQuantity);
 
-        sortedBatches.some(stocktakeBatch => {
-          const batchTotalQuantity = stocktakeBatch.countedTotalQuantity;
+      sortedBatches.some(stocktakeBatch => {
+        const batchTotalQuantity = stocktakeBatch.countedTotalQuantity;
 
-          let thisBatchChangeQuantity = 0;
-          if (isIncreasingQuantity) {
-            thisBatchChangeQuantity = Math.min(stocktakeBatch.snapshotTotalQuantity -
-                                              batchTotalQuantity, difference);
+        let thisBatchChangeQuantity = 0;
+        if (isIncreasingQuantity) {
+          thisBatchChangeQuantity = Math.min(stocktakeBatch.snapshotTotalQuantity -
+                                            batchTotalQuantity, difference);
 
-            if (thisBatchChangeQuantity <= 0) return false;
-          } else {
-            thisBatchChangeQuantity = Math.min(batchTotalQuantity, -difference);
-            thisBatchChangeQuantity = -thisBatchChangeQuantity;
-          }
-
-          stocktakeBatch.countedTotalQuantity = batchTotalQuantity + thisBatchChangeQuantity;
-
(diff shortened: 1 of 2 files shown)


andreievg:
git dif should have been much smaller.. but had to add and else after if (rather then return) .. and everything was indented

src/pages/StockPage.js
```diff
@@ -63,6 +63,15 @@ export class StockPage extends React.Component {
     });
   }
 
+  renderCell = (key, item) => {
+    switch (key) {
+      default:
+        return item[key];
+      case 'totalQuantity':
+        return item[key] || 0;
```
Chris-Petty: This will fix it but why doesn't this bug affect other pages such as `CustomerInvoicePage`?

src/database/DataTypes/StocktakeItem.js
```diff
@@ -95,45 +95,54 @@ export class StocktakeItem extends Realm.Object {
...
-          if (thisBatchChangeQuantity <= 0) return false;
-        } else {
-          thisBatchChangeQuantity = Math.min(batchTotalQuantity, -difference);
-          thisBatchChangeQuantity = -thisBatchChangeQuantity;
+      let difference = quantity - this.countedTotalQuantity;
+      // In case number is entered in Actual Quantity field, but it's the same as
+      // snapshot quantity, we want to remove 'Not Counted' placeholder
+      if (difference === 0) {
+        this.batches.forEach(stocktakeBatch => {
+          stocktakeBatch.countedTotalQuantity = stocktakeBatch.snapshotTotalQuantity;
```
Chris-Petty: Would have thought it'd be cleaner as 
`stocktakeBatch.countedTotalQuantity = quantity;`
But it should literally be the same result and same performance as far as I know, so not really fussed. Your way kind of plays into what the getter for `difference` looks like on stocktakeBatch it could be argued.

src/database/DataTypes/StocktakeItem.js
```diff
@@ -95,45 +95,54 @@ export class StocktakeItem extends Realm.Object {
...
+          database.save('StocktakeBatch', stocktakeBatch);
+        });
+      } else {
```
Chris-Petty: If you want a smaller diff, you could just have an early `return` in the _if_ block and scrap the _else_ block.

Chris-Petty: Oh that might make the database.write throw a wobbly...

src/database/DataTypes/StocktakeItem.js
andreievg: I think it's also meant to set countedTotalQuantity `for each stocktakeBatch` to snapshotTotalQuantity. i.e. if there are two ItemBatches for an item, batchA.totalQuantity = 5, batchB.totalQuantity = 10, when a stocktake is created, two stocktakeBatch will be made for that item with countedTotalQuantity = null and StocktakeBatchA.snapshotTotalQuantity = batchA.totalQuantity and StocktakeBatchB.snapshotTotalQuantity = batchB.totalQuantity. So if the quantity does not change but a number in the actual quantity field, we need to set countedTotalQuantity = snapshotTotalQuantity for each batch (which maybe different from actual quantity).

src/database/DataTypes/StocktakeItem.js
andreievg: I thought it might, but seems fine, less dif now

src/pages/StockPage.js
andreievg: great point, i went to investigate, it's because of this: https://github.com/sussol/react-native-generic-table-page/blob/4803b8295357be98a627ea8f0ccab4c14ab4542f/index.js#L237
when no renderCell method is found, this is displayed: `record[key] || ''`, if record is 0 then display will be ''.

I've changed to just use `renderCell = (key, item) => item[key];`

src/database/DataTypes/StocktakeItem.js
Chris-Petty: Yup looking again that makes sense.

Chris-Petty: Although does this break stocktaking at the batch level at all? A quick test would show.

andreievg: I've tested, seems all good

Chris-Petty: Ages to pick this up, but this is the cause of #626

Emphasises the need for:
a. Better branch management for features
b. Comprehensive testing before releasing to users
c. Automated testing

Issue #584, opened 2017-12-18
1 day bug


andreievg:
Dates are pushed from mobile in 'iso' format https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString i.e. -> 2017-12-05T14:48:00.000Z

When synced to desktop, depending on the current time, will result in + 1 day being added.

Dr Thet in MAM noticed this with expiry dates, but it applies throughout all dates in Mobile.

An easy fix is to remove the time part of the date string.

Replication:

Try setting emulator time to be past 12pm, create and invoice, sync and check on Desktop (i am not sure when exactly + 1 day gets added only tasted it past 12pm which adds a day, or earlier then 5am which keeps the current date).

PR #585, opened 2017-12-18
#584 fix one day bug

andreievg:
Fixes https://github.com/sussol/mobile/issues/584


GaryWilletts:
Wouldn't some comments be helpful in the code to say why you do this? Institutional memory chaps, institutional memory ;-)

Issue #69, opened 2016-07-11
Search filter on Customer: Displays the Number keypad


bijaySussol:
Build used: mSupply_mobile_installer_1.apk (Today)

Should displays the Number pad with Alphabets !

PR #602, opened 2018-02-06
#69-Allow-Finalising-of-zero-difference-stockake

andreievg:
Fixes https://github.com/sussol/mobile/issues/601

### DIFF
--- a/src/database/DataTypes/StocktakeItem.js
+++ b/src/database/DataTypes/StocktakeItem.js
@@ -40,10 +40,6 @@ export class StocktakeItem extends Realm.Object {
    * for StocktakeItem is 0.
    * @return  {boolean} True if StocktakeBatches have adjustments
    */
-  get hasBatchWithQuantityChange() {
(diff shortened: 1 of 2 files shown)


src/database/DataTypes/StocktakeItem.js
```diff
@@ -41,7 +41,7 @@ export class StocktakeItem extends Realm.Object {
    * @return  {boolean} True if StocktakeBatches have adjustments
    */
   get hasBatchWithQuantityChange() {
-    return this.batches.some(stocktakeBatch => stocktakeBatch.difference !== 0);
+    return this.batches.some(stocktakeBatch => stocktakeBatch.hasBeenCounted);
```
Chris-Petty: Would be cleaner to remove this getter all together and replace references with `hasCountedBatches` below seeing as they are now functionally the same.

andreievg: @Chris-Petty -> nice spotting, adjusted.

PR #609, opened 2018-02-12
Better sync errors

Chris-Petty:
Fix `itemBatch` being null in outgoingSyncUtils error. Bit of an edge case from merging records on mSupply versions earlier than ~3.83.

Makes errors trying to parse outgoing sync records easier to diagnose. Decided that it's best not to let sync keep going - we don't know what future bugs might be.

Kinder message to users that tells them to contact support.
image

Also sends more detailed errors to bugsnag.
screen shot 2018-02-12 at 4 02 02 pm

### DIFF
--- a/src/sync/Synchroniser.js
+++ b/src/sync/Synchroniser.js
@@ -193,7 +193,7 @@ export class Synchroniser {
       } catch (error) {
         // If error not safe to continue sync throw it again to pass it up to next handler
         // See outgoingSyncUtils.js
-        if (!error.isSafe) throw error;
+        if (!error.canDeleteSyncOut) throw error;
       }
     };
   };
--- a/src/sync/Synchroniser.js
+++ b/src/sync/Synchroniser.js
@@ -50,13 +50,13 @@ export class Synchroniser {
   /**
    * Redux progress setting functions
    */
-  setTotal = (totalCount) => this.dispatch(setSyncTotal(totalCount));
-  incrementProgress = (increment) => this.dispatch(incrementSyncProgress(increment));
-  setProgress = (currentCount) => this.dispatch(setSyncProgress(currentCount));
-  setProgressMessage = (message) => this.dispatch(setSyncProgressMessage(message));
-  setError = (errorMessage) => this.dispatch(setSyncError(errorMessage));
-  setIsSyncing = (isSyncing) => this.dispatch(setSyncIsSyncing(isSyncing));
-  setCompletionTime = (time) => this.dispatch(setSyncCompletionTime(time));
+  setTotal = totalCount => this.dispatch(setSyncTotal(totalCount));
+  incrementProgress = increment => this.dispatch(incrementSyncProgress(increment));
+  setProgress = currentCount => this.dispatch(setSyncProgress(currentCount));
+  setProgressMessage = message => this.dispatch(setSyncProgressMessage(message));
+  setError = errorMessage => this.dispatch(setSyncError(errorMessage));
+  setIsSyncing = isSyncing => this.dispatch(setSyncIsSyncing(isSyncing));
+  setCompletionTime = time => this.dispatch(setSyncCompletionTime(time));
 
   /**
    * Wipe the current database, check that the given url can be synced against
@@ -99,7 +99,7 @@ export class Synchroniser {
             headers: {
               Authorization: this.authenticator.getAuthHeader(),
             },
-          }
+          },
         );
         // If the initial_dump has been successful, serverURL is valid, and should now have all sync
         // records queued and ready to send. Safe to store as this.serverURL
@@ -115,7 +115,7 @@ export class Synchroniser {
     this.settings.set(SYNC_IS_INITIALISED, 'true');
     this.syncQueue.enable(); // Begin the sync queue listening to database changes
     this.setIsSyncing(false);
-  }
+  };
 
   /**
    * Return whether the synchroniser has been initialised.
@@ -124,7 +124,7 @@ export class Synchroniser {
   isInitialised = () => {
     const syncIsInitialised = this.settings.get(SYNC_IS_INITIALISED);
     return syncIsInitialised && syncIsInitialised === 'true';
-  }
+  };
 
   /**
    * Return whether or not the last sync of the app failed
@@ -133,7 +133,7 @@ export class Synchroniser {
   lastSyncFailed = () => {
     const lastSyncFailed = this.settings.get(SYNC_PRIOR_FAILED);
     return lastSyncFailed && lastSyncFailed === 'true';
-  }
+  };
 
   /**
    * Carry out a synchronization, first pushing any local changes, then pulling
@@ -163,7 +163,7 @@ export class Synchroniser {
       this.setError(error.message);
       this.setIsSyncing(false);
     }
-  }
+  };
 
   /**
    * Push batches of changes to the local database up to the remote server, until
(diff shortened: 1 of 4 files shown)


Chris-Petty:
@andreievg See new approach. It needs testing!

src/sync/outgoingSyncUtils.js
```diff
@@ -51,16 +53,37 @@ export function generateSyncJson(database, settings, syncOutRecord) {
...
+    try {
+      syncData = generateSyncData(settings, recordType, record);
+    } catch (error) {
+      // There was an error with data, often a null object
+      const siteName = settings.get(SYNC_SITE_NAME);
+      const syncUrl = settings.get(SYNC_URL);
+      const originalMessage = error.message;
+      // Change error message to be helpful in bugsnag
+      error.message =
+        `SYNC OUT ERROR. siteName: ${siteName}, serverUrl: ${syncUrl}, ` +
```
andreievg: should this be a json, rather then a string ?

src/sync/outgoingSyncUtils.js
```diff
@@ -51,16 +53,37 @@ export function generateSyncJson(database, settings, syncOutRecord) {
...
+        `syncOutRecord.id: ${syncOutRecord.id}, storeId: ${storeId} changeType: ${changeType}, ` +
+        `recordType: ${recordType}, recordId: ${recordId}, message: ${originalMessage}`;
+      // Ping the error off to bugsnag
+      bugsnagClient.notify(error);
+
+      // Make a nicer message for users and throw it again.
+      error.message =
```
andreievg: From what I understand bugsnag 'syncs' errors, does .notify store it by reference or does a deep copy ? Since the message is changed, would bugsnag's error.message be modified ?

src/sync/Synchroniser.js
```diff
@@ -173,26 +173,37 @@ export class Synchroniser {
...
+
+    const translateRecord = (record) => {
+      try {
+        const translatedRecord = generateSyncJson(this.database, this.settings, record);
+        // If an error was thrown no record is pushed
+        translatedRecords.push(translatedRecord);
+      } catch (error) {
+        // If error not safe to continue sync throw it again to pass it up to next handler
+        // See outgoingSyncUtils.js
+        if (!error.isSafe) throw error;
```
andreievg: If error is thrown here sync will stop and this.syncQyeue.use will not be called, causing BugSnag notifications every 10mins ?

src/sync/outgoingSyncUtils.js
```diff
@@ -233,3 +259,15 @@ function getTimeString(date) {
...
 }
+
+// Tries to get a value that is known to potentially lead to crash
+// If path on the record returns null, throw an error with prototype
+// extended with 'isSafe' set to true so sync knows to continue
+function safeGet(record, path) {
+  const value = _.get(record, path);
+  if (!value) {
+    const error = Error(`Cannot access record value on path: '${path}'`);
+    error.isSafe = true;
```
andreievg: Can we rename isSafe ? To maybe canDeleteSyncOut, or something that is more explanatory?

src/sync/outgoingSyncUtils.js
```diff
@@ -233,3 +259,15 @@ function getTimeString(date) {
   if (!date || typeof date !== 'object') return '00:00:00';
   return date.toTimeString().substring(0, 8);
 }
+
+// Tries to get a value that is known to potentially lead to crash
+// If path on the record returns null, throw an error with prototype
+// extended with 'isSafe' set to true so sync knows to continue
+function safeGet(record, path) {
+  const value = _.get(record, path);
```
andreievg: I am not sure why we want to use underscore library.  It would be quite easy to implement a safeGet(record, ...paths).

Chris-Petty: Axed

src/sync/outgoingSyncUtils.js
Chris-Petty: Yes

src/sync/Synchroniser.js
Chris-Petty: Hmmm

src/sync/outgoingSyncUtils.js
Chris-Petty: That is definitely a possibility. I'll just make copies of the error object.

src/sync/outgoingSyncUtils.js
Chris-Petty: Worked correctly so far. Error message is meant to be a string from how I interpret mdn docs.

andreievg: I was thinking from neatness point of view. I guess it's neat enough as it is (interesting to see string concatenation + template literal, I guess eslint allows multi line concatenation).

Chris-Petty: It's mostly just against concatenating string literals where a template string would be better.

src/sync/outgoingSyncUtils.js
```diff
@@ -51,16 +52,37 @@ export function generateSyncJson(database, settings, syncOutRecord) {
...
+      message =
+        `SYNC OUT ERROR. siteName: ${siteName}, serverUrl: ${syncUrl}, ` +
+        `syncOutRecord.id: ${syncOutRecord.id}, storeId: ${storeId} changeType: ${changeType}, ` +
+        `recordType: ${recordType}, recordId: ${recordId}, message: ${originalMessage}`;
+      // Ping the error off to bugsnag
+      bugsnagClient.notify(Object.assign({}, error, message));
+
+      // Make a nicer message for users and throw it again.
+      message = `There was an error syncing. Contact mSupply mobile support. ${originalMessage}`;
+      throw Object.assign({}, error, message); // doesn't like object literal syntax
```
andreievg: I know i've used Object.assign() quite a bit in other projects, but have been changing it to {...one, ...two} every time I see it now.  

Also I think you are throwing a json here not an Error, and passing json to bugsnag.notify, i think it should still work fine, except no stack trace

src/sync/Synchroniser.js
Chris-Petty: It probably will, but bugsnag should just treat it as 1 _error_ with many _events_:

image

src/sync/outgoingSyncUtils.js
```diff
@@ -233,3 +258,21 @@ function getTimeString(date) {
...
   return date.toTimeString().substring(0, 8);
 }
+
+// Tries to get a value that is known to potentially lead to crash
+// If path on the record returns null, throw an error with prototype
+// extended with 'canDeleteSyncOut' set to true so sync knows to continue
+function safeGet(record, path) {
+  const pathSegments = path.split('.');
+  let nestedProp = record;
+  for (let i = 0; i < pathSegments.length; i++) {
```
andreievg: i was thinking more function safeGet(record, ...paths) can then call it safeGet (record, itemBatch, id) rather then safeGet(record, 'itemBatch.id').
Can then iterate over paths, and if error can paths.join('.') to pass more info to bug snag (by the way that's missing now)

Chris-Petty: I like the single string as that path arg as it more closely resembles what you'd be doing if you were just doing dot notation on an object.

Will make the error better as you suggest though.

src/sync/outgoingSyncUtils.js
Chris-Petty: 1. Eslint was whining about throw + object literal syntax. My comment could explain that better. The object destructuring was my first instinct too. Might revert back and just ignore line it (unless the eslint rules expla

2. Yea I need to test this. Errors are objects, I was hoping that making a copy would include the propertyholding the stack trace but that's a presumption. Hence needing testing! https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/EvalError

andreievg: Object assign or destructuring won't work on Error object.
screen shot 2018-02-14 at 11 37 54 am

Chris-Petty: Alright I've been doing testing. Clone an Error object in anyway and it loses it's special sauce and bugsnag throws a wobbly about it.

Looking at the their implementation [(yay for opensource)](https://github.com/bugsnag/bugsnag-react-native/blob/master/src/Bugsnag.js#L311-L321) it appears it makes it's own internal report object rather than keeping a reference the error.

I'm just going to revert to editing the error message... or use their notify API properly.

src/sync/Synchroniser.js
andreievg: I don't think we should be sending the same error, every 10 minutes to bug snag. Maybe sync out Q should have a 'broken' boolean, then instead of deleting we can flag all records that cause an issue with 'broken'.  In this case we don't need safeGet method at all

Chris-Petty:
Should do a little more testing to make sure records that are skipped are correctly removed from sync queue. Should be easy with realm explorer.

Issue #412, opened 2017-04-07
Sorting current stock can make app unresponsive


andrew-d-uk:
Build Number: 30

Description: Title doesn't really explain it - but if you have a long list of stock in 'Current Stock' and you sort it, it may take a few seconds to sort (which is fine). But if your user is impatient and thinks it hasn't worked they may click it multiple times. And mSupply will sort it multiple times before allowing the user to make any other commands such as moving back a page. (And your user may keep clicking it as it sorts, and you end up waiting for it to sort the list 100 times!)

Reproducible: Yes

Reproduction Steps: Go to a long stock list, click to sort by quantity and keep clicking.

Comments:
Not a major issue as we can tell people not to do it, and even if they do they can terminate the app and restart (not sure whether this may have a negative impact on any processed data in the app though?)

Chris-Petty:
Thanks Andrew,

You are right, if someone gets carried away and spams the sorting (as many would given delay) it'll make jam up the app. There are a few approaches and optimisations that could fix this.

- Debounce the sort function so it doesn't stack up calls when spammed
- Cheeky hack/optimisation, if the sort direction (isAscending) is just being switched on the same column, the array of records being displayed should just be reversed rather than bothering with other sorting/filtering going on.

Chris-Petty:
I think I erroneously tagged this as fixed in a commit

joshxg:
I think this is related to the data table performance issues which has enough issues currently and has a rewrite coming in hot. Closing this issue

PR #627, opened 2018-03-13
#617 #626 fix stocktake actual quantity

### LINKED ISSUE #617: Stocktake crash when sorting by Counted column
Property 'countedTotalQuantity' does not exist on object type 'StocktakeItem' (on bugsnag)

### LINKED ISSUE #626: Stocktake value gets set to snapshot/lost
Build Number: 2.1.0-rc8

Description:
If a existing `Actual Quantity` value is tapped and highlighted, it'll be set to the snapshot quantity if you press next without changing the value to something different

Reproducible: yes

Reproduction Steps:
1. Have a stock take with a few items
2. Enter some values
3. If you select a value you have entered and press next, it'll be lost/set to what ever the snapshot value is.
4. You must enter a different value if you don't want it be set to snapshot quantity.

Comments:
We should test better. This was introduced in Dec and we only now picked it up. https://github.com/sussol/mobile/pull/580/files

Chris-Petty:
fixes #617 
fixes #626 

Fix not #412, tried to fix it using vscode but just totally messed up my commits as below...

So:

- Sorting by actual quantity no longer crashes
- Editing a actual quantity and setting it to previous actual quantity value doesn't set the countedTotalQuantity to snapshotQuantity
- Finalised stocktakes that have 0/no batches for countedTotalQuantity show 0 rather than blank.

### DIFF
--- a/src/database/DataTypes/StocktakeItem.js
+++ b/src/database/DataTypes/StocktakeItem.js
@@ -98,6 +98,16 @@ export class StocktakeItem extends Realm.Object {
       }
 
       let difference = quantity - this.countedTotalQuantity;
+      // In case number is entered in Actual Quantity field, but it's the same as
+      // snapshot quantity, we want to remove 'Not Counted' placeholder
+      if (quantity === this.snapshotTotalQuantity) {
+        this.batches.forEach(stocktakeBatch => {
+          stocktakeBatch.countedTotalQuantity = stocktakeBatch.snapshotTotalQuantity;
+          database.save('StocktakeBatch', stocktakeBatch);
+        });
+        return;
+      }
+
       const isIncreasingQuantity = difference > 0;
       const sortedBatches = this.batches.sorted('expiryDate', isIncreasingQuantity);
 
--- a/src/database/DataTypes/StocktakeItem.js
+++ b/src/database/DataTypes/StocktakeItem.js
@@ -98,16 +98,6 @@ export class StocktakeItem extends Realm.Object {
       }
 
       let difference = quantity - this.countedTotalQuantity;
-      // In case number is entered in Actual Quantity field, but it's the same as
-      // snapshot quantity, we want to remove 'Not Counted' placeholder
-      if (difference === 0) {
-        this.batches.forEach(stocktakeBatch => {
-          stocktakeBatch.countedTotalQuantity = stocktakeBatch.snapshotTotalQuantity;
-          database.save('StocktakeBatch', stocktakeBatch);
-        });
-        return;
-      }
-      // Actual Quantity is not snaphot quantity
       const isIncreasingQuantity = difference > 0;
       const sortedBatches = this.batches.sorted('expiryDate', isIncreasingQuantity);
 
--- a/src/pages/StocktakeEditPage.js
+++ b/src/pages/StocktakeEditPage.js
@@ -101,6 +101,7 @@ export class StocktakeEditPage extends React.Component {
         sortDataType = 'string';
         break;
       case 'snapshotTotalQuantity':
+      case 'countedTotalQuantity':
       case 'difference':
         sortDataType = 'number';
         break;
--- a/src/pages/StocktakeEditPage.js
+++ b/src/pages/StocktakeEditPage.js
@@ -116,12 +116,22 @@ export class StocktakeEditPage extends React.Component {
     switch (key) {
       default:
         return stocktakeItem[key];
-      case 'countedTotalQuantity':
+      case 'countedTotalQuantity': {
+        let cellContents;
+        if (!stocktakeItem.hasCountedBatches) {
+          // If not editable (i.e. finalised) we want to show 0 rather than '' as
+          // placeholder isn't shown
+          cellContents = isEditable ? '' : 0;
+        } else {
+          cellContents = stocktakeItem.countedTotalQuantity;
+        }
+        // If finalised we want to show 0 rather than '' as placeholder isn't shown
         return {
           type: isEditable ? 'editable' : 'text',
-          cellContents: stocktakeItem.hasCountedBatches ? stocktakeItem.countedTotalQuantity : '',
+          cellContents: cellContents,
           placeholder: tableStrings.not_counted,
         };
+      }
       case 'difference': {
         const difference = stocktakeItem.difference;
         const prefix = difference > 0 ? '+' : '';
@@ -136,7 +146,7 @@ export class StocktakeEditPage extends React.Component {
       onPress={() => this.props.navigateTo('stocktakeManager',
         navStrings.manage_stocktake,
         { stocktake: this.props.stocktake },
-        )}
+      )}
       isDisabled={this.props.stocktake.isFinalised}
     />
   );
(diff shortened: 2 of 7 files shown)


src/database/DataTypes/StocktakeItem.js
```diff
@@ -98,16 +98,6 @@ export class StocktakeItem extends Realm.Object {
...
-      // In case number is entered in Actual Quantity field, but it's the same as
-      // snapshot quantity, we want to remove 'Not Counted' placeholder
-      if (difference === 0) {
-        this.batches.forEach(stocktakeBatch => {
-          stocktakeBatch.countedTotalQuantity = stocktakeBatch.snapshotTotalQuantity;
-          database.save('StocktakeBatch', stocktakeBatch);
-        });
-        return;
-      }
-      // Actual Quantity is not snaphot quantity
```
Chris-Petty: I've tested this, the app correctly shows "Not Counted" when no value was set, not if the value is the same as snapshot.

If snapshot was 30, you counted 25. Then selected that field and entered 25 again, it'd see difference 0 and set it to 30.

src/pages/StocktakeEditPage.js
```diff
@@ -101,6 +101,7 @@ export class StocktakeEditPage extends React.Component {
         sortDataType = 'string';
         break;
       case 'snapshotTotalQuantity':
+      case 'countedTotalQuantity':
```
Chris-Petty: Fixes crash on sorting. Realm can't sort by getters.

src/pages/StocktakeEditPage.js
```diff
@@ -115,12 +116,22 @@ export class StocktakeEditPage extends React.Component {
...
+      case 'countedTotalQuantity': {
+        let cellContents;
+        if (!stocktakeItem.hasCountedBatches) {
+          // If not editable (i.e. finalised) we want to show 0 rather than '' as
+          // placeholder isn't shown
+          cellContents = isEditable ? '' : 0;
+        } else {
+          cellContents = stocktakeItem.countedTotalQuantity;
+        }
+        // If finalised we want to show 0 rather than '' as placeholder isn't shown
```
Chris-Petty: Fixes blank "ACTUAL QUANTITY" being blank when finalised line has no batches.

Chris-Petty:
image

No content if you select the field and enter nothing and change focus (press next or tap nav bar)

src/pages/StocktakeEditPage.js
```diff
@@ -115,12 +116,22 @@ export class StocktakeEditPage extends React.Component {
...
         return {
           type: isEditable ? 'editable' : 'text',
-          cellContents: stocktakeItem.hasCountedBatches ? stocktakeItem.countedTotalQuantity : '',
+          cellContents: cellContents,
```
andreievg: Could just be cellContents,

src/pages/StocktakeEditPage.js
```diff
@@ -115,12 +116,22 @@ export class StocktakeEditPage extends React.Component {
     switch (key) {
       default:
         return stocktakeItem[key];
-      case 'countedTotalQuantity':
+      case 'countedTotalQuantity': {
+        let cellContents;
+        if (!stocktakeItem.hasCountedBatches) {
+          // If not editable (i.e. finalised) we want to show 0 rather than '' as
+          // placeholder isn't shown
+          cellContents = isEditable ? '' : 0;
```
andreievg: I am pretty sure if stocktakeItem does not have counted batches it gets removed during finalisation.

src/database/DataTypes/StocktakeItem.js
andreievg: If there is more then one batch, and actual quantity is set to snapshot quantity we get this:
screen shot 2018-03-19 at 8 37 18 pm

src/pages/StocktakeEditPage.js
Chris-Petty: Nah the item remains annoyingly. If you expand it there are simply no rows. Maybe that's the bug I should be fixing!

andreievg: I've tried numerous times (on master), If item has no counted quantity it get's removed from finalised stock take.  Can you please give a use case of where it remains please (on master).

src/database/DataTypes/StocktakeItem.js
```diff
@@ -98,16 +98,6 @@ export class StocktakeItem extends Realm.Object {
       }
 
       let difference = quantity - this.countedTotalQuantity;
-      // In case number is entered in Actual Quantity field, but it's the same as
-      // snapshot quantity, we want to remove 'Not Counted' placeholder
-      if (difference === 0) {
-        this.batches.forEach(stocktakeBatch => {
-          stocktakeBatch.countedTotalQuantity = stocktakeBatch.snapshotTotalQuantity;
-          database.save('StocktakeBatch', stocktakeBatch);
```
andreievg: I've made a comment which you might have missed:

With this change, if there is more then one batch, and actual quantity is set to snapshot quantity we get this:
screen shot 2018-03-19 at 8 37 18 pm

src/database/DataTypes/StocktakeItem.js
Chris-Petty: Ah I see! Will fix.

src/database/DataTypes/StocktakeItem.js
Chris-Petty: Had missed, thanks!

src/pages/StocktakeEditPage.js
Chris-Petty: So here are some items getting stocktake.
image
Result:
image

So if the snapshot was 0, then it'll be included with no stocktakeItemBatches. If snapshot was more than 0 it works fine

andreievg: So true. Nicely spotted

src/database/DataTypes/StocktakeItem.js
andreievg: (changing back to countedTotalQuantity) @Chris-Petty, that was totally my bad. 
`if (difference === 0)` should ready:  `if (quantity === this.snapshotTotalQuantity)`

andreievg:
@Chris-Petty, happy to merge if you reinstate:
```
if (quantity === this.snapshotTotalQuantity) {
        this.batches.forEach(stocktakeBatch => {
          stocktakeBatch.countedTotalQuantity = stocktakeBatch.snapshotTotalQuantity;
          database.save('StocktakeBatch', stocktakeBatch);
        });
        return;
      }
```
but with `if (quantity === this.snapshotTotalQuantity)`

src/pages/StocktakeEditPage.js
andreievg: I don't think it's a bug though, i have a distant recollection we wanted to keep them ?

Chris-Petty: @andreievg I just spotted while sifting through code

(stocktake.js:adjustInventory)
```
// Delete all 'fresh' StocktakeBatches that have been created by stocktake
// but have not been changed
database.delete('StocktakeBatch', stocktakeBatches.filter(stocktakeBatch =>
  stocktakeBatch.snapshotTotalQuantity === 0 && stocktakeBatch.difference === 0));
```
So yea, it does check but doesn't touch the stocktakeItems, just clears out the batches.

Chris-Petty:
I've put it back in @andreievg and seems to work as intended.

But I found a different unrelated problem hehe

Same as snapshot works correctly:
image

Reduced works correctly:
image

Increased works correctly:
image

But if you recount or correct a mistake, reducing the increased value
image

I think this anomaly is caused by the fact none of the 3 batches have a date. Giving them dates it works correctly.
image

Chris-Petty:
I'm happy to make an issue for that so that this PR can move on in life.
#639

andreievg:
@Chris-Petty thank you for all the testing!

Issue #629, opened 2018-03-14
Finalise stocktake making invalid stock adjustments (into negative on ledger)


Chris-Petty:
Build Number: 2.1.0-rc8

Description: 
Through a fun sequence of editing you can make stocktakes that show a negative ledger total for an item but show's "fine" on mobile. Both are wrong. Fortunately it's not totally easy to achieve.

Reproducible: Yes

Reproduction Steps:
1. Make a new stocktake with an item
2. Make sure it has some quantity for snapshot (say 100)
3. Make a customer invoice for the same item and issue some amount (say 100)
4. Go back to the stocktake and set the actual quantity to 0. You cannot finalise because it'd make negative stock.
5. Set actual quantity >= 100
6. Edit actual quantity to 0 AND hit finalise before anything else
7. Confirm and viola, you have messed the ledger.

In mSupply mobile, it'll say there is 0 stock. In mSupply Desktop for that store it'll say 0 stock too, but by the ledger it should have -100 stock.

(I've done it a couple times here, so -250)
![image](https://user-images.githubusercontent.com/7684221/37379532-bb65fdf0-2798-11e8-88ac-3ae3b1cfeac4.png)

Comments: 🙈

anildahalsussol:
Yah I also faced this problem in medical action myanmar with negative stock which is creating ledger problem. Sorry i didnot made the issue but i informed to andrei about it.

andreievg:
@Chris-Petty 
I think the really issue is in react-native-data table and what happens when cell loses focus.  We can also finalise CustomInvoice/Stocktake with 0 quantities, i.e. set quantity to something other then zero, then back to zero then immediately finalise... 

I thought this PR fixed it https://github.com/sussol/react-native-data-table/pull/9/files
I've tried to explain here: https://github.com/sussol/mobile/issues/427
Obviously still an issue. 

I like your fix here: https://github.com/sussol/mobile/pull/634 , I think it can be extended, after checking 'checkForFinaliseError' and before record.finalise.. call checkForFinalisationError again.

andreievg:
[...]

PR #634, opened 2018-03-15
#629 Throw errors on erroneous finalise

Chris-Petty:
Sort of a fix for #629 

Stops it finalising but still a problem with race condition. Potential for a big refactor to fix it properly.

### DIFF
--- a/src/widgets/modals/FinaliseModal.js
+++ b/src/widgets/modals/FinaliseModal.js
@@ -5,9 +5,12 @@
 
 import React from 'react';
 import PropTypes from 'prop-types';
+import { Client as BugsnagClient } from 'bugsnag-react-native';
 import { ConfirmModal } from './ConfirmModal';
 import globalStyles, { DARK_GREY } from '../../globalStyles';
 import { modalStrings } from '../../localization';
+
+const bugsnagClient = new BugsnagClient();
 
 /**
  * Presents a modal allowing the user to confirm or cancel finalising a record.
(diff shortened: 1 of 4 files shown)


Chris-Petty:
Might be potential here for a better approach https://realm.io/docs/javascript/2.2.0/api/Realm.html#isInTransaction

I've updated the realm version for giggles and didn't try the above function.

src/widgets/modals/FinaliseModal.js
```diff
@@ -30,6 +33,27 @@ export function FinaliseModal(props) {
...
   const errorText = !record.isFinalised && checkForError && checkForError(record);
+
+  // Wrapped in try-catch block so that finalise methods in schema can throw an error
+  // as last line of defence
+  const tryFinalise = () => {
+    props.runWithLoadingIndicator(() => {
+      try {
+        if (record) {
+          props.database.write(() => {
+            record.finalise(props.database, props.user);
```
andreievg: I think just before calling record.finalise() we should check this.isFinalise and again check checkForError(record). This way we can remove all isFinalised checks in stocktake, transact etc.. and remove the extra this.itemsBelowMinimum.length > 0 check that's you've added in this PR.  I've added some extra comments to https://github.com/sussol/mobile/issues/629, should describe why.

Chris-Petty: I like this idea but I also like keeping the checks in the finalise methods so that juuuust in case someone sometime adds a call to it outside of the finalise button, it's still safe. It's negligible overhead wise checking it twice. Though redundant. But redundancy is kind of the point I'm going for.

andreievg: Ok sure, then we should do the full pre finalise check in record.finalise right?

Chris-Petty:
@andreievg Given this the new issue you made #638, I'm happy to close #629

PR #635, opened 2018-03-16
#xxxx item cross reference

### LINKED ISSUE #622: Mobile does not handle cross reference items
Build Number: 2.1.0-rc8

Description: 
mSupply desktop has support for cross reference items that when selected actually select a different preferred item (say a brand name goes to a generic name e.g. voltaren -> diclofenac ). mSupply server syncs these items out like any other item and mobile handles them like any other item!

Reproducible: Yes

Reproduction Steps:
1. Make a cross reference item in mSupply (that is visible to a mobile store)
2. Open mobile on device/emulator and see that the cross reference item is visible
3. Stock take that item and observe that it has a total quantity separate to the reference item
4. (bonus) sync the tablet and see if server barfs trying to receive item lines for a cross reference item.

Comments: 
Bug and feature! mSupply mobile should handle like mSupply desktop. When adding items to things in mobile, the cross reference item should show the stock of the referenced item and should add the reference item if selected.
Stocktake should not show the cross reference items.

Chris-Petty:
fixes #622 

If existing sites have already used cross reference items and made item batches for them... it's a bit late. The most solid resolution I have is to merge the cross reference items with the items they reference, then recreate the cross reference item (as it definitely should have been the one deleted in the merge). This should sync out to mobile safely. @GaryWilletts @ujwalSussol @craigdrown @bijaySussol @andreievg @adrianwb any other ideas or caveats?

Hey you can actually tag groups @sussol/msupplyadmin 

Alternative is a more manual version of the same thing, making inventory adjustments for the real items adding the stock and deleting the stock on the cross reference items.

# Testing
Cross reference item to be referred to as CRI below.
- [ ] Syncs CRIs safely and sets crossReferenceItem (can check in realm explorer)
- [ ] Adding an CRI in customer invoice works adds the referenced item
- [ ] Doing it twice does nothing! (no duplicates)
- [ ] CRIs are ignored adding master list items to customer invoice (no duplicates, no CRI line)
- [ ] Adding CRI in Supplier Requisition adds the referenced item
- [ ] Doing it twice does nothing! (no duplicates)
- [ ] CRIs are ignored adding master list items to supplier invoice (no duplicates, no CRI line)
- [ ] CRIs are hidden in current stock table
- [ ] CRIs are hidden in manage stock take table

### DIFF
--- a/src/database/DataTypes/Item.js
+++ b/src/database/DataTypes/Item.js
@@ -9,7 +9,8 @@ export class Item extends Realm.Object {
   }
 
   get totalQuantity() {
-    return getTotal(this.batches, 'totalQuantity');
+    const item = this.crossReferenceItem ? this.crossReferenceItem : this;
+    return getTotal(item.batches, 'totalQuantity');
   }
 
   get dailyUsage() {
@@ -124,5 +125,6 @@ Item.schema = {
     category: { type: 'ItemCategory', optional: true },
     defaultPrice: { type: 'double', optional: true },
     isVisible: { type: 'bool', default: false },
+    crossReferenceItem: { type: 'Item', optional: true },
   },
 };
--- a/src/database/DataTypes/Item.js
+++ b/src/database/DataTypes/Item.js
@@ -8,9 +8,17 @@ export class Item extends Realm.Object {
     database.delete('ItemBatch', this.batches);
   }
 
+  /**
+   * A useful getter to ensure that if the item is a cross reference item,
+   * we use the item it references (the real item) rather than the cross
+   * reference item
+   */
+  get realItem() {
+    return this.crossReferenceItem ? this.crossReferenceItem : this;
+  }
+
   get totalQuantity() {
-    const item = this.crossReferenceItem ? this.crossReferenceItem : this;
-    return getTotal(item.batches, 'totalQuantity');
+    return getTotal(this.realItem.batches, 'totalQuantity');
   }
 
   get dailyUsage() {
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   },
   "name": "mSupplyMobile",
   "//": "version must be in the format ${majorNumber}.${minorNumber}.${patchNumber}-rc${releaseCandidateNumber}",
-  "version": "2.1.0-rc7",
+  "version": "2.1.0-rc10",
   "private": false,
   "license": "MIT",
   "description": "Mobile app for use with the mSupply medical inventory control software",
--- a/src/database/DataTypes/Requisition.js
+++ b/src/database/DataTypes/Requisition.js
@@ -52,7 +52,7 @@ export class Requisition extends Realm.Object {
   }
 
   hasItem(item) {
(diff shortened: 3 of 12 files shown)


ujwalSussol:
Chris sounds like a good plan when we come across such a problem.

adrianwb:
It sounds like a reasonable plan to merge the cross-reference items out - but presumably they can be added in again after the mobile fix to handle them properly?  From the description of the problem, it sounds like it might even be possible for the total stock issued for an item and its cross-reference equivalent(s) to be more than the actual stock??

It should be possible to run a query on the primary to determine how many cross-reference items are actually used in existing mobile stores. At the very least that should reduce the scope of the manual patch-up, but it might even be possible to use that information to run some code to automatically apply a fix (but note that we can't sync item_lines on the server back to mobile in case they have changed in the meantime).

Chris-Petty:
@adrianwb A manual stocktake for the affected items would be sufficient to correct the stock values for the "real" items. But existing item lines on the cross reference items would still be sitting around and on the desktop side they behave a little weird: 
- In item list on desktop when logged into the mobile store, it shows the current stock based off them rather than the referred item) 
image

- Not sure what affects on reporting or consumption calculations this might have

So a manual stocktake + a footrunner that just deletes item lines on cross reference items should tie up any loose ends.

Do many places use the cross reference items _and_ mobile? Cooks do, Timor? Sols? Kiribati?

src/database/DataTypes/Item.js
```diff
@@ -9,7 +9,8 @@ export class Item extends Realm.Object {
   }
 
   get totalQuantity() {
-    return getTotal(this.batches, 'totalQuantity');
+    const item = this.crossReferenceItem ? this.crossReferenceItem : this;
```
andreievg: we do this check quite often, i think we should make and use a getter item.realItem => this.crossReferenceItem || this.

Chris-Petty:
@andreievg as mobile didn't store `crossReferenceItem` before, there is no way to migration until after said items have been re-synced. So we can't do migration unless we do it in the schema for when it receives the crossReferenceItem on sync.

@adrianwb I was just thinking - It's still safer to do the merge/recreate of the cross reference items. That way it's accounted for if they were ever used in customer invoices/requisitions. This miiiight be better yet if mobile handled merging proper.

andreievg:
@Chris-Petty, ahh that's true, can't do migration for those items at all.. And during desktop sync process, sync is fully disabled (so merging won't sync to mobile).. Maybe we need upgrade/re-sync those tablets after the merge/recreation on desktop.

adrianwb:
@Chris-Petty if you think it's better to do merge & recreate, that's fine with me

Chris-Petty:
@andreievg I made changes you requested. I've tested it pretty extensively I think, but it's always worth another person having a pass 😛

Cook islands are itching for this!

Chris-Petty:
Cool. Can sit for a bit, I might try to smash out mobile merge as that'll allow handling this more cleanly.

Chris-Petty:
@andreievg Not gonna bother with mobile merge. Users will just have to stocktake to correct.

PR #643, opened 2018-03-28
#531 better sync string

### LINKED ISSUE #531: Sync progress message starts as 'All records updated'
Even when it is not true, while sync is waiting for the first response from the server, the message in the sync dashboard is a very confusing 'All records updated'. This is possibly only true when there are no records to push...not sure.

Chris-Petty:
fixes #531 
screen shot 2018-03-28 at 14 28 12


README.md
```diff
@@ -31,8 +31,16 @@ As an open source software project, we welcome contributions from external devel
...
 6. Code
 7. Make a pull request ensuring you have
+  * Made it to the approach branch
   * Adhered to our conventions/style
   * Respected module boundaries
   * Tested that your additions work and haven't caused regressions
 8. We'll review the PR and merge it if it adds value for a general use case (not just one user)
 9. Sit back and enjoy the warm glow :-)
+
+#### Branch Structure
```
Chris-Petty: This came from master. Herp.

src/sync/SyncModal.js
```diff
@@ -26,7 +26,7 @@ export function SyncModal({ database, isOpen, onClose, onPressManualSync, state
       const recordsToSyncCount = database.objects('SyncOut').length;
       message = recordsToSyncCount > 0 ? `${recordsToSyncCount} Records Waiting` : 'Sync Complete';
     } else if (progress >= total) {
-      message = 'All records updated.';
+      message = 'Checking server for records...';
```
Chris-Petty: This is the only proper change.

andreievg:
@Chris-Petty, I am not sure where I've mentioned this, but should we localise these strings ?

PR #653, opened 2018-04-12
#638 fix stocktake ledger issue

### LINKED ISSUE #638: Stocktake, adjusts based on snapshot quantity difference not on real quantity difference
Stock-takes by their nature need to be done synchronously to other actions like receiving/issuing stock. This is explained in training but is not restricted in the app. Sound like we need to handle the cases where users issues/receives stock while stocktake is in progress, to minimise a chance of ledger issues.
Currently when stocktake is finalise we create an inventory adjustment based on difference between snapshot quantity, not actual quantity. This can result in ledger discrepancy. 
i.e.

Item A -> Starting Quantity: 1000
* Stocktake is created with snapshot quantity = 1000
* Then a customer invoice is created, issuing 100 of stock (Item A Quantity = 900, Transaction Line A = -100)
* User goes back to stocktake and enters Actual quantity = 900
* Stocktake is finalised (Item A Quantity = 900, and stock take adjustment, based on Snapshot Quantity = 1000 - Real Quantity = 900, Transaction Line B = -100)

So our actual quantity = 900 but our ledger reads:
Starting Quantity = 1000
Transaction Line A -100
Transaction Line B -100
Expected Quantity = 800

Even though Quantity is correct, ledger is not.

This is how this example looks in mSupply Desktop Item->Stock->Ledger:

screen shot 2018-03-26 at 4 42 40 pm

Prior to https://github.com/sussol/mobile/pull/634 with some trickery it was possible to finalise an a stocktake that show a negative ledger.

Solution:

* As per mSupply desktop, when creating adjustment, make sure to use real quantity not snapshot quantity
* When opening a stocktake, update the snapshot quantity = real quantity. (makes more sense from data integrity point of view, i.e. adjustment = difference between snapshot quantity and counted/actual quantity).
* Disable issuing and receiving items that are in un-finalised stocktake (a bit of a mission to implement + restricts users from some tasks they may wish to do) 

@GaryWilletts @ujwalSussol @Chris-Petty @craigdrown, which one do you prefer ?

* Can we adjust possible ledger problems with migration code ?
  * Will have to implement forward/backwards ledger calculations, and adjust inventory adjustment lines accordingly.

Chris-Petty:
Not quite a fix for #638 

Stocktake correctly applies the difference of `counted quantity - snapshot quantity` through transaction mechanism, rather than separately editing the stock on hand directly to be counted quantity.

Also, reducing stock below minimum check is done on a batch basis. Previously it checked if item total quantity would be made negative, rather than if any individual batch quantity would be made negative.

### DIFF
--- a/src/database/DataTypes/StocktakeBatch.js
+++ b/src/database/DataTypes/StocktakeBatch.js
@@ -24,7 +24,7 @@ export class StocktakeBatch extends Realm.Object {
    * stock on hand for this batch would result in a negative, return true. Can occur
    * when stock has been issued in a customer invoice for this batch before the parent
    * stocktake was finalised.
-   * @return {boolean} True if (stock on hand - stocktakebatch.difference) is negative
+   * @return {boolean} True if (stock on hand + stocktakebatch.difference) is negative
    */
   get isReducedBelowMinimum() {
     const stockOnHand = this.itemBatch.totalQuantity;
--- a/src/database/DataTypes/StocktakeBatch.js
+++ b/src/database/DataTypes/StocktakeBatch.js
@@ -71,6 +71,8 @@ export class StocktakeBatch extends Realm.Object {
       transactionItem, this.itemBatch);
 
     // Apply difference from stocktake to actual stock on hand levels.
+    // Whether stock is increased or decreased is determined by the transaction
+    // so we need to use the absolute value of difference (i.e. always treat as positive)
     const snapshotDifference = Math.abs(this.snapshotTotalQuantity - this.countedTotalQuantity);
     transactionBatch.setTotalQuantity(database, snapshotDifference);
 
--- a/src/database/DataTypes/StocktakeBatch.js
+++ b/src/database/DataTypes/StocktakeBatch.js
@@ -19,6 +19,11 @@ export class StocktakeBatch extends Realm.Object {
     return this.countedNumberOfPacks * this.packSize;
   }
 
+  get isReducedBelowMinimum() {
+    const stockOnHand = this.itemBatch.totalQuantity;
+    return (stockOnHand - Math.abs(this.difference)) < 0;
+  }
+
   get hasBeenCounted() {
     return this.countedNumberOfPacks !== null;
   }
--- a/src/database/DataTypes/StocktakeBatch.js
+++ b/src/database/DataTypes/StocktakeBatch.js
@@ -19,11 +19,13 @@ export class StocktakeBatch extends Realm.Object {
     return this.countedNumberOfPacks * this.packSize;
   }
 
-    /**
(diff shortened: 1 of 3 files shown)


src/database/DataTypes/StocktakeItem.js
```diff
@@ -73,10 +59,7 @@ export class StocktakeItem extends Realm.Object {
    * @return {Boolean} Whether the counted quantity is below the minimum for this item
    */
   get isReducedBelowMinimum() {
-    const countedTotalQuantity = this.countedTotalQuantity;
-    return countedTotalQuantity !== undefined &&
-           countedTotalQuantity !== null &&
-           countedTotalQuantity < this.minimumTotalQuantity;
+    return this.batches.some(batch => batch.isReducedBelowMinimum);
```
andreievg: nice, by batch checking looks like the way to go (and it's much clearer)

src/database/DataTypes/StocktakeBatch.js
```diff
@@ -45,26 +57,33 @@ export class StocktakeBatch extends Realm.Object {
...
-    // Create TransactionItem, TransactionBatch to store inventory adjustment in this Stocktake
-    const item = this.itemBatch.item;
-    const transactionItem = createRecord(database, 'TransactionItem', inventoryAdjustment, item);
-    const transactionBatch = createRecord(database, 'TransactionBatch',
-                                          transactionItem, this.itemBatch);
-    transactionBatch.numberOfPacks = Math.abs(this.snapshotTotalQuantity
-                                              - this.countedTotalQuantity);
 
+    // Make inventory adjustments if there is a difference to apply
+    if (this.difference !== 0) {
```
Chris-Petty: Check for this now, that way if user just updates batch details it's still included in the finalisation (these remain as lines 63-64 above).

src/database/DataTypes/StocktakeBatch.js
```diff
@@ -45,26 +57,33 @@ export class StocktakeBatch extends Realm.Object {
...
+
+      // Apply difference from stocktake to actual stock on hand levels.
+      // Whether stock is increased or decreased is determined by the transaction
+      // so we need to use the absolute value of difference (i.e. always treat as positive)
+      const snapshotDifference = Math.abs(this.snapshotTotalQuantity - this.countedTotalQuantity);
+      transactionBatch.setTotalQuantity(database, snapshotDifference);
+      database.save('TransactionBatch', transactionBatch);
+    }
     database.save('ItemBatch', this.itemBatch);
-    database.save('TransactionBatch', transactionBatch);
```
Chris-Petty: Transaction batch only happens in the scope of the above if statement, so moved this a few lines up.

Chris-Petty:
@andreievg Should be fixed!

PR #663, opened 2018-05-01
#xxxx fix supplier invoice nav bugs and onePressButton

Chris-Petty:


Chris-Petty:
Fixes some issues that exist?

PR #665, opened 2018-05-02
Update data migration

andreievg:
I’ve checked, migration code works, rc1 to rc10

### DIFF
--- a/src/utilities/compareVersions.js
+++ b/src/utilities/compareVersions.js
@@ -1,0 +1,69 @@
+import { Client as BugsnagClient } from 'bugsnag-react-native';
+const bugsnagClient = new BugsnagClient();
+
+export function compareVersions(versionOne, versionTwo) {
+  try {
+    const result = versionToInteger(versionOne) - versionToInteger(versionTwo);
+    if (result > 0) return 1;
+    if (result < 0) return -1;
+    return 0;
+  } catch (error) {
+      // Errors thrown in migration method in dataMigration.js do not trickle down to bugsnag
+      // so will need manually notify bug snag.
+    error.message = `error verifying versions, versionOne: ${versionOne}, versionTwo ${versionTwo}`;
+    bugsnagClient.notify(error);
+  }
+  return 0;
+}
+// versionToInteger logic is the same as ./android/app/build.gradle
+function versionToInteger(version) {
+  const [majorMinorPatch, provisional] = version.split('-');
+  const [major, minor, patch] = majorMinorPatch.split('.');
+  return Number(major) * 10000000 +
+         Number(minor) * 100000 +
+         Number(patch) * 100 +
+         Number(!provisional ? '99' : provisional.match(/[0-9]+/)[0]);
+         // Regex /[0-9]+/ matches only numbers and at least on digit number
+}
+
+export function test() {
+  const testHelper = (versionOne, versionTwo, expectedResult) => {
+    const testResult = compareVersions(versionOne, versionTwo);
+    const testMessage = testResult === expectedResult ? 'pass' : 'FAIL';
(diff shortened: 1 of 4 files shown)


src/utilities/compareVersions.js
```diff
@@ -0,0 +1,69 @@
...
+    error.message = `error verifying versions, versionOne: ${versionOne}, versionTwo ${versionTwo}`;
+    bugsnagClient.notify(error);
+  }
+  return 0;
+}
+// versionToInteger logic is the same as ./android/app/build.gradle
+function versionToInteger(version) {
+  const [majorMinorPatch, provisional] = version.split('-');
+  const [major, minor, patch] = majorMinorPatch.split('.');
+  return Number(major) * 10000000 +
```
Chris-Petty: I think casting using `Number` is my favourite, but out of curiosity I went searching for why to use it over parseInt or parseFloat. Found a brilliant table on the matter!

https://stackoverflow.com/questions/17106681/parseint-vs-unary-plus-when-to-use-which/17106702#17106702

Some people debate the performance of each method, but from what I've read most discussion is naive as it's platform specific.

src/utilities/compareVersions.js
```diff
@@ -0,0 +1,69 @@
...
+         Number(minor) * 100000 +
+         Number(patch) * 100 +
+         Number(!provisional ? '99' : provisional.match(/[0-9]+/)[0]);
+         // Regex /[0-9]+/ matches only numbers and at least on digit number
```
Chris-Petty: This sentence might not be worded correctly?

src/utilities/compareVersions.js
```diff
@@ -0,0 +1,69 @@
...
+}
+
+export function test() {
```
Chris-Petty: Would be better in a Jest test but don't bother. We don't have any level of automated tests sadly.

src/utilities/compareVersions.js
```diff
@@ -0,0 +1,69 @@
...
+    const testResult = compareVersions(versionOne, versionTwo);
+    const testMessage = testResult === expectedResult ? 'pass' : 'FAIL';
+    console.log(`checking ${versionOne} vs ${versionTwo}, status: ${testMessage}`);
+    console.log(`expected result: ${expectedResult}, actual result: ${testResult}`);
+  };
+
+  testHelper('1.0.0', '1.0.1', -1);
+  testHelper('1.0.0', '1.1.0', -1);
+  testHelper('1.0.0', '2.0.0', -1);
+  testHelper('2.0.0', '2.0.0', 0);
```
Chris-Petty: Missed some example of
testHelper('2.0.0', '1.0.0', 1);

src/utilities/compareVersions.js
Chris-Petty: This is the only comment I actually want addressed haha

src/utilities/compareVersions.js
andreievg: Nice table, looking at it i just realised if someone there is no number after RC, Number('') will return zero, that should not be a problem or ?

src/utilities/compareVersions.js
andreievg: fixed

src/utilities/compareVersions.js
Chris-Petty: Why assume when you can test 😛

src/utilities/compareVersions.js
Chris-Petty: Shouldn't be `x + 0 = x` after all. More test cases! haha

src/utilities/compareVersions.js
andreievg: Sure testing is the key. To me it seemed like 1 + 2 = 2 + 1

PR #692, opened 2018-07-05
Copy realm data file from appdata into Download folder

Chris-Petty:


src/pages/MenuPage.js
```diff
@@ -33,14 +26,20 @@ export class MenuPage extends React.Component {
...
     );
   }
 
   componentWillUnmount() {
     this.props.database.removeListener(this.databaseListenerId);
   }
 
+  exportData = () => {
+    const { settings, database } = this.props;
+    const syncSiteName = settings.get(SYNC_SITE_NAME);
```
Chris-Petty: The settings could easily come from within the exportData method in UIDatabase, I should probably do that

Chris-Petty: Actually I won't accessing settings is a bit janky in UIDatabase. Ugh.

src/database/UIDatabase.js
```diff
@@ -6,6 +9,32 @@ export class UIDatabase {
...
+    const exportFolder = `${RNFS.ExternalStorageDirectoryPath}/Download/mSupplyMobile_data`;
+
+    const date = new Date();
+    const dateString = `${date.getFullYear()}-${date.getMonth()}-${date.getDay()}T${date.getHours()}-${date.getMinutes()}-${date.getSeconds()}`;
+    const copyFileName = `${filename} ${dateString}`;
+
+    // If the database is not closed, there is a small chance of corrupting the data
+    // if it's currently in a transaction
+    realm.close();
+    RNFS.mkdir(exportFolder)
```
andreievg: would making export dir fail if it already exists ?

src/database/UIDatabase.js
```diff
@@ -6,6 +9,32 @@ export class UIDatabase {
...
+      .then(() => RNFS.copyFile(realmPath, `${exportFolder}/${copyFileName}.realm`)
+        .catch(() => RNFS.copyFile(realmPath, `${exportFolder}/msupply-mobile-data.realm`))
```
andreievg: hmm, I don't see why this catch is necessary

src/database/UIDatabase.js
Chris-Petty: Nah

>Create a directory at filepath. Automatically creates parents and does not throw if already exists (works like Linux mkdir -p).

https://github.com/itinance/react-native-fs#mkdirfilepath-string-options-mkdiroptions-promisevoid

src/pages/MenuPage.js
```diff
@@ -106,14 +103,22 @@ export class MenuPage extends React.Component {
...
-            }
+            {isInAdminMode && (
+              <View>
+                <Button
+                  style={globalStyles.menuButton}
+                  textStyle={globalStyles.menuButtonText}
+                  text="Realm Explorer"
+                  onPress={() => navigateTo('realmExplorer', 'Database Contents')}
+                />
+                <Button
```
andreievg: I am not sure this will fit on some devices, I remember seeing fourth button in one column disappearing of screen

src/database/UIDatabase.js
Chris-Petty: If I recall correctly it's due to `copyFileName` being derived from store name, it can have invalid characters for a filename/path. Added a comment.

Issue #699, opened 2018-07-26
Can make customer invoice to own store


Chris-Petty:
Build Number: 2.1.0

Description: title

Reproducible: yes

Reproduction Steps:
Make the mobile store's name visible to itself on server
Make invoice to self
Show's up as supplier invoice
Infinite supply?

Comments:

PR #702, opened 2018-07-31
#xxxx sync delete

Chris-Petty:
Allow deleting a transaction through sync. This is gross, but should only occur when we manually send a delete record (as primary wouldn't normally allow deleting in this way).

Required at this point as a regression in mobile allowed sending customer invoices and requisitions to self (and corresponding supplier invoices), so to clean up we need to be able to run scripts on server to delete transactions.


Chris-Petty:
Required for some janky fixes of data due to #699

PR #704, opened 2018-08-02
#xxxx sync with while loop

Chris-Petty:
Tests with data file of 3363 records of initial sync pull time (ignores initialisation time).

Change test | run 1 | run 2 | run 3 | average | % speed increase
-- | -- | -- | -- | -- | --
Master (recursion) | 61.462 | 56.639 | 56.747 | 58.28266667 |  
While loop (less waiting record requests) | 44.465 | 47.196 | 43.848 | 45.16966667 | 22.49897053


src/sync/Synchroniser.js
```diff
@@ -238,65 +249,45 @@ export class Synchroniser {
...
+      this.incrementProgress(incomingRecords.length);
+      progress += incomingRecords.length;
 
-    // Recurse to get the next batch of records from the server
-    await this.recursivePull(serverURL, thisSiteId, serverId, newTotal);
+      // Check no new records added to incoming queue on last iteration.
+      if (progress >= total) {
+        waitingRecordCount = await this.getWaitingRecordCount();
+        if (waitingRecordCount > 0) {
+          total += waitingRecordCount;
```
andreievg: I think this should be `total = progress + waitingRecordCount`, otherwise you can get progress exceeding total by maximums of INITIAL_BATCH_SIZE

andreievg: Otherwise the PR looks very good

Chris-Petty: I'm not sure what you mean, as I understand it:
Currently:
1. 19/20
2. 39/20 (very briefly)
3. 39/60 (there was another 40 records on server)
4. 59/60
5. 60/60 (incomingRecords.length was 1)

Your suggestion:
1. 19/20
2. 39/20
3. 39/79 (there were another 40 records on server)
4. 59/79
5. 60/79 (incomingRecords.length was 1)
6. Falsely claiming an extra 19 records waiting, server never returns them, we end up in infinite loop? 

Step 2 can be improved by checking and updating total before calling incrementProgress, this way in the UI total is updated before progress.

Chris-Petty: Correction (I think):
Currently:
1. 19/20
2. 39/20 (very briefly)
3. 39/41 (there was another ~~40~~ 21 records on server, as we've retrieved 20 of the total 60 since the last query on waiting)
4. 59/41 (
5. 59/60 (incomingRecords.length was 1)

Your suggestion:
1. 19/20
2. 39/20
3. 39/60 (there were another ~~40~~ 21 records on server)
4. 59/60
5. 60/60 (incomingRecords.length was 1)
6. ~~Falsely claiming an extra 19 records waiting, server never returns them, we end up in infinite loop?~~

andreievg: Maybe should also check for incomingRecords.length = 0 and exit loop, this is before integrating records

andreievg: @Chris-Petty that requested change is not really necessary I was just being pedantic.
But I do think we need a check to make sure the loop is not infinite (if server for some reason has less records then the total)

Issue #740, opened 2018-10-29
Update dependancies


Chris-Petty:
Build Number: 2.1.1

Description: 
As you can read [here](http://facebook.github.io/react-native/blog/2018/07/04/releasing-react-native-056#modernizing-android-support) in an official RN blog from July, to publish on the android store we'll need to upgrade mSupply Mobile to run on at least RN 0.56 if we are to update the app published on the app store next month, or possibly already need to because it's currently unpublished (I think that's their terminology).

So:
- [x] Update react-native to at least 0.56 (0.57 is the latest). You'll need to follow the release notes of each release of RN to make sure any require changes (if any) are met!
- [x] Update the rest of dependancies where necessary or convenient. Some will likely have peer dependancies tied to react/RN versions. Yarn has a bunch of useful cli commands such as `yarn outdated` and `yarn upgradeInteractive`.

Comments:
- Probably be incremental with your commits, so you can peel back if certain sets of modules are incompatible.
- ListView is deprecated, but probably still available in RN 0.56. If it's fully cut from RN API we'll need to commit to rewriting react-native-data-table with FlatList, too.

sworupSussol:
https://github.com/sussol/mobile/pull/741

sworup:
I think we can close this.

PR #741, opened 2018-10-30
#9127 - Upgrade React Native to v0.57.4

sworupSussol:
Resolves issue #740 

**Things Done**

- First upgrade React native to v0.56.1 and then to v0.57.4
- Upgrade react native with `react-native-git-upgrade`
- Resolve conflict
- Force set android Build tool & Compile SDK to 26 
- Upgraded minimal number of packages 
  - "prop-types": "^15.6.2",
  - "react": "16.6.0-alpha.8af6728"
  - "react-native": "0.57.4"
  - "react-native-data-table": "^0.2.0"
  - "react-native-generic-table-page": "^0.3.0"
  - "react-native-ui-components": "^0.5.0"
  - "react-native-vector-icons": "~6.0.2"
  - "redux": "^4.0.1"
  - "redux-persist": "^5.10.0"
  - "redux-persist-error-handler": "^0.1.1"
- Upgrade Gradle version to gradle-4.10.2-bin.zip
- Upgrade Android-gradle-plugin to 3.2.0
- Replace deprecated View.propTypes.style with ViewPropTypes.style at various files
- Change redux-persist bootstrap code for newer version
- Remove trailing slash after the rest element

**Important information**

Babel no longer allows you to add trailing comma after the rest element. I have removed it from the code but we are using some SusSol packages which still has them ([react-native-data-table](https://github.com/sussol/react-native-data-table)) . 

For testing right now I removed them manually in the package but we need to update the package.
 
> "A trailing comma is not permitted after the rest element"


sworupSussol:
@Chris-Petty 
> Seems editing the node modules to remove the commas isn't working for me, any other steps?

If you compiled the code and then changed the node_module package to recompile then I was having the same error too.

**Possible Solution:** Apparently the code is cached by React Native, so you have to close the debugger and the console React Native packager so that the cached code would be released.

Chris-Petty:
I updated react-native-data-table and published to npm. Tested this a bit, seems good.

Issue #755, opened 2018-11-28
Automate debugging db from device


wlthomson:
Build Number: 2.1.1
Description: When testing database logic changes, it is useful to view the database using Realm Browser on the local machine. This could be done programmatically (e.g. hosting on the local network) or externally (e.g. bash script wrapping `adb pull`).

PR #757, opened 2018-11-28
#755 automate db debug

wlthomson:


dev_scripts/get_db.sh
```diff
@@ -15,4 +15,5 @@ if [ ! -d "$DIR_LOCAL" ]; then
     mkdir "$DIR_LOCAL"
 fi
 
+adb root > /dev/null
```
andreievg: does this work in Genymotion as well as ADM ?

PR #766, opened 2018-12-09
Add script for automating mSupply site login

wlthomson:
Debugging mobile functionality can involve repeatedly uninstalling and reinstalling the app on the testing device. This currently involves a time-consuming process of manually entering the login credentials for the test server site. This script uses `adb` commands to automatically populate these fields and login. Parameters can be set using defaults or provided as command line args.

### DIFF
--- a/dev_scripts/site_login.sh
+++ b/dev_scripts/site_login.sh
@@ -1,0 +1,32 @@
+#!/usr/bin/env bash
+
+# mSupply Mobile
+# Sustainable Solutions (NZ) Ltd. 2018.
+#
+# Automate mSupply site login.
+
+KEYEVENT_ENTER=66
+
+SERVER_URL_INPUT_X=750
+SERVER_URL_INPUT_Y=630
+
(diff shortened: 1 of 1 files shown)


Chris-Petty:
Normally I'd just edit the initial state in code to have the required details (and again in login page), seems a bit easier than this. Does this have other benefits?

wlthomson:
The main benefit is that you are entering details in the same way as the user, so you can be sure to catch any errors that you might not if you take shortcuts to manually change the state. Also, I've been working across a few branches and I find this easier than manually editing the state for each branch. Plus, having a script will prevent people accidentally committing hardcoded state changes to the repo (and then reverting them and messing up the git log).

andreievg:
@wlthomson 
The script is very smart, and I can see this functionality being useful for test interface. I do agree somewhat, that the same can be achieved with code.

I think we need an issue label 'things to make dev live easier'. And have make some issues, one being being hot reload, another one will be the objective of this script. Then we can all input in the design of 'how to solve this'.

For now, not sure if we want to merge this one. Happy to do so if @Chris-Petty approves

Issue #762, opened 2018-12-04
(S)2 tablets 1 store STOP it happening


Chris-Petty:
**Problem:** Can currently make 2 tablets sync to the same store. This causes sync issues, breaking the whole system as it's not design for it.

**Expected:** Cannot sync two tablets to same store, hardware (or something) is tracked in some manner and blocks it happening. As this will likely need server side work, could be applied to desktop sync too.

**Comments:**

ujwalSussol:
We have a mantis issue that @anildahalsussol  wants to work on : https://bugs.msupply.org.nz/view.php?id=8898

andreievg:
- [ ] - checking test

joshxg:
Realised the desktop PR for this https://github.com/sussol/msupply/pull/2280 has not been merged, so obviously is not working.

joshxg:
# TESTING

- [x] Try to sync the same store to two tablets

There's an issue with the hardware ID though. Issue here: #1290

Issue #680, opened 2018-05-15
Push records needs better errors and bugsnag


Chris-Petty:
Build Number: 2.1.0

Description: Currently error messages are lacklustre on record push with sync. Below are the only messages:
```javascript
try {
  responseJson = await response.json();
} catch (error) {
  throw new Error('Unexpected response from sync server');
}
if (responseJson.error.length > 0) {
  throw new Error('Server rejected pushed records');
}
```

Reproducible: yes

Reproduction Steps:
I don't have something specific at hand. Get the server to send back an error.

Comments:
Refactor: make bugsnag/utilities.js, pull the outgoingSyncUtils.js use of bugsnag out and put it there. Then use appropriately and improve the errors in description snippet.

nmadruga:
This problem happened again on 3rd July 2018 on Taro tablet. And it ended up being just a very week or maybe unavailable mobile data signal to sync. One idea is to improve this message, if it is covering more than one type of problem. A better message for the case on Taro is: "No connection with the server" to make it clear that the problem is with the connectivity and not mSupply server.

Chris-Petty:
It's actually getting an error response from the server, but the connection is just too flaky. I think it'd be better to actually display the error, or handle the http status codes (presumably coming through) with appropriate messages.

Chris-Petty:
Cook Islands now suffering a similar pain regarding this vague error.

raviSussol:
@Chris-Petty @nmadruga Since Im working on this #775 issue (where both have similar codes) to edit - so I'll branch off from that issue and update the error message to more verbose (or handle http response code) to clarify the current unknown error.

joshxg:
Going to close this and use #1025 - the solution there has problems and might not be suitable as a final solution, but we can reopen and/or make new issues with different solutions afterwards

PR #775, opened 2018-12-24
#762 prohibit 2tablets 1store stop happening

raviSussol:
Added device unique id in the `authenticateAsync` package method

### DIFF
--- a/src/authentication/UserAuthenticator.js
+++ b/src/authentication/UserAuthenticator.js
@@ -3,6 +3,7 @@ import {
   AUTH_ERROR_CODES,
   hashPassword,
 } from 'sussol-utilities';
+import DeviceInfo from 'react-native-device-info';
 
 import { SETTINGS_KEYS } from '../settings';
 const { SYNC_URL, THIS_STORE_ID } = SETTINGS_KEYS;
@@ -14,6 +15,7 @@ const {
 
 const AUTH_ENDPOINT = '/sync/v3/user';
 const CONNECTION_TIMEOUT_PERIOD = 10 * 1000; // 10 second timeout for authenticating connection
+const uniqueId = DeviceInfo.getUniqueID();
 
 export class UserAuthenticator {
   constructor(database, settings) {
@@ -55,7 +57,7 @@ export class UserAuthenticator {
       // Race condition promise, if connection is taking to long will be rejected
       // with CONNECTION_FAILURE error
       const userJson = await Promise.race([
-        authenticateAsync(authURL, username, passwordHash),
+        authenticateAsync(authURL, username, passwordHash, uniqueId),
         createConnectionTimeoutPromise()]);
       if (!userJson || !userJson.UserID) throw new Error('Unexpected response from server');
       else { // Success, save user to database
--- a/src/sync/Synchroniser.js
+++ b/src/sync/Synchroniser.js
@@ -8,6 +8,7 @@ import { SyncDatabase } from './SyncDatabase';
 import { generateSyncJson } from './outgoingSyncUtils';
 import { integrateRecord } from './incomingSyncUtils';
 import { SETTINGS_KEYS } from '../settings';
+import DeviceInfo from 'react-native-device-info';
 import {
   incrementSyncProgress,
   setSyncProgress,
@@ -30,6 +31,7 @@ const {
 const MIN_SYNC_BATCH_SIZE = 10;
 const MAX_SYNC_BATCH_SIZE = 500;
 const OPTIMAL_BATCH_SPEED = 5;
+const uniqueId = DeviceInfo.getUniqueID();
 
 /**
  * Provides core synchronization functionality, initilising the database with an
@@ -111,9 +113,14 @@ export class Synchroniser {
           {
             headers: {
               Authorization: this.authHeader,
+              msupply_site_uuid: uniqueId,
             },
           },
-        );
+        )
+          .then(response => response.json()
+            .then(responseJson => {
+              if (responseJson.error) throw new Error(responseJson.error);
+            }));
         // If the initial_dump has been successful, serverURL is valid, and should now have all sync
         // records queued and ready to send. Safe to store as this.serverURL
         this.serverURL = serverURL;
@@ -233,6 +240,7 @@ export class Synchroniser {
         method: 'POST',
         headers: {
           Authorization: this.authHeader,
+          msupply_site_uuid: uniqueId,
         },
         body: JSON.stringify(records),
       },
@@ -244,7 +252,11 @@ export class Synchroniser {
       throw new Error('Unexpected response from sync server');
     }
     if (responseJson.error.length > 0) {
-      throw new Error('Server rejected pushed records');
+      if (responseJson.error.startsWith('Site registration doesn\'t match.')) {
+        throw new Error(responseJson.error);
+      } else {
+        throw new Error('Server rejected pushed records');
+      }
     }
   };
 
@@ -298,6 +310,7 @@ export class Synchroniser {
       {
         headers: {
           Authorization: this.authHeader,
+          msupply_site_uuid: uniqueId,
         },
       },
     );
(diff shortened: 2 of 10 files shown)


src/sync/Synchroniser.js
```diff
@@ -233,6 +240,7 @@ export class Synchroniser {
         method: 'POST',
         headers: {
           Authorization: this.authHeader,
+          msupply_site_uuid: uniqueId,
```
Chris-Petty: The closest thing to a convention for HTTP header keys would be to use hyphens rather than under scores 
```suggestion
          msupply-site-uuid: uniqueId,
```

yarn.lock
```diff
@@ -5,14 +5,12 @@
 "@babel/code-frame@^7.0.0":
   version "7.0.0"
   resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.0.0.tgz#06e2ab19bdb535385559aabb5ba59729482800f8"
-  integrity sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==
```
Chris-Petty: What happened to your yarn? haha

src/sync/Synchroniser.js
raviSussol: Done.

yarn.lock
raviSussol: Ah, my yarn was old and so couldn't upgrade sha512 automatically (in which case it does in the new version). Upgraded to the new version now along with `yarn.lock`

andreievg:
@raviSussol, I think it would be better to store hardware UUID in settings and access it from there. (reduces imports in each file it's used and also I am not sure about over-head of calling DeviceInfo.getUniqueID()).

You can store it during app startup in MobileAppSettings constructor

src/sync/Synchroniser.js
```diff
@@ -111,9 +113,14 @@ export class Synchroniser {
...
             headers: {
               Authorization: this.authHeader,
+              'msupply-site-uuid': uniqueId,
             },
           },
-        );
+        )
+          .then(response => response.json()
+            .then(responseJson => {
+              if (responseJson.error) throw new Error(responseJson.error);
```
andreievg: @raviSussol 

We've been using async, await instead of then(), catch etc..

in this case the code will look like this:
```
 const response = await fetch(
          `${this.serverURL}/sync/v3/initial_dump/` +
            `?from_site=${this.thisSiteId}&to_site=${this.serverId}`,
          {
            headers: {
              Authorization: this.authHeader,
              'msupply-site-uuid': uniqueId,
            },
          },
        );
        const json = await response.json();
        if(json.error && json.error != '')...
```

src/sync/Synchroniser.js
```diff
@@ -244,7 +252,11 @@ export class Synchroniser {
       throw new Error('Unexpected response from sync server');
     }
     if (responseJson.error.length > 0) {
-      throw new Error('Server rejected pushed records');
+      if (responseJson.error.startsWith('Site registration doesn\'t match.')) {
+        throw new Error(responseJson.error);
+      } else {
+        throw new Error('Server rejected pushed records');
```
andreievg: We don't seem to inform of error here. I would suggest instead of 'Server rejected pushed record' -> `Server rejected pushed record (${responseJson.error)` <- would work for both cases + extra information in case there is error.

src/sync/Synchroniser.js
```diff
@@ -368,10 +382,15 @@ export class Synchroniser {
         method: 'POST',
         headers: {
           Authorization: this.authHeader,
+          'msupply-site-uuid': uniqueId,
         },
         body: JSON.stringify(requestBody),
       },
-    );
+    )
+      .then(response => response.json()
```
andreievg: similar here, as per first comment, await please

src/sync/Synchroniser.js
Chris-Petty: @andreievg this related #680

src/authentication/UserAuthenticator.js
```diff
@@ -55,7 +57,7 @@ export class UserAuthenticator {
       // Race condition promise, if connection is taking to long will be rejected
       // with CONNECTION_FAILURE error
       const userJson = await Promise.race([
-        authenticateAsync(authURL, username, passwordHash),
+        authenticateAsync(authURL, username, passwordHash, uniqueId),
```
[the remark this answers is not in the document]
raviSussol: @andreievg This may also leads to do changes in other places like `GET, POST` method inside `Synchroniser.js` file? Do you think the above change (passing extra header object rather than a single header text `msupply-site-uuid`) should also require inside `Synchroniser.js` rest api methods? If yes then I may also need to update on 4D mSupply backend part to parse and make readable for site uuid to check against the database.

src/sync/Synchroniser.js
raviSussol: No worries. Done.

src/sync/Synchroniser.js
raviSussol: Covered with https://github.com/sussol/mobile/pull/775/commits/85534a5104a4ce7fece2750ddf063cd6ce830c2b

src/authentication/UserAuthenticator.js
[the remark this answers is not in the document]
andreievg: @rabiSussol, i mean to write it this way, sorry:

`authenticateAsync(authURL, username, passwordHash, {'msupply-site-uuid':uniquID})`

or even better (since uniqueId should be stored in settings)

```
constructor(database, settings) {
this.extraHeaders = {'msupply-site-uuid': settings.get(HARDWARE_UUID)};
....
...
....
authenticateAsync(authURL, username, passwordHash, this.extraHeaders)
```

same for Synchroniser: 
```
headers: {
   Authorization: this.authHeader,
   ...this.extraHeaders
}

```
what do you think ?

src/sync/Synchroniser.js
raviSussol: Thanks @andreievg I've updated comments there about to handle the codes for these both issues.

src/settings/MobileAppSettings.js
```diff
@@ -19,6 +20,19 @@ export class MobileAppSettings extends Settings {
     }
   }
 
+  get(key) {
+    switch (key) {
+      case 'Hardware_UUID': return this.getHardwareUUID();
```
andreievg: Hi @raviSussol, I was suggesting that DeviceInfo.getUniqueID() is called only once, on app startup (in MobileAppSettings constructor) and the value is stored in the settings (then retrieved from settings).
Can you please remove the extra `get(key)` and `getHardwareUUID()`. And instead just have one line in the constructor: 
```
  constructor(database) {
    super(database);
    this.load();
    this.set(SETTINGS_KEYS.HARDWARE_UUID, DeviceInfo.getUniqueID())
  }

```

Chris-Petty: `load()` method defined at the bottom was intended for this kind of stuff, though in looking at it I think it's a bit silly, the content of `load` would be fine to have inline in the constructor.

raviSussol: @Chris-Petty @andreievg - should I remove `load()` method and replace it by putting inline code inside constructor too?

joshxg:
@andreievg @Chris-Petty @raviSussol 

Just checking before I continue:

With this issue, you should not be able to login to the same store with two tablets. I have set up a build on a tablet, initialized and logged into a store and also the same on an emulator, to the same store (Everything is fine with this). Next step is to try with two apk's on tablets, but before I crack open another box I thought I'd check that using an emulator would stop this from working

Chris-Petty:
I think emulator should be able to show this. Possible that the server isn't handling it.

joshxg:
@raviSussol - for the sake of time, since you seem to know all about this would you be able to double check if this is working? I did end up using two tablets to log into the same store, and I was able to. Maybe I did something wrong, though. If you don't have the time right now I'm happy to look into it, let me know

joshxg:
sorry @raviSussol - I just now realised https://github.com/sussol/msupply/pull/2280 the desktop PR is still not merged, so obviously it is not working yet

PR #781, opened 2019-01-07
#780 Sync Error: Value not convertible to a number

### LINKED ISSUE #780: Sync Error: Value not convertible to a number
During initial sync, getting this error:

screen shot 2019-01-07 at 6 25 17 pm

Caused by:

screen shot 2019-01-07 at 6 06 13 pm

screen shot 2019-01-07 at 6 20 20 pm

andreievg:
fixed with isNaN check

fixes #780


andreievg:
Realm seems to support infinity, so I don't envision any further breaks here.

screen shot 2019-01-08 at 11 01 37 am

Not a fan of storing the 'Infinity' number though, so not keen to translate 4D 'INF' to js Infinity.

PR #787, opened 2019-01-09
Upgrade all packages (Realm 2.22.0 being focus) to the latest version

### LINKED ISSUE #753: Update to realm@2.20.0
Build Number: 2.1.1
Description:  Currently using realm@1.13.0, multiple errors when running `yarn install` in newly cloned repo.
Reproducible: Yes.
Reproduction Steps: Clone the repo and run `yarn install` .
Comments: Test after updating, latest stable realm version is backwards-incompatible with current version.

sworup:
Fixes #753 

# Description
- Realm 2.21.1 only supports upto Node v10 so please use this version of node
- React Navigation has lots of problems with out code, I tried to make it work but v3, v2 and even the later revision of v1 would not play nice with our code
- I got v1.5 of React Navigation working (which has similar code to v2) but the back button would not work. So I decided to lock the version to v1.0 beta.
> Everything else is updated to the latest version


android/app/build.gradle
```diff
@@ -105,20 +101,12 @@ android {
         applicationId "com.msupplymobile"
         minSdkVersion rootProject.ext.minSdkVersion
         targetSdkVersion rootProject.ext.targetSdkVersion
-        versionCode appVersionCode
-        versionName getAppVersion()
+        versionCode 1
+        versionName "1.0"
```
Chris-Petty: Wiping out a bunch of our code in this file, losing app versioning of android

.buckconfig
```diff
@@ -1,6 +1,6 @@
 
 [android]
-  target = Google Inc.:Google APIs:26
+  target = Google Inc.:Google APIs:23
```
Chris-Petty: This appears to be a downgrade

android/app/src/main/java/com/msupplymobile/MainActivity.java
```diff
@@ -1,18 +1,8 @@
 package com.msupplymobile;
 
 import com.facebook.react.ReactActivity;
-import com.babisoft.ReactNativeLocalization.ReactNativeLocalizationPackage;
-import io.realm.react.RealmReactPackage;
```
Chris-Petty: I think you need to relink native modules. `react-native link` or something along those lines should do it. Unless it's simply all removed from here and only in `MainApplication.java` now.

android/app/src/main/java/com/msupplymobile/MainActivity.java
```diff
@@ -1,18 +1,8 @@
...
-import com.oblador.vectoricons.VectorIconsPackage;
-import com.bugsnag.BugsnagReactNative;
-import android.os.Bundle;
 
 public class MainActivity extends ReactActivity {
-    @Override
-    public void onCreate(Bundle savedInstanceState) {
-      super.onCreate(savedInstanceState);
-      BugsnagReactNative.start(this);
```
Chris-Petty: RIP bugsnag

android/build.gradle
```diff
@@ -1,12 +1,20 @@
...
+        supportLibVersion = "27.1.1"
+    }
     repositories {
-        jcenter()
         google()
+        jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.2.0'
+        classpath 'com.android.tools.build:gradle:3.1.4'
```
Chris-Petty: downgrade?

app.json
```diff
@@ -0,0 +1,4 @@
+{
+  "name": "mSupplyMobile",
+  "displayName": "mSupplyMobile"
```
Chris-Petty: Should this have a space in it? I presume this is what shows under the icon on a device.

android/gradle.properties
```diff
@@ -16,10 +16,3 @@
 # This option should only be used with decoupled projects. More details, visit
 # http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
 # org.gradle.parallel=true
-
-android.useDeprecatedNdk=true
-
-MYAPP_RELEASE_STORE_FILE=my-release-key.keystore
-MYAPP_RELEASE_KEY_ALIAS=my-key-alias
-MYAPP_RELEASE_STORE_PASSWORD=sussolpikelets
-MYAPP_RELEASE_KEY_PASSWORD=sussolpikelets
```
Chris-Petty: The keystore details should be somewhere, where are they now?

android/app/src/main/java/com/msupplymobile/MainApplication.java
```diff
@@ -32,11 +28,11 @@ public boolean getUseDeveloperSupport() {
     protected List<ReactPackage> getPackages() {
       return Arrays.<ReactPackage>asList(
           new MainReactPackage(),
-          new RNFSPackage(),
-          new RealmReactPackage(),
-          new VectorIconsPackage(),
-          new ReactNativeLocalizationPackage(),
-          BugsnagReactNative.getPackage()
+            new RealmReactPackage(),
```
Chris-Petty: These lines changes are just a reordering and excessive indentation?

android/app/src/main/java/com/msupplymobile/MainActivity.java
sworup: I did link the modules, which was supposed to add these lines to the MainActivity.java but it didn't. Both realm and localization are working without these lines so I did not manually add them.

android/app/src/main/java/com/msupplymobile/MainActivity.java
sworup: Sorry about this, missed putting it back.

android/build.gradle
sworup: gradle:3.1.4 comes with the latest version of React Native, I ungraded it in previous version upgrade but everything seems to work fine with gradle:3.1.4.

app.json
sworup: Yes displayname should have space. But this json file is not being used right now. We should add variable from this json file in places where names are hard coded right now but for now I did not touch the code unless it was absolutely necessary.

Chris-Petty:
@sworup looks good to go, just need the conflicts cleaned up

sworup:
@Chris-Petty I have resolved the conflicts and also updated react and other packages to the latest version. I have done the test and it works.

sworup:
EsLint now adds a lot of extra rules that out app does not comply with, so it will give you a lot of errors and warnings. I will bump it down a few versions for now tomorrow morning.

sworup:
From this merge on, our code will use Node v10.

android/app/src/main/AndroidManifest.xml
```diff
@@ -15,6 +15,7 @@
       android:allowBackup="true"
       android:label="@string/app_name"
       android:icon="@mipmap/ic_launcher"
+      android:roundIcon="@mipmap/ic_launcher_round"
```
joshxg: Hey @sworup  - adding this line seems to have changed the android icon

Chris-Petty:
@sworup could you please address Josh's comment and package/yarn conflicts? I want to try get this merged this week.

android/app/src/main/AndroidManifest.xml
sworup: @joshxg @Chris-Petty 

It seems android 8 or higher add a rounder launcher icon and adds a background and foreground layer. 

I have two choices here:

1. Remove roundIcons
2. Add png images in resources for round icons

I decided to go with option 2 and add round icons. 

I use Android Studio's image asset generator to generate the icons and xml configuration files. 

Then I replaced the transparent files that were already there in our system with the newly generated while background icons.

Chris-Petty:
@joshxg I'm happy enough with this if you are. The icons should work. If they don't we can fix it later.

joshxg:
They work thanks @sworup

PR #790, opened 2019-01-13
Import createOrUpdateRecord in mergeRecords.js

### LINKED ISSUE #778: Merge Test - createOrUpdateRecord not imported in mergeRecords.js
In develop branch

wlthomson:
Fixes #778.  

As merge logic is relatively bound to incoming sync, it may be more suitably located in the sync folder. This would avoid the ugly '../../' import statement.


andreievg:
@wlthomson, fair point. I guess merging is a unique use case, it's done rarely, so it's probably cleaner to keep it in `database/utilities`, away from everyday sync logic

PR #795, opened 2019-01-17
Customer Invoice - Add Master List Items - Exclude Items from combined MasterList that are not visible in the store

### LINKED ISSUE #783: Customer invoice > "Add master list items" adds items tablet store can't see
Build Number: 2.2.0-rc0 ( probably exists earlier )

Description: 
untitled

Here are a bunch of items in the customer invoice visible to the customer store, but not this store. So naturally there is no available stock

Reproducible: Yes

Reproduction Steps:
1. Create a customer invoice for a store with different item visibility than the store being worked on
2. See all the items that this store can't see and therefore has no available stock for

Comments:
Should only add the intersection of items between supplying store and customer name to the customer invoice I reckon.

sworup:
Fixes #783 

- Customer Invoice > Add Master List Items 
- Exclude Items from combined MasterList that are not visible in the store
- Done as per discussion with @ujwalSussol

### DIFF
--- a/src/database/DataTypes/Transaction.js
+++ b/src/database/DataTypes/Transaction.js
@@ -114,9 +114,7 @@ export class Transaction extends Realm.Object {
       this.otherParty.masterLists.forEach(masterList => {
         const itemsToAdd = complement(masterList.items, this.items, item => item.itemId);
         itemsToAdd.forEach(masterListItem => {
-          if (!masterListItem.item.crossReferenceItem
-            && database.objects('Item')
-            .filtered('id == $0 AND isVisible == true', masterListItem.item.id).length > 0) {
+          if (!masterListItem.item.crossReferenceItem && masterListItem.item.isVisible) {
             // Don't add cross reference items or we'll get duplicates
             createRecord(database, 'TransactionItem', this, masterListItem.item);
           }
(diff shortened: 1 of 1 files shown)


Chris-Petty:
I'm pretty sure that this approach will work (just filter what isn't visible in this store) but don't quite have a moment to test it today.

src/database/DataTypes/Transaction.js
```diff
@@ -114,7 +114,9 @@ export class Transaction extends Realm.Object {
       this.otherParty.masterLists.forEach(masterList => {
         const itemsToAdd = complement(masterList.items, this.items, item => item.itemId);
         itemsToAdd.forEach(masterListItem => {
-          if (!masterListItem.item.crossReferenceItem) {
+          if (!masterListItem.item.crossReferenceItem
+            && database.objects('Item')
```
andreievg: masterListItem.item.isVisible

sworup: I did not realize masterlist had visible field. Will do this.

andreievg: When you get a realm db object, you can access related records just like object

sworup:
@andreievg Requisition code also has add from masterlist like transaction does
https://github.com/openmsupply/mobile/blob/a5c005a4cbb37dcda5cb8b863874048047f1628f/src/database/DataTypes/Requisition.js#L96-L101

sworup:
> mobile/src/database/DataTypes/Requisition.js

If the Requisition checks the store's master lists then the items in the list should by default be visible to the store. Maybe we do not need to change this code, but leaving the comment so that someone could confirm.

PR #797, opened 2019-01-20
#768 eslint prettier configs

### LINKED ISSUE #768: Prettier config
We need one. Needs to play ball with eslint rules too.

wlthomson:
Fixes #768. Everything linted and prettifed using existing configs.


wlthomson:
Added husky for pre-commit hooks. All `*.{js,json,css,md}` files will be prettified and all `js` files also linted. In case of conflict, rules specified in `.eslintrc` take precedence over `.prettierrc`. This will cause a loop if prettier is applied on save (e.g. using a vscode extension), and the `eslint --fix` conflicts with this and requires manual intervention. If this occurs, either the prettier or eslint rules should be updated for consistency.

wlthomson:
The existing `.eslintrc` enforced the `no-confusing-arrow` rule with the `allow-parens` exception set to `true`. As prettier removes unnecessary parentheses, these parens will always be stripped, causing linter errors. This can be fixed by returning these expressions in a block body. In order to prevent this from also causing a linter error, the `arrow-body-style` rule is turned off. 

The other option is to turn off `no-confusing-arrow` rule entirely, and retain the default `arrow-body-style` rule. However, I went this route as it prioritises a style which is understandable rather than concise. 

e.g. old style:

```
export const getCurrentRouteName = state => state.routes[state.index] ? state.routes[state.index].routeName : undefined;
```

e.g. new style:

```
export const getCurrentRouteName = state => {
  return state.routes[state.index]
    ? state.routes[state.index].routeName
    : undefined;
};
```

wlthomson:
Following the installation instructions at [prettier.io](https://prettier.io/docs/en/install.html), we should pin a version of prettier to the project. I've pinned the latest version, which is spitting out a lot of style errors. These can all be fixed with a simple `prettier --write` (and some testing afterwards). 

The `eslint` packages we are using are also out of date, and updating spits out quite a lot of issues with the codebase which are probably important to look at. Updating all this means we can also update the linting and prettier configs and bring them all up-to-date.

This will be a more time consuming thing to do, but I think it might be worth it. The current versions are pretty far behind (~3-4 major versions), and if we leave it this is going to get bigger.

Chris-Petty:
@wlthomson can you clear up the conflicts, then should be good to go 👍

wlthomson:
@Chris-Petty Ready to go and conflicts merged. Quite a lot of changes.

I think it would be a good idea to enforce some style rules for mobile. I've added pre-commit hooks using `husky` for `prettier` and `eslint`.  good idea for everyone working on mobile to install the relevant editor plugins,

Aside from the enforced rules, there are a few conventions I have applied, which I think would be good to follow in future.

- Order imports alphabetically, with imports closer in the file hierarchy listed first (e.g. ../fileA` comes before `../../fileB`).
- Separate logical groups of imports with a newline (e.g. group of `react` imports, group of `react-native` imports, etc.)
- Use imperative voice for commenting functions, methods. E.g. `// Create a new requisition` not `// Creates a new requisition` for a function `createRequisition()`.
- Use consistent docstrings for functions, methods (example below):
  - Short description followed by newline.
  - Parameter declarations, with fields separated by two lines and aligned vertically.
  - Return value, omit if no return value (i.e. do not specify no return value).
- If object fields can be destructured without making code confusing or ambiguous, they should be, including `this`. e.g. `{ someField } = this;`
- Use "arrow" functions (no `function` keyword allowed!)
- Separate clusters of code within methods with newlines. This is down to individual preference, but I prefer a bit more use of newlines that mobile currently uses, especially around inline functions and callbacks.
- Last (and probably most controversial)... When referring to identifiers in a comment, enclose the variable in pipes. This was a style I was taught a while ago from a couple of Google guys. E.g. `// This is a comment string referring to |aVariable|.` Pipes are distinguished from absolute value, e.g. `// This is a comment string referring to |abs(aVariable)|`.

``
/**
 * This describes the function.
 * 
 * @param   {paramType}  paramName  This describes the parameter.
 * @return  {returnType}            This describes the returned value.
 */
```

PR #802, opened 2019-01-31
Fix ESLint error & warning squiggles

sworup:
- Continuation of PR #787 and is based on branch `#753-Upgrade-react-native-package-versions`
- Continuation of #753

# Description

- This makes ESLink compliant with our code.
- Some of the rules are suppressed, please check .eslintrc file to see which ones.
- Has the latest version of all of the packages except react-naviation


wlthomson:
Hey @sworup. Thanks for picking this up. I think it's best if we update the code rather than disable the rules, just in case we disable something important. This also means we can enforce the rules for later commits. I've done this in PR #797, if you want to have a look and leave any comments. 

@Chris-Petty @andreievg just keeping you guys in the loop.

I can see you did some work on this, so I'd be keen to get any feedback you have!

sworup:
> Hey @sworup. Thanks for picking this up. I think it's best if we update the code rather than disable the rules, just in case we disable something important. This also means we can enforce the rules for later commits. I've done this in PR #797, if you want to have a look and leave any comments. I can see you did some work on this, so I'd be keen to get any feedback you have!
> 
> @Chris-Petty @andreievg just keeping you guys in the loop.

These code were to removed the squiggles after the version was bumped up. I hid most of the squiggles by disabling the rules because I wanted them to just go away for now. 
I did not realise you had worked on ESLint too when I was working on this. This was more of a ESLint thing and I thought you worked on Prettier.
Our code should be compatible with each other I think, shouldn't take much of fixing to merge your code into this branch.
I am wondering if we could merge your branch here because I am working on updating React Navigate which originates from this branch.

sworup:
The branch with this pr is now merged to #787

Issue #793, opened 2019-01-15
Pressing 'done' on keyboard in comment edit, 'locks' modal


andreievg:
When pressing 'done' on they tablet keyboard in comment modal locks the modal. Can't close it with 'X' or 'Done' button on the modal. Only way to close the modal is to press 'Back' on the modal, which navigates back one extra level.

Related: https://github.com/sussol/mobile/issues/646

joshxg:
fixed by #938

Issue #646, opened 2018-04-03
Back navigation doesn't affect modals


Chris-Petty:
Build Number: 2.1.0-rc8

Description:
Originally this issue was for (in app and hardware) back button still navigating pages rather than closing modals. After #805 (I guess) the navigation problem is fixed, but the back button should still close modals.

- Add item
- New invoice/ any other selector modal.
- Batch Editing in stocktakes
- Editing comments/references on invoices (in pageInfo)
- Months of stock on requisition

Reproducible: yes

Reproduction Steps:
1. Open/create in progress requisition
2. Tap "Months stock" number to edit
3. Press hardware back button
The modal is closed, but app also navigates back to the requisition list.

Comments:

andreievg:
Related: https://github.com/sussol/mobile/issues/793

joshxg:
This just requires adding the prop `onRequestClose={onClose}` to `ModalContainer`

PR #805, opened 2019-02-07
#804 Upgrade react navigation package to v3.1.5

### LINKED ISSUE #804: Upgrade react-navigation from version 1.0.0-beta.11 to version 3.X.X (latest)
# Description

React navigation in our application has been struck at version 1.0.0-beta.11 for a long time. After the said version, the developers changed the code bootstrapping and most of the code to a large extent and our application became no longer compatible for a version upgrade.

We need to change our code so that the latest version of react-navigation, v3.1.5 at the time of this writing, would work with our code.

sworup:
Fixes #804

# Description

React navigation in our application has been struck at version 1.0.0-beta.11 for a long time. After the said version the package changed the code bootstrapping and most of the code to a large extent and our application became no longer compatible for a version upgrade.

We need to change our code so that the latest version of react-navigation, v3.1.5 at the time of this writing, would work with our code.

# Prerequisite 

- This code branches off of PR https://github.com/openmsupply/mobile/pull/787
- After the changes of #753


src/navigation/index.js
```diff
@@ -3,6 +3,6 @@
  * Sustainable Solutions (NZ) Ltd. 2019
  */
 
-export { Navigator } from './Navigator';
-export { reducer } from './reducer';
+export AppNavigator from './Navigator';
```
joshxg: Think default exports need to be imported and exported to keep the linter happy?

```
import AppNavigator from './Navigator';
import reducer from './reducer';

export { AppNavigator, reducer };
```

src/navigation/reducer.js
```diff
@@ -3,23 +3,15 @@
...
 
-  return Object.entries(params1).every(([key, value]) => {
-    return value === params2[key];
-  });
-};
-
-export const reducer = (state, action) => {
-  // Prevent the same route being pushed twice (e.g. double tap of button).
-  // See https://github.com/react-community/react-navigation/issues/135.
+const navReducer = (state = initialState, action) => {
```
joshxg: It seems a lot of this code is not longer needed since this react-navigation PR [here](https://github.com/react-navigation/react-navigation/pull/3393)

We can remove the code handling quick double tapping, which includes the `paramsEqual` method?

nit: could make this `const navReducer = (state = initialState, { type }) => { ... }` 
Also I much prefer navigation over nav

src/mSupplyMobileApp.js
```diff
@@ -310,7 +312,8 @@ MSupplyMobileAppContainer.propTypes = {
   syncState: PropTypes.object.isRequired,
 };
 
-function mapStateToProps({ navigation: navigationState, sync: syncState }) {
+function mapStateToProps(state) {
```
joshxg: nit: Should come before prop types

src/mSupplyMobileApp.js
```diff
@@ -21,11 +20,15 @@ import {
 } from 'react-native';
 
 import { Scheduler } from 'sussol-utilities';
+import { NavigationActions } from 'react-navigation';
+import globalStyles, {
```
joshxg: nit: I preferred the other format with styles being the last import

src/Store.js
```diff
@@ -0,0 +1,26 @@
...
+  blacklist: ['nav'],
+};
+
+const persistedReducer = persistReducer(persistConfig, reducers);
+
+const store = createStore(persistedReducer, {}, applyMiddleware(thunk, appNavigatorMiddleware));
+
+const persistedStore = persistStore(store);
+
+export { store, persistedStore };
```
joshxg: I like this refactor of the store.  I wonder if the file structure could use an improvement though? Personally I like having all of the redux related files consolidated in a 'store' or 'redux' folder. Currently it seems just so spread, makes things a little confusing to understand. What do you think?

src/reducers.js
```diff
@@ -5,12 +5,10 @@
...
-import { reducer as navigation } from './navigation';
-import { reducer as sync } from './sync';
+import { reducer as appReducer } from './navigation';
+import { reducer as syncReducer } from './sync';
 
-export const reducers = combineReducers({
-  navigation,
-  sync,
+export default combineReducers({
+  nav: appReducer,
```
joshxg: are we able to return this to navigation again? Then we can stop using nav everywhere?

src/reducers.js
```diff
@@ -5,12 +5,10 @@
...
+  sync: syncReducer,
```
joshxg: seems this is just renamed to `syncState` when it is used? Should we just name it that here?

src/reducers.js
```diff
@@ -5,12 +5,10 @@
 
 import { combineReducers } from 'redux';
 
-import { reducer as navigation } from './navigation';
-import { reducer as sync } from './sync';
+import { reducer as appReducer } from './navigation';
```
joshxg: This reducer seems to be renamed a lot, going from `navReducer` to `reducer` to `appReducer`. Could we just stick with one? Maybe `navigationReducer`?

src/mSupplyMobileApp.js
```diff
@@ -310,7 +312,8 @@ MSupplyMobileAppContainer.propTypes = {
   syncState: PropTypes.object.isRequired,
```
joshxg: I think when we are editing files we can slowly start to introduce default props while we're there rather than disabling, what do you think?

src/mSupplyMobileApp.js
```diff
@@ -118,8 +119,8 @@ class MSupplyMobileAppContainer extends React.Component {
     // If we are on base screen (e.g. home), back button should close app as we can't go back.
     if (!this.getCanNavigateBack()) BackHandler.exitApp();
     else {
-      const { navigation } = this.navigator.props;
-      navigation.goBack();
+      const { dispatch } = this.navigator.props;
+      dispatch(NavigationActions.back());
```
joshxg: Just an link to #793 and #646 - once this is in we should fix these problems

src/navigation/index.js
sworup: @joshxg yes, that is how it should be if I was using this `navigation/index.js` reference. Which I am not.

https://github.com/openmsupply/mobile/blob/9cf90124f4aed17eb533fd845572757b68505a06/src/navigation/ReduxNavigation.js#L5

src/reducers.js
sworup: @joshxg navigation is used by react navigation by default. Because we are using redux, we are avoiding react navigation's navigation variable and setup our own. If we use nav the variable names will conflict.

I would have preferred calling it navigation as well.

src/reducers.js
sworup: @joshxg They are state of the store, so they can be named syncState but that would become state.syncState which is not desirable either. One can even argue that after mapStateToProps they become props. :D

I would have preferred just naming them without `state` as word, but navigation is taken by react navigation plugin and only renaming sync does not sound too consistent.

src/reducers.js
sworup: Resolved with e129dc5

src/mSupplyMobileApp.js
sworup: I agree.

src/Store.js
sworup: I agree.

src/mSupplyMobileApp.js
Chris-Petty: I also agree, though I think prettier/eslint setup is currently going to force an order of:
1. Node module imports alphabetically
2. Relative imports alphabetically

src/navigation/index.js
Chris-Petty: My stance is to make the linter happy. If what makes the linter happy should be changed make a new issue to discuss that refactor so we can get this PR going.

sworup: These exports of AppNavigator and reducer in `src/navigation/index.js` are there but are never used since the import is happening directly (as in the example code I have left above). 

Linter is happy with the entire project, there is no error shown.

sworup:
@Chris-Petty These exports of AppNavigator and reducer in `src/navigation/index.js` are there but are never used since the import is happening directly (as in the example code I have left above). 

Linter is happy with the entire project, there is no error shown.

src/navigation/reducer.js
sworup: Resolved with 7011276

src/mSupplyMobileApp.js
sworup: Resolved with cb95329

src/mSupplyMobileApp.js
sworup: Yes this order is forced by eslint

src/Store.js
sworup: store.js and reducers.js are composition files that would put all the other files together. That's why they go to the root of src folder. They can be like a table of content.

I am not fan of creating redux folder. Store folder also does not sound good to me. We can create actions and reducers folders and keep all the actions and reducers there. That would be the standard redux way to do it.

sworup: Resolved with dfa7207

src/navigation/index.js
sworup: Resolved with dfa7207

sworup:
@joshxg @Chris-Petty This should be about it. Ready to merge.

Chris-Petty:
I just want to put 2c on the actions/reducers in their own folders. Not against it, pretty sure I did the same in Tupaia (in fact dunno if it's still the case, but all actions were in one file at the root of the project. Works fine till you have like 100 actions lol).

Particularly for components there is the pattern to keep the action/reducer/selector/component.js all together, so you don't have open several folders to find the logic for 1 thing. Can also crudely copy the component folder to another repo and get everything it needs to work.

sworup:
> I just want to put 2c on the actions/reducers in their own folders. Not against it, pretty sure I did the same in Tupaia (in fact dunno if it's still the case, but all actions were in one file at the root of the project. Works fine till you have like 100 actions lol).

@Chris-Petty we are not really using redux in our mobile application though. Redux is only there for sync and navigation part is to make react-navigation plugin happy, we are barely using redux in our code. For the functionality we are trying to get, I think this setup would be fine. For future, it's quite easy to change.

> Particularly for components there is the pattern to keep the action/reducer/selector/component.js all together, so you don't have open several folders to find the logic for 1 thing. Can also crudely copy the component folder to another repo and get everything it needs to work.

The modular approach is what I would also do. Specially after the emergence of [Bit](https://bit.dev/) and [Storybook](https://storybook.js.org/). But our use of redux (for now) is limited and complete with sync and probably we do not need to do the code overhaul required for this approach. I :+1: on doing this if our requirement changes in future.

PR #821, opened 2019-03-08
#820 Item DefaultPrice is synced incorrectly

### LINKED ISSUE #820: Item defaultPrice is synced incorrectly
Build Number: 2.2.0

# Description

https://github.com/openmsupply/mobile/blob/9dbaf8a9f9115edc2583c3b23c715b81f34b4f8b/src/sync/incomingSyncUtils.js#L70

This uses buy_price. Buy_price field in mSupply is actually only for "service" items. The default sell price of an item is actually per store and stored on item_store_join. Ha.

## Expected behaviour
When syncing item_store_joins it should set the default price on the related item record based on 		`[item_store_join]default_price`.

## Reproducible
It's just wrong. Symptoms would be that no default price anywhere is actually correct. Main run on affect from that is stocktakes creating new item batches will set them up with erroneous sell/cost price. Could happen with supplier invoices, too.

## Comment:
Develop this on master, patch and release with v2.2.1

sworup:
Fixes #820

Set defaultPrice through ItemStoreJoin instead of using item. buy_price which is not correct.


src/database/DataTypes/ItemStoreJoin.js
```diff
@@ -9,6 +9,7 @@ export class ItemStoreJoin extends Realm.Object {
     // Make the item no longer visible in this store
     const item = itemResults[0];
     item.isVisible = false;
+    item.defaultPrice = 0;
```
Chris-Petty: Not sure why this is necessary? Does mSupply desktop do this? Given visibility is controlled server side, I'd have thought the server would send any update to default price anyway?

sworup: This is not necessary, defaultPrice would be set to 0 by Item schema declaration. This code is also doing default value set.  
I did this (I think) for two reasons:
1. I was replicating the isVisible code in line 11.
2. It does not hurt the code and acts as a fail safe

sworup:
@Chris-Petty About `defaultPrice/packSize`: I removed the `parseNumber(record.buy_price) / packSize` code and replaced it with defaultPrice. I do not see any instance of `defaultPrice/packSize` in code.

src/database/DataTypes/ItemStoreJoin.js
Chris-Petty: Yea I guess if the join is recreated then it'll set the value again. mSupply doesn't actually usually delete these joins anyway, on desktop isVisible is on the join and just switches it True or False.

Chris-Petty:
image

Blocking issue when building dev on to genymotion device that already had a mobile version installed. Cheeky fix would be to make defaultPrice optional again, but I think it's probably better with the default as you have done. So that means bumping the schema version in this PR. Possibly also worth a bit of migration code that sets `defaultPrice` to 0 if currently undefined/null/NaN/falsy

src/database/DataTypes/ItemStoreJoin.js
Chris-Petty:
```suggestion
```
I still think it's not correct. 
1. Line 11 is necessary, being the core purpose of this destructor due to how mobile differs from desktop in handling visibility (it's recorded on item, not ItemStoreJoin).
2. If schema is handling it then it's going into redundant zone.

I'm going to push this along so I can merge it.

Chris-Petty:
Gah I just butchered this by merging develop into it. Changing base. Derp.

Issue #823, opened 2019-03-20
[IC][Requisitions][Stocktake] byProgramModal


joshxg:
Epic: https://github.com/sussol/org-issues/issues/18
Related issues: #815 #816 

Description:
Require a modal for gathering details for a requisition and stock take by programs. Should be generic enough to cater to both use cases.

Full design details #815

DhanyaHerath:
# TESTING
#### Requisition
- [x] The sequential steps work correctly
    - [ ] The step numbers are correct .. (1,2,3,4..) - _there are no sequence numbers_ 
    - [x] Each step is ticked as you step through the process
    - [x] When going back a step, the process is reset to the correct state
    - [x] When a step is unavailable, i.e. no periods, the interface clearly shows this
    - [x] All modals that appear resulting in a button press display the correct information

#### Stocktakes
- [x] Process steps display the correct state
- [x] Can type in a name and have it shown correctly

Not a bug as such, but seems inconsistent that if you have no programs you can create a stocktake without a name (and it will assign a name to it) but it's not possible to do so once you have programs associated.

#### All
- [x] Navigating between general and program types correctly display the right information
- [x] Navigating between general and program types correctly reset the state

Issue #818, opened 2019-03-06
[Ivory Coast] Add MOS (months of stock) to requisitions


wlthomson:
Epic:  sussol/org-issues#18

### Description:

MOS for a particular item represents that item's quantity divided by average monthly consumption: `currentMOS = stockOnHand / AMC`. The max MOS is taken into account when making a requisition,  anything above this should not be allowed to order: `requestedMOS + currentMOS <= maxMOS`. 

Threshold MOS is used to identify which items to show on a requisition (mainly for emergency requisitions). Anything above this will be hidden, i.e. only show items which are below the required threshold: `currentMOS <= thresholdMOS`.

### Requirements:

Requisitions for stock should satisfy `requestedMOS <= maxMOS - currentMOS`. 

Items displayed for a requisition should satisfy `currentMOS <= thresholdMOS`.

joshxg:
Essentially/quickly: If it is a program based requisition - use `MaxMOS` and `thresholdMOS` in `orderType` - use these to set the current MOS UI field and make this not editable. Have a threshold MOS field next to it somewhere which shows only on program based requisitions. Also a checkbox(?) for applying threshold MOS visibility

DhanyaHerath:
# TESTING:

- [x] Clicking `Hide Over Stock` button on a program requisition will hide all of the items whose current stock on hand exceeds the threshold MOS set in the order type used to create the requisition
- [x] Clicking automatic order will set the quantities to order using the maxMOS for the order type used to create the requisition
Nb. 'Automatic Order' is only in general orders, program requisitions have 'use suggested amount' only - both work. 
It does confuse me a bit to have the toggle set to 'Hide overstock' when the requisition is finalised. If you only order overstocked items, it looks like an empty requisition. But that might be on purpose and I'm just easily confused?

joshxg:
Thanks! Issue for here for the confusion: #1042

PR #836, opened 2019-04-01
#818 IC Requisition Customisations

### LINKED ISSUE #815: [Ivory Coast] Create requisitions by program
Epic:  sussol/org-issues#18

### Description:
Require functionality to create a requisition linked to a program

### Requirements:
- Creating a requisition should have the option to create a "program" requisition
- When these are linked, a user should not be able to add or remove items.
- Should require a period to be created (?)
- All the items in the program will have a requisition line related to the requisition

### Comments:
This has been implemented on desktop already issue [here](https://github.com/sussol/msupply/issues/2524).

#### TLDR of Desktops implementation:
- When a new requisition is to be created, *if there exists a `masterList` with `isProgram` [being added in this [issue](link here)*, prompt the user to create a `general` or `program` requisition. 
- The requisition is then populated with `requisitionLines` for each item in the `masterList`
- The user then selects the supplier and period after the requisition is created.

**I might be missing things here**

#### Thoughts on mobile implementation (from what I think andrei has described)

- When a user clicks create requisition and there exists a `masterList` with `isProgram = true`, a modal opens prompting if it is a `general` order or `program` order.
- If `general` order, nothing different.
- if `program` order - display a new modal prompting for:
    - supplier
    - program
    - period
    - order type
- Pre-populate the requisition with this data

#### things to note
- Period limitations on creating a requisition [ Ensuring multiple requisitions aren't created for a certain store/program in the same period.

Need more discussion on design and implementation. This is potentially a reasonable sized issue also, so could be broken down further once discussed and assigned a little further

joshxg:
Epic:  sussol/org-issues#18
Fixes #815 #823 #818 

Primary additions:
- `ByProgramModal` component -  Modal for creating a program requisition
- `SequentialSteps` component - component similar to `PageInfo` which when passed the 'steps' will disable/enable other components to allow a sequential progress flow

Minor additions:
- Realm model Helper methods 
    - Item: `isLessThanThresholdMOS`
    - MasterList: `get parsedProgramSettings`
    - Period: `removeRequisition`, `requisitionsForOrderType`, `toString()`
    - PeriodSchedule: `getPeriodsForOrderType`
- createRecord method
    - createProgramRequisition
- Updates/changes to:
    - `SupplierRequisitionPage`: To support program requisitions
    - `SupplierRequisitionsPage`: To support the `ByProgramModal` when a store uses programs
    - `AutocompleteSelector`: Refactored slightly to allow for arrays and results objects, as well as rendering an empty list when `data === null`
    - `PageInfo`: Allow for a new properties for in a 'title/info' property/config object - `shouldShow` and `canEdit`.
- New Utilitiy methods (`byProgram.js`)
    - `thisStoreUsesPrograms`
    - `getAllPrograms`
    - `getAllPeriodsForProgram`

Logic I think is pretty convoluted. I think extracting more of the logic from `ByProgramDialog` to `SequentialSteps` and creating new children components for that would help, however the cost of creating more components doesn't seem to be worth it, especially with even further prop drilling. 

Might need further testing from someone with more experience with mobile on the side effects of changes to `PageInfo`/`AutocompleteSelector`/`ToggleBar` on other components.

Potential issues from this (Can do them with this, if better)
- Still need to do localization.
- UI isn't good (`ByProgramModal`)
- Convoluted logic could be simplified, abstracted/extracted to more components.


Chris-Petty:
I assume if you've requested a review it's ready for it? Should this reviewed/merged before or after #837 (which PRs into the changes here)?

joshxg:
Yes. merged after #837 is merged into this would most likely be best. I was trying to reduce the diff so that it can be reviewed easier

src/database/DataTypes/MasterList.js
```diff
@@ -68,16 +72,24 @@ export class MasterList extends Realm.Object {
...
    * @param  {string}  tags            Current stores tags field
    * @param  {string}  orderTypeName   Name of the orderType to search for
    * @return {object} The matching orderType object
    */
   getOrderType(tags, orderTypeName) {
-    this.getStoreTagObject(tags).orderTypes.reduce(orderType => {
-      if (!(orderType.name === orderTypeName)) return null;
-      return orderType;
-    });
+    return this.getStoreTagObject(tags).orderTypes.filter(
```
Chris-Petty: Array.find() would be slightly neater

src/database/DataTypes/Options.js
```diff
@@ -4,7 +4,16 @@
...
 import Realm from 'realm';
-
+/**
+ * An Options object is a simple explanation of why a change or
+ * value was used in a situation. Used for stock takes, to
+ * provide a reason for having a stock take batch quantity differ
+ * from the stock take batch snapshot quantity.
+ * @property  {string}  id
+ * @property  {string}  title     A generic name field
+ * @property  {string}  type      The type of option. See below for types
```
Chris-Petty: See below for types?

src/database/DataTypes/Period.js
```diff
@@ -5,19 +5,55 @@
...
   }
+
+  getFormattedPeriod() {
+    return `${this.startDate.toLocaleDateString('en-US')} - ${this.endDate.toLocaleDateString(
+      'en-US'
+    )} `;
+  }
+
+  toString() {
+    return `${this.startDate.toLocaleDateString('en-US')} - ${this.endDate.toLocaleDateString(
```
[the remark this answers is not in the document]
joshxg: Just removed it 5b102033a6444ed930f321f2b0634571d984b812

src/database/DataTypes/Period.js
```diff
@@ -5,19 +5,55 @@
...
+ */
 export class Period extends Realm.Object {
   numberOfRequisitions() {
     return this.requisitions.length;
   }
 
-  numberOfRequisitionsForProgram(program) {
-    return this.requisitions.filtered('program.id = $0', program.id).length;
+  removeRequisition(requisition, database) {
+    const indexToRemove = this.requisitions.indexOf(requisition);
```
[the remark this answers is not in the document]
joshxg: Removed 62f79d382403c5df4850c98e87ff0a034b1f7fde

src/database/DataTypes/MasterList.js
joshxg: Fixed c0c9b0c6db00ab473b9009c4aeb0fe45b200e9b1

src/database/DataTypes/Options.js
joshxg: fixed 5eff88b97ac298594602cfeeb649dc403f2c8406

src/database/DataTypes/MasterList.js
```diff
@@ -44,18 +46,20 @@ export class MasterList extends Realm.Object {
...
-    return Object.entries(JSON.parse(this.programSettings).storeTags).reduce(
-      ([programStoreTag, storeTagObject]) => {
-        if (!(storeTags.indexOf(programStoreTag) >= 0)) return null;
-        return storeTagObject;
-      }
-    );
+    const firstParse = JSON.parse(this.programSettings);
+    if (!firstParse) return null;
+
+    return Object.entries(JSON.parse(firstParse).storeTags).filter(
```
andreievg: Would be safer is you check if storeTags exist. Also it's fine to break this down further so it's more readable, i.e;
* Use keys instead of entries 
* Then return foundKey ?  parsedProgramSettings[foundKey] : null

src/database/DataTypes/MasterList.js
andreievg: @Chris-Petty agree

src/database/DataTypes/MasterList.js
```diff
@@ -68,16 +72,24 @@ export class MasterList extends Realm.Object {
...
-      if (!(orderType.name === orderTypeName)) return null;
-      return orderType;
-    });
+    return this.getStoreTagObject(tags).orderTypes.filter(
+      orderType => orderType.name === orderTypeName
+    )[0];
+  }
+
+  get parsedProgramSettings() {
+    if (!this.programSettings) return null;
```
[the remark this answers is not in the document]
andreievg: btw should use this method instead of parsing in getStoreTagObject

src/database/DataTypes/Options.js
```diff
@@ -4,7 +4,16 @@
...
+ * @property  {boolean} isActive  If the option is currently available to use
+ */
 export class Options extends Realm.Object {}
 
 export default Options;
```
andreievg: should the export be after options schema defined below ? I guess this is done in Stocktake PR ?

src/database/DataTypes/Requisition.js
```diff
@@ -39,11 +42,14 @@ export class Requisition extends Realm.Object {
   }
 
   /**
-   * Delete requisition and associated requisition items.
+   * Delete requisition and associated requisition items/period.
    *
    * @param {Realm} database
    */
   destructor(database) {
+    if (this.period) {
```
andreievg: once again, no need to removeRequisition from period

src/database/utilities/createRecord.js
```diff
@@ -341,6 +341,28 @@ const createTransactionItem = (database, transaction, item) => {
   return transactionItem;
 };
 
+const createProgramRequisition = (database, requisitionValues) => {
```
andreievg: can destructure like this (database, {period, currentUser, supplier, orderType})

Also, i am not sure if this needs to be another method. Can either: 
* Keep it as createProgramRequisition, but use createRequisition inside this method (to prevent repeating code)
* Or extend createRequisition: `const createRequisition = (database, user, otherStoreName, programSettings)`

With the second option:
```
const programValues = !programSettings ? {} : {
    otherStoreName: programSettings.supplier,
    daysToSupply: programSettings.orderType.maxMOS * 30,
    ...programSettings,
    orderType: programSettings.orderType.name,
}
const requisition = database.create('Requisition', {
  ...programValues
 ....
```

by the way, what extra comes from `...requisitionValues` ?

There is another option (my preferred), and that is: `const createRequisition = (database, user, otherStoreName, programSettings = {})`, and then make sure programSetting can be destructured directly into requisition

src/database/DataTypes/MasterList.js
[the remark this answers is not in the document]
joshxg: Forgot about this - I was thinking that 4D is stringifying incorrectly. We don't need it nicely formatted now, so we could just change the option?

src/database/utilities/createRecord.js
joshxg: I have this as one of the potential issues with the PR. Wasn't sure if I should fix this up here or make an issue afterward

andreievg: Hmm, since we are not particularly time pressed, i would suggest fixing in this PR

src/pages/SupplierRequisitionPage.js
```diff
@@ -49,9 +53,17 @@ export class SupplierRequisitionPage extends React.Component {
     };
   }
 
+  componentDidMount = () => {
+    const { requisition, settings } = this.props;
+    if (requisition.program) {
+      const tags = settings.get(SETTINGS_KEYS.THIS_STORE_TAGS);
```
andreievg: I've just noticed that thresholdMOS is not in mobile schema, and doesn't sync to mobile.
I think this needs to happen. The only reason you need to find orderType object here is to get thresholdMOS. It should be assigned as requisition is created.
If orderTypes change, existing requisitions should not be affected.

andreievg: as per above, we don't need componentDidMount method.

src/pages/SupplierRequisitionPage.js
```diff
@@ -41,6 +42,9 @@ export class SupplierRequisitionPage extends React.Component {
       modalKey: null,
       modalIsOpen: false,
       selection: [],
+      isProgramOrder: false,
+      orderType: null,
```
andreievg: isProgramOrder, orderType, thresholdMOS (as per below comments) should be set based on props.requisition

src/pages/SupplierRequisitionPage.js
```diff
@@ -63,10 +75,14 @@ export class SupplierRequisitionPage extends React.Component {
 
   onCreateAutomaticOrder = () => {
```
andreievg: if we use `Use Suggested Quantities` button, logic in the method is not needed

src/pages/SupplierRequisitionPage.js
```diff
@@ -163,14 +179,22 @@ export class SupplierRequisitionPage extends React.Component {
    */
   refreshData = (newSearchTerm, newSortBy, newIsAscending) => {
     const { requisition } = this.props;
+    const { useThresholdMOS, orderType } = this.state;
 
-    this.updateDataFilters(newSearchTerm, newSortBy, newIsAscending);
-    const { searchTerm, sortBy, isAscending } = this.dataFilters;
+    if (newSearchTerm && newSortBy && newIsAscending) {
+      this.updateDataFilters(newSearchTerm, newSortBy, newIsAscending);
+    }
```
andreievg: hmm not sure why this is changed, this.updateDataFilter checks `newSearchTerm && newSortBy && newIsAscending`

src/pages/SupplierRequisitionPage.js
```diff
@@ -163,14 +179,22 @@ export class SupplierRequisitionPage extends React.Component {
...
 
-    const data = requisition.items.filtered(
+    const { searchTerm, sortBy, isAscending } = this.dataFilters;
+    let data = requisition.items.filtered(
       'item.name BEGINSWITH[c] $0 OR item.code BEGINSWITH[c] $0',
       searchTerm
     );
+    if (useThresholdMOS) {
+      data = data.filter(requisitionItem =>
```
andreievg: i think this if statement should be the only change in this method ?

src/pages/SupplierRequisitionPage.js
```diff
@@ -163,14 +179,22 @@ export class SupplierRequisitionPage extends React.Component {
...
+        requisitionItem.item.isLessThanThresholdMOS(orderType.thresholdMOS)
```
andreievg: not a required change, but i wonder if realm can do this much quicker (and maybe we don't need that getter) don't know if this would work or not `data.filtered('stockOnHand < (monthlyUsage * $0)';thresholdMos)`

src/pages/SupplierRequisitionsPage.js
```diff
@@ -218,6 +251,16 @@ export class SupplierRequisitionsPage extends React.Component {
           onClose={() => this.setState({ isCreatingRequisition: false })}
           title={modalStrings.search_for_the_supplier}
         />
+        {byProgramModalOpen && (
+          <ByProgramModal
+            isOpen={byProgramModalOpen}
+            onConfirm={this.onNewRequisition}
+            onCancel={this.onCancelByProgram}
```
andreievg: For onConfirm and onCancel I think it's better to call `this.setState({ byProgramModalOpen: false })` here, i.e:
```
onConfirm={({supplier, ...requisitionValues}) =>
   this.setState({byProgramModalOpen: false}, () => this.onNewRequisition(supplier, requisitionValues))
}
```
And if you change createRecord to be generic for both normal and program requisition, then onNewRequisition would just be:
```
onNewRequisition = (otherStoreName,  requisitionValues) =>{
...
      requisition = createRecord(database, 'Requisition', currentUser, otherStoreName, requisitionValues);	  
...
}
```

src/pages/SupplierRequisitionsPage.js
```diff
@@ -38,6 +41,17 @@ export class SupplierRequisitionsPage extends React.Component {
     };
   }
 
+  // Need to add filter to find if there are programs [Not in the schema of this branch]
+  componentDidMount() {
+    const { database } = this.props;
+    const usesPrograms = database.objects('MasterList');
```
andreievg: Should check master lists for this store and isProgram (check out SupplierRequisitionPage.onAddMasterItems for hints)
Originally, when i was thinking about mobile logic for figuring out programs, I thought this way:
When newRequisition button is pressed, try to find all programs that have at least one order type for current store tags, then build an object with all relevant information (keeps all of the programSetting object parsing logic in one place, and only parse it once when creating requisition). If no programs with order types are found then use the previous modal, if they are found use ByProgramModal <- passing all needed information to it:
The object passed to ByProgramModal would be something like:
```
{
 programOneName: {
  orderType: { periods: [period1, period2, period3, etc..] // periods here can be realmObject
      maxMos, thresholdMos etc..
    }
   }
}
```
I haven't fully reviewed ByProgramModal, but it looks like it would be doing the queries to figure out order types etc.. No need for refactor on this, but would like to hear your thoughts

src/sync/incomingSyncUtils.js
```diff
@@ -373,7 +373,7 @@ export const createOrUpdateRecord = (database, settings, recordType, record) =>
         name: record.description,
         note: record.note,
         isProgram: parseBoolean(record.isProgram),
-        programSettings: JSON.stringify(record.programSettings),
+        programSettings: record.programSettings ? JSON.stringify(record.programSettings) : '',
```
andreievg: As per comments up the track, `record.programSettings && record.programSettings.replace(/\\/g, '')`

src/utilities/byProgram.js
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Finds if the currently logged in store is a store which uses programs.
+ * If true then: there exists a master list which is a program, which has
+ * a store tag in programSettings which also exists in the tags for this
+ * store.
+ * @param  {Object}  settings
+ * @param  {Realm}   database
+ * @return {boolean} indicator of the currently logged in store using programs
+ */
+const thisStoreUsesPrograms = (settings, database) => {
```
andreievg: Don't see this method being used, can use in SupplierRequisitionPage ?

src/utilities/byProgram.js
```diff
@@ -0,0 +1,55 @@
...
+  const thisStoresTags = settings.get('ThisStoreTags');
+  const programs = database.objects('MasterList').filtered('isProgram = $0', true);
```
andreievg: also 'MasterList' that is joined to current store

src/utilities/byProgram.js
```diff
@@ -0,0 +1,55 @@
...
+  const hasAProgram = programs.find(program => program.canUseProgram(thisStoresTags));
+  return !!hasAProgram;
```
andreievg: canUseProgram should also check that at least one order type exists

src/utilities/byProgram.js
```diff
@@ -0,0 +1,55 @@
...
+ * @param  {Realm}              database
+ * @param  {Object/MasterList}  program             An object consisting of a programID
+ * @param  {string}             periodScheduleName  the name of a periodSchedule
+ * @param  {Object}             orderType           an order Type object
+ * @return {array}
+ */
+const getAllPeriodsForProgram = (database, program, periodScheduleName, orderType) =>
+  database
+    .objects('PeriodSchedule')
+    .filtered('name = $0', periodScheduleName)[0]
```
andreievg: PeriodSchedule may not exist

src/pages/SupplierRequisitionsPage.js
joshxg: Ah that is a bad check for programs existing, huh.

As for passing props: This is how I had it originally, passing all the program values: programs, suppliers, ordertypes and periods to `ByProgramModal`. It turned out to be a lot messier, in my opinion. Supplier requisitions page was already a sizeable component and adding the logic to manage these state transformations (which is quite a lot, if you are accounting for managing at which stage in the process you are in to reduce the amount of querying that is going on) seemed to be quite a nightmare. Although in hindsight this was also when I was trying to use the `PageInfo` prop, so that could have been a part of the problem also. 

Basically it just ended up being a bit cleaner to manage state within the modal component rather than having the modal uncontrolled and simply passing in props for values, event handlers etc. I think I would still prefer to abstract even more of the state management into more components, though

src/widgets/AutocompleteSelector.js
```diff
@@ -74,7 +103,7 @@ export class AutocompleteSelector extends React.PureComponent {
         {data.length > 0 && (
           <FlatList
             data={data}
-            keyExtractor={item => item.id}
+            keyExtractor={item => item.id || item.name}
```
andreievg: can use a prop identifying which 'key' to use, but not required atm

src/widgets/PageButton.js
```diff
@@ -21,9 +21,10 @@ export function PageButton(props) {
   if (isDisabled) defaultTextStyle.push(globalStyles.disabledButtonText);
   return (
     <Button
-      {...buttonProps}
+      isDisabled={isDisabled}
```
andreievg: hmm, does this mean disabled buttons where not actually disabled

src/widgets/PageInfo.js
```diff
@@ -69,7 +72,7 @@ const renderInfoComponent = (isEditingDisabled, columnIndex, rowData, rowIndex)
       {infoString}
     </Text>
   );
-  if (rowData.onPress && !isEditingDisabled) {
+  if (rowData.onPress && !isEditingDisabled && (!isEditingDisabled && canEdit !== false)) {
```
andreievg: I don't think we need canEdit at all.
In SupplierRequisition:
```
        {
          title: `${pageInfoStrings.months_stock_required}:`,
          info: requisition.monthsToSupply,
          ...(isProgramOrder ? {} : {
          onPress: !isProgramOrder &&  this.openMonthsSelector,
          editableType: !isProgramOrder && 'selectable',
          })
        },
```

or does that look dirty ?

src/widgets/PageInfo.js
```diff
@@ -129,14 +137,14 @@ export const PageInfo = props => {
             }
           >
             <View>
-              {columnData.map((...args) =>
-                renderTitleComponent(isEditingDisabled, columnIndex, ...args)
-              )}
+              {columnData
+                .filter(data => data.shouldShow !== false)
```
andreievg: If am not a fan of default value = true for optional props/keys. In my opinion it' better to use falsy, i.e. shouldHide, then you don't need to !== false. Keeping optionals as falsy would avoid ppl making mistakes when changing code in the future

src/utilities/byProgram.js
joshxg: I will add this, just a tangent: Do you think a there should be an enforcement of an orderType in 4D when creating a store tag? (When you now set a master list to be a program, there is validation that there exists a store tag, do you think extending this to also include requiring an order type is to much?)

src/widgets/PageButton.js
joshxg: I believe so

src/widgets/PageInfo.js
joshxg: It is my preferred way, my only problem is that page info configs are so big that adding this rather than the prop does make it look a bit dirty. Not sure!

src/widgets/PageInfo.js
joshxg: This was to avoid side effects in the rest of the app, not having to go and set `shouldShow` on everywhere that uses `PageInfo`. I think best to keep it like this, make a new issue to fix it?

src/widgets/SequentialSteps.js
```diff
@@ -0,0 +1,131 @@
...
+import { StyleSheet, View, Text, TouchableOpacity } from 'react-native';
+
+import { WARM_GREY, SUSSOL_ORANGE } from '../globalStyles';
+
+export default class SequentialSteps extends React.PureComponent {
+  renderStepIcon = ({ isCurrentStep, isLessThanCurrentStep, step }) => {
+    const { error } = step;
+    return (
+      <View style={{ minWidth: 40 }}>
+        {isCurrentStep && error && <Ionicons name="ios-close" color="red" size={30} />}
```
andreievg: what the use case for an error ?

src/widgets/modals/ByProgramModal.js
```diff
@@ -0,0 +1,406 @@
...
+
+import { StyleSheet, View } from 'react-native';
+
+import { PageContentModal } from './PageContentModal';
+
+import globalStyles, { DARK_GREY, WARM_GREY, SUSSOL_ORANGE } from '../../globalStyles';
+import { AutocompleteSelector, ToggleBar, PageButton, TextEditor } from '..';
+
+import { SETTINGS_KEYS } from '../../settings';
+import { getAllPrograms, getAllPeriodsForProgram } from '../../utilities/byProgram';
```
andreievg: byProgram should be exported in utilities/index

src/widgets/modals/ByProgramModal.js
```diff
@@ -0,0 +1,406 @@
...
+        orderType: {},
+        orderTypes: null,
+        modalIsOpen: false,
+        currentStep: 1,
+      };
+      break;
+    case 'SELECT_ORDER_TYPE':
+      newState = { period: {}, periods: null, modalIsOpen: false };
+      break;
+    case 'SELECT_NAME':
```
andreievg: what's the difference between name and supplier ?

src/widgets/modals/ByProgramModal.js
```diff
@@ -0,0 +1,406 @@
...
+  orderTypeTitle: 'Select a Order Type to use',
+  periodTitle: 'Select a Period to use',
+  nameTitle: 'Give your stocktake a name',
+  programError: 'No programs available',
+  supplierError: 'No suppliers available',
+  orderTypeError: 'No order types available',
+  periodError: 'No periods available',
+};
+
+const getNewState = command => {
```
andreievg: this can just be an object 
```
return { 
   'RESET_ALL': { 
       program: {},
        supplier: {},
        period: {},
        periods: null,
        orderType: {},
        orderTypes: null,
        name: null,
        currentStep:  
    }, ...
}[command] ||  { modalIsOpen: false };
...

```

src/widgets/SequentialSteps.js
joshxg: currently using it for an empty result - eg no periods are defined for this combination of program and order type

src/widgets/modals/ByProgramModal.js
joshxg: supplier is the name object - I think I should have named this `otherStoreName` to be more consistent with mobile. Name is the name of a stock take

src/widgets/modals/ByProgramModal.js
```diff
@@ -0,0 +1,406 @@
...
+      default:
+        break;
+    }
+    return steps;
+  };
+
+  getSequentialStepsProps = () => {
+    const { name } = this.state;
+    const stepNumbers = this.getStepNumbers();
+    const programValues = this.getProgramValues();
```
andreievg: maybe program values should be in state.programValues ? this way don't need this.getProgramValues method

src/widgets/modals/ByProgramModal.js
```diff
@@ -0,0 +1,406 @@
...
+
+  // Initial set up or update the current steps property to reflect the current state
+  setCurrentSteps = () => {
+    const sequentialSteps = this.getSequentialStepsProps();
+    const steps = this.getSteps().map(step => sequentialSteps[step]);
+    this.setState({ steps });
+  };
+
+  // Gets the available steps this modal can handle
+  getSteps = () => {
```
andreievg: I think it's neater to use objects here too:

return { 'requisition': 
             isProgramBased ? ['program', 'supplier', 'orderType', 'period'] : ['supplier'],
              'stocktake': 
             isProgramBased ? ['program', 'name'] : ['name'] 
           } [type]

src/widgets/modals/ByProgramModal.js
```diff
@@ -0,0 +1,406 @@
...
+  };
+
+  getStepNumbers = () => {
+    const { isProgramBased } = this.state;
+    return {
+      program: 0,
+      supplier: isProgramBased ? 1 : 0,
+      orderType: 2,
+      period: 3,
+      name: isProgramBased ? 1 : 0,
```
andreievg: not a huge fan of keeping track of steps as number btw

src/pages/SupplierRequisitionPage.js
```diff
@@ -41,6 +42,9 @@ export class SupplierRequisitionPage extends React.Component {
       modalKey: null,
       modalIsOpen: false,
       selection: [],
+      isProgramOrder: false,
+      orderType: null,
+      useThresholdMOS: false,
```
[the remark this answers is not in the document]
joshxg: true when program requisition? The toggle is only visible for program requisitions - 11610b84b2db57a37d938606a335860145e4b77a

src/widgets/modals/ByProgramModal.js
joshxg: changed to use keys only 4ce5f8b35bc68f8ca5cd7a2d47e42bd383aa7e87

src/widgets/modals/ByProgramModal.js
joshxg: ended up removing this method as using an object for this made it unnecessary 3816a439a08b7de1910db19e0178c88558d06d45

src/widgets/modals/ByProgramModal.js
joshxg: Done - couldn't quite understand your pseudo though c00254b837aeed77c1808616114b36d7cf209fee

src/widgets/modals/ByProgramModal.js
joshxg: Actually noticed supplier is used a lot also throughout the app. I think `otherStoreName` is actually kind of a bad variable name, so sticking with `supplier` unless there are objections

src/widgets/modals/ByProgramModal.js
joshxg: done b983e71ac72281f8329e9de43dd87dbdf1ce6aac

src/widgets/SequentialSteps.js
joshxg: added a doc string 4d7664bb0599396d3acd2563272644fd459c6ff7

src/widgets/SequentialSteps.js
```diff
@@ -0,0 +1,131 @@
...
+      <View style={textContent}>
+        <Text numberOfLines={1} ellipsizeMode="tail" style={style}>
+          {(error && errorText) || name || placeholder}
+        </Text>
+      </View>
+    );
+  };
+
+  renderEditIcon = ({ canEdit, isInput, error }) => (
+    <View style={{ marginTop: 4, minWidth: 60, display: 'flex', alignItems: 'flex-end' }}>
```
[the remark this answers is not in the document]
joshxg: done bbf393f43245c7716663824b35884b15015ddf1e

src/widgets/SequentialSteps.js
```diff
@@ -0,0 +1,131 @@
...
+
+SequentialSteps.defaultProps = {
+  type: 'select',
+};
+SequentialSteps.propTypes = {
+  onPress: PropTypes.func.isRequired,
+  steps: PropTypes.array.isRequired,
+  type: PropTypes.string,
+  currentStep: PropTypes.number.isRequired,
+};
```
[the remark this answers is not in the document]
joshxg: done 4d7664bb0599396d3acd2563272644fd459c6ff7

src/widgets/PageInfo.js
joshxg: done 20821301679666ffe2b38fb80267fac54d73c1b9

src/widgets/PageInfo.js
joshxg: done 20821301679666ffe2b38fb80267fac54d73c1b9

src/utilities/byProgram.js
joshxg: Done 9175d2a7ec9f1aa36f3a972df640e87ef1341393

src/utilities/byProgram.js
joshxg: done b983e71ac72281f8329e9de43dd87dbdf1ce6aac

src/utilities/byProgram.js
joshxg: done b983e71ac72281f8329e9de43dd87dbdf1ce6aac

src/utilities/byProgram.js
joshxg: done b983e71ac72281f8329e9de43dd87dbdf1ce6aac - using 'getAllPrograms' as a check for if this store uses programs

src/sync/incomingSyncUtils.js
joshxg: done as above

src/pages/SupplierRequisitionsPage.js
joshxg: Done here b983e71ac72281f8329e9de43dd87dbdf1ce6aac

src/pages/SupplierRequisitionsPage.js
joshxg: I consolidated these to use the same method. Even though it's a probably insignificant performance enhancement I think it's best to stay away from creating new functions in the render method as a general rule/convention? 7c48ac37482d327dc51f0f8275ff91a4484445e5

src/pages/SupplierRequisitionsPage.js
```diff
@@ -7,11 +7,12 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 import { GenericPage } from './GenericPage';
+import { ByProgramModal } from '../widgets/modals/index';
```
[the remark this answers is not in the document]
joshxg: done b983e71ac72281f8329e9de43dd87dbdf1ce6aac

src/pages/SupplierRequisitionPage.js
```diff
@@ -320,38 +364,63 @@ export class SupplierRequisitionPage extends React.Component {
...
-            isDisabled={requisition.isFinalised}
-          />
+          {isProgramOrder && (
+            <ToggleBar style={globalStyles.toggleBar} toggles={this.getToggleBarProps()} />
+          )}
+
+          {!isProgramOrder && (
+            <PageButton
+              style={globalStyles.leftButton}
+              text={buttonStrings.use_suggested_quantities}
```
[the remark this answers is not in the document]
joshxg: done as above 4dab3938830fa467815b1fc17fbe120d8781c7a7

src/pages/SupplierRequisitionPage.js
```diff
@@ -320,38 +364,63 @@ export class SupplierRequisitionPage extends React.Component {
     }
   };
 
+  getToggleBarProps = () => {
```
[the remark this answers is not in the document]
joshxg: done 4b24270c726dde3b771207d9bc679d468a929655

src/pages/SupplierRequisitionPage.js
joshxg: I couldn't get a similar query to this to work. Using `data = data.filter(requisitionItem => requisitionItem.isLessThanThresholdMOS);` still 9095693bb8516259926361a7bf6d4ca464bdad4f

src/pages/SupplierRequisitionPage.js
joshxg: done 9095693bb8516259926361a7bf6d4ca464bdad4f

src/pages/SupplierRequisitionPage.js
joshxg: done 9095693bb8516259926361a7bf6d4ca464bdad4f

src/pages/SupplierRequisitionPage.js
joshxg: done 4dab3938830fa467815b1fc17fbe120d8781c7a7

src/pages/SupplierRequisitionPage.js
joshxg: - Removed `isProgramOrder` 4dab3938830fa467815b1fc17fbe120d8781c7a7
- Removed `orderType` state b582583b51aa8e56582f4c80a5ede306971861b0
- Using `ThresholdMOS` from requisition 9095693bb8516259926361a7bf6d4ca464bdad4f

src/pages/SupplierRequisitionPage.js
joshxg: - Added `thresholdMOS` 5e15d9b71efa180f31d83a288659732de7d25e40
- Removed `componentDidMount` b582583b51aa8e56582f4c80a5ede306971861b0

src/database/utilities/createRecord.js
joshxg: Did this over a few commits
- removed the createProgramRequisition - 27787c22a42b37a8e1f3b3a16f992e202eb9e19b
- changed parameters - d05e85dde3ebfa2266291cd993a753f44c68ebdb

src/database/DataTypes/Requisition.js
joshxg: done 62f79d382403c5df4850c98e87ff0a034b1f7fde

src/database/DataTypes/Options.js
joshxg: done a1ea4e47e7ac0b926851fe89977a67cda9730d65

src/database/DataTypes/MasterList.js
joshxg: done b983e71ac72281f8329e9de43dd87dbdf1ce6aac

src/database/DataTypes/MasterList.js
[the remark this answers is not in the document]
joshxg: - removing extra stringifying 60b96aec358d560c0032da965637f01e5c8449af 
- using this in `getStoreTagObject` b983e71ac72281f8329e9de43dd87dbdf1ce6aac

src/pages/SupplierRequisitionPage.js
```diff
@@ -41,6 +42,7 @@ export class SupplierRequisitionPage extends React.Component {
       modalKey: null,
       modalIsOpen: false,
       selection: [],
+      useThresholdMOS: !!props.requisition.program,
```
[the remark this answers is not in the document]
andreievg: meaning there is a use case where thresholdMOS will be set to 0

src/pages/SupplierRequisitionPage.js
```diff
@@ -320,38 +343,69 @@ export class SupplierRequisitionPage extends React.Component {
...
+            style={{
+              ...globalStyles.topButton,
+              ...(program ? { marginLeft: 5 } : {}),
+            }}
             text={buttonStrings.use_suggested_quantities}
             onPress={this.onUseSuggestedQuantities}
             isDisabled={requisition.isFinalised}
           />
+
+          {program && this.renderThresholdMOSToggle()}
```
andreievg: Here again i think, program && requisition.thresholdMOS // if requisition.thresholdMOS is missing or is 0, don't show the toggle

src/pages/SupplierRequisitionsPage.js
```diff
@@ -38,36 +40,44 @@ export class SupplierRequisitionsPage extends React.Component {
...
-
     let requisition;
+    if (!requisitionValues) {
+      this.setState({ byProgramModalOpen: false, isCreatingRequisition: false });
+      return;
+    }
     database.write(() => {
-      requisition = createRecord(database, 'Requisition', currentUser, otherStoreName);
+      requisition = createRecord(database, 'Requisition', currentUser, requisitionValues);
+      if (requisition.program) requisition.addItemsFromProgram(database);
```
andreievg: do we close the modal if !!requisitionValues ?

src/pages/SupplierRequisitionsPage.js
```diff
@@ -133,17 +142,17 @@ export class SupplierRequisitionsPage extends React.Component {
...
+    return (
+      <PageButton text={buttonStrings.new_requisition} onPress={() => this.setState(newState)} />
+    );
+  };
 
   render() {
-    const { database, genericTablePageStyles, topRoute } = this.props;
-    const { data, isCreatingRequisition, selection } = this.state;
-
+    const { database, genericTablePageStyles, topRoute, settings } = this.props;
```
andreievg: where do settings come from ?

joshxg: redux - all of these props are

src/pages/SupplierRequisitionsPage.js
joshxg: Gets closed in `navigateToRequisition` [ does cleanup, clears the selection etc. ]

src/pages/SupplierRequisitionsPage.js
```diff
@@ -38,36 +40,44 @@ export class SupplierRequisitionsPage extends React.Component {
     };
   }
 
-  onDeleteConfirm = () => {
+  componentDidMount() {
+    const { settings, database } = this.props;
+    this.setState({ usesPrograms: !!getAllPrograms(settings, database) });
```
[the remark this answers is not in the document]
joshxg: Fixed 13cc7a4dc368832eaa0dc68d184f9235a1e02f08

src/pages/SupplierRequisitionPage.js
joshxg: fixed 47e28093d7fd4a9e6496fccbdab1bfb94d9a2220

src/pages/SupplierRequisitionPage.js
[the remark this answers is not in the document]
joshxg: fixed 47e28093d7fd4a9e6496fccbdab1bfb94d9a2220 [ includes next comment about mos also ]

src/database/utilities/createRecord.js
```diff
@@ -134,7 +134,7 @@ const createItemBatch = (database, item, batchString) => {
  * @param   {Name}         otherStoreName  Name of other store (e.g. supplying store).
  * @return  {Requisition}
  */
-const createRequisition = (database, user, otherStoreName) =>
+const createRequisition = (database, user, { otherStoreName, program, period, orderType }) =>
```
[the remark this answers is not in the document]
joshxg: fixed d27b1f54ca30efbf341cb55bdf1146581f96039b

src/database/DataTypes/RequisitionItem.js
```diff
@@ -23,6 +24,18 @@ import { parsePositiveInteger } from '../../utilities';
...
  */
 export class RequisitionItem extends Realm.Object {
+  /**
+   * Calculates if this items current stock is less then the supplied
+   * threshold
+   * @return {bool} true if this item has total quantity of stock less than the threshold
+   */
+  get isLessThanThresholdMOS() {
+    return (
+      getTotal(this.item.realItem.batches, 'totalQuantity') <
```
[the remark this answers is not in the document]
joshxg: fixed 9ae6d28a4b4e6e62ef776c4902415bb172794cea

src/database/DataTypes/Period.js
```diff
@@ -5,19 +5,43 @@
...
+    ).length;
   }
 
   addRequisitionIfUnique(requisition) {
     if (this.requisitions.filtered('id == $0', requisition.id).length > 0) return;
     this.requisitions.push(requisition);
   }
+
+  toString() {
+    return `${this.startDate.toLocaleDateString('en-US')} - ${this.endDate.toLocaleDateString(
```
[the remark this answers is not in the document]
joshxg: On my machine it gives the same format. It will use the locale date and zone if no parameters are passed.
- Not sure if this should be consistent or something?
- Not sure if you're looking for a different format? (Oddly `toLocalDateString` is pretty customizable, but I can't find a way to change separators). Can make a method or use the other parsers we have?

src/database/DataTypes/MasterList.js
```diff
@@ -44,40 +54,39 @@ export class MasterList extends Realm.Object {
...
   /**
-   * Find a specifically named order type in the store tag object for this store and masterlist
+   * Find a specific order type by name field in the store tag object
+   * associated to this store and master list.
+   *
    * @param  {string}  tags            Current stores tags field
    * @param  {string}  orderTypeName   Name of the orderType to search for
    * @return {object} The matching orderType object
    */
   getOrderType(tags, orderTypeName) {
```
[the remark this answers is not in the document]
joshxg: fixed a89e3e7505290c0e2999e2b00a125deeb9d53cc3

src/database/DataTypes/Period.js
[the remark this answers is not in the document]
andreievg: I think omitting parameter her would be ideal, it should default to system date:

Below, running `node` in terminal

Screen Shot 2019-04-12 at 10 12 10 PM

src/pages/SupplierRequisitionsPage.js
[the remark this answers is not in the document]
andreievg: @joshxg 
sorry i didn't add a suggestions, getAllPrograms should return an array or null IMO.
This def looks a bit out of the ordinary `!!getAllPrograms(settings, database).length`
 
Thinking more deeply about this, we may we should add a 'coding convention'
* if a method, may return a 'fail' it should be 'null'
or
* method should always return type in header comment docs

Second one sounds like the way to go, but i don't fully agree with it. i.e.
* integer returning method, that fails to find and integer, we would have to use special integer (-1 or something ) to know that method failed, null as generic would be preferred

This is up for discussion but def need a change from `!!getAllPrograms(settings, database).length`

src/pages/SupplierRequisitionPage.js
```diff
@@ -37,10 +38,15 @@ const MODAL_KEYS = {
 export class SupplierRequisitionPage extends React.Component {
   constructor(props) {
     super(props);
+
+    const { requisition } = props;
```
andreievg: @joshxg 
introduced unused variable ? this should be enforced in pre commit eslint error right ?

src/pages/SupplierRequisitionsPage.js
andreievg: does it work with 'original' way of creating requisition (it might, but I don't see it immediately in code as state.isCreatingRequisition = false is not set as far as i can see, could be wrong)

src/pages/SupplierRequisitionsPage.js
andreievg: ahh, and getAllPrograms in ByProgramModal is called with settings, now it makes sense

src/database/DataTypes/Period.js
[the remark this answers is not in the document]
joshxg: done 2b9b707fc2635ca3a035716b75c97763ba36859b

src/pages/SupplierRequisitionsPage.js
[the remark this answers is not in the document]
joshxg: done : b81089e992be835526c01c2def469ae87324bd89 (I think?)

---

I'm not very consistent with return values, I need to be. 

Both sides of the argument have merit - I think in JS, it could be better to return null as it is somewhat easier to deal with than in other languages, but not sure.

The argument against your integer example is `Array.prototype.findIndex()`, and other various `findIndex`  methods in almost every language which returns `-1`.

These are not directly related being an OOP patterns: [Null-object pattern](https://en.wikipedia.org/wiki/Null_object_pattern) and [polymorphism over conditionals](https://refactoring.guru/replace-conditional-with-polymorphism) but somewhat relevant I think. Conclusions from these seem to argue for returning an object of the same type. I think quite different for JS though!

src/pages/SupplierRequisitionPage.js
joshxg: It's used on the next line

src/pages/SupplierRequisitionsPage.js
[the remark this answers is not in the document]
andreievg: > The argument against your integer example is Array.prototype.findIndex(), and other various findIndex methods in almost every language which returns -1 (Although one counter-example isn't great. Also I have always found it strange that that this returns -1).

I was never a fan of -1 returns btw.

> These are not directly related being an OOP patterns: Null-object pattern and polymorphism over conditionals but somewhat relevant I think. Conclusions from these seem to argue for returning an object of the same type. I think quite different for JS though!

Would be good to have a group discussion about thi

> done : b81089e (I think?)

looks good

src/pages/SupplierRequisitionPage.js
andreievg: oops ma bad

src/widgets/SequentialSteps.js
```diff
@@ -0,0 +1,159 @@
...
+
+  renderInputField = ({ canEdit, step }) => {
+    const { textContent, text } = localStyles;
+    const { name, placeholder } = step;
+    const textStyle = { ...text, color: canEdit ? 'white' : WARM_GREY, marginBottom: 10 };
+    const containerStyle = { borderBottomColor: 'white', borderBottomWidth: 1, ...textContent };
+    return (
+      <View style={containerStyle}>
+        <Text numberOfLines={1} ellipsizeMode="tail" style={textStyle}>
+          {name || placeholder}
```
andreievg: @joshxg, was wondering if we should display placeholder if !isLessThanCurrentStep

joshxg: I thought the same trying to think of the best way to give feedback to the user on what they should be doing. 

Currently there is an arrow pointing to the step which is the 'current step', and the sequential steps have a lower opacity. When steps are complete they get a tick and their opacity is kept high. Not completely happy, but thought this would be sufficient.

I also couldn't think of a good placeholder
- 'Please do the step above first' (3 times)
- 'Please do X step' 3 times 
- blank/nothing

Open to suggestions!

Keeping them visible also kinda gives feedback to what is 'next' to do

src/widgets/modals/ByProgramModal.js
```diff
@@ -0,0 +1,360 @@
...
+
+  getSequentialStepsProps = () => {
+    const { name } = this.state;
+
+    const getBaseProps = key => ({
+      name: this.state[key] && this.state[key].name,
+      placeholder: localization[`${key}Title`],
+      key,
+      error: this.state[`${key}s`] ? this.state[`${key}s`].length === 0 : false,
+      errorText: localization[`${key}Error`],
```
andreievg: `${key}s`, `${key}Title`, `${key}Error` is pretty crude, you may think so yourself.
Letting this be just a comment, not requirement, this maybe the only example of how sequential steps component is used, we should think about refactor when we have more time, and some headspace.

src/widgets/modals/ByProgramModal.js
```diff
@@ -0,0 +1,360 @@
...
+    else this.setState({ modalIsOpen: false });
+  };
+
+  getSequentialStepsProps = () => {
+    const { name } = this.state;
+
+    const getBaseProps = key => ({
+      name: this.state[key] && this.state[key].name,
+      placeholder: localization[`${key}Title`],
+      key,
```
andreievg: is key only needed because we use indexOf instead of object.entries, then indexOf in sequential step modal ?

src/widgets/modals/ByProgramModal.js
```diff
@@ -0,0 +1,360 @@
...
+import { PageContentModal } from './PageContentModal';
+
+import globalStyles, { DARK_GREY, WARM_GREY, SUSSOL_ORANGE } from '../../globalStyles';
+import { AutocompleteSelector, ToggleBar, PageButton, TextEditor } from '..';
+
+import { SETTINGS_KEYS } from '../../settings';
+import { getAllPrograms, getAllPeriodsForProgram } from '../../utilities/byProgram';
+import SequentialSteps from '../SequentialSteps';
+
+const localization = {
```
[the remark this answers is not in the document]
joshxg: import: 65303ce5bcfa5505b2fa25d3764754659af7b196
TODO: 1f24e592df0699948140aa2d4cbb6d4f45c16a74

src/widgets/modals/ByProgramModal.js
```diff
@@ -0,0 +1,360 @@
...
+    this.state = {
+      // Values used for deriving current UI state
+      steps: null,
+      modalIsOpen: false,
+      isProgramBased: true,
+      currentStepKey: null,
+      // New requisition or stocktake values
+      program: null,
+      supplier: null,
+      orderType: null,
```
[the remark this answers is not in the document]
joshxg: I'd argue that this also works as 'docs' for the component. Showing the full explicit shape of the state used by a component? (It does look ugly though :( )

src/widgets/modals/ByProgramModal.js
```diff
@@ -0,0 +1,360 @@
...
+  }
+
+  // Store unchanging select modal options and set the default
+  // steps array.
+  componentDidMount = () => {
+    const { settings, database } = this.props;
+    const programs = getAllPrograms(settings, database);
+    const suppliers = database.objects('Name').filtered('isSupplier = $0', true);
+    const steps = this.setCurrentSteps();
+    this.setState({ programs, suppliers, currentStepKey: steps[0].key });
```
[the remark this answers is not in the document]
joshxg: Oops 3a265e82761b059959df9c6c0c4791319b7166bf

src/widgets/modals/ByProgramModal.js
```diff
@@ -0,0 +1,360 @@
...
+  };
+
+  onConfirmRequisition = () => {
+    const { onConfirm } = this.props;
+    const { supplier: otherStoreName, orderType } = this.state;
+    onConfirm({ ...this.state, otherStoreName, orderType: orderType && orderType.name });
+    this.setState({ ...newState.RESET_ALL }, () => this.setCurrentSteps());
```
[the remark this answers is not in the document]
joshxg: Hm, leftover from before the modal was unmounted on `modalOpen` 421b34c9b0704a0220772af240221fb976c1e071

src/widgets/modals/ByProgramModal.js
```diff
@@ -0,0 +1,360 @@
...
+          style={localStyles.okButton}
+          textStyle={{ ...globalStyles.buttonText, color: 'white' }}
+        />
+
+        {modalIsOpen && (
+          <PageContentModal
+            isOpen={modalIsOpen}
+            onClose={this.onModalTransition}
+            title={localization[`${currentStepKey}Title`]}
+            coverScreen
```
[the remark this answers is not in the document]
joshxg: full screen (above nav bar etc)

src/widgets/modals/ByProgramModal.js
joshxg: had to 5ee672eca0a586924679acdb5fd39e4def8c5942

src/widgets/modals/ByProgramModal.js
```diff
@@ -0,0 +1,360 @@
...
+    const { isProgramBased } = this.state;
+    this.setState({ ...newState.RESET_ALL, isProgramBased: !isProgramBased }, () =>
+      this.setCurrentSteps(true)
+    );
+  };
+
+  // Open a modal - in case it is an earlier step in the process,
+  // set the current step to the step number corresponding with
+  // the onPress event.
+  onModalTransition = ({ key: currentStepKey }) => {
```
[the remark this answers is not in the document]
joshxg: done 9632c845dba3297bcc90cb042c2ec5d24b686b8a

src/widgets/modals/ByProgramModal.js
joshxg: Not 100% on what you mean. Key is used mainly for determining which step is the current step and to pass to the `onPress` to open a particular modal

Chris-Petty:
In a store not using these changes (i.e. had no store tags) if you closed the new requisition dialog using the 'X' at the top right, it'd still make a requisition but with no supplier selected.

Chris-Petty:
Program bug:
- I can select my own store as a supplier

Chris-Petty:
So like this PR is pretty meaty now. @andreievg @joshxg we going _all the way_?

PR #857, opened 2019-04-13
#856 generic choose modal

### LINKED ISSUE #856: Generic Choose Modal
We need a generic component which can be used as a drop down - similar to the language modal, but more generic, usable by anything

joshxg:
Fixes #856 

- Adds `GenericChooseModal` component
- Updates `AutoCompleteSelector` to allow for arrays and empty arrays

### DIFF
--- a/src/widgets/modals/GenericChooseModal.js
+++ b/src/widgets/modals/GenericChooseModal.js
@@ -19,7 +19,7 @@ import { COMPONENT_HEIGHT } from '../../globalStyles';
  * which was selected, the index/row within the data array and
  * the field which was displayed to the user.
  * @prop  {array}   data            Array of objects from which a choice must be made from
- * @prop  {string}  field           The object field which should be displayed to the user
+ * @prop  {string}  keyToDisplay    The object property which should be displayed to the user
  * @prop  {string}  title           The title of the modal
  * @prop  {func}    onPress         Function to call on selection returns {item, index, field}
  * @prop  {boolean} isOpen          Indicator for if the modal should be open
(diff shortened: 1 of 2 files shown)


src/widgets/modals/GenericChooseModal.js
```diff
@@ -0,0 +1,97 @@
...
+ * @prop  {array}   data            Array of objects from which a choice must be made from
+ * @prop  {string}  field           The object field which should be displayed to the user
+ * @prop  {string}  title           The title of the modal
+ * @prop  {func}    onPress         Function to call on selection returns {item, index, field}
+ * @prop  {boolean} isOpen          Indicator for if the modal should be open
+ * @prop  {int}     highlightIndex  Index of the data array for which object should be 'highlighted'
+ */
+export default class GenericChooseModal extends React.PureComponent {
+  keyExtractor = (item, index) => {
+    const { field } = this.props;
```
andreievg: not in jdocs

andreievg: and maybe should be `keyToShow` ?

joshxg: done c126e3107de90964228d7d724e4a0c8a3096b378

src/widgets/modals/GenericChooseModal.js
```diff
@@ -0,0 +1,97 @@
...
+    const content = field && item ? item[field] : item;
+    return `${content}${index}`;
+  };
+
+  renderRow = ({ item, index }) => {
+    const { onPress, field, highlightIndex } = this.props;
+    const { row, text } = localStyles;
+
+    const indexToHighlight = highlightIndex >= 0 ? highlightIndex : 0;
```
andreievg: keyToHighlight and indexToHighlight ?
```
let shouldHighlight = false;
if(keyToDisplay && valueToHighlight) shouldHighlight= item[keyToDisplay] === valueToHighlight 
else if (indexToHighlight)  index === indexToHighlight // null !== 0, undefined !== 0
```

joshxg: done 2e47c9604fb499646169a7b3497724fcda948086

Defaults to highlighting the first row if no `highlightIndex` or `highlightValue` is used. 

This will highlight multiple rows if there are rows with the same value. I think that this is fine to leave as the responsibility of the caller as there are other problems with having multiple same values in a drop down

PR #860, opened 2019-04-14
#859 transaction batch missing getter

### LINKED ISSUE #859: [Vaccine Module] Broken TransactionBatch.splitBatch() method broken
Description: `TransactionBatch.splitBatch()` used a getter `get transactionItem`. I removed this because I thought I wasn't using it, but I was.

Comments: Currently on `vaccine-development` dose splitting just freezes

joshxg:
Fixes #859 

Just adding a getter back that I removed.

Note on safeness:
- Realm list properties are always defined
- Every object must have an id
- array method is safe to run on empty lists

### DIFF
--- a/src/database/DataTypes/TransactionBatch.js
+++ b/src/database/DataTypes/TransactionBatch.js
@@ -169,6 +169,14 @@ export class TransactionBatch extends Realm.Object {
     return `${this.itemBatch} in a ${this.transaction.type}`;
   }
 
+  get transactionItem() {
+    const { transaction } = this;
+    return transaction.items.find(transactionItem => {
+      const { batches } = transactionItem;
+      return batches.some(batch => batch.id === this.id);
+    });
+  }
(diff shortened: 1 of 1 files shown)


src/database/DataTypes/TransactionBatch.js
```diff
@@ -169,6 +169,11 @@ export class TransactionBatch extends Realm.Object {
     return `${this.itemBatch} in a ${this.transaction.type}`;
   }
 
+  get transactionItem() {
+    const { transaction } = this;
+    return transaction.items.find(({ batches }) => batches.some(batch => batch.id === this.id));
```
andreievg: I think this works:
transaction.items.filtered('batches.id = $0', this.id)

joshxg: couldn't get that query to work. Would be wicked if it did though

PR #866, opened 2019-04-14
#865 class helper methods

### LINKED ISSUE #865: [Vaccine Module] Item/ItemBatch Helper methods
Description:

Need quite a few helper methods in general for VM, but specifically for #858 

Helper methods:

- Number of batches in a location
- Quantity of stock in a location
- Quantity of stock in a breach, in a location
- Temperature exposure in a location
- If an item has an ItemBatch that is in a breach.

Comments:

New issue for this so the review is easier!

joshxg:
Fixes #865 

- Some of these could be slightly more efficient (e.g. `getBreachedBatches` could filter on location before breach status) for uglier code
- Some may be unnecessary e.g just getting the length. Personally I like them as they are used a bit, but I can see the argument for them not being useful.
- Not sure about the names
- The two removed getters were duplicates

### DIFF
--- a/src/database/DataTypes/Item.js
+++ b/src/database/DataTypes/Item.js
@@ -223,9 +223,9 @@ export class Item extends Realm.Object {
    * are returned.
    * @param {Location} location
    */
-  getBatchesInLocation({ id } = {}) {
-    if (!id) return this.batches;
-    return this.batches.filtered('location.id = $0', id);
+  getBatchesInLocation({ id: locationId } = {}) {
+    if (!locationId) return this.batches;
+    return this.batches.filtered('location.id = $0', locationId);
   }
 
   /**
@@ -257,10 +257,10 @@ export class Item extends Realm.Object {
    * in breaches are returned.
    * @param {Location} location
    */
-  getBreachedBatches({ id } = {}) {
+  getBreachedBatches({ id: locationId } = {}) {
     const breachedBatches = this.batches.filter(({ hasBreached } = {}) => hasBreached);
-    if (!id) return breachedBatches;
-    return breachedBatches.filter(({ location } = {}) => location && location.id === id);
+    if (!locationId) return breachedBatches;
+    return breachedBatches.filter(({ location } = {}) => location && location.id === locationId);
   }
 
   /**
--- a/src/database/DataTypes/Item.js
+++ b/src/database/DataTypes/Item.js
@@ -245,10 +245,7 @@ export class Item extends Realm.Object {
    * @param {Location} location
    */
   getQuantityInLocation(location) {
-    return this.getBatchesInLocation(location).reduce(
-      (sum, { numberOfPacks = 0, packSize = 1 } = {}) => sum + numberOfPacks * packSize,
-      0
-    );
+    return getTotal(this.getBatchesInLocation(location), 'totalQuantity');
   }
 
   /**
@@ -280,10 +277,7 @@ export class Item extends Realm.Object {
    * @param {Location} location
    */
   getQuantityInBreach(location) {
-    return this.getBreachedBatches(location).reduce(
-      (sum, { numberOfPacks = 0, packSize = 1 } = {}) => sum + numberOfPacks * packSize,
-      0
-    );
+    return getTotal(this.getBreachedBatches(location), 'totalQuantity');
   }
 
   /**
--- a/src/database/DataTypes/Item.js
+++ b/src/database/DataTypes/Item.js
@@ -287,16 +287,18 @@ export class Item extends Realm.Object {
   getTemperatureExposure(location) {
     let { batches } = this;
     if (location) batches = this.getBatchesInLocation(location);
+
     const temperatures = batches.map(({ temperatureExposure } = {}) => temperatureExposure);
-    const maxTemperatures = temperatures.reduce(
-      (maxTemp, { max = 0 } = {}) => Math.max(maxTemp, max),
+
+    const maxTemperature = temperatures.reduce(
+      (maxTemp, { maxTemperature: max = -Infinity } = {}) => Math.max(maxTemp, max),
       0
     );
-    const minTemperatures = temperatures.reduce(
-      (minTemp, { min = Infinity } = {}) => Math.min(minTemp, min),
+    const minTemperature = temperatures.reduce(
+      (minTemp, { minTemperature: min = Infinity } = {}) => Math.min(minTemp, min),
       Infinity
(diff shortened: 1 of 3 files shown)


src/database/DataTypes/Item.js
```diff
@@ -224,6 +216,96 @@ export class Item extends Realm.Object {
...
     return `${this.code} - ${this.name}`;
   }
+
+  /**
+   * Finds all this items batches in a specific location.
+   * If no location is passed, all of this items ItemBatches
+   * are returned.
+   * @param {Location} location
+   */
+  getBatchesInLocation({ id } = {}) {
```
andreievg: id: locationId ? (i know the jdocs say it all, but seems more appropriate (not required for PR approval though

src/database/DataTypes/Item.js
```diff
@@ -224,6 +216,96 @@ export class Item extends Realm.Object {
...
+
+  /**
+   * Returns the sum of each ItemBatch.totalQuantity in a specific
+   * location. If no location is passed, finds the totalQuantity
+   * for all batches.
+   * @param {Location} location
+   */
+  getQuantityInLocation(location) {
+    return this.getBatchesInLocation(location).reduce(
+      (sum, { numberOfPacks = 0, packSize = 1 } = {}) => sum + numberOfPacks * packSize,
```
andreievg: itemBatch does have totalQuantity
can use getTotal utility method getTotal(this.getBatchesInLocation(location), 'totalQuantity').

or this.getBatchesInLocation(location).sum('numberOfPacks') <- we enforce packSize of 1 in mobile

src/database/DataTypes/Item.js
```diff
@@ -224,6 +216,96 @@ export class Item extends Realm.Object {
...
+  }
+
+  /**
+   * Returns all ItemBatches which have been in a breach in a specific
+   * location. If no location is passed, all of this items ItemBatches
+   * in breaches are returned.
+   * @param {Location} location
+   */
+  getBreachedBatches({ id } = {}) {
+    const breachedBatches = this.batches.filter(({ hasBreached } = {}) => hasBreached);
```
andreievg: do't need { hasBreached } `= {}`, this.batches should only contain realm objects.

by the way, can you not use getBatchesInLocation(location).filter ...

src/database/DataTypes/Item.js
```diff
@@ -224,6 +216,96 @@ export class Item extends Realm.Object {
...
+  }
+
+  /**
+   * Returns the sum of all ItemBatches in breaches related to this item.
+   * If no location is passed, returns the sum for all ItemBatches related
+   * to this item.
+   * @param {Location} location
+   */
+  getQuantityInBreach(location) {
+    return this.getBreachedBatches(location).reduce(
```
andreievg: once again the getTotal would be cleaner here

src/database/DataTypes/Item.js
```diff
@@ -224,6 +216,96 @@ export class Item extends Realm.Object {
...
+   * related to this Item in a specific location. If no location is passed,
+   * uses all ItemBatches related to this item to find the values.
+   * @param {Location} location
+   */
+  getTemperatureExposure(location) {
+    let { batches } = this;
+    if (location) batches = this.getBatchesInLocation(location);
+    const temperatures = batches.map(({ temperatureExposure } = {}) => temperatureExposure);
+    const maxTemperatures = temperatures.reduce(
+      (maxTemp, { max = 0 } = {}) => Math.max(maxTemp, max),
```
andreievg: temperatureExposure returns: maxTemperature, minTemperature but destructured to max and min.

can do -Infinity instead of max = 0.

something to be mindful, this method could return infinities batches.length = 0 or no sensorLogs for any of the bathes, what should we return if not temperatures ? Btw what does realmLst.max and min return if there are not object in the list

return has `minTemperatures, maxTemperatures`, plural ?

src/database/DataTypes/ItemBatch.js
```diff
@@ -125,6 +125,31 @@ export class ItemBatch extends Realm.Object {
...
+  get hasBreached() {
+    return this.breaches.length > 0;
+  }
+
+  /**
+   * Returns an object {maxTemperature, minTemperature} for all
+   * recorded temperatures for this ItemBatch.
+   */
+  get temperatureExposure() {
+    const maxTemperature = this.sensorLogs.max('temperature');
```
andreievg: what would this return if nothing in sensorLogs ?

src/database/DataTypes/Item.js
joshxg: done 7629c92aa4f52605295b355a311f7885d3ab49a1

Still trying to find my balance between explicit and implicit code, how clear etc.

src/database/DataTypes/Item.js
joshxg: done d4e4cce3f84bf0e9a9e36da471d51ac2c18c6995

src/database/DataTypes/Item.js
joshxg: done 7a71e8d33c0263d93c30efd490959976131d834c

src/database/DataTypes/Item.js
joshxg: done d4e4cce3f84bf0e9a9e36da471d51ac2c18c6995

src/database/DataTypes/ItemBatch.js
joshxg: hmm, it returns undefined. I have set the default case to be  `{ maxTemperature: -Infinity, minTemperature: Infinity }` as above.

src/database/DataTypes/Item.js
joshxg: I think it's correct to return `{ maxTemperature: -Infinity, minTemperature: Infinity }` on incorrect cases, although I do see the benefit of returning null, also. 

-- Also: `minTemperature` or `min`

andreievg: @joshxg, `minTemperature` or `min`, not sure minTemperature sounds like a safe way

src/database/DataTypes/Item.js
andreievg: @joshxg the only reason i mentioned { locationId } is because of ambiguity of 'id' (all record have it)

src/database/DataTypes/Item.js
andreievg: -Infinity and +Infinity sounds all good to me as a failure result

Issue #867, opened 2019-04-14
[Vaccine Module][Menu Page] - customisable menu page through store custom_data


andreievg:
From epic: https://github.com/sussol/org-issues/issues/23

#### Description:

Need to customise MenuPage to show modules when one of the modules is enabled (starting with vaccineModule).

Screenshot from Mike on how he sees it looking:
![1822798F702E421F90168AA92AB6026F-1](https://user-images.githubusercontent.com/26194949/56091057-0af45680-5efe-11e9-9eaf-6d6ed9c7714a.png)

PR #868, opened 2019-04-14
#867 add customisations to menu page

andreievg:
Fixes: #867 

This branch will only work (sync wise) when https://github.com/openmsupply/mobile/pull/864 is merged into vaccine-development and then this branch is rebased, can still check it working if you are not syncing Location or LocationType. -> just set as per below screenshot in store[input] of the mobile store then then restart app;

Screen Shot 2019-04-14 at 8 41 14 PM

with useVaccineModule == "true" (below)
Screen Shot 2019-04-14 at 8 40 41 PM

otherwise:

Screen Shot 2019-04-14 at 8 42 28 PM

Export and Explorer buttons are still only showing when in admin mode (just wanted to show where they sit now)


src/pages/MenuPage.js
```diff
@@ -15,7 +15,47 @@ import { SETTINGS_KEYS } from '../settings';
 
 import globalStyles, { APP_FONT_FAMILY, SHADOW_BORDER, GREY, WARMER_GREY } from '../globalStyles';
 
-const { SYNC_SITE_NAME } = SETTINGS_KEYS;
+/* eslint-disable global-require */
+
+const { SYNC_SITE_NAME, THIS_STORE_CUSTOM_DATA } = SETTINGS_KEYS;
```
joshxg: I'm not sure about holding store custom data. I can see why we are storing this data in `custom_data` in 4D. I don't see why bringing this into mobile is a good idea. 

My ideal:
`THIS_STORE_MODULES: ['VACCINES', 'DISPENSARY']`

Although I can see there could *potentially* be a use-case for having additional information about a certain stores modules.

`THIS_STORE_MODULES: { VACCINE: { }, DISPENSARY: { } }`

--

Keep with the current way: I think:
`THIS_STORE_CUSTOM_DATA: { modules: ['VACCINE'] }`
or 
`THIS_STORE_CUSTOM_DATA: { vaccine: { enabled : true }}`

The current shape/names seem a little odd

src/pages/MenuPage.js
```diff
@@ -24,7 +64,24 @@ export class MenuPage extends React.Component {
...
+
+    if (!usesModules) {
+      this.layout = ORIGINAL_LAYOUT;
+      this.imageStyle = localStyles.originalImage;
+      this.sectionStyle = localStyles.section;
+    } else {
+      this.layout = VACCINE_MODULE_LAYOUT;
+      this.imageStyle = localStyles.VM_Image;
+      // Image will go on the left side as layout changes when modules are enabled
+      this.sectionStyle = { ...localStyles.section, flexDirection: 'row' };
```
joshxg: Am liking having `localStyles` be a function. Pass it values which can determine what styles are retrieved.  Abstracts a lot of style logic out

src/pages/MenuPage.js
```diff
@@ -38,96 +95,66 @@ export class MenuPage extends React.Component {
...
   }
 
+  /**
+   * Method will try to match stringified customData object for current store
+   * against keys in MODULES, will return values of keys in MODULES
+   * that match: customData[keyInModules]data === 'true'
+   * @param {string}  customDataAsString  Store custom data as string
+   * @return {array}  result              The maxOrdersPerPeriod of the orderType
+   */
+  getUsedModules = customDataAsString => {
```
joshxg: Do you think this would ever be used elsewhere? If so maybe a utility method? (would require some reshuffling. Maybe having the method return a string of modules, iterate through afterward adding the button from `MODULES` to `modules`.

src/pages/MenuPage.js
```diff
@@ -38,96 +95,66 @@ export class MenuPage extends React.Component {
...
+    if (!customDataAsString || customDataAsString.length === 0) return [];
```
joshxg: `''` is falsey already

src/pages/MenuPage.js
```diff
@@ -38,96 +95,66 @@ export class MenuPage extends React.Component {
...
+
+    const customData = JSON.parse(customDataAsString);
+    const result = [];
+    const checkAndAddToResult = ([moduleName, moduleValues]) => {
+      if (!customData[moduleName]) return;
+      if (!customData[moduleName].data) return;
+      if (customData[moduleName].data === 'true') result.push(moduleValues);
```
joshxg: Seems excessive. Checking if `customData[moduleName]` is falsey, and then checking if it's equal to a specific string.

Could also `.reduce()` and append to the accumulator with an `initialValue = []` rather than pushing

src/pages/MenuPage.js
```diff
@@ -38,96 +95,66 @@ export class MenuPage extends React.Component {
...
+    };
+
+    Object.entries(MODULES).forEach(checkAndAddToResult);
+
+    return result.length > 0 && result;
```
joshxg: Inconsistent returns. If `result = []`, we return `false` and then the caller above calls `false.length`. Still undefined and falsey so fine but unintended?

src/pages/MenuPage.js
```diff
@@ -38,96 +95,66 @@ export class MenuPage extends React.Component {
...
     const syncSiteName = settings.get(SYNC_SITE_NAME);
     database.exportData(syncSiteName);
   };
 
+  renderButton = ({ title, page }) => (
+    <Button
+      style={globalStyles.menuButton}
+      textStyle={globalStyles.menuButtonText}
+      text={title}
+      // eslint-disable-next-line react/destructuring-assignment
```
joshxg: don't like destructuring :(?

src/pages/MenuPage.js
```diff
@@ -38,96 +95,66 @@ export class MenuPage extends React.Component {
...
+    />
+  );
+
+  renderSection = sectionName => {
+    const { imageStyle, sectionStyle } = this;
+    const section = this.sections[sectionName];
+    return (
+      <View style={sectionStyle}>
+        <Image style={imageStyle} resizeMode="contain" source={section.icon} />
+        <View>{section.buttons.map(this.renderButton)}</View>
```
joshxg: `this.sections.modules.buttons`  is assigned `undefined` above if `result = []`. Seems fine as it can only be `ORIGINAL_LAYOUT` though?

src/pages/MenuPage.js
```diff
@@ -24,7 +64,24 @@ export class MenuPage extends React.Component {
...
 
   componentWillMount() {
-    const { database } = this.props;
+    const { database, settings } = this.props;
+
+    this.sections = SECTIONS;
+    // check is there are any modules enabled in this store customData
+    const usableModules = this.getUsedModules(settings.get(THIS_STORE_CUSTOM_DATA));
+    const usesModules = usableModules.length > 0;
+    this.sections.modules.buttons = usableModules;
```
joshxg: Would be good to have these initialized to null in the constructor. I think it makes it a lot clearer rather than random class scoped variables popping up

src/pages/MenuPage.js
Chris-Petty: Seems odd that the property 'data' would be the string 'true' anyway.

Chris-Petty:
Not doing proper review, @joshxg's is solid. I do like the modules layout.

src/pages/MenuPage.js
andreievg: it does look neater: https://github.com/openmsupply/mobile/pull/868/commits/cdc1d9034ff584383b6407ad421f324e2a2bf9f2, is this what you meant ?

src/pages/MenuPage.js
andreievg: what if it's null/undefined though ? using it here to safely check length

src/pages/MenuPage.js
andreievg: removed excessive checking https://github.com/openmsupply/mobile/pull/868/commits/cdc1d9034ff584383b6407ad421f324e2a2bf9f2

I think the syntax of for each and 'external' array is much neater

src/pages/MenuPage.js
andreievg: fixed: https://github.com/openmsupply/mobile/pull/868/commits/cdc1d9034ff584383b6407ad421f324e2a2bf9f2

src/pages/MenuPage.js
andreievg: not sure if i like this if style getter is called frequently (i.e. def not for a table cell)

src/pages/MenuPage.js
andreievg: fixed: https://github.com/openmsupply/mobile/commit/cdc1d9034ff584383b6407ad421f324e2a2bf9f2

src/pages/MenuPage.js
andreievg: was trying to avoid making that method have `return` statement, but ended up doing it anyways: destructure

src/pages/MenuPage.js
andreievg: done: https://github.com/openmsupply/mobile/commit/cdc1d9034ff584383b6407ad421f324e2a2bf9f2

src/pages/MenuPage.js
andreievg: good point, i'll make an issue

andreievg:
@joshxg, I've made the requested changes, will make an issue for how we handle custom data

src/pages/MenuPage.js
andreievg: i'll try to deal with this also in that issue, will create it later today

src/pages/MenuPage.js
joshxg: Sorry what I meant was

```
const {someStyle} = localStyles({someProps})

const localStyles = ({someValueWhichAltersStyles}) => StyleSheet.create({
    someStyle: {
        color: someValueWhichAltersStyles ? 'white' : 'black'
    }
})
```

But they way you've done it is nice.

I'd also argue that doing it the above way, a simple function returning a value and evaluating simple conditionals could be faster than creating and mutating complex objects

src/pages/MenuPage.js
joshxg: I mean `!customDataAsString` covers all three but this is more expliciti

src/pages/MenuPage.js
```diff
@@ -38,107 +86,109 @@ export class MenuPage extends React.Component {
...
-                textStyle={globalStyles.menuButtonText}
-                text="Realm Explorer"
-                onPress={() => navigateTo('realmExplorer', 'Database Contents')}
-              />
-            )}
-          </View>
-        </View>
-        <View style={globalStyles.bottomContainer}>
+      <View style={pageContentContainer}>
+        {this.layout.map(row => (
```
joshxg: need a `key` property on all of these `.map()` rendering I think? (how do you feel about using `generateUUID` for `key` props?)

src/pages/MenuPage.js
andreievg: @Chris-Petty that's the format custom data comes in as from mSupply primary

src/pages/MenuPage.js
andreievg: good point, fixed: https://github.com/openmsupply/mobile/pull/868/commits/166d3695533183bb3e28133c994b054d7510490e

src/pages/MenuPage.js
andreievg: ohh true, i completely missed that one, the only way this would fail if customDataAsString is a true boolean or a number, which should never be the case, fixed: https://github.com/openmsupply/mobile/pull/868/commits/ece8f366218fe2d66debe189f0ed3247d0564a1b

src/pages/MenuPage.js
andreievg: https://github.com/openmsupply/mobile/issues/880

src/pages/MenuPage.js
andreievg: @joshxg, sorry I don't quite understand, how would your method deal with localStyles.section ? 
i.e. use the full localStyle.section but use flexDirection as row ?

I do see the use case for it with VM_Image, where we just want hight and width changed.

By the way, i don't see how the original way (which i think was the quickest way) and the get style mutate any object, they just create new objects by destructing localStyle and replacing optional fields

src/pages/MenuPage.js
andreievg: https://github.com/openmsupply/mobile/issues/880

andreievg:
@joshxg @Chris-Petty looks like I've dealt with all of the requested changes.

src/pages/MenuPage.js
joshxg: Sorry originally it was more of just how I like to use `localStyles` as an actual function that abstracts away a lot of the style logic.

That's true about objects, just I think we're talking about tiny processor performance increases at this point, in which case dealing with primitive data types are pretty much always going to be faster than creating complex objects? 

Anyway this is more just off topic sorry

PR #872, opened 2019-04-16
Simple table component

### LINKED ISSUE #871: [Vaccine Module] Simple table component
## Description

Require a simple, responsive table which renders simple scalar values.

## Requirements

- Uses FlatList to be a little more performant
- Highly responsive
- Optional title
- Similar look and feel to the current data tables.

joshxg:
Fixes #871 

Just a simple table component. Can change colours etc.

Also adds a 'StackedTables' component which evenly spreads `SimpleTable` components. Could be overkill.

image

### DIFF
--- a/src/widgets/SimpleTable.js
+++ b/src/widgets/SimpleTable.js
@@ -10,7 +10,6 @@ import PropTypes from 'prop-types';
 
 import { StyleSheet, FlatList, View, Text } from 'react-native';
 import { ROW_BLUE, APP_FONT_FAMILY, SUSSOL_ORANGE } from '../globalStyles/index';
-import { generateUUID } from '../database';
 
 /**
  * Simple table component only rendering scalar values in a cell.
@@ -23,7 +22,7 @@ import { generateUUID } from '../database';
  * title: Title of the table.
  */
 export class SimpleTable extends React.PureComponent {
-  keyExtractor = () => generateUUID();
+  keyExtractor = ({ index }) => index;
 
   renderTitle = () => {
     const { title } = this.props;
@@ -38,7 +37,7 @@ export class SimpleTable extends React.PureComponent {
   renderCell = ({ content, index, header, width }) => {
     const { cell, cellFont } = localStyles({ ...this.props, index, header, width });
     return (
-      <View style={cell} key={this.keyExtractor()}>
+      <View style={cell} key={this.keyExtractor({ index })}>
         <Text style={cellFont}>{content}</Text>
       </View>
     );
--- a/src/widgets/SimpleTable.js
+++ b/src/widgets/SimpleTable.js
@@ -34,10 +34,10 @@ export class SimpleTable extends React.PureComponent {
     );
   };
 
-  renderCell = ({ content, index, header, width }) => {
-    const { cell, cellFont } = localStyles({ ...this.props, index, header, width });
+  renderCell = ({ content, rowIndex, header, width, cellIndex }) => {
+    const { cell, cellFont } = localStyles({ ...this.props, rowIndex, header, width });
     return (
-      <View style={cell} key={this.keyExtractor({ index })}>
+      <View style={cell} key={this.keyExtractor({ index: cellIndex })}>
         <Text style={cellFont}>{content}</Text>
       </View>
     );
@@ -56,15 +56,15 @@ export class SimpleTable extends React.PureComponent {
     );
   };
 
-  renderRow = ({ item, index }) => {
+  renderRow = ({ item, index: rowIndex }) => {
     const { columns } = this.props;
     const { row } = localStyles(this.props);
     return (
       <View style={row}>
-        {columns.map(column => {
+        {columns.map((column, cellIndex) => {
           const { key, width } = column;
           const content = item[key];
-          return this.renderCell({ content, column, index, width });
+          return this.renderCell({ content, column, width, cellIndex, rowIndex });
         })}
       </View>
     );
@@ -88,7 +88,7 @@ export class SimpleTable extends React.PureComponent {
 }
 
 const localStyles = ({
-  index,
+  rowIndex,
   header,
   width,
   cellFontColor,
@@ -151,7 +151,7 @@ const localStyles = ({
       alignItems: 'center',
       borderRightColor: columnSeperatorColor,
       borderRightWidth: 2,
-      backgroundColor: index % 2 === 0 ? ROW_BLUE : 'white',
+      backgroundColor: rowIndex % 2 === 0 ? ROW_BLUE : 'white',
     },
   });
 
--- a/src/widgets/SimpleTable.js
+++ b/src/widgets/SimpleTable.js
@@ -71,13 +71,17 @@ export class SimpleTable extends React.PureComponent {
   };
 
   render() {
-    const { data, title } = this.props;
+    const { data, title, keyExtractor } = this.props;
     const { container } = localStyles(this.props);
     return (
       <View style={container}>
         {title && this.renderTitle()}
         {this.renderHeaderRow()}
-        <FlatList data={data} renderItem={this.renderRow} keyExtractor={this.keyExtractor} />
+        <FlatList
+          data={data}
+          renderItem={this.renderRow}
+          keyExtractor={keyExtractor || this.keyExtractor}
+        />
       </View>
     );
   }
@@ -164,6 +168,7 @@ SimpleTable.defaultProps = {
   cellFontColor: 'black',
   backgroundColor: 'white',
   columnSeperatorColor: '#ecf3fc',
+  keyExtractor: null,
 };
 
 SimpleTable.propTypes = {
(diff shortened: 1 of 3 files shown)


Chris-Petty:
Stacked tables seems to be quite similar to https://facebook.github.io/react-native/docs/sectionlist

src/widgets/SimpleTable.js
```diff
@@ -0,0 +1,187 @@
...
+      <View style={titleContainer}>
+        <Text style={titleFont}>{title}</Text>
+      </View>
+    );
+  };
+
+  renderCell = ({ content, index, header, width }) => {
+    const { cell, cellFont } = localStyles({ ...this.props, index, header, width });
+    return (
+      <View style={cell} key={this.keyExtractor()}>
```
Chris-Petty: Often in this context you can be more basic and get away with ```key={index}```. Unless the cell order might change. At that point a better thought out keyExtractor is probably sensible (and best chance through props). UUID is still not the right approach, as it'll be different every time you render the cells.

As explained [here](https://reactjs.org/docs/lists-and-keys.html#keys), they keys only have to be unique among its siblings in the DOM tree, and they should be stable if possible. If the key is the same as some parent row it shouldn't matter, so long as it's not the same as another cell in the same row.

src/widgets/SimpleTable.js
```diff
@@ -0,0 +1,187 @@
...
+
+/**
+ * Simple table component only rendering scalar values in a cell.
+ * Required props
+ * data = [ { a: 1, b: 2}, {a: 3, b: 4}, .. ]
+ * columns = [
+ * { key: 'a', width: 1, title: 'A COLUMN' },
+ * { key: 'b', width: 1, title: 'B COLUMN' } ..
+ * ]
+ * title: Title of the table.
```
Chris-Petty: Bit of a nit: This comment block should follow the format of the rest of the codebase.

src/widgets/StackedTables.js
```diff
@@ -0,0 +1,65 @@
...
+ * mSupply Mobile
+ * Sustainable Solutions (NZ) Ltd. 2019
+ */
+
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { StyleSheet, View } from 'react-native';
+
+import MiniTable from './SimpleTable';
```
Chris-Petty: Why not just import it as SimpleTable? I haven't done ES6 for so long I can't remember if you have to do more than this to alias a default import lol.

src/widgets/StackedTables.js
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * Simple component which renders SimpleTable components
+ * on top one one another. By default will use 100% of
+ * the width and height of it's parent. To give space
+ * between each table, the tablesHeight prop is the space
+ * all tables should occupy, leaving room between each if
+ * < 100%. i.e : 90% will occupy 90% of the height and use
+ * flexbox space-between, so 10% of the height is split evenly
+ * between each table.
```
Chris-Petty: Not following format of other doc blocks

src/widgets/SimpleTable.js
```diff
@@ -0,0 +1,187 @@
+/* eslint-disable react/no-unused-prop-types */
```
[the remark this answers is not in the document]
joshxg: I use all the props, I just pass them to a method (`localStyles`) which is not in the class. I can see how this is different from normal use and can revert all the styles if needed ( I just like this as it makes the component so much clearer without having all the styles logic ).

src/widgets/SimpleTable.js
```diff
@@ -0,0 +1,187 @@
...
+ * Required props
+ * data = [ { a: 1, b: 2}, {a: 3, b: 4}, .. ]
+ * columns = [
+ * { key: 'a', width: 1, title: 'A COLUMN' },
+ * { key: 'b', width: 1, title: 'B COLUMN' } ..
+ * ]
+ * title: Title of the table.
+ */
+export class SimpleTable extends React.PureComponent {
+  keyExtractor = () => generateUUID();
```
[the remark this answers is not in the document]
joshxg: done 00e5077acff17e5acf6a0ba9f9540d6c54cae9b9

src/widgets/SimpleTable.js
joshxg: done here https://github.com/openmsupply/mobile/pull/872#discussion_r276085908

src/widgets/SimpleTable.js
joshxg: done 58b5855d83a44af0d710c3b6e03effc6f94cb806

src/widgets/StackedTables.js
joshxg: Mistake, leftover from renaming and messy dev branch not trying to do anything tricky. Done 3486dd3a76dcdd68eefe62a2aabcd92d980bba34

src/widgets/StackedTables.js
joshxg: done 58b5855d83a44af0d710c3b6e03effc6f94cb806

src/widgets/SimpleTable.js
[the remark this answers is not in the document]
andreievg: i don't get this linter error in vscode, does it come up when you try to commit ?

src/widgets/SimpleTable.js
```diff
@@ -0,0 +1,186 @@
...
+      <View style={titleContainer}>
+        <Text style={titleFont}>{title}</Text>
+      </View>
+    );
+  };
+
+  renderCell = ({ content, index, header, width }) => {
+    const { cell, cellFont } = localStyles({ ...this.props, index, header, width });
+    return (
+      <View style={cell} key={this.keyExtractor({ index })}>
```
andreievg: https://reactjs.org/docs/lists-and-keys.html#extracting-components-with-keys

src/widgets/SimpleTable.js
```diff
@@ -0,0 +1,186 @@
...
+ * @prop    {string} title   Title for this table.
+ * data = [ { a: 1, b: 2}, {a: 3, b: 4}, .. ]
+ * columns = [ { key: 'a', width: 1, title: 'A COLUMN' }, { key: 'b', width: 1, title: 'B COLUMN' }]
+ */
+export class SimpleTable extends React.PureComponent {
+  keyExtractor = ({ index }) => index;
+
+  renderTitle = () => {
+    const { title } = this.props;
+    const { titleContainer, titleFont } = localStyles(this.props);
```
andreievg: as per our conversation, not a fan of recreating styles for every cell etc.. 
can leave this be as longs as you make an issue to refactor it somehow (it does look super clear and simple and I do like the neatness of the code). 
the only way i can think of refactoring at this stage (with the least impact on performance is, say for a cell: 
`cell = this.adjustStyle(cell, index, header, width)`
...
```
// when component mounts
this.styles = localStyles(this.props)
```
```
adjustmentStyles = ({index, width} = {}) = { 
cell: { flex: width, backgroundColor: index % 2 === 0 ? ROW_BLUE : 'white' })
```
...
```
adjustStyle = (type, ...extra) => {
   return { ...this.styles[type], ...this.adjustmentStyles(...extra)[type]}
}
```

src/widgets/SimpleTable.js
andreievg: from what i understand, it's should be enough to just pass key extractor to list, don't need them on cells.

Also you should be able to pass keyExtractor to this page (can use index by default, but apparently it has some performance issues when sorting)

src/widgets/SimpleTable.js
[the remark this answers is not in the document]
joshxg: image


Because the props aren't used in  the method

src/widgets/SimpleTable.js
joshxg: The example given is 2D, just rows, whereas this is 3D, rows and cells. Basically whenever something is rendered from an array (rows and cells for that row) it needs a key. I also get key warnings without it.

Had a conversation with @Chris-Petty  about keys. Basically in the end couldn't find a better unique value to use

Added keyExtractor prop 9ca521998257d360c189f1ced9a62611d48fb7d9

src/widgets/SimpleTable.js
joshxg: #885

Chris-Petty:
@joshxg @andreievg I'm happy with this given the issue to fix the localStyles exists.

Unrelated, I should probably update this repo with good issue/PR templates with appropriate labels and what not.

src/widgets/SimpleTable.js
andreievg: Screen Shot 2019-04-18 at 2 00 30 PM

I am not a react expert, but i am suspicious about having the same key for siblings. And since every cell in a row has the same key, that seems to be the fact here. btw, map also calls callback with 'index' as second optional parameter, so you can pass it to render cell as the key

src/widgets/SimpleTable.js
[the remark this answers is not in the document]
joshxg: d07b0e478c4810b25deee727f559ecf2ec2d4fc6

src/widgets/SimpleTable.js
joshxg: d07b0e478c4810b25deee727f559ecf2ec2d4fc6

Issue #878, opened 2019-04-17
[Vaccine Module] aggregate sensor data


wlthomson:
## Description

Helper method for vaccine modals and charts to extract aggregated temperature values from sensor data.

## Requirements

Utility method to aggregate sensorLog minimum or maximum, temperatures into `n = numberOfIntervals` data points, spanning a period of time bound by `startDate` and `endDate`.

`aggregateLogs({sensorLogs, numberOfIntervals, isMax, startDate, endDate})`

- `sensorLogs`: `sensorLogs` (required)
- `numberOfIntervals`: `int` (required)
- `isMax`: `boolean` (if true aggregate maximum temperatures, if false minimum temperatures, optional)
- `startDate`: `date` (starting bound for date, optional)
- `endDate`: `date` (starting bound for date, optional)

Issue #843, opened 2019-04-02
[Vaccine Module][Temperature Breach] - Helper methods to extract breach information.


andreievg:
As per epic: https://github.com/sussol/org-issues/issues/23

### NOTE

After breaking this down, and writing pseudo logic for breaches, it seems it's better as a DataBase entity. Should be quicker for users, and cleaner for developers too. Please ignore any logic written here for calculating breaches, there will be a Realm Table called `TemperatureBreach`. Also all of the helper methods should be named with `TemperatureBreach` not just breach as described below. (`link to TemperatureBreach, creation and updates issue here`)

#### Description

This issue should probably be broken down into smaller issues, here I will try to summarise all requirements for calculating breach.

##### What is a breach ?

Temperature breach occurs when temperature (sensorLog) for a particular fridge is breached for a certain time period.

LocationType.customData holds the setting for identifying a breach, in this format:

```
{
    breaches: [{
      upperBound: 1000,
      lowerBound: 8,
      duration: 60,
    }, {
      lowerBound: -10000,
      upperBound: 2,
      duration: 15
    }]
}
```

There are a number of places where breach calculation needs to be done, in summary:
* In some cases just need to display the ⚠️ in a row, for item or batch
  * .hasBreach() should be sufficient in this case
* Another case is to display ⚠️ on the graph at particular point, here we need more fine grained information (min/max temperature, and from, to time/date)
* Lastly, even more info is needed in the Breach Modal, and Breach Modal
  * Here we need a list of breaches with item/batches associated with them

#####  Breach for Vaccine Module Page

Need to find breaches for particular fridge, fromDate and toDate. Breaches need to be displayed in the graph as a ⚠️ (red one of those)
`Proposing`
* Fridge.getBreaches(fromDate, toDate) 
  * Just need min or max temperature and fromDate, toDate

##### Breach for Vaccine Manage Item Page

In this page, non-zero batches for a particular Vaccine Item are shown.
Display ⚠️ in a batch row
`Proposing`
* ItemBatch.hasBreaches
  * Return true on first match

##### Breach for Customer Invoice/Manage Vaccine Stock Page

In this page, non-zero batches for a particular Vaccine Item are shown.
Display ⚠️ in a batch row
`Proposing`
* Item.hasBreaches
  * Goes through each non-zero batch for item 
  * Return true on first match

##### Breach Modal

[...]

PR #881, opened 2019-04-17
#843 helper methods

joshxg:
Fixes half of #843 

Other methods will be a part of #878 

`extractBreaches` : 
- Allows for either an array of `SensorLog` objects or a `Realm.results`. *Could* enforce that it must be a `Realm.results` so that we can use Realm sorting. Wanted to be safe for now, if it turns out we always are calling with a `Realm.Results`, I think change the method then. It is annoying to 'create' `Realm.Results` objects, imo.
- Potentially there is a different way to use `Results.filtered` by using the timestamps of the delimiter `SensorLog` objects, but this would be time consuming to find the surrounding `SensorLog` objects. Would require a similar method, but would use the timestamps of the delimiters to find all `SensorLog` between. Would then need to `find()` the first and last `SensorLog` in an array of all `SensorLog`'s, and use the index before and after. I don't think this offers any improvement to performance, as you would still not have a query which returns a `Realm.Results` of `SensorLog`'s.

`sensorLogExtractBreaches`
- Don't like this name
- Return object has a slightly different shape to what was described in the issue. Have added a couple of fields because of not having any negatives, but could come in handy.
- Using the object shape during the method to make searching for `Item` and `ItemBatch` objects are done in constant time, rather than having to search an array each time you want to update.
- Could not bother creating a different shape at the end of the method, but this means it can be passed straight to `BreachModal` after adding columns for the tables easily.

### DIFF
--- a/src/utilities/modules/vaccines.js
+++ b/src/utilities/modules/vaccines.js
@@ -9,24 +9,6 @@
  */
 
 /**
- * Method used with Array.prototype.sort for sensorLogs
- * Sorts from oldest to newest.
- * @param  {SensorLog} sensorLogA
- * @param  {SensorLog} sensorLogB
- * @param  {Date}      sensorLogA.timestamp
- * @param  {Date}      sensorLogB.timestamp
- * @return {Number}    1 for A>B, -1 for A<B, 0 for A=B
- */
-const sortSensorLogs = (
-  { timestamp: timestampA = new Date(-8640000000000000) },
-  { timestamp: timestampB = new Date(-8640000000000000) }
-) => {
-  if (timestampA > timestampB) return 1;
-  if (timestampA < timestampB) return -1;
-  return 0;
-};
-
-/**
  * Extracts breaches from a set of sensor logs.
  * Breach: Sequential sensorLog objects whose temperature is outside
  * the lower or uppound limits, form a breach. Each array returned
@@ -40,7 +22,7 @@ const sortSensorLogs = (
  * All breached sensorLogs are pushed, until a non breached log, where this
  * marks a breach has been found. Store these sensorLogs and reset the stack.
  *
- * @param  {Realm.results/array} sensorLogs array of sensorLog objects
+ * @param  {Realm.results} sensorLogs array of sensorLog objects
  * @return {Array<Realm.results<SensorLog>} A 2D array of Realm.results. Example:
  * [ [ sensorLog1, sensorLog2 .. ], [ sensorLog1, sensorLog2, sensorLog3, .. ]]
  * where each element is a Realm.results<SensorLog>
@@ -52,7 +34,7 @@ export const extractBreaches = ({ sensorLogs = [], database }) => {
   const breaches = [];
 
   // Sort the sensor logs by timestamp.
-  sensorLogs.sort(sortSensorLogs);
+  sensorLogs.sorted('timestamp', true);
 
   sensorLogs.forEach(sensorLog => {
     const { isInBreach } = sensorLog;
@@ -72,7 +54,7 @@ export const extractBreaches = ({ sensorLogs = [], database }) => {
       // onto the stack as the new potential delimiter for the next breach.
     } else {
       sensorLogStack.push(sensorLog);
-      breaches.push(sensorLogStack);
+      breaches.push([...sensorLogStack]);
       sensorLogStack.length = 0;
       sensorLogStack.push(sensorLog);
     }
@@ -82,13 +64,11 @@ export const extractBreaches = ({ sensorLogs = [], database }) => {
   if (sensorLogStack.length > 0) breaches.push(sensorLogStack);
 
   // Create Realm.result objects for each breach.
-  breaches.map(breach => {
+  return breaches.map(breach => {
     const sensorIds = breach.map(log => log.id);
     const queryString = sensorIds.map((_, i) => `id = $${i}`).join(' OR ');
     return database.objects('SensorLog').filtered(queryString, ...sensorIds);
   });
-
-  return breaches;
 };
 
 /**
--- a/src/utilities/modules/vaccines.js
+++ b/src/utilities/modules/vaccines.js
@@ -34,17 +34,16 @@ export const extractBreaches = ({ sensorLogs = [], database }) => {
   const breaches = [];
   const sensorLogsByLocation = {};
 
-  // Sort the sensor logs by timestamp.
-  sensorLogs.sorted('timestamp', true);
-
-  // Group all sensorLogs by location
-  sensorLogs.forEach(sensorLog => {
-    const { location } = sensorLog;
-    // If this sensorLog doesn't have a location,
-    // ignore it
-    if (!location) return;
-    if (!sensorLogsByLocation[location.id]) sensorLogsByLocation[location.id] = [sensorLog];
-    else sensorLogsByLocation[location.id].push(sensorLog);
+  sensorLogs.forEach(({ location }) => {
+    // If no location or it's is is falsey, ignore it.
+    // to avoid a 'null' grouping.
+    if (!(location && location.id)) return;
+    // If this lcoation has been seen, ignore it.
+    if (sensorLogsByLocation[location.id]) return;
+    // Create a grouping of sensorlogs for this locations
+    sensorLogsByLocation[location.id] = sensorLogs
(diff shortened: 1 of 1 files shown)


src/utilities/modules/vaccines.js
```diff
@@ -0,0 +1,172 @@
...
+ * where each element is a Realm.results<SensorLog>
+ */
+export const extractBreaches = ({ sensorLogs = [], database }) => {
+  if (!(database && sensorLogs.length > 0)) return [];
+
+  const sensorLogStack = [];
+  const breaches = [];
+
+  // Sort the sensor logs by timestamp.
+  sensorLogs.sort(sortSensorLogs);
```
wlthomson: I don't think `.sort` is defined for Realm collections?

src/utilities/modules/vaccines.js
```diff
@@ -0,0 +1,152 @@
...
+    else if (!sensorLogStack[stackLength - 1].isInBreach && !isInBreach) {
+      sensorLogStack.pop();
+      sensorLogStack.push(sensorLog);
+      // In all other cases, a breach has been found. Push the last sensorlog,
+      // store the breach and reset the stack. Push this sensorLog back
+      // onto the stack as the new potential delimiter for the next breach.
+    } else {
+      sensorLogStack.push(sensorLog);
+      breaches.push([...sensorLogStack]);
+      sensorLogStack.length = 0;
```
andreievg: that's very cool, but can you please `sensorLogStack = []`, i just not sure if .length = 0 will aways be `supported`

src/utilities/modules/vaccines.js
```diff
@@ -0,0 +1,152 @@
...
+    }
+  });
+  // Push any remaining sensorlogs. If any are left, they form a breach
+  // with no delimiting non-breached sensorlog.
+  if (sensorLogStack.length > 0) breaches.push(sensorLogStack);
+
+  // Create Realm.result objects for each breach.
+  return breaches.map(breach => {
+    const sensorIds = breach.map(log => log.id);
+    const queryString = sensorIds.map((_, i) => `id = $${i}`).join(' OR ');
```
andreievg: why not breach.map({id} => `id = "${id}"`).join(' OR ')

src/utilities/modules/vaccines.js
joshxg: 69db3a26b57771c4457fff893da0579fdf5901b6 

(I *think* it works though, I can't remember if I actually tested this explicitly)

src/utilities/modules/vaccines.js
joshxg: c43838fcaeb90c506c44543a599b7c4b38fafb47

src/utilities/modules/vaccines.js
joshxg: I tried this at first. It didn't work, can't remember the error exactly. Was something about having static predicates

joshxg: wanted to make sure: 
image

From what I can tell - queries were changed, if you had done this in the past. There are a few very old issues on the realm repo where this is suggested,  but doesn't work anymore.

joshxg:
39aad5825db9f1bc35afb7e23b9494744de31f5c location groupings

joshxg:
@andreievg - Should this also check if a breach length >= 3 ? e.g. must have at least 3 sensorLogs?

- Only case it wouldn't be, that I can think of is for the most recent log, which may have 1 or 2

src/utilities/modules/vaccines.js
andreievg: @joshxg, you are missing "

Screen Shot 2019-04-19 at 8 15 48 PM
Screen Shot 2019-04-19 at 8 15 14 PM

andreievg: if itemBatches.length is zero though, it crashes (empty query)

src/utilities/modules/vaccines.js
```diff
@@ -0,0 +1,170 @@
...
+  sensorLogs.sorted('timestamp', true);
+
+  // Group all sensorLogs by location
+  sensorLogs.forEach(sensorLog => {
+    const { location } = sensorLog;
+    // If this sensorLog doesn't have a location,
+    // ignore it
+    if (!location) return;
+    if (!sensorLogsByLocation[location.id]) sensorLogsByLocation[location.id] = [sensorLog];
+    else sensorLogsByLocation[location.id].push(sensorLog);
```
andreievg: change not required, but food for though
```
 sensorLogs.forEach(({ location }) => {  
    if(!location || sensorLogsByLocation[location]) return;
   sensorLogsByLocation[location.id] = sensorLogs.filtered('location.id = $0', location.id).sorted('timestamp')
} 
```

src/utilities/modules/vaccines.js
joshxg: .. 063e98340484f70afdc20b473b1edcab2936fdb0

- Won't ever have an empty query

src/utilities/modules/vaccines.js
```diff
@@ -0,0 +1,170 @@
...
+    // { duration, id, code, enteredDate, totalQuantity, expiryDate }
+    // If already stored, increment the duration in the breach by this logs interval.
+    batchesToUse.forEach(
+      ({ item: batchItem, id: batchId, code, enteredDate, expiryDate, totalQuantity }) => {
+        // Premature return if this ItemBatch does not have an Item associated
+        if (!batchItem) return;
+        const { id: itemId } = batchItem;
+        // If this batches item hasn't been encountered yet, create
+        // a grouping object. groupedBatches = { itemId: { item, batches: {}  }}
+        if (!groupedBatches[itemId]) groupedBatches[itemId] = { item: batchItem, batches: {} };
```
andreievg: not sure why you would destrcuture item: to batchItem earlier

src/utilities/modules/vaccines.js
```diff
@@ -0,0 +1,170 @@
...
+ * [
+ *  {
+ *     item,
+ *     batches: [ { id, code, expiryDate, enteredDate, duration, totalQuantity }]
+ *  }
+ *  ...
+ * ]
+ */
+export const sensorLogsExtractBatches = ({ sensorLogs = [], itemBatch, item } = {}) => {
+  const groupedBatches = {};
```
andreievg: once again change not required, you can do this however:
filter sensorLogs by item or batch
for each sensorLogs, if groupedBatches,item.batch exists return 
otherwise add it and fine 'duration' by :
if not find duration, by sensorLogs.filtered by batch .max('timestamp') - .min('timestamp')

andreievg:
> Only case it wouldn't be, that I can think of is for the most recent log, which may have 1 or 2

there would be at least 2 sensorLogs in breach

src/utilities/modules/vaccines.js
```diff
@@ -0,0 +1,170 @@
...
+  sensorLogs.forEach(({ itemBatches, logInterval, isInBreach }) => {
+    // Premature exit if log has no batches, or is not a breached sensorLog
+    if (itemBatches.length === 0 || !isInBreach) return;
+    // Ensure each batch has stock
+    let batchesToUse = itemBatches.filtered('totalQuantity > $0', 0);
+    // If an Item has been passed, only find batches for this item.
+    if (item) batchesToUse = batchesToUse.filtered('batch.id = $0', item.id);
```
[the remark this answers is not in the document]
joshxg: Change the method as per comment below

src/utilities/modules/vaccines.js
joshxg: done, forgot to do a commit for this, sorry: b8919e997007a723bf88c9af19ab8e12a5807416

src/utilities/modules/vaccines.js
joshxg: parameter named item

src/utilities/modules/vaccines.js
joshxg: 1+ (3+2+1-5+2-3 ) - b8919e997007a723bf88c9af19ab8e12a5807416

src/utilities/modules/vaccines.js
andreievg: sure, sorry i wasn't clear enough, was just advising that the only crash with query would be if itemBatches.length is zero (which won't be the case)

PR #884, opened 2019-04-17
#878 aggregate sensor data

### LINKED ISSUE #878: [Vaccine Module] aggregate sensor data

wlthomson:
Fixes #878.

Test branch `#878-aggregate-sensor-data-test`.

Takes a Realm collection of `sensorLog` objects, `numberOfIntervals` (integer representing the number of intervals to "split" the sensor logs into), and optional parameters: `isMax`, `startDate` and `endDate`.

 Returns an array of custom objects consisting of:

```
{
  intervalStartDate,
  intervalEndDate,
  medianDate,
  sensorLogs,
  sensorLog 
}
```

`medianDate` is the date representing the aggregated data point, and `sensorLog` is the log in the interval with the maximum/minimum temperature.


src/utilities/aggregateLogs.js
```diff
@@ -0,0 +1,117 @@
...
+ *                                                         return.
+ * @param   {Date}                     startDate           Start date of temperature range to
+ *                                                         aggregate over.
+ * @param   {Date}                     endDate             End date of temperature range to
+ *                                                         aggregate over.
+ * @return  {Object[]}                                     Aggregated sensor logs, array of objects
+ *                                                         in form [{timestamp}, {temperature},...].
+ */
+export const aggregateLogs = ({ data, isMax, numberOfDataPoints, startDate, endDate }) => {
+  // If start date is later than earliest date in data, update.
```
andreievg: Hmm i was thinking from the start to try and use the simplicity of realm.
* Sort sensorLogs with realm
* if startDate is not defined use sortedSensorLogs[first]
* if endData is not defined use sortedSensorLogs[last]
* (endDate - startDate) / numberOfDataPoints -> interval in milliseconds
* from 0 to numberOfDataPoints, make an array of object ( [{ startDate, endDate, middleDate}, { startDate, endDate, middleDate}, ... ]
* use realm, data.filtered, (timestamp > startDate and endDate =< endDate).min('temperature') for every element of that array.

done

at the end call a method that checks the size of the interval and formats each middle date accordingly (i.e. hours or minutes etc..) can leave that to the last

wlthomson: No worries, I'll have another go at it the Realm way 😆 .

src/utilities/aggregateLogs.js
```diff
@@ -0,0 +1,80 @@
...
+ *                                               over.
+ * @param   {Date}           endDate             End date of temperature range to aggregate over.
+ * @return  {Object[]}                           Aggregated sensor logs, array of objects in form
+ *                                              [{timestamp}, {temperature},...].
+ */
+export const aggregateLogs = ({
+  sensorLogs,
+  numberOfIntervals,
+  isMax = true,
+  startDate = new Date(),
```
joshxg: If no `startDate` is passed, the `startDate` will always be today, rather than the first `timestamp` of `sensorLogs` unless your `sensorLogs` are from the future.

src/utilities/aggregateLogs.js
```diff
@@ -0,0 +1,80 @@
...
+}) => {
+  // Break out if no sensorLogs.
+  if (!(sensorLogs.length > 0)) return [];
+
+  // Sort sensor logs ascending by date.
+  sensorLogs.sorted('timestamp');
+
+  // If start date is later than earliest log, update.
+  const startLog = sensorLogs[0];
+  startDate = startDate > startLog.timestamp ? startLog.timestamp : startDate;
```
joshxg: This is a downside of passing objects as parameters. Assigning new values to parameters causes more side effects than usual. 

In any case it's usually discouraged to reassign any value for the usual reasons and apparently most JS engines are not optimized for reassignment. Also I've read some say it is worse to reassign parameters (as this can mutate the `arguments` object, but I think this argument isn't very valid as it's also discouraged to access the `arguments` object).

src/utilities/aggregateLogs.js
```diff
@@ -0,0 +1,80 @@
...
+
+  // If end date is earlier than latest date in data, update.
+  const endLog = sensorLogs[sensorLogs.length - 1];
```
joshxg: can also `const { timestamp:lastTimestamp } = ..` I think you don't use it anywhere else. I don't *think* we have to be careful of not having a timestamp?

src/utilities/aggregateLogs.js
```diff
@@ -0,0 +1,80 @@
...
+    const intervalStartDate = new Date(startDate.getTime() + intervalDuration * i);
+    const intervalEndDate = new Date(startDate.getTime() + intervalDuration * (i + 1) - 1);
+    aggregatedLogs.push({ intervalStartDate, intervalEndDate });
+  }
+
+  // Map intervals to aggregated objects.
+  aggregatedLogs.forEach(({ intervalStartDate, intervalEndDate }, index) => {
+    // Calculate median date.
+    aggregatedLogs[index].medianDate = new Date(
+      intervalStartDate.getTime() + (intervalEndDate.getTime() - intervalStartDate.getTime()) / 2
```
joshxg: Just a little clearer
`const median = (intervalEndDate.getTime() - intervalStartDate.getTime()) / 2`
`aggregatedLogs[index].medianDate = intervalStartDate.getTime() + median`

src/utilities/aggregateLogs.js
```diff
@@ -0,0 +1,80 @@
...
+      intervalEndDate
+    );
+
+    // Get maximum or minimum log.
+    const getMinMaxLog = (previousLog, currentLog) => {
+      if (isMax) return previousLog.temperature > currentLog.temperature ? previousLog : currentLog;
+      return previousLog.temperature < currentLog.temperature ? previousLog : currentLog;
+    };
+
+    aggregatedLogs[index].sensorLog = aggregatedLogs[index].sensorLogs.reduce(
```
joshxg: `aggregatedLogs[index].sensorLogs.max(temperature)`
`aggregatedLogs[index].sensorLogs.min(temperature)`

src/utilities/aggregateLogs.js
```diff
@@ -0,0 +1,80 @@
...
+
+    // Get sensor logs.
+    aggregatedLogs[index].sensorLogs = sensorLogs.filtered(
+      'timestamp >= $0 && timestamp <= $1',
+      intervalStartDate,
+      intervalEndDate
+    );
+
+    // Get maximum or minimum log.
+    const getMinMaxLog = (previousLog, currentLog) => {
```
joshxg: at this point the difference between returning min or max is so little - one query - that we may as well return both. The performance hit to calling this method for just a min or max is so little relative to the performance hit of wanting both and having to call this method twice?

src/utilities/aggregateLogs.js
```diff
@@ -0,0 +1,80 @@
...
+  endDate = new Date(),
+}) => {
+  // Break out if no sensorLogs.
+  if (!(sensorLogs.length > 0)) return [];
+
+  // Sort sensor logs ascending by date.
+  sensorLogs.sorted('timestamp');
+
+  // If start date is later than earliest log, update.
+  const startLog = sensorLogs[0];
```
joshxg: `const [startLog,,endLog] = sensorLogs`

src/utilities/aggregateLogs.js
wlthomson: I thought it would initialise as epoch time (which still would have been a problem, so not defence!) Fixed by 6f9f29dd.

src/utilities/aggregateLogs.js
wlthomson: I didn't like reassigning either, I was actually surprised the args weren't constants. Fixed by 6f9f29dd.

src/utilities/aggregateLogs.js
wlthomson: Fixed by 6f9f29dd.

src/utilities/aggregateLogs.js
wlthomson: Fixed by 6f9f29dd.

src/utilities/aggregateLogs.js
wlthomson: Living that Realm life. Fixed by 6f9f29dd.

wlthomson:
> Nice. Couple comments, maybe need input from @andreievg on return values.
> 
> Opinion: Most of your comments explain _what_ the code is doing, which is easy to see and relatively useless. e.g. 'sort by x', 'get sensor logs'. IMO comments should explain _why_ this is necessary and what the result or purpose is

All recommended changes made and comments updated.

src/utilities/aggregateLogs.js
wlthomson: Much nicer. Fixed by 6f9f29dd.

src/utilities/aggregateLogs.js
wlthomson: I agree. Returning `minSensor` and `maxSensor` seems reasonable to me.

joshxg:
I messed up a comment, or I can't see it again:

I thought that this method was returning `{date, temp}` for each aggregated log, so that it could be input straight into a chart component. If it's not, we should have a wrapper method which will do this, as it will be code that is repeated at least a few times

wlthomson:
> I messed up a comment, or I can't see it again:
> 
> I thought that this method was returning `{date, temp}` for each aggregated log, so that it could be input straight into a chart component. If it's not, we should have a wrapper method which will do this, as it will be code that is repeated at least a few times
 
I was thinking a better idea might be to refactor the chart component to take `sensorLog` or objects with the same property names, e.g. `timestamp`, `temperature`. What do you think?

joshxg:
Objects with those properties sounds good. I would say that the data keys should be a prop to pass in - defaultProp: temperature, timestamp.

src/utilities/aggregateLogs.js
joshxg: These methods return scalar values, not the object

wlthomson:
> Objects with those properties sounds good. I would say that the data keys should be a prop to pass in - defaultProp: temperature, timestamp.

Agree. Here an example of the current structure of the return values from `aggregateLogs`:

```
{ 
    intervalStartDate: Sun Feb 03 2019 08:00:00 GMT+1300 (NZDT),
    intervalEndDate: Tue Feb 05 2019 15:59:59 GMT+1300 (NZDT),
    sensorLogs: [ { '0': { id: 'c71f9192626311e998a5553603cab91a',
                           sensor: null,
                           location: null,
                           pointer: null,
                           timestamp: Mon Feb 04 2019 00:00:00 GMT+1300 (NZDT),
                           temperature: 1,  
                           logInterval: null,
                           isInBreach: false,
                           itemBatches: {} } }, ... ],
    minimumLog: { timestamp: Mon Feb 04 2019 11:59:59 GMT+1300 (NZDT),
                  temperature: 1 },
    maximumLog: { timestamp: Mon Feb 04 2019 11:59:59 GMT+1300 (NZDT),
                  temperature: 4 } 
}
```

andreievg:
@wlthomson @joshxg i didn't really read through earlier comments. But did catch a glimps of Wills reply to
> Objects with those properties sounds good. I would say that the data keys should be a prop to pass in - defaultProp: temperature, timestame

That would be pretty good, btw at this stage parameters become keys of one object and get destructured with default values in method header.

joshxg:
@wlthomson  There is now a folder `utilities/modules` with a `vaccines.js` file this should go in please

andreievg:
@wlthomson, on second though it's fine to return min and max.

src/utilities/aggregateLogs.js
```diff
@@ -0,0 +1,79 @@
...
+  const endBoundary = new Date(Math.max(endTimestamp, endDate) || endTimestamp);
+
+  const totalDuration = endBoundary - startBoundary;
+  const intervalDuration = totalDuration / numberOfIntervals;
+
+  const aggregatedLogs = [];
+  for (let i = 0; i < numberOfIntervals; i += 1) {
+    const intervalStartDate = new Date(startBoundary.getTime() + intervalDuration * i);
+    const intervalEndDate = new Date(startBoundary.getTime() + intervalDuration * (i + 1) - 1);
+    aggregatedLogs.push({ intervalStartDate, intervalEndDate });
```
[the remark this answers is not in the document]
wlthomson: Fixed by 6b0fce7.

src/utilities/aggregateLogs.js
```diff
@@ -0,0 +1,79 @@
...
+
+  const startBoundary = new Date(Math.min(startTimestamp, startDate) || startTimestamp);
+  const endBoundary = new Date(Math.max(endTimestamp, endDate) || endTimestamp);
+
+  const totalDuration = endBoundary - startBoundary;
+  const intervalDuration = totalDuration / numberOfIntervals;
+
+  const aggregatedLogs = [];
+  for (let i = 0; i < numberOfIntervals; i += 1) {
+    const intervalStartDate = new Date(startBoundary.getTime() + intervalDuration * i);
```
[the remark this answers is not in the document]
wlthomson: Fixed by 6b0fce7.

src/utilities/aggregateLogs.js
```diff
@@ -0,0 +1,79 @@
...
+    const intervalEndDate = new Date(startBoundary.getTime() + intervalDuration * (i + 1) - 1);
+    aggregatedLogs.push({ intervalStartDate, intervalEndDate });
+  }
+
```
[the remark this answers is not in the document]
wlthomson: Fixed by 6b0fce7, a5df9df.

src/utilities/aggregateLogs.js
```diff
@@ -0,0 +1,79 @@
...
+  // Map intervals to aggregated objects.
+
+  aggregatedLogs.forEach(({ intervalStartDate, intervalEndDate }, index) => {
```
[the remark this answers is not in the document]
wlthomson: Fixed by 6b0fce7.

wlthomson:
> @wlthomson, on second though it's fine to return min and max.

Return value has been simplified to a basic `sensorLog` object with max and min temperatures, e.g.:

```
[
  { 
    timestamp: Wed Feb 06 2019 20:00:00 GMT+1300 (NZDT),
    minTemperature: 1,
    maxTemperature: 6
  }
]
```

andreievg:
> Return value has been simplified to a basic sensorLog object with max and min temperatures, e.g.:

That means they need to be re-mapped for charts to show them?

wlthomson:
> > Return value has been simplified to a basic sensorLog object with max and min temperatures, e.g.:
> 
> That means they need to be re-mapped for charts to show them?

I thought it would be good to return more info, and we are returning both max and min. Will update to:

```
{
  minLine: [ { timestamp, temperature }, ... ],
  maxLine: [ { timestamp, temperature }, ... ]
}
```

That way can feed straight to the chart, destructuring if only one required.

PR #897, opened 2019-04-19
#870 breach modal

### LINKED ISSUE #870: [Vaccine Module] Breach Modal
### Description:

A modal which displays information about a breach for a particular batch, item or fridge.

Prototype:

image
image

- Will open when a "Breach Hazard" icon is pressed anywhere in the app.
- The "Breach Hazard" icons appear in charts or tables for
    - Items
    - ItemBatches
    - Fridges
- Will display a list of breaches with the above columns

### Requirements

- Modal usable for all three different data sets
- Rows are expandable which will display, for each data set
    - Items -  Chart with (date, temp) for the breach | ItemBatches affected by the breach
    - ItemBatches - Chart with (date, temp) for the breach | ??
    - Fridges -  chart with (date, temp) for the breach | Items and batches within the fridge

### Tasks
- SimpleTable component for rendering the smaller table [ Just rendering scalar values in cells using FlatList to be a little more performant and easier styling ]
- Chart component - #854 
- Expansion component - Render the chart and items side by side (?) will make state management a little easier?
- Helper methods
    - Transform sensorlog / sensorLogs to objects with just the data for the columns needed(?)
    - Method which fetches all ItemBatches that were in a location for a specific item, preferrably returning an object with only the data needed(?)
    - Method which returns temperatures for a particular breach, and another which transforms this into a format for the chart (?)
- Modal - bringing all components together (Should this be an actual modal or a component intended to be wrapper in PageContentModal ?)

joshxg:
Fixes #870 

[messy dev branch](https://github.com/openmsupply/mobile/tree/breach-modal)

image

#### Ignore the actual data

Ended up making this very dumb.

Expectation is that data will just be passed. Data will be fetched inside of a `runWithLoadingIndicator`. (Could cache it to save time if needed?)

Title is set in the `PageContentModal` this will be wrapped in

- I really hate the `formattedDifferenceBetweenDates` method
- Could change this to fetch some data when expanding a row quite easily
- To fetch the data:
    - `extractBreaches()` - passing `sensorLogs` from a `Location`, `ItemBatch` or `Item`
    - `sensorLogsExtractBreaches()` passing each `breach` from `extractBreaches`
    - `extractBreach()` - on return value from above
    -  `aggregateLogs` - will return the chartData needed

### DIFF
--- a/src/utilities/formatters.js
+++ b/src/utilities/formatters.js
@@ -17,3 +17,22 @@ export const formatExposureRange = ({ minTemperature, maxTemperature } = {}) => 
   const degree = String.fromCharCode(176);
   return `${minTemperature}${degree}C to ${maxTemperature}${degree}C`;
 };
+
+export const formattedDifferenceBetweenDates = (dateA, dateB) => {
+  const differenceInMs = new Date(dateB.getTime() - dateA.getTime()).getTime();
+
+  const msPerMinute = 1000 * 60;
+  const msPerHour = msPerMinute * 60;
+  const msPerDay = msPerHour * 24;
+
+  const numberOfDays = Math.floor(differenceInMs / msPerDay);
+  const numberOfHours = Math.floor((differenceInMs - numberOfDays * msPerDay) / msPerHour);
+  const numberOfMinutes = Math.floor((differenceInMs - numberOfHours * msPerHour) / msPerMinute);
+
+  let formattedDifference = '';
+  if (numberOfDays) formattedDifference += `${numberOfDays} Days `;
+  if (numberOfHours) formattedDifference += `${numberOfHours} Hours `;
+  if (numberOfMinutes) formattedDifference += `${numberOfMinutes} Minutes `;
+
+  return formattedDifference;
+};
--- a/src/utilities/formatters.js
+++ b/src/utilities/formatters.js
@@ -18,8 +18,9 @@ export const formatExposureRange = ({ minTemperature, maxTemperature } = {}) => 
   return `${minTemperature}${degree}C to ${maxTemperature}${degree}C`;
 };
 
-export const formattedDifferenceBetweenDates = (dateA, dateB) => {
-  const differenceInMs = new Date(dateB.getTime() - dateA.getTime()).getTime();
+export const formattedDifferenceBetweenDates = ({ startDate, endDate }) => {
+  if (!(startDate || endDate)) return 'Unavailable';
+  const differenceInMs = new Date(endDate.getTime() - startDate.getTime()).getTime();
 
   const msPerMinute = 1000 * 60;
   const msPerHour = msPerMinute * 60;
--- a/src/utilities/modules/vaccines.js
+++ b/src/utilities/modules/vaccines.js
@@ -1,5 +1,3 @@
-import { formattedDifferenceBetweenDates, formatExposureRange } from '../formatters';
-
 /**
  * mSupply Mobile
  * Sustainable Solutions (NZ) Ltd. 2019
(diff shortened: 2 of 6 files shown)


andreievg:
@joshxg, about to review this one. conflicts

src/utilities/modules/vaccines.js
```diff
@@ -0,0 +1,41 @@
...
+ *    ...
+ * ]
+ */
+export const setBreachData = (data = []) => {
+  const dataCopy = [];
+  data.forEach(dataSet => {
+    const { breach } = dataSet;
+    // A breach must have a minimum of 3 sensorlogs to be considered
+    // a breach.
+    if (!breach || breach.length < 4) return;
```
andreievg: i don't think you need to restrict to 3 sensorLogs ?

src/utilities/modules/vaccines.js
```diff
@@ -0,0 +1,41 @@
...
+    const [firstSensorLog = null, , { timestamp: lastTimestamp } = {}] = breach;
```
andreievg: that's cool, didn't know `, ,` lets you deconstruct first and last, why null for fistSensorLog and { timestamp: lastTimestamp } = {} for last one, why not both ?

src/utilities/modules/vaccines.js
```diff
@@ -0,0 +1,41 @@
...
+    // If there is no first or last sensor logs, or if either does not have a timestamp,
+    // don't create data for this breach.
+    if (!(firstSensorLog && firstSensorLog.timestamp && lastTimestamp)) return;
+    dataCopy.push({
+      ...dataSet,
+      breachData: {
+        affectedQuantity: firstSensorLog.sum('itemBatches.numberOfPacks'),
```
andreievg: should this not come from dataSet.items.batches ?
can modify sensorLogsExtractBatches to also have all batches (ungrouped) in dataSet, that will make this easy

src/utilities/modules/vaccines.js
```diff
@@ -0,0 +1,41 @@
...
+        location: (firstSensorLog.location && firstSensorLog.location.description) || 'Unavailable',
+        duration: formattedDifferenceBetweenDates(firstSensorLog.timestamp, lastTimestamp),
+        exposureRange: formatExposureRange(breach.min('temperature'), breach.max('temperature')),
+        numberOfAffectedBatches: firstSensorLog.itemBatches.length,
```
andreievg: same as above comment

src/utilities/formatters.js
```diff
@@ -17,3 +17,22 @@ export const formatExposureRange = ({ minTemperature, maxTemperature } = {}) =>
...
+  const msPerDay = msPerHour * 24;
+
+  const numberOfDays = Math.floor(differenceInMs / msPerDay);
+  const numberOfHours = Math.floor((differenceInMs - numberOfDays * msPerDay) / msPerHour);
+  const numberOfMinutes = Math.floor((differenceInMs - numberOfHours * msPerHour) / msPerMinute);
+
+  let formattedDifference = '';
+  if (numberOfDays) formattedDifference += `${numberOfDays} Days `;
+  if (numberOfHours) formattedDifference += `${numberOfHours} Hours `;
+  if (numberOfMinutes) formattedDifference += `${numberOfMinutes} Minutes `;
```
andreievg: this method is fine for now, but i would envision showing something like this:
1 week/s, 2 day/s, 2 day/s 1 hour/s, 6 hour/s 12 minute/s, 200 minutes
we would need to set a constants (number of milliseconds, to switch between day hour or minutes)

src/utilities/modules/vaccines.js
joshxg: hm, there is one case where we won't have 3 `sensorLogs` - if the first two `sensorLogs` are breached, and the one after isn't.

(I just realized I need to add some more logic here to find the first sensor log to use)

The surrounding `sensorLogs` of a breach will also be in this array, so will have `4` in most cases, `3` if it is the very last breach that has occured (min. 2 sensor logs which have been breached?).

I needed to set a minimum number of `SensorLogs` to calculate values (or set these values as null if there just *happened* to be 1 `SensorLog` in this array.) and if there really was less than 3 (other than the first ever 2 `sensorLogs`, something must have gone wrong. 

Can reduce this to two, or set values to null if < 3?

andreievg:
@joshxg 

So let me get my head around how this component is called:

in backwards order,
* this component get data form setBreachData
* before `setBreachData`, we use Wills methods, `aggregateLogs` fore each breach row and store this in chartData: `{ minLine, minTemperature }`, `{ maxLine, maxTemperature }`, still need to make method to format dates and numberOfDataPoints will be a constant say 7 or numberOfSensorLogs of sensorLogs is below 7
* before that we use `sensorLogsExtractBatches` (with appropriate filters or no filters) on every breach that we get from `extractBreaches`

src/utilities/modules/vaccines.js
joshxg: I suppose it should. I think I figured it was going to be pretty inaccurate anyway you slice it but that does seem closer and less lazy on my part. Another option is to just calculate some of these statistics in `sensorLogsExtractBreaches` and pass them rather than having the extra field with all the batches?

src/utilities/modules/vaccines.js
joshxg: Needed the timestamp from the last one but other things from the first one

joshxg:
Yes exactly. I am not 100% if `chartData` should be called before the component is rendered or when a row is pressed. I think during `runWithLoadingIndicator` is probably the right move.

> still need to make method to format dates
>

Is this for the chart labels?

andreievg:
> Is this for the chart labels?

yeah.. given difference between timestamps, format timestamps differently

src/utilities/modules/vaccines.js
andreievg: if items.batches are not used from params it would be very inaccurate for a breach for a certain ItemBatch.

sure i am happy if you add the aggregation it to `sensorLogsExtractBreaches` method

src/utilities/modules/vaccines.js
andreievg: as you've suggested, extraAggregation in sensorLogsExtractBreaches should resolve the need for this

src/utilities/modules/vaccines.js
andreievg: > hm, there is one case where we won't have 3 sensorLogs - if the first two sensorLogs are breached, and the one after isn't.

data here comes from extractBreaches, which appends sensorLogs at both ends, i guess it's possible if we don't actually have logs at both ends, but still that's enough for us ?

joshxg:
@andreievg 

Just replying here for all the comments:

Method was getting a little over the top so had to refactor a little bit. Have basically merged `setBreachData` and `sensorLogsExtractBatches` 9eeed541c3c640adbc7da374490d358d65089015

Updated `BreachTable` to account for this - had to add a little cell logic as I thought it best to still return soewhat generic data from `sensorLogsExtractBatches` rather than formatted strings. 74fb5e63c962dac711032877889ef514a37676ce

And just a little update to account for null values in dates for this utility helper 1a70ab854b590d993a55e26a12b290828a4d6057

src/utilities/modules/vaccines.js
```diff
@@ -82,6 +82,70 @@ export const extractBreaches = ({ sensorLogs = [], database }) => {
...
+  location: sensorLogs.location,
+  numberOfAffectedBatches: itemBatches.length,
+  affectedQuantity: itemBatches.sum('numberOfPacks'),
+  exposureRange: {
+    minTemperature: itemBatches.min('temperature') || Infinity,
+    maxTemperature: itemBatches.max('temperautre') || -Infinity,
+  },
+  breachDuration: {
+    startDate: itemBatches.max('timestamp') || null,
+    endDate: itemBatches.min('timestamp') || null,
```
andreievg: sensorLogs.max/min, for both exposureRange and breachDuration

src/utilities/modules/vaccines.js
```diff
@@ -158,10 +187,16 @@ export const sensorLogsExtractBatches = ({ sensorLogs = [], itemBatch, item } =
       batches: Object.values(batches).map(batchObject => batchObject),
     };
   });
+  // Get all ItemBatch objects associated with these sensorLogs, used in
+  // calculating statistics of this breach.
+  const allItemBatches = database
+    .objects('ItemBatch')
+    .filtered(sensorLogs.map(({ id }) => `sensorLogs.id = "${id}"`).join(' OR '));
```
andreievg: you've mentioned you forgot to account for surrounding batches.
Just a filter at the start of this method should do it (filter by isInBreach). 

Method should still return unfiltered sensor logs though.

joshxg: 89e3f18862158caacbebe4e4f2ce61fdba90047c

src/utilities/modules/vaccines.js
joshxg: 89e3f18862158caacbebe4e4f2ce61fdba90047c

PR #900, opened 2019-04-19
#895 custom data undefined

### LINKED ISSUE #895: [Vaccine Module] custom_data undefined
When a store has no `custom_data`, the value will be 'undefined' (the string). This then is passed to `validateJson` (in incomingSync) which will return `null`. 

`Setting` requires a string and will throw an error

joshxg:
Fixes #895 

Null was being returned from `validateJson`, which caused a runtime error as `Settings` values must be a string.

#### Alternatives
- Changing `validateJson` to just return `''`
- Checking for 'undefined' earlier

#### Downsides
Incoming `custom_data` or `tags` which are empty strings will not be set

### DIFF
--- a/src/sync/incomingSyncUtils.js
+++ b/src/sync/incomingSyncUtils.js
@@ -677,7 +677,7 @@ export const createOrUpdateRecord = (database, settings, recordType, record) => 
         if (validateJson(custom_data)) {
           database.update('Setting', {
             key: THIS_STORE_CUSTOM_DATA,
-            value: validateJson(custom_data),
+            value: custom_data,
           });
         }
         if (validateJson(tags)) {
--- a/src/sync/incomingSyncUtils.js
+++ b/src/sync/incomingSyncUtils.js
@@ -674,11 +674,15 @@ export const createOrUpdateRecord = (database, settings, recordType, record) => 
     case 'Store': {
       const { tags, custom_data } = record;
       if (settings.get(THIS_STORE_ID) === record.ID) {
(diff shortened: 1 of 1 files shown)


src/sync/incomingSyncUtils.js
```diff
@@ -674,11 +674,15 @@ export const createOrUpdateRecord = (database, settings, recordType, record) =>
...
-          key: THIS_STORE_CUSTOM_DATA,
-          value: validateJson(custom_data),
-        });
+        if (validateJson(custom_data)) {
+          database.update('Setting', {
+            key: THIS_STORE_CUSTOM_DATA,
+            value: custom_data,
+          });
+        }
+        if (validateJson(tags)) {
```
andreievg: tags are not json btw, just a text string. could also be null though

Also your idea in the PR description sounds like the way to go validateJson -> to return empty string.

joshxg: b23db3d9df301ae8d3341571b58767f6b17fcf68

The real problem is 4D sending a string 'undefined' which throws and error during parsing. 

My reluctance was setting fields as an empty string rather than being ignored. As well as `''` being falsey

andreievg: sure, I was just thinking of a use case where customData is fully removed, was wondering if 4D would send null or 'undefined', then nothing will change in settings sync if it's just ignored

andreievg:
@joshxg, when you are tidying up IC next week, can you make sure this issue is dealt with (as customData for stores was introduces there too)

PR #902, opened 2019-04-19
#882 refactor item getters

### LINKED ISSUE #882: [Vaccine Module] - Refactor Item getters (and maybe batches getters)
@joshxg 
Getters should use realm as much as possible and found some possible issues:

getBatchesInLocation -> should only count batches with non zero numberOfPacks

getTemperatureExposure -> can be smarter and check temperature exposure in location and should use realm really:
```
let sensorLogs = database
      .objects('sensorLogs')
      .filtered('batches.numberOfPacks > 0 AND AND batches.item.id = $0, this.id')

if(location) sensorLogs.filtered('batches.location.id = $0', location.id)

if(sensorLogs.length === 0 ) return { minTemperature: Infinity, maxTemperature: -Infinity };
 
return { minTemperature: sensorLogs.min('temperature'), maxTemperature: sensorLogs.max('temperature') };
```

joshxg:
Fixes #882 

Still some methods that could be removed, but think I should leave them till later.

The `Item.getTemperatureExposure` example in issue gives the temperature exposure for an Item while it was in a certain location. I don't think we should not count the exposure for certain batches while they were in different locations? Happy to change this but it doesn't seem correct to throw away that a batch has potentially been in a lower or higher temperature, just in a different location?

### DIFF
--- a/src/database/DataTypes/Item.js
+++ b/src/database/DataTypes/Item.js
@@ -225,7 +225,7 @@ export class Item extends Realm.Object {
    */
   getBatchesInLocation({ id: locationId } = {}) {
     if (!locationId) return this.batches;
-    return this.batches.filtered('location.id = $0', locationId);
+    return this.batches.filtered('location.id = $0 & numberOfPacks > 0', locationId);
   }
 
   /**
@@ -255,7 +255,7 @@ export class Item extends Realm.Object {
    * @param {Location} location
    */
   getBreachedBatches(location) {
-    return this.getBatchesInLocation(location).filter(hasBreached => hasBreached);
+    return this.getBatchesInLocation(location).filter(({ hasBreached }) => hasBreached);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/database/DataTypes/Item.js
```diff
@@ -225,7 +225,7 @@ export class Item extends Realm.Object {
    */
   getBatchesInLocation({ id: locationId } = {}) {
     if (!locationId) return this.batches;
-    return this.batches.filtered('location.id = $0', locationId);
+    return this.batches.filtered('location.id = $0 & numberOfPacks > 0', locationId);
```
andreievg: do wee need to filter by numberOfPacks first, and then return either this.batches or batches filtered by location ?

joshxg: Good point. Used an existing helper method also instead of explicit query: 8ea020f5999f806219c2e1df4b733f7a6a5aa8a9 not sure if that's the best way to go, but I imagine it's equivalent in the long run

andreievg:
> The Item.getTemperatureExposure example in issue gives the temperature exposure for an Item while it was in a certain location. I don't think we should not count the exposure for certain batches while they were in different locations? Happy to change this but it doesn't seem correct to throw away that a batch has potentially been in a lower or higher temperature, just in a different location?

Hmm, good point, i posed that question to self once, but didn't really have a clear answer. if the column specifically says `Temperature Exposure In ${location}`, then it should be fine to show temp exposure in just that location, but since we are also showing breaches for item (regardless of the location), i think it's fine to show all temperature exposure. User may think item is not breached when looking at specific location where breach did not happen

src/database/DataTypes/Item.js
andreievg: great

Issue #874, opened 2019-04-16
[Vaccine Module] Vaccine Module Page


andreievg:
From epic: sussol/org-issues#23

#### Description:
Create vaccine module page, as per designs (menu buttons on top should be Customer Invoice, Supplier Invoice, Order stock and Manage Stock

![AC4DB839CB41471AACAB805253B6C2F0](https://user-images.githubusercontent.com/26194949/56245292-2f655400-60f3-11e9-981d-ca8346f15b77.png)

PR #903, opened 2019-04-21
#874 [Refinements] vaccine modul page

andreievg:
@joshxg, this pr will have other additions, opening it so that comments can be made

### DIFF
--- a/src/pages/VaccineModulePage.js
+++ b/src/pages/VaccineModulePage.js
@@ -159,7 +159,7 @@ export class VaccineModulePage extends React.Component {
           </View>
           {isFridgeSelected && (
             <View style={{ height: 250, alignSelf: 'stretch' }}>
-              {<VaccineChart {...fridgeChartData} />}
+              {<VaccineChart {...fridgeChartData} hazardPress={this.onHazardPress} />}
             </View>
           )}
         </View>
--- a/src/pages/VaccineModulePage.js
+++ b/src/pages/VaccineModulePage.js
@@ -19,6 +19,8 @@ import globalStyles, {
   HAZARD_RED,
 } from '../globalStyles';
 
+const MILLISECONDS_IN_DAY = 24 * 60 * 60 * 10000;
+const MAX_LOOKBACK_DAYS = 30 * MILLISECONDS_IN_DAY;
 const CHERVON_ICON_STYLE = { size: 18, color: SUSSOL_ORANGE };
 const BREACH_ICON_STYLE = { size: 25, color: HAZARD_RED };
 
@@ -126,6 +128,18 @@ export class VaccineModulePage extends React.Component {
     );
   };
 
+  onHazardPress = breach => {
+    console.log(breach.temperature);
+    // open modal with params, data: [getDataForBreachModal(breach, MAX_BREACH_CHART_DATAPOINS )]
+  };
+
+  extractChartInfo = fridge => {
+    console.log(fridge.description, MAX_LOOKBACK_DAYS);
+    // return reult of extractDataForFridgeChart(fridge, MAX_LOOKBACK_DAYS)
+  };
+
+  renderChart = fridge => <VaccineChart {...this.extractChartInfo(fridge)} />;
+
   /* Fridge and all of it's components */
   renderFridge = fridge => {
     const { sectionStyle, fridgeInfoSectionStyle } = localStyles;
@@ -146,9 +160,7 @@ export class VaccineModulePage extends React.Component {
             </View>
           </View>
           {isFridgeSelected && (
-            <View style={{ height: 250, alignSelf: 'stretch' }}>
-              <VaccineChart {...fridge.getTemperaturePoints()} />
-            </View>
+            <View style={{ height: 250, alignSelf: 'stretch' }}>{this.renderChart(fridge)}</View>
           )}
         </View>
       </View>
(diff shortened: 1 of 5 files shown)


src/utilities/modules/vaccines.js
```diff
@@ -165,3 +165,87 @@ export const sensorLogsExtractBatches = ({ sensorLogs = [], itemBatch, item } =
...
+    isMax: true, // sensorLogs.max('temperature') > temperatureRange.maxTemperature;
+    ...temperatureRange,
+  },
+];
+
+/**
+ * Returns aggregated data for breach modal, based on passed array
+ * of sensorLogs (breaches)
+ */
+export const extractDataForBreachModal = ({
```
andreievg: breaches in this method refers to output of extractBreachPoints. Would be good if it's generic for all use cases though.

src/pages/VaccineModulePage.js
```diff
@@ -19,6 +19,8 @@ import globalStyles, {
   HAZARD_RED,
 } from '../globalStyles';
 
+const MILLISECONDS_IN_DAY = 24 * 60 * 60 * 10000;
```
joshxg: `1000` not `10000` - also I just found this constant in `database/utilities` also

src/utilities/modules/vaccines.js
```diff
@@ -165,3 +165,87 @@ export const sensorLogsExtractBatches = ({ sensorLogs = [], itemBatch, item } =
...
+  breaches,
+  numberOfChartDataPoints,
+  itemFilter,
+  itemBatchFilter,
+}) => {
+  const result = [];
+  breaches.forEach(breach => {
+    const { sensorLogs, maxTemperature, minTemperature, isMax } = breach;
```
joshxg: if we do as above, `extractBreachPoints` can return `isMax`, `maxTemperature`, `minTemperature` ?

src/utilities/modules/vaccines.js
```diff
@@ -165,3 +165,87 @@ export const sensorLogsExtractBatches = ({ sensorLogs = [], itemBatch, item } =
...
+    const numberOfDataPoints =
```
joshxg: There a use case for displaying less data points than breaches?

src/utilities/modules/vaccines.js
```diff
@@ -165,3 +165,87 @@ export const sensorLogsExtractBatches = ({ sensorLogs = [], itemBatch, item } =
...
+const MILLISECONDS_IN_DAY = 24 * 60 * 60 * 10000;
+
+/**
+ * Returns aggregated data for breach modal, based on passed array
+ * of sensorLogs (breaches)
+ */
+export const extractDataForFridgeChart = ({ fridge, MAX_LOOKBACK_DAYS }) => {
+  const { temperatureRange } = fridge;
+
+  const fromTimestamp = new Date(new Date() - MAX_LOOKBACK_DAYS);
```
joshxg: bit of duplication, I think we could either remove timestamp filtering here, add it to `extractBreaches`, or remove `startDate`, `endDate` from `aggreagateLogs`?

src/utilities/modules/vaccines.js
```diff
@@ -165,3 +165,87 @@ export const sensorLogsExtractBatches = ({ sensorLogs = [], itemBatch, item } =
...
+  const sensorLogs = fridge.sensorLogs.filtered('timestamp => $0', fromTimestamp);
+  const chartRangeMilliseconds = sensorLogs.max('timestamp') - sensorLogs.min('timestamp');
+  const numberOfDataPoints = Math.floor(chartRangeMilliseconds / MILLISECONDS_IN_DAY);
+
+  const minLine = aggregateLogs(sensorLogs, numberOfDataPoints, true);
```
joshxg: this will become one call

src/utilities/modules/vaccines.js
```diff
@@ -165,3 +165,87 @@ export const sensorLogsExtractBatches = ({ sensorLogs = [], itemBatch, item } =
...
+  const breaches = [
+    ...extractBreachPoints(minLine, fullBreaches, temperatureRange),
+    ...extractBreachPoints(maxLine, fullBreaches, temperatureRange),
+  ];
+
+  return {
+    minLine,
+    maxLine,
+    breaches,
+    onHazardPress: this.onHazardPress,
```
joshxg: hm, not sure what this will be. I don't think passing a `hazardPress` function to this method is right. I think just setting the prop in the component as per normal is probably best

src/utilities/modules/vaccines.js
andreievg: i scrapped that

src/pages/VaccineModulePage.js
andreievg: fixed in last commit

src/utilities/modules/vaccines.js
andreievg: this for breach modal chart, if there are lot's of sensorLogs, we only really want to show 'X' amount number of points on graph

src/utilities/modules/vaccines.js
andreievg: that was changes in last commit

src/utilities/modules/vaccines.js
andreievg: ok true

src/utilities/modules/vaccines.js
andreievg: adding to do in next commit

src/utilities/modules/vaccines.js
andreievg: that was copy paste from VaccineModulePage btw

PR #906, opened 2019-04-21
few fixes

andreievg:


andreievg:
i used this file with sensor logs and some batches for main fridge:

[default.realm.zip](https://github.com/openmsupply/mobile/files/3101643/default.realm.zip)


Also the sensor log creation: 

https://github.com/openmsupply/mobile/compare/VM-post-merge-fixes...VM-sensor-log-creation

src/database/DataTypes/Item.js
```diff
@@ -284,12 +284,16 @@ export class Item extends Realm.Object {
...
+    let sensorLogs = location ? location.getSensorLogs(database) : database.objects('SensorLog');
+    sensorLogs = sensorLogs.filtered(
+      'itemBatches.item.id = $0 && itemBatches.numberOfPacks > 0',
+      this.id
+    );
+
     return {
-      maxTemperature: batches.max('sensorLog.temperature') || -Infinity,
-      minTemperature: batches.min('sensorLog.temperature') || Infinity,
+      maxTemperature: sensorLogs.max('temperature') || -Infinity,
```
joshxg: Looks like I misspelled `sensorlog` -> `sensorLogs`. Did `batches.max('sensorLogs.temperature')` not work either? I remember testing this type of query.

Either way I think this will return the incorrect value when the `temperature` is `0`

andreievg: sure, will change

looks like max and min only works on immidiate properties.

PR #910, opened 2019-04-22
#909 customer invoice breach column

### LINKED ISSUE #909: [VaccineModule] Customer invoice breach column
Require a column `Breach` on customer invoices when a `Vaccine` item has been added to the invoice.

Will open a `BreachTable` in a modal for that particular item.

joshxg:
Fixes #909 

- Adds a Breach column to `CustomerInvoicePage`
- Small update to `vaccines.js`
    - Was not filtering the `ItemBatches` by `Item` or `ItemBatch`
    - Added data for the date column for a breach
    - Double use of variable `Item` (Used as a parameter

I couldn't not trace it properly, but if I did not set `breachData` as state, and instead just used a `getBreachData()` method for the `BreachModal` to get it's data, on navigating back from `CustomerInvoicePage`, when the modal was open - some infinite recursion was happening? Not a fan of putting this in state, so will take another look most likely to make sure there isn't a way to not have this data in state.

### DIFF
--- a/src/database/DataTypes/Item.js
+++ b/src/database/DataTypes/Item.js
@@ -301,7 +301,9 @@ export class Item extends Realm.Object {
   }
 
   getAllSensorLogs(database) {
-    return database.objects('SensorLog').filtered('itemBatches.item.id = $0', this.id);
+    return database
+      .objects('SensorLog')
+      .filtered('itemBatches.item.id = $0 && itemBatches.numberOfPacks > 0', this.id);
   }
 }
 
--- a/src/database/DataTypes/Item.js
+++ b/src/database/DataTypes/Item.js
@@ -299,6 +299,10 @@ export class Item extends Realm.Object {
       minTemperature: sensorLogs.min('temperature'),
     };
   }
+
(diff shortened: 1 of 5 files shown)


src/database/DataTypes/Item.js
```diff
@@ -299,6 +299,10 @@ export class Item extends Realm.Object {
       minTemperature: sensorLogs.min('temperature'),
     };
   }
+
+  getAllSensorLogs(database) {
+    return database.objects('SensorLog').filtered('itemBatches.item.id = $0', this.id);
+  }
```
andreievg: Do we need to check for itemBatches.numberOfPacks not zero here?

joshxg: fabcdb468a1da7cae9d7fe9fd6501e19f0eee28f

joshxg:
@andreievg Will give it a try after next APK - I have a feeling the timing may be off (rendering the spinner midway through the modal opening animation/slide in which is not great) but seems like the way to go if I can't work out why the call stack is overflowing

Issue #904, opened 2019-04-21
[Vaccine Module] Manage vaccine item disposal


joshxg:
#851 introduces a vaccine item page, with a `dispose` column. This needs to be hooked up.

Disposal will require a `reason`  - coming from the `Options` table.

Disposal via a failed VVM test will be `vvmFailed` (or similar).

PR #912, opened 2019-04-22
Vaccine item add breach modal

### LINKED ISSUE #911: [Vaccine Module] Breach modals on remaining pages
`VaccineManageStockPage` and `VaccineManageItemPage` both have a column with a breach indicator. This indicator is pressable, where a breach modal will render and show details for that breach, for that particular `Item` or `ItemBatch`

Require both to be hooked up.

joshxg:
Fixes #911 

#### NOTE: Branched from #904 

- Adds the breach modal to `ManageVaccineItemPage` and `ManageVaccineStockPage`.
- Still using `breachData` in state for now - going to be branching off this so I think refactoring afterwards is the best way to move forward
- Adds a title to the breach modal is `VaccineModulePage`


joshxg:
closed to do fixes now rather than later

PR #923, opened 2019-04-26
#920 fix gradle

### LINKED ISSUE #920: [DevOps] Wrong ARM architecture in build.gradle
Description: In the big upgrade of '19, we added code to `build.gradle` which tries to create a build for ARM64-v8a, which is not supported by our dependencies.

- react-native 0.59 supports ARM64-v8a & x86-64, we are currently on 0.58.4 https://github.com/facebook/react-native/commit/f3e5cce4745c0ad9a5c697be772757a03e15edc5
- Realm does not support 64 bit apps at the moment https://github.com/realm/realm-js/issues/2221

This issue is to fix the current build problems, however we need to keep realm and react-native upgrades in mind as no new apps will be accepted to google play store if they are not 64 bit.

https://www.androidpolice.com/2019/01/16/play-store-to-phase-out-32-bit-apps-on-64-bit-devices-by-2021-all-new-apps-from-august-1-must-be-64-bit/

Comments:

Need to remove all traces of `arm65-v8a` in `build.gradle`

joshxg:
Fixes #920 

#### NOTE: This should probably be merged into IC branch also

- Ensures only a 32bit build is compiled

Without this, the app fails to load on certain tablets, e.g. Lenovo. I haven't had a look yet at the Lenovo architecture so I am not sure exactly why the builds were working on other 64 bit tablets.

As per the issue, our dependencies do not currently support compiling for 64bit architectures regardless

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -115,7 +115,7 @@ android {
         versionCode appVersionCode
         versionName getAppVersion()
         ndk {
-            abiFilters "armeabi-v7a", "x86", "arm64-v8a"
+            abiFilters "armeabi-v7a", "x86"
         }
     }
     signingConfigs {
@@ -131,7 +131,7 @@ android {
             reset()
             enable enableSeparateBuildPerCPUArchitecture
             universalApk false  // If true, also generate a universal APK
-            include "armeabi-v7a", "x86", "arm64-v8a"
+            include "armeabi-v7a", "x86"
         }
     }
     buildTypes {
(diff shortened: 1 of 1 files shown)


joshxg:
@sworup  What do you mean these changes will refuse to compile?

and yes, we need to upgrade to react-native 0.59 - and realm is looking to be 64bit compatible before august, when google play will stop letting new 32 bit apps on google play store

sworup:
> @sworup What do you mean these changes will refuse to compile?

Oh it will compile but wouldn't run, got it. :)

joshxg:
@sworup you mean before this change? Sorry I am just making sure I am not introducing more bugs! :)

sworup:
> @sworup you mean before this change? Sorry I am just making sure I am not introducing more bugs! :)

I have not compiled the code yet.

I presume the code will compile as usual but after these changes wouldn't even install on 64bit application. Previously it would install but fail to load. I do not think these changes would affect 32bit system. 

We should do a test compile and install on 32bit tablet but I am occupied with something else, I can only do that on Monday.

PR #932, opened 2019-04-27
Vm supplier invoice ia

### LINKED ISSUE #842: [Vaccine Module][VVM Inventory Adjustment] - Create generic Inventory Adjustment helper for VVM statuses
As per epic: https://github.com/sussol/org-issues/issues/23

#### Description

Both Supplier Invoices and Manage Vaccine Item will have VVM toggles. VVM Status needs to be recorded in inventory adjustment  `transactionBatch.isVVMPassed`.

We need generic way of creating these adjustments.

Input will be an array of transactionBatches (some with isVVMPassed = null)

### LINKED ISSUE #926: [VaccineModule] Location not set
Description: When adding a vaccine item in `SupplierInvoicePage`, a default fridge is applied. This fridge is however not set in `TransactionBatch.location`. A user must choose a different fridge, and then that original one to actually set it.

Reproducible: Yes

Reproduction Steps:
1) Add an item to a supplier invoice.
2) Change the quantity and finalize
3) Viewing vaccine stock by the location which was set will not show that batch, however showing all locations will

Comments:

joshxg:
Fixes #842 Fixes #926 

- Adds a generic inventory adjustment method for vaccines
    - It would be easy to add an additional adjustment here for location changes when/if needed
    - Maybe in a different place?
- Adds a `navigateBack` method to redux, passed to every page
- Changes `ManageVaccineItem` to ensure each split batch has a property `parentBatch`, which is the `ItemBatch` split from.
- Adds `GenericChoiceList` to `ManageVaccineItemPage` for location selection
- Fixes a bug for `LocationFilter` on `ManageVaccineStockPage`


joshxg:
Added commits:

8cad75fbc10c7beb06cc1102fd29670ac43eaf69

c280ff5fd53b0ae8387b663a4ca51485cddc18cd

PR #937, opened 2019-04-28
#936 price and unit columns

### LINKED ISSUE #936: [Ivory Coast] Price and Unit columns on program requisitions
Description:

Require price and unit columns on a program requisition.

Comments:

Prices should come from the suppliers master list, for the supplier of the requisition.
Units need to come from `[Units]` table linked by `[Item]unitID`.

- I cannot see a better solution than to just add the `Unit` table to mobile. Could just store the `unitID` in a `Item.unitId` field, and check all `Items` for any `Item.unit = Unit.ID` on incoming sync, and update it with the actual unit, but I think this is a lot of work and heavy strain on performance for avoiding adding a simple table.

joshxg:
Fixes #936 

### NOTE: This finds a price from the Supplier masterlist - the cost price. Should this be finding the price from the current stores masterlist?

- Add `Unit` table
- Add `Item.unit` field
- Add `MasterListItem.price` field
- Add Price column to a program requisition
- Add a unit column to a program requisition (?)

#### NOTE: Does NOT use `[item]default_price` - waiting for #821 ? Just assigns 'N/A' to any unit or price which has not been found.

Could maybe? be more performant. At the moment creates the`ITEM_PRICE_MAPPING` table for all `MasterListItems` (`n`). Could potentially be faster by iterating through items for the requisition (`m`) and performing a realm query for each, finding the max price within the `MasterListItems` related. Essentially trading `n` assignments for `m` queries. I dunno if it's worth it - I think the simpler logic is best.

### DIFF
--- a/src/database/DataTypes/Item.js
+++ b/src/database/DataTypes/Item.js
@@ -231,6 +231,7 @@ Item.schema = {
     defaultPrice: { type: 'double', optional: true },
     isVisible: { type: 'bool', default: false },
     crossReferenceItem: { type: 'Item', optional: true },
+    UnitId: { type: 'Unit', optional: true },
   },
 };
 
--- a/src/database/DataTypes/Item.js
+++ b/src/database/DataTypes/Item.js
@@ -231,7 +231,7 @@ Item.schema = {
     defaultPrice: { type: 'double', optional: true },
     isVisible: { type: 'bool', default: false },
     crossReferenceItem: { type: 'Item', optional: true },
-    unitId: { type: 'Unit', optional: true },
+    unit: { type: 'Unit', optional: true },
   },
 };
 
--- a/src/database/DataTypes/Item.js
+++ b/src/database/DataTypes/Item.js
@@ -231,7 +231,7 @@ Item.schema = {
     defaultPrice: { type: 'double', optional: true },
     isVisible: { type: 'bool', default: false },
     crossReferenceItem: { type: 'Item', optional: true },
-    UnitId: { type: 'Unit', optional: true },
+    unitId: { type: 'Unit', optional: true },
   },
 };
 
--- a/src/database/DataTypes/Item.js
+++ b/src/database/DataTypes/Item.js
@@ -149,7 +149,7 @@ export class Item extends Realm.Object {
 
   /**
    * Returns a string representing the units for this item.
-   * @return {string} the unit for this item, or N/A if none has been assgined.
+   * @return {string} the unit for this item, or N/A if none has been assigned.
    */
   get unitString() {
     return (this.unit && this.unit.units) || 'N/A';
--- a/src/database/DataTypes/Item.js
+++ b/src/database/DataTypes/Item.js
@@ -148,6 +148,14 @@ export class Item extends Realm.Object {
   }
 
   /**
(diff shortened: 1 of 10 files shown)


Chris-Petty:
My understanding is that using the supplier's master list pricing is correct. @craigdrown might chime in but I'm going to go ahead assuming that's correct.

I like simplier logic. If performance is found to be an issue then we can look at optimisations. Even then, we know big bottle necks already that'd likely be more worthwhile.

joshxg:
@Chris-Petty after talking with Craig, the customer price should be used. It seems because there is a `bool` `price_list` on `[list_master_name_join]`, if there was a master list related to a store which was delegated as their 'price' list, if there were any other related master list with this `bool` related to the store, it wouldn't be possible to determine which was the price list? **I could be wrong here, I've only just had a quick look at pricing etc**

So I decided to just take the max price of all `list_master_item` for a specific `Item`.

Chris-Petty:
Sounds solid enough to get out of the gate. Can review it at a later stage once it's been live.

src/sync/incomingSyncUtils.js
```diff
@@ -233,6 +234,10 @@ export const sanityCheckIncomingRecord = (recordType, record) => {
       cannotBeBlank: ['name', 'startDate', 'endDate', 'periodScheduleID'],
       canBeBlank: [],
     },
+    Unit: {
+      cannotBeBlank: [],
+      canBeBlank: ['units', 'comment', 'order_number'],
```
Chris-Petty: Similar to my comment on the schema, I think 'units' should be 'cannotBeBlank' cause the record is screwed without it.

src/database/DataTypes/Unit.js
```diff
@@ -0,0 +1,24 @@
...
+ * TODO
+ */
+export class Unit extends Realm.Object {}
+
+Unit.schema = {
+  name: 'Unit',
+  primaryKey: 'id',
+  properties: {
+    id: 'string',
+    units: { type: 'string', default: 'Each' },
```
Chris-Petty: I'm wondering if this should actually be optional rather than defaulted. In mSupply desktop there doesn't seem to be a default "Each" set anywhere? As far as I can tell you cannot make/save a unit with this field blank.

joshxg: oops I forgot about this - I didn't know what to do so I did that!

joshxg: Done in commits:

- 257ec37fe8a3161ce10fa8ed0b51f8e67f6c3df3
- 14de00471670b039d0c2a74e6a0793bc3f83eb51
- c0d00463540ac713ca67477db8d206f144ab896a

src/database/DataTypes/Item.js
```diff
@@ -147,6 +147,14 @@ export class Item extends Realm.Object {
     return this.batchesWithStock.length;
   }
 
+  /**
+   * Returns a string representing the units for this item.
+   * @return {string} the unit for this item, or N/A if none has been assgined.
```
Chris-Petty:
```suggestion
   * @return {string} the unit for this item, or N/A if none has been assigned.
```

PR #942, opened 2019-04-30
Add bug fix for requisitions nto being cancelled

### LINKED ISSUE #924: [Ivory Coast] Requisition made after cancelling supplier modal
Description:

Cancelling the supplier modal for a requisition creates the requisition anyway

Reproducible:
yes?

Reproduction Steps:
1. Go to supplier requisitions
2. Click new requisition
3. cancel the modal
4. see your requisition

Comments:
`onNewRequisition` method is called on both selecting a supplier and cancelling the modal. Should cancel if no parameters are passed, however it is currently passed the event object, which makes the conditional return true, not closing the modal

joshxg:
Fixes #924 

- Event object was being passed, which caused the check for a parameter being passed return true, creating the requisition.

### DIFF
--- a/src/pages/SupplierRequisitionsPage.js
+++ b/src/pages/SupplierRequisitionsPage.js
@@ -239,7 +239,7 @@ export class SupplierRequisitionsPage extends React.Component {
           queryString="name BEGINSWITH[c] $0"
           sortByString="name"
           onSelect={name => this.onNewRequisition({ otherStoreName: name })}
-          onClose={this.onNewRequisition}
+          onClose={() => this.onNewRequisition()}
           title={modalStrings.search_for_the_supplier}
         />
 


src/pages/SupplierRequisitionsPage.js
```diff
@@ -239,7 +239,7 @@ export class SupplierRequisitionsPage extends React.Component {
           queryString="name BEGINSWITH[c] $0"
           sortByString="name"
           onSelect={name => this.onNewRequisition({ otherStoreName: name })}
-          onClose={this.onNewRequisition}
+          onClose={() => this.onNewRequisition()}
```
Chris-Petty: This seems functionally equivalent?

joshxg: as above:

Event object was being passed, which caused the check for a parameter being passed return true, creating the requisition.

This results in no parameter being passed, and the check to then evaluate to false, closing the modal

Chris-Petty: Ah right. I think I raised somewhere (maybe I just thought it) when looking at this in the past that passing the same method `onNewRequisition` into onClose wasn't well separated, that is to say another method would be cleaner:
```javascript
onCancelNewRequisition = () => this.setState({ byProgramModalOpen: false, isCreatingRequisition: false });
```
Seeing as I just looked at it,  I think `byProgramModalOpen`, `isCreatingRequisition` and `usesPrograms` could be used a bit differently less complexly.

PR #944, opened 2019-04-30
Add correct check for having to have a reason

### LINKED ISSUE #943: [Ivory Coast][Stocktake] reason modal opening
Description:

When an item already has a reason, the reason modal still opens. Not game breaking but annoying

Reproducible:

Assign a reason to an item, edit the quantity again to be different to the snapshot - modal opens

### LINKED ISSUE #946: [Ivory Coast] Fields disabled/enabled in incorrect state
Description:

- Couple of fields are editable during finalized transactions
    - i.e reason in a finalized stocktake
- Couple of fields are disabled incorrectly - e.g. a reason for a stocktake line when it has a reason already (can't change a reason)

joshxg:
Fixes #943 Fixes #946 

- Does a correct check for every `StocktakeBatch` having a reason where needed and doesn't open a modal when one doesn't need to be assigned

- Adds correct checking for finalized state for both `StocktakeEditPage` & `StocktakeBatchModal`


src/database/DataTypes/StocktakeItem.js
```diff
@@ -305,7 +305,16 @@ export class StocktakeItem extends Realm.Object {
...
+    return this.batches.every(batch => {
+      if (batch.snapshotTotalQuantity !== batch.countedTotalQuantity) {
+        return !!batch.option;
+      }
+      return true;
+    });
+  }
+
+  get hasAnyReason() {
+    return this.batches.some(({ option }) => !!option);
```
Chris-Petty:
```suggestion
    return this.batches.some(batch => !!batch.option);
```
I personally find this clearer, but maybe I'm just behind the times. It just reads slightly more obviously to me that it's "in `batches` some `batch` has an option".

src/database/DataTypes/StocktakeItem.js
```diff
@@ -305,7 +305,16 @@ export class StocktakeItem extends Realm.Object {
    */
   get hasReason() {
     if (!this.batches) return false;
-    return this.batches.some(batch => batch.option);
+    return this.batches.every(batch => {
+      if (batch.snapshotTotalQuantity !== batch.countedTotalQuantity) {
+        return !!batch.option;
+      }
+      return true;
+    });
```
Chris-Petty: I feel like the getter name and description don't match this anymore.

I wonder if it's useful semantically to not have these as getters. Idk:

src/database/DataTypes/StocktakeItem.js
```diff
@@ -305,7 +305,16 @@ export class StocktakeItem extends Realm.Object {
    */
   get hasReason() {
```
Chris-Petty: The 4 getters at the end of this file should be above the class methods with the rest of the getters.

src/database/DataTypes/StocktakeItem.js
```diff
@@ -305,7 +305,16 @@ export class StocktakeItem extends Realm.Object {
...
     if (!this.batches) return false;
-    return this.batches.some(batch => batch.option);
+    return this.batches.every(batch => {
```
Chris-Petty: While I say in `get hasAnyReason` I like just using `batch`, I think here destructuring into `{ snapshotQuanity, countedTotalQuantity, option }` could tidy this up:

```javascript
    return this.batches.every(({ snapshotTotalQuantity, countedTotalQuantity, option }) =>
      snapshotTotalQuantity !== countedTotalQuantity ? !!option : true
    );
```

src/database/DataTypes/StocktakeItem.js
```diff
@@ -315,9 +324,9 @@ export class StocktakeItem extends Realm.Object {
    * @return {bool}
    */
   get shouldHaveReason() {
-    const { snapshotTotalQuantity, countedTotalQuantity } = this;
+    const { snapshotTotalQuantity, countedTotalQuantity, hasReason } = this;
     const equalQuantities = snapshotTotalQuantity === countedTotalQuantity;
-    return !equalQuantities;
+    return !(equalQuantities || hasReason);
```
Chris-Petty: I feel like `hasReason` isn't accurate to its name anymore and its new implementation makes `equalQuantities` redundant. So I'd just nix `hasReason` and change `shouldHaveReason` do what `hasReason` does in this PR.
```javascript
  /**
   * Returns true if the snapshot and counted quantities on any batch differ.
   * Does not account for if reasons are used by the user, caller
   * needs to account for this.
   * @return {bool}
   */
  get shouldHaveReason() {
    if (!this.batches) return false;
    return this.batches.every(({ snapshotTotalQuantity, countedTotalQuantity, option }) =>
      snapshotTotalQuantity !== countedTotalQuantity ? !!option : true
    );
  }
```

Chris-Petty: I got wondering why you changed from `some` to `every`. Interestingly a key difference between `some` and `every` is that on an empty array the former will always return `falsy` and the latter `truthy`. So I think you could kill the early return 

```javascript
get shouldHaveReason() {
  return this.batches.some(({ snapshotTotalQuantity, countedTotalQuantity, option }) =>
    snapshotTotalQuantity !== countedTotalQuantity ? !option : false
  );
}
```

joshxg: 930e6108c27d8e4bf53bbd50af875b09b4c05758 
8caa0a429e1e2844ada6d04da1797770bac010b5

src/database/DataTypes/StocktakeItem.js
joshxg: can I disagree with this and just not do it?

Chris-Petty: Yup totally

PR #951, opened 2019-05-01
#950 customer requisition null checks 

### LINKED ISSUE #950: Customer requisition finalisation crash
Build Number: 2.3.0-rc0 dev+apk

Description: 
image

Reproducible: yes

Reproduction Steps:
1. receive a customer requisition
2. enter a value
3. finalise it
4. some loading spinner and a RSoD (in dev, crash in apk)

Comments: There were items in the customer requisition that my store didn't actually have visible, but were visible in the requisition regardless.

Chris-Petty:
Fixes #950

### DIFF
--- a/src/database/DataTypes/RequisitionItem.js
+++ b/src/database/DataTypes/RequisitionItem.js
@@ -93,11 +93,10 @@ export class RequisitionItem extends Realm.Object {
    * @return  {TransactionItem}
    */
   get linkedTransactionItem() {
-    // TODO: RequisitionItem.Requisition is optional field, unhandled possible TypeError.
-    // TODO: isRequest is Requisition field, will always return undefined.
-    if (this.isRequest || !this.requisition.linkedTransaction) return null;
+    if (this.requisition.isRequest || !this.requisition.linkedTransaction) {
+      return null;
+    }
 
-    // TODO: Requisition.linkedTransaction is optional field, unhandled possible TypeError.
     return this.requisition.linkedTransaction.items.filtered('item.id == $0', this.item.id)[0];
   }
 
@@ -108,15 +107,11 @@ export class RequisitionItem extends Realm.Object {
    * @return  {number}
    */
   get ourStockOnHand() {
-    // TODO: this.linkedTransactionItem and this.item can both be null or undefined, unhandled
-    // possible TypeError.
+    if (!this.linkedTransactionItem || !this.item) return null;
     const { availableQuantity } = this.linkedTransactionItem;
     const { totalQuantity } = this.item;
 
-    // TODO: this.linkedTransactionItem and this.item can both be falsey, unhandled possible return
-    // undefined.
-    if (availableQuantity) return availableQuantity;
-    return totalQuantity;
+    return availableQuantity || totalQuantity;
   }
 
   /**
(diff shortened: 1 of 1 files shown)


src/database/DataTypes/RequisitionItem.js
```diff
@@ -93,11 +93,10 @@ export class RequisitionItem extends Realm.Object {
...
   get linkedTransactionItem() {
-    // TODO: RequisitionItem.Requisition is optional field, unhandled possible TypeError.
-    // TODO: isRequest is Requisition field, will always return undefined.
-    if (this.isRequest || !this.requisition.linkedTransaction) return null;
+    if (this.requisition.isRequest || !this.requisition.linkedTransaction) {
+      return null;
+    }
 
-    // TODO: Requisition.linkedTransaction is optional field, unhandled possible TypeError.
     return this.requisition.linkedTransaction.items.filtered('item.id == $0', this.item.id)[0];
```
wlthomson: Think this can be better written (more consistent with existing codebase which uses boolean short-circuits for this stuff).

```
if (this.requisition.isRequest) return null;
return this.requisition.linkedTransaction && this.requisition.linkedTransaction.items.filtered('item.id == $0', this.item.id)[0];
```

src/database/DataTypes/RequisitionItem.js
```diff
@@ -108,15 +107,11 @@ export class RequisitionItem extends Realm.Object {
    * @return  {number}
    */
   get ourStockOnHand() {
-    // TODO: this.linkedTransactionItem and this.item can both be null or undefined, unhandled
-    // possible TypeError.
+    if (!this.linkedTransactionItem || !this.item) return null;
```
wlthomson: Is this correct? We want to return null if either `linkedTransactionItem` OR `item` are null? Shouldn't this be return null only if BOTH are null?

src/database/DataTypes/RequisitionItem.js
Chris-Petty: This would change the behaviour a little bit, if `this.requisition.linkedTransaction` is falsy then the method will return `false` rather than `null`. I'd rather keep it returning null

Chris-Petty: I think? I want to test this now.

Chris-Petty: I'm wrong, though if x is `undefined`, it'll return that instead of `null` which still is a bit inconsistent.

src/database/DataTypes/RequisitionItem.js
Chris-Petty: Using AND would not return if item is defined but linkedTransactionItem is not. Then the next line would TypeError crash the app?

src/database/DataTypes/RequisitionItem.js
wlthomson: Shouldn't be an issue, optional fields in Realm should be null if not defined otherwise.

src/database/DataTypes/RequisitionItem.js
wlthomson: Yeah, the next lines would have to change. What I'm asking is, it looks like we returns values based on which fields are defined, but we return prematurely if neither one of their parent objects are defined. Shouldn't we check the parent items in the same way, rather than returning prematurely if neither are defined?

Chris-Petty: Ah I see, will address

Chris-Petty:
@wlthomson pls merge, tested in build, mission accomplished

PR #978, opened 2019-05-16
#977 full aggregation rewrite

### LINKED ISSUE #977: Full aggregator rewrite
Build Number:
2.2.0 rc-0?
Description:
The full aggregator currently used in the vaccines module is a little convoluted, needs to be rewritten to be a little more simple.

Reproducible:
N/A

Reproduction Steps:
N/A

Comments:

joshxg:
Fixes #977 

- Adds a new full aggregation method
- Does not use the most recent 2 hours of sensor logs.
- Deletes all other pre-aggregated logs

- I think the time constants should be in their own file
- Could use the generic sensorlog create method in other methods when theyre cleaned up

- think I need to take another go over to make sure it's as defensive as possible

### DIFF
--- a/src/utilities/modules/temperatureSensorHelpers.js
+++ b/src/utilities/modules/temperatureSensorHelpers.js
@@ -12,7 +12,6 @@ const ONE_HOUR_MILLISECONDS = ONE_MINUTE_MILLISECONDS * 60;
 const EIGHT_HOURS_MILLISECONDS = ONE_HOUR_MILLISECONDS * 8;
 const FIVE_MINUTES_MILLISECONDS = ONE_MINUTE_MILLISECONDS * 5;
 const NO_FULL_AGGREGATE_PERIOD = EIGHT_HOURS_MILLISECONDS * 4 + FIVE_MINUTES_MILLISECONDS;
-const SENSOR_LOG_NO_DELETION_INTERVAL = ONE_HOUR_MILLISECONDS * 24 * 3;
 // Include a small offset on the interval to account for each timestamp being
 // small amounts of time off.
 const FULL_AGGREGATION_INTERVAL = EIGHT_HOURS_MILLISECONDS + FIVE_MINUTES_MILLISECONDS;
@@ -475,7 +474,6 @@ function doFullAggregation({ result, sensor, database }) {
   const mostRecentPreAggregateTimestamp = sensorLogs
     .filtered('aggregation == $0', SENSOR_LOG_PRE_AGGREGATE_TYPE)
     .max('timestamp');
-  const maxTimestampForDeletion = new Date().getTime() - SENSOR_LOG_NO_DELETION_INTERVAL;
   // Starting timestamp is either from the first fullAggregate timestamp, or beginning of time
   const aggregationIntervalStart = (mostRecentFullAggregateTimestamp || -Infinity) + 1;
   // Ending timestamp is either from the most recent preAggregate or from now, less the
--- a/src/utilities/modules/temperatureSensorHelpers.js
+++ b/src/utilities/modules/temperatureSensorHelpers.js
@@ -448,7 +448,7 @@ function doFullAggregation({ result, sensor, database }) {
     sensorLogs.filtered('aggregate == $0', 'preAggregate').max('timestamp') -
     TWO_HOURS_MILLISECONDS;
   const sortedSensorLogs = sensorLogs
-    .filtered('aggregate == $0 && timestamp < $1', 'preAggregate', maxTimeToAggregate)
+    .filtered('aggregation == $0 && timestamp < $1', 'preAggregate', maxTimeToAggregate)
     .sorted('timestamp');
   // Initialized for creating a timestamp of the first sensor log for
   // a sequential group of logs, which are delimited by a sensorlog
--- a/src/utilities/modules/temperatureSensorHelpers.js
+++ b/src/utilities/modules/temperatureSensorHelpers.js
@@ -470,10 +470,10 @@ function doFullAggregation({ result, sensor, database }) {
   // aggregate the most recently added preAggregate logs, from the most
   // most recent fullAggregation log.
   const mostRecentFullAggregateTimestamp = sensorLogs
-    .filtered('aggregate == $0', SENSOR_LOG_FULL_AGGREGATE_TYPE)
+    .filtered('aggregation == $0', SENSOR_LOG_FULL_AGGREGATE_TYPE)
     .max('timestamp');
   const mostRecentPreAggregateTimestamp = sensorLogs
-    .filtered('aggregate == $0', SENSOR_LOG_PRE_AGGREGATE_TYPE)
+    .filtered('aggregation == $0', SENSOR_LOG_PRE_AGGREGATE_TYPE)
     .max('timestamp');
   const maxTimestampForDeletion = new Date().getTime() - SENSOR_LOG_NO_DELETION_INTERVAL;
   // Starting timestamp is either from the first fullAggregate timestamp, or beginning of time
--- a/src/utilities/modules/temperatureSensorHelpers.js
+++ b/src/utilities/modules/temperatureSensorHelpers.js
@@ -391,13 +391,15 @@ function linkSensorLogToItemBatches({ sensorLog, database }) {
     ...sensorLog,
     itemBatches,
   });
-  itemBatches.forEach(itemBatch =>
+  itemBatches.forEach(itemBatch => {
+    itemBatch.addSensorLog(sensorLog);
+    itemBatch.save();
     database.update('SensorLogItemBatchJoin', {
       id: generateUUID(),
       itemBatch,
       sensorLog,
-    })
-  );
+    });
+  });
 }
 
 function createGenericSensorLog({
@@ -524,7 +526,9 @@ function doFullAggregation({ result, sensor, database }) {
   const aggregatedSensorLogs = sensorLogGroups
     .map(group => createFullAggregateSensorLogs(group))
     .flat();
-  const sensorLogsToDelete = sortedSensorLogs.filtered('timestamp < $0', maxTimestampForDeletion);
+  const deleteFromTimestamp =
+    sensorLogGroup.length > 0 ? sensorLogGroup[0].timestamp : aggregationIntervalEnd;
+  const sensorLogsToDelete = sortedSensorLogs.filtered('timestamp < $0', deleteFromTimestamp);
   // Store each aggregated sensorlog in the database and link the item batches
   // currently in the same location. Also create a sensorLogItemBatchJoin record.
   database.write(() => {
@@ -540,7 +544,7 @@ function doFullAggregation({ result, sensor, database }) {
   return {
     ...result,
     fullAggregateAdditions: aggregatedSensorLogs.length,
-    preAggregateDeletions: sortedSensorLogs.length,
+    preAggregateDeletions: sensorLogsToDelete.length,
   };
 }
 
--- a/src/utilities/modules/temperatureSensorHelpers.js
+++ b/src/utilities/modules/temperatureSensorHelpers.js
@@ -1,3 +1,4 @@
+/* eslint-disable no-console */
 /* eslint-disable no-continue */
 import { generateUUID } from 'react-native-database';
 import { NativeModules } from 'react-native';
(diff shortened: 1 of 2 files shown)


src/utilities/modules/temperatureSensorHelpers.js
```diff
@@ -370,87 +378,119 @@ function addHeadAndTrailingLogToBreach({ result, sensor, database }) {
...
+ */
+function doFullAggregation({ result, sensor, database }) {
+  const { sensorLogs } = sensor;
+  if (!sensorLogs || sensorLogs.length === 0) return null;
+  // Want to leave the last 2 hours of logs un-aggregated. Filter them
+  // out.
+  const maxTimeToAggregate =
+    sensorLogs.filtered('aggregate == $0', 'preAggregate').max('timestamp') -
+    TWO_HOURS_MILLISECONDS;
+  const sortedSensorLogs = sensorLogs.filtered(
```
andreievg: Can you get this sorted

src/utilities/modules/temperatureSensorHelpers.js
```diff
@@ -370,87 +378,119 @@ function addHeadAndTrailingLogToBreach({ result, sensor, database }) {
...
+    }
   });
-
-  return { ...result, fullAggregateAdditions: logsToAdd.length, fullAggregationDeletions };
+  // Create an array of aggregated sensor log objects
+  const aggregatedSensorLogs = sensorLogGroups
+    .map(group => createFullAggregateSensorLogs(group))
+    .flat();
+  // Store each aggregated sensorlog in the database.
+  aggregatedSensorLogs.forEach(aggregatedLog => database.update('SensorLog', aggregatedLog));
```
andreievg: I think we are not in write transaction atm

joshxg: Have added this here: cdff488fe5270c2e3e767ddbe43af57e331f171c

I actually wonder if write transactions shouldn't be used within methods. Means you cannot use this method at all if the caller is already in a write transaction (creates situations as with settings)

src/utilities/modules/temperatureSensorHelpers.js
joshxg: oops 3a18cca9ff6858e75629009e3d9d62776b07319d

src/utilities/modules/temperatureSensorHelpers.js
```diff
@@ -370,87 +378,119 @@ function addHeadAndTrailingLogToBreach({ result, sensor, database }) {
...
-
-    logsToAdd.push({
-      id: generateUUID(),
-      location: sensor.location,
-      sensor,
-      temperature: thisPeriodLogs.max('temperature'),
-      timestamp: aggregateMiddlePoint,
-      aggregation: 'aggregate',
-    });
+function createGenericSensorLog({
```
joshxg: UUID's are generated here

src/utilities/modules/temperatureSensorHelpers.js
```diff
@@ -370,87 +378,119 @@ function addHeadAndTrailingLogToBreach({ result, sensor, database }) {
...
+function doFullAggregation({ result, sensor, database }) {
+  const { sensorLogs } = sensor;
+  if (!sensorLogs || sensorLogs.length === 0) return null;
+  // Want to leave the last 2 hours of logs un-aggregated. Filter them
+  // out.
+  const maxTimeToAggregate =
+    sensorLogs.filtered('aggregate == $0', 'preAggregate').max('timestamp') -
+    TWO_HOURS_MILLISECONDS;
+  const sortedSensorLogs = sensorLogs
+    .filtered('aggregate == $0 && timestamp < $1', 'preAggregate', maxTimeToAggregate)
```
andreievg: @joshxg 'aggregation'

src/utilities/modules/temperatureSensorHelpers.js
```diff
@@ -370,87 +378,119 @@ function addHeadAndTrailingLogToBreach({ result, sensor, database }) {
...
+  const sensorLogGroups = [];
+  // A collection of sensor logs, grouped for the currently being operated on,
+  // < 8 hour period.
+  let sensorLogGroup = [];
+  sortedSensorLogs.forEach(sensorLog => {
+    const { timestamp } = sensorLog;
+    // Premature return if this sensorlog is malformed
+    if (!timestamp) return;
+    // Check if this sensorLog is in the current interval.
+    const isInSameInterval = logGroupStartTimestamp - timestamp < FULL_AGGREGATION_INTERVAL;
```
andreievg: @joshxg is logGropuStartTimestamp set for first iteration ?

src/utilities/modules/temperatureSensorHelpers.js
```diff
@@ -370,87 +378,119 @@ function addHeadAndTrailingLogToBreach({ result, sensor, database }) {
...
+  // Store each aggregated sensorlog in the database.
   database.write(() => {
-    logsToAdd.forEach(sensorLog => {
-      const newSensorLog = database.update('SensorLog', sensorLog);
-      sensor.sensorLogs.push(newSensorLog);
-      linkToBatches(newSensorLog, database);
-    });
+    aggregatedSensorLogs.forEach(aggregatedLog => database.update('SensorLog', aggregatedLog));
+    // Delete each pre-aggregated sensorlog.
+    sortedSensorLogs.forEach(preAggregatedLog => database.delete(preAggregatedLog));
```
andreievg: was wondering, because sortedSensorLog is realm collection, is it safe to iterate it while deleting the logs ? i.e. would it mutate during iteration causing adverse effects ?

joshxg:
@andreievg Have done the changes requested as well as the changes as per the other issue

src/utilities/modules/temperatureSensorHelpers.js
```diff
@@ -370,87 +383,167 @@ function addHeadAndTrailingLogToBreach({ result, sensor, database }) {
...
+  const mostRecentPreAggregateTimestamp = sensorLogs
+    .filtered('aggregation == $0', SENSOR_LOG_PRE_AGGREGATE_TYPE)
+    .max('timestamp');
+  const maxTimestampForDeletion = new Date().getTime() - SENSOR_LOG_NO_DELETION_INTERVAL;
+  // Starting timestamp is either from the first fullAggregate timestamp, or beginning of time
+  const aggregationIntervalStart = (mostRecentFullAggregateTimestamp || -Infinity) + 1;
+  // Ending timestamp is either from the most recent preAggregate or from now, less the
+  // no full aggregate period (32 hours)
+  const aggregationIntervalEnd =
+    (mostRecentPreAggregateTimestamp || new Date().getTime()) - NO_FULL_AGGREGATE_PERIOD;
```
andreievg: i think this should just be maxTimestampForDeletion ?

src/utilities/modules/temperatureSensorHelpers.js
```diff
@@ -370,87 +383,167 @@ function addHeadAndTrailingLogToBreach({ result, sensor, database }) {
...
+      sensorLogGroup = [sensorLog];
+      logGroupStartTimestamp = timestamp;
+    }
   });
-
+  // Create an array of aggregated sensor log objects
+  const aggregatedSensorLogs = sensorLogGroups
+    .map(group => createFullAggregateSensorLogs(group))
+    .flat();
+  const sensorLogsToDelete = sortedSensorLogs.filtered('timestamp < $0', maxTimestampForDeletion);
```
andreievg: this could mean the logs at the end of sortedSensorLogs, that are not aggregated yet (i.e. time between them is less then FULL_AGGREGATION_INTERVAL), would be deleted ?

src/utilities/modules/temperatureSensorHelpers.js
```diff
@@ -370,87 +383,167 @@ function addHeadAndTrailingLogToBreach({ result, sensor, database }) {
...
+    // Somewhat arbitrary three days to ensure there is always enough data and
+    // a little bit defensive in case something goes wrong.
+    database.delete('SensorLog', sensorLogsToDelete);
   });
-
-  return { ...result, fullAggregateAdditions: logsToAdd.length, fullAggregationDeletions };
+  return {
+    ...result,
+    fullAggregateAdditions: aggregatedSensorLogs.length,
+    preAggregateDeletions: sortedSensorLogs.length,
```
andreievg: sensorLogsToDelete

src/utilities/modules/temperatureSensorHelpers.js
```diff
@@ -370,87 +383,167 @@ function addHeadAndTrailingLogToBreach({ result, sensor, database }) {
...
+function linkSensorLogToItemBatches({ sensorLog, database }) {
+  const { location } = sensorLog;
+  if (!location) return;
+  const itemBatches = location.getItemBatchesWithQuantity(database);
+  if (!itemBatches || itemBatches.length === 0) return;
+  database.update('SensorLog', {
+    ...sensorLog,
+    itemBatches,
+  });
+  itemBatches.forEach(itemBatch =>
```
andreievg: also add sensorLog to itemBatches please

src/utilities/modules/temperatureSensorHelpers.js
joshxg: 7fb11aeda7c1f2c9c78024d9f123eb6166993196 - I think using the date of the last preaggregate is right - if we have large intervals between syncing it could incorrectly aggregate

src/utilities/modules/temperatureSensorHelpers.js
joshxg: da613bfbbc6d41ab60e2c213d675915ab62295fe

src/utilities/modules/temperatureSensorHelpers.js
joshxg: da613bfbbc6d41ab60e2c213d675915ab62295fe - sorry screwed my commits up

src/utilities/modules/temperatureSensorHelpers.js
joshxg: da613bfbbc6d41ab60e2c213d675915ab62295fe


Added a helper to itemBatches to be consistent with other realm models?

src/utilities/modules/temperatureSensorHelpers.js
andreievg: true

andreievg: actually, full aggregation would happen after a successful sync, at this stage we should have all of the info pulled from the sensor, so either way would work. i like the pre-emptive thinking though

src/utilities/modules/temperatureSensorHelpers.js
```diff
@@ -370,87 +382,170 @@ function addHeadAndTrailingLogToBreach({ result, sensor, database }) {
...
+    }
   });
-
+  // Create an array of aggregated sensor log objects
+  const aggregatedSensorLogs = sensorLogGroups
+    .map(group => createFullAggregateSensorLogs(group))
+    .flat();
+  const deleteFromTimestamp =
+    sensorLogGroup.length > 0 ? sensorLogGroup[0].timestamp : aggregationIntervalEnd;
+  const sensorLogsToDelete = sortedSensorLogs.filtered('timestamp < $0', deleteFromTimestamp);
```
andreievg: @joshxg, i think this is not quite right yet, shouldn't it it just be sensorLogGroup.flat() ?

joshxg: `sensorLogGroup` is a 1D array, filled with sensorLog objects, not sure what you mean. Essentially it is a stack which builds an 8 hour interval of `sensorLogs`. If we take the first log from that, it is the earliest in the last set of `sensorLogs` which have not been aggregated yet and we can delete all the pre aggregate logs left over and delete them, as we have already filtered `sortedSensorLogs` to be a collection of `sensorLogs` that are to-be-aggregated. If `sensorLogGroup` does not have any sensorLogs in it, we can just delete everything as there are no more `sensorLogs` to aggregate (this case shouldn't occur, maybe it is to safe?)

andreievg: @joshxg I missed the 's' on sensorLogGroups.
`sensorLogsToDelete = sensorLogGroups.flat()`
 
Your code all good apart from when 0 sortedSensorLogs?

joshxg: Oh right, sorry. I think I was not 100% that actual realm objects were required or not for the delete method.

Technically the method does nothing when `sortedSensorLogs.length = 0`, but can add an early return.

On that note, I forgot to ask if you think the early return on no `sensorLogs` should return an object in the same form, or if null is fine. Further, I kinda hate the result parameter and return value, heh

andreievg: @joshxg I think you need to loop through each element to delete it of not realm array. I thought that would be cleaner and quicker then query realm in this case

andreievg: I am not a fan of returns either, can discuss tomorrow

PR #1003, opened 2019-05-28
#987 Fix sync error caused by invalid [transact]entry_date

### LINKED ISSUE #987: Transaction sync failure on invalid entry date
Build Number: 2.3.3

Description:

Currently it is possible that a `Transaction` has no `entryDate` on incoming sync.
When this occurs, parsing the date returns `null`, which is then used to set the field which is not optional, causing an error and failing to integrate this record. This makes initialization impossible.

Reproducible:

Yes

Reproduction Steps:
Create a `transact` record in mSupply with no `entry_date` field (00/00/000).
Try to sync this record

Comments:

Although this should be fixed in mSupply, it exists and needs to be handled.

The default field is to just set the current date, so the easiest fix is to simply set that as the `entryDate` on incoming sync.

@andreievg @Chris-Petty - just double chechking this is the solution we should go with and not others such as ignoring the record?

wlthomson:
Fixes #987.

## Change summary

Simple fix for bug caused by syncing invalid required date fields.

## Testing

- [ ] Syncing a `transact` record with an invalid date does not cause a sync error.
  - Update a `transact` record on the server to have an invalid `entry_date`, e.g. `00/00/00`.
  - Update the sync out queue for mobile (e.g. `sync_create_syncout_record_UUID`).
  - Manually resync mobile
  - If all OK, test pass :)

Not sure if this is the best solution, but works well for the short term. A more tidy solution would be to include this logic in `sanityCheckIncomingRecord`. May also be a good idea to add similar logic to handle invalid dates for other records in incoming sync.

### DIFF
--- a/src/sync/incomingSyncUtils.js
+++ b/src/sync/incomingSyncUtils.js
@@ -608,18 +608,26 @@ export const createOrUpdateRecord = (database, settings, recordType, record) => 
       break;
     }
     case 'Transaction': {
-      if (record.store_ID !== settings.get(THIS_STORE_ID)) break; // Not for this store
+      // Only sync transactions for this store.
+      const { store_ID } = record;
+      if (store_ID !== settings.get(THIS_STORE_ID)) break;
+
+      // Do not sync transactions without valid entry date.
+      const entryDate = parseDate(record.entry_date);
+      if (entryDate == null) break;
+
       const otherParty = database.getOrCreate('Name', record.name_ID);
       const enteredBy = database.getOrCreate('User', record.user_ID);
       const linkedRequisition = record.requisition_ID
         ? database.getOrCreate('Requisition', record.requisition_ID)
         : null;
       const category = database.getOrCreate('TransactionCategory', record.category_ID);
+
       internalRecord = {
         id: record.ID,
         serialNumber: record.invoice_num,
         comment: record.comment,
-        entryDate: parseDate(record.entry_date),
+        entryDate,
         type: TRANSACTION_TYPES.translate(record.type, EXTERNAL_TO_INTERNAL),
         status: STATUSES.translate(record.status, EXTERNAL_TO_INTERNAL),
         confirmDate: parseDate(record.confirm_date),
@@ -629,6 +637,7 @@ export const createOrUpdateRecord = (database, settings, recordType, record) => 
         otherParty,
         linkedRequisition,
       };
+
       const transaction = database.update(recordType, internalRecord);
       if (linkedRequisition) {
         database.update('Requisition', {


src/sync/incomingSyncUtils.js
```diff
@@ -608,18 +608,26 @@ export const createOrUpdateRecord = (database, settings, recordType, record) =>
       break;
     }
     case 'Transaction': {
-      if (record.store_ID !== settings.get(THIS_STORE_ID)) break; // Not for this store
+      // Only sync transactions for this store.
+      const { store_ID } = record;
+      if (store_ID !== settings.get(THIS_STORE_ID)) break;
```
Chris-Petty: In cases like this I don't really get the improvement. What was readable on 1 line is now 2 lines. I'd understand if store_ID was used again or if you were looking to use other properties of `record`.

To follow this through properly shouldn't you grab every other property used in this case statement (i.e. all properties of `record`)?

src/sync/incomingSyncUtils.js
```diff
@@ -608,18 +608,26 @@ export const createOrUpdateRecord = (database, settings, recordType, record) =>
...
+
+      // Do not sync transactions without valid entry date.
+      const entryDate = parseDate(record.entry_date);
+      if (entryDate == null) break;
```
Chris-Petty: I think this is a dangerous path without logging and mSupply Hub to scream at us.

src/sync/incomingSyncUtils.js
wlthomson: Yeah, I think its good practice in general to destructure instead of dot notation. It's consistent throughout the code base and it also has the nice self-documenting property of showign at the top of the method what properties it uses. Will add the other destructured variables.

src/sync/incomingSyncUtils.js
wlthomson: This PR is a short term band aid for a bug which is breaking core functionality. We need to have a bigger discussion about how we handle violations of data constraints in sync, but right now this seems like the best solution to get things working again.

joshxg: just a quick one to be annoying - `===` over `==` ?

src/sync/incomingSyncUtils.js
andreievg: I think we should refrain from refactoring code full stop

src/sync/incomingSyncUtils.js
andreievg: i think it's better to set entryDate = confirmDate.

And yes, we need to make our sync code more resilient (by-pass possible breaks), with logging to hub or bug-snag.

andreievg: Ohh and something else to discuss in an issue re improving resilience of sync -> requesting resync of a table (i.e when options were added on primary, sites that were upgraded to VM had to resync the options)

Chris-Petty:
@andreievg @wlthomson I want to flag here that I don't think this approach should be merged at all, as per discussions that occurred on the issue #987

wlthomson:
Closing as per @Chris-Petty request.

PR #1005, opened 2019-05-28
#994 Incorrect reasons being applied

### LINKED ISSUE #994: Reason displayed as Not applicable rather than the reason you chose
Build Number:2.3.3

Description: When entering a reason in a stocktake, from time to time the reason is displayed as Not Applicable even if they selected a reason from the list. They reckon it might be because when changing field the cursor moves to the line below rather than the next field to the right. It doesn't happened all the time  

Reproducible: Yes but not all the time 

Reproduction Steps:  Start a stocktake, put a quantity different than the snapshot quantity, add a reason and if you are unlucky the eason will display Not applicable rather than the reason you chose.

Comments: It seems to display properly in the cloud server though.

joshxg:
#### NOTE: @Chris-Petty develop or master ?

Fixes #994 

## Change summary
- Reproduced by focusing a different row in the stocktakes data table between `onEndEditing` triggering and the reasons modal opening.

When the modal finally opens, focus is still on the cell which was last set by the user. When clicking on the modal to select a reason, focus then leaves the cell, triggering a new `onEndEditing`, which then updates the `currentStocktakeItem`. 

This still leaves a problem: If focusing a new row AND entering a value, the `currentStocktakeItem` will still update.

To resolve the second issue: Essentially required to force focus in the current cell after `onEndEditing` *if* it needs a reason. This requires a ref matrix, updated on each column sort (The react-native-data-table already creates one of these each `refreshData` I believe). I think this is really excessive for right now. This is only really possible on larger stock takes. It's pretty hard to reproduce on a emulator with keyboard and mouse on a smaller stocktake, let alone on a tablet. I think the more time consuming solution will just need to be a factor of consideration when we redesign the data table (enforcement of fields)

Unless anyone can think of a better solution!

## Testing
Reproduction steps above and in issue
- [ ] When entering a quantity !== snapshot quantity in a stock take line and then focusing a different line, the stocktake line still has a reason applied [ If you don't enter a value into the line that is focused next ]

Just adding another factor to think about for the data table (I imagine this type of thing may come up again one day), enforcing values for cells and triggering events from entering values in cells (i.e. if we use `onKeyPressed`, that might be weird for having to apply a reason!)

### DIFF
--- a/src/pages/StocktakeEditPage.js
+++ b/src/pages/StocktakeEditPage.js
@@ -161,7 +161,7 @@ export class StocktakeEditPage extends React.Component {
     const { database } = this.props;
     const { usesReasons } = this.state;
 
-    if (key !== 'countedTotalQuantity' || newValue === '') return;
+    if (key !== 'countedTotalQuantity' || !newValue) return;
     const quantity = parsePositiveInteger(newValue);
     if (quantity === null) return;
 


wlthomson:
Just sticking my beak in here. 

Correct me if I'm wrong, but when `onEndEditing` fires, doesn't the component state still contain the "correct" stocktake in `this.state.currentStocktakeItem`? Could you check:

- if this item exists
- whether it should have a reason assigned 
- whether it does in fact have a reason assigned

If the first two are true and the last is false, could you then decide to make the decision not to update the `currentStocktakeItem`, ensuring that the reason is correctly assigned? I'm not sure if this would then need some cleanup afterwards (like resetting the selected row to that of the currently selected `stocktakeItem`.

src/pages/StocktakeEditPage.js
```diff
@@ -161,7 +161,7 @@ export class StocktakeEditPage extends React.Component {
     const { database } = this.props;
     const { usesReasons } = this.state;
 
-    if (key !== 'countedTotalQuantity' || newValue === '') return;
+    if (key !== 'countedTotalQuantity' || !newValue) return;
```
wlthomson: There should probably be a comment here explaining this logic and that it is a temporary workaround.

joshxg:
I believe @Chris-Petty is having a quick look through this also - don't merge yet

Chris-Petty:
Bug fixes to be released in patches like this should be into master. Basically priority immediate and high should almost always be in master. They should also be tested properly before merge to make sure they do what they say they do without breaking related stuff. It's pretty easy to fix one case and break 3 others, and we want to catch that! 

Actually want to make a push to change the mobile project "Dev" a bit on this front, I'll try get an email/discussion going before the end of the day.

Chris-Petty:
Speaking of testing, I test with on onscreen keyboard as clunky as it may be as that is what users actually use.

In my testing @wlthomson @joshxg:
1. Make a new stocktake and go from top to bottom replacing "Not counted" values with counts and adding reasons, all is well.
Screen Shot 2019-05-28 at 17 13 00
Screen Shot 2019-05-28 at 17 13 42
2. Edit line 1 to have no diff. Reason will be reset to Not applicable.
Screen Shot 2019-05-28 at 17 14 05
3. Edit line 1 again to have a diff, hitting next or selecting the next line manually. Selecting a reason will edit the next line's reason
Screen Shot 2019-05-28 at 17 17 41

I'm happy to call this a separate/new issue to what is being solved in this PR, but the code here is still problematic.

joshxg:
Definitely problematic - my suggestion is to fix this in the data table rewrite, unless anyone else has a better idea or thinks creating a new ref map/changing code in the current data table is a better solution

Chris-Petty:
Ye, gonna merge this then make issue.

Chris-Petty:
Lol it was still based on develop, fml

PR #1006, opened 2019-05-28
#999 Uncaught NaN error in createRecord

### LINKED ISSUE #999: Bugsnag: Crash on create new program requisition
## Error in mSupply Mobile

**Error** in **MainActivity**
Value 'NaN' not convertible to a number.

[View on Bugsnag](https://app.bugsnag.com/sustainable-solutions-nz-ltd/msupply-mobile/errors/5ce69144f72a31001ca41fd7?event_id=5ce691440039ef3f28be0000&i=gh&m=ci)

## Stacktrace

    src/database/UIDatabase.js:149 - value
    src/database/utilities/createRecord.js:150 - createRequisition
    src/database/utilities/createRecord.js:393 - createRecord
    src/pages/SupplierRequisitionsPage.js:92 - 
    src/database/UIDatabase.js:157 - readFileGeneric
    src/pages/SupplierRequisitionsPage.js:83 - onNewRequisition
    src/widgets/modals/ByProgramModal.js:124 - onConfirm

[View full stacktrace](https://app.bugsnag.com/sustainable-solutions-nz-ltd/msupply-mobile/errors/5ce69144f72a31001ca41fd7?event_id=5ce691440039ef3f28be0000&i=gh&m=ci)

joshxg:
#### NOTE: Merge into Develop - not sure if that should be master or not

Fixes #999

## Change summary
If `maxMOS` is undefined, doing arithmetic on it results in `NaN`. This is the only way that I can see a `NaN` result anywhere in this method. Please check to see if you can see one!

This solution is defensive against `maxMOS` being `undefined`, `null` or `0`.. I did use a default value in the destructuring, but that does not protect against `0` or `null`.

From what I can tell, if any other `orderType` fields are `undefined` then we are covered by default realm fields and optional fields.

## Testing
I could only reproduce this error by having `maxMOS` be undefined. So:
- [ ] No error is thrown when `maxMOS` is undefined

Please do check if there is anyway, other than from `maxMOS` that a `NaN` error could be thrown in `createRequisition`

### DIFF
--- a/src/database/utilities/createRecord.js
+++ b/src/database/utilities/createRecord.js
@@ -145,7 +145,7 @@ const createRequisition = (
   const { name: orderTypeName, maxMOS, thresholdMOS } = orderType;
   const regimenData =
     program && program.parsedProgramSettings ? program.parsedProgramSettings.regimenData : null;
-  const daysToSupply = monthsLeadTime + (maxMOS || 1) * 30;
+  const daysToSupply = (monthsLeadTime || 0) + (maxMOS || 1) * 30;
 
   const requisition = database.create('Requisition', {
     id: generateUUID(),
--- a/src/database/utilities/createRecord.js
+++ b/src/database/utilities/createRecord.js
@@ -142,10 +142,10 @@ const createRequisition = (
   user,
   { otherStoreName, program, period, orderType = {}, monthsLeadTime = 0 }
 ) => {
-  const { name: orderTypeName, maxMOS, thresholdMOS } = orderType;
+  const { name: orderTypeName, maxMOS = 1, thresholdMOS } = orderType;
   const regimenData =
     program && program.parsedProgramSettings ? program.parsedProgramSettings.regimenData : null;
-  const daysToSupply = monthsLeadTime ? (monthsLeadTime + maxMOS) * 30 : 30;
+  const daysToSupply = monthsLeadTime + maxMOS * 30;
 
   const requisition = database.create('Requisition', {
     id: generateUUID(),
--- a/src/database/utilities/createRecord.js
+++ b/src/database/utilities/createRecord.js
@@ -142,10 +142,10 @@ const createRequisition = (
   user,
   { otherStoreName, program, period, orderType = {}, monthsLeadTime = 0 }
 ) => {
-  const { name: orderTypeName, maxMOS = 1, thresholdMOS } = orderType;
+  const { name: orderTypeName, maxMOS, thresholdMOS } = orderType;
   const regimenData =
     program && program.parsedProgramSettings ? program.parsedProgramSettings.regimenData : null;
-  const daysToSupply = monthsLeadTime + maxMOS * 30;
+  const daysToSupply = monthsLeadTime + (maxMOS || 1) * 30;
 
   const requisition = database.create('Requisition', {
     id: generateUUID(),
--- a/src/database/utilities/createRecord.js
+++ b/src/database/utilities/createRecord.js
@@ -145,7 +145,7 @@ const createRequisition = (
   const { name: orderTypeName, maxMOS, thresholdMOS } = orderType;
   const regimenData =
     program && program.parsedProgramSettings ? program.parsedProgramSettings.regimenData : null;
-  const daysToSupply = (monthsLeadTime || 0) + (maxMOS || 1) * 30;
+  const daysToSupply = ((monthsLeadTime || 0) + (maxMOS || 1)) * 30;
 
   const requisition = database.create('Requisition', {
     id: generateUUID(),


src/database/utilities/createRecord.js
```diff
@@ -145,7 +145,7 @@ const createRequisition = (
   const { name: orderTypeName, maxMOS, thresholdMOS } = orderType;
   const regimenData =
     program && program.parsedProgramSettings ? program.parsedProgramSettings.regimenData : null;
-  const daysToSupply = monthsLeadTime ? (monthsLeadTime + maxMOS) * 30 : 30;
+  const daysToSupply = monthsLeadTime + (maxMOS || 1) * 30;
```
wlthomson: On the subject of defensive programming, should probably add a similar check for `monthsLeadTime`. It has a default value, but we can't guarantee that another component might not pass it a `NaN` or similar value also. Could make it consistent with:

`const daysToSupply = ((monthlyLeadTime || 0) + (maxMOS || 1)) * 30;`

wlthomson: I think there's an associativity bug here. Think it should be:

```
const daysToSupply = (monthsLeadTime + (maxMOS || 1)) * 30;
```

wlthomson: One last small thing which may be a stupid question... should we assume that if `maxMOS` is `NaN` or similar, that it should be `1`? Is it possible that `maxMOS` should ever be `0` or fractional?

joshxg: 62b4f46be71e714d80e6de78047613566f625c15

joshxg: cbb01c806e6913dd641d808fc9688b9322461462

joshxg: I think it is fine to assume maxMOS being 0 is an error state and can be defaulted to 1. Fractions will be chosen before the default, still

Chris-Petty:
@wlthomson Just to be clear, did you test the area and confirm that you couldn't reproduce the bug (and couldn't easy find another way to break it)?

Do we need a label to be absolutely clear a PR has been tested?

wlthomson:
> @wlthomson Just to be clear, did you test the area and confirm that you couldn't reproduce the bug (and couldn't easy find another way to break it)?
> 
> Do we need a label to be absolutely clear a PR has been tested?

Bit late but yessir.

wlthomson:
> @wlthomson Just to be clear, did you test the area and confirm that you couldn't reproduce the bug (and couldn't easy find another way to break it)?
> 
> Do we need a label to be absolutely clear a PR has been tested?

Yes and yes I think so. All PRs should be tested (I think should be a rule that the branch is checked out and actually run). I also think that to aid this we should make it a rule to upload your structure or realm file. Some functionality is hard to set up if you aren't already working on it, so if we want to encourage proper testing practices (i.e. testing what you haven't worked on) I think this is a must.

Chris-Petty:
By structure do you actually mean the 4D structure, or just the datafile? The former probs not on a public repo. I don't like Realm as it'll be an excuse to test mobile without covering sync.

I think having a core server datafile that has config for everything and is reasonably consistent is probably more useful than various people's random barf data. That was the original idea with the demo/hogwarts datafile. If your PR implicates changes for that, then I think include it and it becomes the new master. Generally won't be necessary I hope.

I suppose now we have 2 configs though, datafiles with reasons and those without 😕. Those with programs and those without.

wlthomson:
> By structure do you actually mean the 4D structure, or just the datafile? The former probs not on a public repo. I don't like Realm as it'll be an excuse to test mobile without covering sync.
> 
> I think having a core server datafile that has config for everything and is reasonably consistent is probably more useful than various people's random barf data. That was the original idea with the demo/hogwarts datafile. If your PR implicates changes for that, then I think include it and it becomes the new master. Generally won't be necessary I hope.
> 
> I suppose now we have 2 configs though, datafiles with reasons and those without 😕. Those with programs and those without.

Good point re data. I like the Hogwarts file, but the problem is it still requires a fair bit of setting up for testing different functionality, which can be a pain when you are testing lots of different PRs in different areas. I think the most reliable way of improving our testing processes is making testing easier for the tester by giving them clear tests and preconfigured datafiles (tests can obv include turning on and off preferences, but having some things already setup will make things a lot more straightforward and make it easier to define the expected coverage of our tests). One idea would be to maintain a variety of test datafiles for given customisations (e.g. Cambodian Hogwarts, Ivory Coast Hogwarts, etc.). Thoughts?

Chris-Petty:
> One idea would be to maintain a variety of test datafiles for given customisations (e.g. Cambodian Hogwarts, Ivory Coast Hogwarts, etc.). Thoughts?

That'll easily get fragmented though, if you want to test a PR but it may affect 4 different configs then you need to update all the datafiles.

Dates are another big problem, as they drift into the past on testing data files. There is an issue on mSupply repo to make a script that just pulls them forward. Maybe the better way is to take advantage of XML import/export in mSupply. Have XML files for various testing environments in mSupply repo. Base XML + additional flavours on configs. Post import, shift dates forward based on (date now - export time stamp). A friendly admin window interface (or API for automated tests) for selecting what config you want.

This could be pretty handy for demo servers too. Although, maybe super safe to limit it all to interpreted.

In the interim I think a few pointers to config on a PR is reasonable though. Datafile if it's a dog to setup (say a sync site with particular data set required).

PR #1007, opened 2019-05-28
#1002 Supplier requisition sorting error

### LINKED ISSUE #1002: Bugsnag: Crash viewing supplier requisition
## Error in mSupply Mobile

TypeError: undefined is not a function (evaluating 't.sorted(n,!u)')

This error is located at:
    in p
    in f
    in p
    in s
    in Connect(s)
    in Unknown
    in v
    in RCTView
    in f
    in RCTView
    in f
    in C
    in n
    in P
    in RCTView
    in n
    in RCTView
    in f
    in b
    in y
    in L
    in RCTView
    in h
    in C
    in k
    in v
    in P
    in Unknown
    in Connect(Component)
    in RCTView
    in n
    in Connect(n)
    in s
    in c
    in v
    in RCTView
    in RCTView
    in c

[View on Bugsnag](https://app.bugsnag.com/sustainable-solutions-nz-ltd/msupply-mobile/errors/5cebfb4c37ebda0019f89ee9?event_id=5cebfb4c003e68b535d50000&i=gh&m=ci)

## Stacktrace

    src/utilities/sortDataBy.js:36 - sortDataBy
    src/pages/SupplierRequisitionPage.js:319 - refreshData

[View full stacktrace](https://app.bugsnag.com/sustainable-solutions-nz-ltd/msupply-mobile/errors/5cebfb4c37ebda0019f89ee9?event_id=5cebfb4c003e68b535d50000&i=gh&m=ci)

joshxg:
#### NOTE: PR Into develop

Fixes #1002

## Change summary
Simple fix that has been done before. I will check for anymore of these errors in other pages also.

## Testing

- [ ] Sorting by the stock on hand column does not crash the app!

I need to also check all the other pages for this type of error. Would be good to double check that this is sufficient though, before I go looking

### DIFF
--- a/src/pages/SupplierRequisitionPage.js
+++ b/src/pages/SupplierRequisitionPage.js
@@ -307,6 +307,7 @@ export class SupplierRequisitionPage extends React.Component {
       case 'itemName':
         sortDataType = 'string';
         break;
+      case 'stockOnHand':
       case 'monthlyUsage':
       case 'suggestedQuantity':
       case 'requiredQuantity':


Chris-Petty:
I think this would be better into master

Chris-Petty:
`stockOnHand` is a field on `requisitionItem` schema. Sorting by realm should be correct. Were you able to replicate this issue before you fixed it?

joshxg:
yeha I could. I can't remember if it was this issue but it could have been only when the hide over stock radio button was triggered

Chris-Petty:
Just had a test and poke around. In refresh data, if using hide over stock it'll set useThresholdMOS to true, then it'll change the data to an array rather than realm.result. This causes the crash when doing the realm based sort.

```javascript
if (useThresholdMOS) {
    data = data.filter(requisitionItem => requisitionItem.isLessThanThresholdMOS);
}
```

Can't do much about it as you have to filter on a getter like this. Could optimise the useThresholdMOS being false but probably not worth it.

Chris-Petty:
Lol whoops

PR #1008, opened 2019-05-28
#1000 Fix stocktake reason sort crash

### LINKED ISSUE #1000: Bugsnag: Type error crash on table sort in stocktakes
## Error in mSupply Mobile

TypeError: undefined is not an object (evaluating 't[n].localeCompare')

This error is located at:
    in p
    in f
    in p
    in s
    in Connect(s)
    in Unknown
    in v
    in RCTView
    in f
    in RCTView
    in f
    in C
    in n
    in P
    in RCTView
    in n
    in RCTView
    in f
    in b
    in y
    in L
    in RCTView
    in h
    in C
    in k
    in v
    in P
    in Unknown
    in Connect(Component)
    in RCTView
    in n
    in Connect(n)
    in s
    in c
    in v
    in RCTView
    in RCTView
    in c

[View on Bugsnag](https://app.bugsnag.com/sustainable-solutions-nz-ltd/msupply-mobile/errors/5ce6b64381e082001a39f229?event_id=5ce6b643003d8cdbc3c40000&i=gh&m=ci)

## Stacktrace

    src/utilities/sortDataBy.js:26 - 
    src/utilities/sortDataBy.js:26 - sortDataBy
    src/pages/StocktakeEditPage.js:240 - refreshData

[View full stacktrace](https://app.bugsnag.com/sustainable-solutions-nz-ltd/msupply-mobile/errors/5ce6b64381e082001a39f229?event_id=5ce6b643003d8cdbc3c40000&i=gh&m=ci)

Chris-Petty:
Fixes #1000

## Change summary
Change stocktakeItem `mostUsedReason` to return `option.title` rather than `option` itself. This allows string based sorting to work on the getter. Changed code using `mostUsedReason` to use `mostUsedReasonTitle`. 

A bit of rework on the choosing modal to make sure the right reason is highlighted, but should be a bit better using the realm result of reasons saved in state rather than querying the db for them every time the modal was rendered. I think in the old implementation the modal choices wouldn't have been filtered by just stocktakeReason either.

## Testing
- [ ] Sort the reasons column in a stocktake and not crash the app
- [ ] Sort sorts alphabetically
- [ ] Sort handles "Not applicable" (`''`/empty string) values
- [ ] Select reason modal highlights none if the current row didn't have one prior
- [ ] Select reason modal highlights correct option if the row did have one prior

Pretty sure I covered them all. If you're really keen, test stocktakes without reasons too.

### DIFF
--- a/src/database/DataTypes/StocktakeItem.js
+++ b/src/database/DataTypes/StocktakeItem.js
@@ -166,7 +166,7 @@ export class StocktakeItem extends Realm.Object {
     // Sort (ASC) the options by count, return the first option
     const sortedOptions = Object.values(options).sort((a, b) => b.count - a.count);
 
-    return sortedOptions[0] ? sortedOptions[0].option : '';
+    return sortedOptions[0] ? sortedOptions[0].option.title : '';
   }
 
   /**
--- a/src/database/DataTypes/StocktakeItem.js
+++ b/src/database/DataTypes/StocktakeItem.js
@@ -142,8 +142,8 @@ export class StocktakeItem extends Realm.Object {
    * Finds the mode option within this stocktake items batches
    * @return {string} The title of the reason with the highest frequency
    */
-  get mostUsedReason() {
-    if (!this.batches.length) return false;
+  get mostUsedReasonTitle() {
+    if (!this.batches.length) return '';
 
     // Mapping table for ranking reasons by usage
     // {option.id: {option: OptionObject, count: X}, ... }
@@ -164,11 +164,9 @@ export class StocktakeItem extends Realm.Object {
     });
 
     // Sort (ASC) the options by count, return the first option
-    const sortedOptions = Object.values(options).sort(
-      ({ count: valueA }, { count: valueB }) => parseInt(valueB, 10) - parseInt(valueA, 10)
-    );
+    const sortedOptions = Object.values(options).sort((a, b) => b.count - a.count);
 
-    return sortedOptions[0] && sortedOptions[0].option;
+    return sortedOptions[0] ? sortedOptions[0].option : '';
   }
 
   /**
--- a/src/database/DataTypes/StocktakeItem.js
+++ b/src/database/DataTypes/StocktakeItem.js
@@ -139,7 +139,7 @@ export class StocktakeItem extends Realm.Object {
   }
 
   /**
-   * Finds the mode option within this stocktake items batches
+   * Returns the title of the most common option within this stocktakeItem's batches
    * @return {string} The title of the reason with the highest frequency
    */
   get mostUsedReasonTitle() {
--- a/src/database/DataTypes/StocktakeItem.js
+++ b/src/database/DataTypes/StocktakeItem.js
@@ -165,7 +165,7 @@ export class StocktakeItem extends Realm.Object {
 
     // Sort (ASC) the options by count
     const sortedOptions = Object.values(options).sort((a, b) => b.count - a.count);
-    // Return the first option or empty string if there aren't any options
+    // Return the first option title or empty string if there aren't any options
     return sortedOptions[0] ? sortedOptions[0].option.title : '';
   }
 
--- a/src/database/DataTypes/StocktakeItem.js
+++ b/src/database/DataTypes/StocktakeItem.js
@@ -163,9 +163,9 @@ export class StocktakeItem extends Realm.Object {
       }
     });
 
-    // Sort (ASC) the options by count, return the first option
+    // Sort (ASC) the options by count
     const sortedOptions = Object.values(options).sort((a, b) => b.count - a.count);
-
+    // Return the first option or empty string if there aren't any options
     return sortedOptions[0] ? sortedOptions[0].option.title : '';
   }
 
--- a/src/pages/StocktakeEditPage.js
+++ b/src/pages/StocktakeEditPage.js
@@ -99,7 +99,7 @@ export class StocktakeEditPage extends React.Component {
       isResetModalOpen: false,
       stocktakeItem: null,
       isStocktakeEditModalOpen: false,
-      usesReasons: false,
+      reasons: [],
       isReasonsModalOpen: false,
       currentStocktakeItem: null,
     };
@@ -120,7 +120,7 @@ export class StocktakeEditPage extends React.Component {
     const { database } = this.props;
     const queryString = 'type == $0 && isActive == true';
     const reasons = database.objects('Options').filtered(queryString, 'stocktakeLineAdjustment');
-    this.setState({ usesReasons: reasons.length !== 0 });
+    this.setState({ reasons });
   };
 
   reasonModalConfirm = ({ item: option }) => {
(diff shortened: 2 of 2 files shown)


src/database/DataTypes/StocktakeItem.js
```diff
@@ -142,8 +142,8 @@ export class StocktakeItem extends Realm.Object {
    * Finds the mode option within this stocktake items batches
    * @return {string} The title of the reason with the highest frequency
    */
-  get mostUsedReason() {
-    if (!this.batches.length) return false;
+  get mostUsedReasonTitle() {
```
joshxg: Just wondering if it's worth using realm to filter batches without an option before iterating?

```
if (!this.batches.length) return '';
const batchesWithAReason = this.batches.filtered('option !== null')(???);
if (!batchesWithAReason) return '';
```

bit ugly but could be worth it?

src/database/DataTypes/StocktakeItem.js
```diff
@@ -142,8 +142,8 @@ export class StocktakeItem extends Realm.Object {
    * Finds the mode option within this stocktake items batches
```
joshxg: could use an update

src/database/DataTypes/StocktakeItem.js
```diff
@@ -164,11 +164,9 @@ export class StocktakeItem extends Realm.Object {
     });
 
     // Sort (ASC) the options by count, return the first option
```
joshxg: technically this is incorrect?

src/pages/StocktakeEditPage.js
```diff
@@ -120,7 +120,7 @@ export class StocktakeEditPage extends React.Component {
     const { database } = this.props;
     const queryString = 'type == $0 && isActive == true';
     const reasons = database.objects('Options').filtered(queryString, 'stocktakeLineAdjustment');
-    this.setState({ usesReasons: reasons.length !== 0 });
+    this.setState({ reasons });
```
joshxg: Much better. Wondering how you feel about having just a constant in the constructor: `this.USES_REASONS = null/false` and assigning `!!reasons.length` in CDM - saves a bit of fluffing around throughout the component. As it would be static it seems right to have it outside of state, even though I don't really like it!

src/pages/StocktakeEditPage.js
```diff
@@ -422,8 +422,26 @@ export class StocktakeEditPage extends React.Component {
     this.setState({ stocktakeItem: item, isStocktakeEditModalOpen: true });
   };
 
+  renderReasonModal = () => {
+    const { currentStocktakeItem, isReasonsModalOpen, reasons } = this.state;
+    const currentReasonIndex = reasons.findIndex(
```
joshxg: I'm wondering - I have added a `highlightValue` property to `GenericChooseModal` (as well as a couple of other things) in the vaccines branch. Wondering if it is better for me to make the changes to this component sooner rather than later? Not a huge difference here, but is something

src/database/DataTypes/StocktakeItem.js
Chris-Petty: Interestingly, the return definition is already correct! Was the intent to return the string originally?

joshxg: lets go with that!

Chris-Petty: https://github.com/openmsupply/mobile/pull/1008/commits/d04a3ebf1bdee015e1e35b1aabf0aad43cd01f23

src/database/DataTypes/StocktakeItem.js
Chris-Petty: The following existing code will pretty quickly evaluate the same thing without hitting up realm with a filter query
```javascript
this.batches.forEach(batch => {
      const { option } = batch;
      if (!option) return;
...
```

Chris-Petty: I'd put this in the premature optimisation bucket but I'm going to do some science anyway

Chris-Petty: For a stocktakeItem with 30 batches, 15 with a reason, 15 without. 10000 runs averages:
(got surprising variance, but I'll just take worst run of 4 and not spend more time on this lol)

Without this tweak: 0.9065ms
With this tweak: 1.0453ms

30 batches, no diff no reasons:

Without this tweak: 0.6661ms
With this tweak: 0.1110ms

**Take away:**
- Unless we have a stocktake covering over say 1000 batches, probably no point optimising here. Unless something is noticeably slow, not worth investigating (I wasted an hour... maybe 2 now doing this, even if it was fun)

Hilariously, somehow, the OG optimisation slows things down? `if (!this.batches.length) return '';`

Without:
Screen Shot 2019-05-29 at 12 45 05
Without:
Screen Shot 2019-05-29 at 12 46 11

I suspect that might be the JS JIT compiler is having trouble optimising the function with our optimisation in there.

Anyway, I'm removing the old one and adding yours and moving on Jesus christ me

Chris-Petty: I'd argue that it's probably worse adding this change:
- Places that don't use reasons and therefore don't render the column, it makes no difference
- Most stocktakeItems are going to have a few batches with at least one reason, so this will likely just slow down the stocktake as they fill it out
- Unless it's a freaking huge stocktake, it's negligible. And still nothing on the time wasted on unnecessary re-renders!
- If I spend a single keystroke more on this put me in the naughty corner to think about what I've done.

Oh god I couldn't stop:
On an item with 1 batch, one reason, 10000 run average:

Without: ~0.075ms
With: ~0.185ms

Yea it's ded let's stop pls god

joshxg: I wonder though if this should be calculated in the component cached and updated as reasons are added to help in the render cy.... hehehhee

src/pages/StocktakeEditPage.js
Chris-Petty: I think it's fine as is

src/pages/StocktakeEditPage.js
Chris-Petty: Wouldn't rush, though if the API is changing then will have to test reason modal with vaccine PR for sure!

src/database/DataTypes/StocktakeItem.js
Chris-Petty: https://github.com/openmsupply/mobile/pull/1008/commits/7aba7b1141240a40701adc9ef9d1fed49dbe85ae

Chris-Petty: https://github.com/openmsupply/mobile/pull/1008/commits/02fa945319c4e7702ecbf2c018645b502a7e754a

src/database/DataTypes/StocktakeItem.js
Chris-Petty: If the shallow compare saw that the object hasn't changed, shouldn't rerender/re-get the cell/value I think. Blah blah new table things

src/pages/StocktakeEditPage.js
joshxg: Last thing is just the `highlightIndex` - I think it just has a nicer look if the `highlightIndex` is set to `0` when there is no reason selected. Couldn't think of a cool JS way :( `highlightIndex={currentReasonIndex >= 0 ? currentReasonIndex : 0}`. 

This defaulting to 0 is not part of the `GenericChooseModal` (Neither this version nor the vaccine version....) as I thought being 'generic' it should have as little behaviour pre-defined as possible although that line of thinking might be wrong!

i.e. this view when no reasons have been applied yet
image

Over this view (what it is with this changes)
image

I can see how this could *potentially* be confusing, but I think it's convention at this point for a drop down to have the first item selected

Chris-Petty: Currently with my changes I get neither of those pictures. There is just no highlighting or shadow at all.

image

Chris-Petty:
@joshxg what can I do to get your approval?

src/pages/StocktakeEditPage.js
joshxg: maybe I did something to make the second pic highlight come up - I am suggesting that the first highlight should be used

PR #1013, opened 2019-05-29
#1011 fix reason saving to wrong row

### LINKED ISSUE #1011: Stocktake reason saved on wrong row
Build Number: 2.3.3

Description:

1. Make a new stocktake and go from top to bottom replacing "Not counted" values with counts and adding reasons, all is well.
Screen Shot 2019-05-28 at 17 13 00
Screen Shot 2019-05-28 at 17 13 42
2. Edit line 1 to have no diff. Reason will be reset to Not applicable.
Screen Shot 2019-05-28 at 17 14 05
3. Edit line 1 again to have a diff, hitting next or selecting the next line manually. Selecting a reason will edit the next line's reason
Screen Shot 2019-05-28 at 17 17 41

Reproducible: Yes

Reproduction Steps: Description

Comments:

Chris-Petty:
Fixes #1011

## Change summary
TL;DR: Small tweak. Basically stops the the `currentStocktakeItem` from changing when the reason modal is already up.

After editing a row to have a diff, hitting next or selecting the next row changes focus from the edited row and triggers endEditing. This will show the reason modal. Pressing anywhere on modal to change focus from _next row_ would trigger endEditing on next row. Modal is already up but the currentStocktakeItem would update to be the next row. Selecting a reason therefore updates that row erroneously.

## Testing
- [ ] Make a new stocktake and go from top to bottom replacing "Not counted" values with counts and adding reasons, all is well.
Screen Shot 2019-05-28 at 17 13 00
Screen Shot 2019-05-28 at 17 13 42
- [ ] Edit line 1 to have no diff. Reason will be reset to Not applicable.
Screen Shot 2019-05-28 at 17 14 05
- [ ] Edit line 1 again to have a diff, hitting next or selecting the next line manually. Selecting reason saves correctly
- [ ] Do above, but click on search bar instead to change focus. Selecting reason saves correctly

- There is also a state variable `stocktakeItem`. Seems a bit redundant. Bunch of the logic around reason modal and `assignReason()` seems a bit more spaghetti than needed but not refactoring today.

### DIFF
--- a/src/pages/StocktakeEditPage.js
+++ b/src/pages/StocktakeEditPage.js
@@ -159,11 +159,15 @@ export class StocktakeEditPage extends React.Component {
    */
   onEndEditing = (key, stocktakeItem, newValue) => {
     const { database } = this.props;
-    const { reasons } = this.state;
+    const { reasons, isReasonsModalOpen } = this.state;
 
     if (key !== 'countedTotalQuantity' || newValue === '') return;
     const quantity = parsePositiveInteger(newValue);
     if (quantity === null) return;
+    // If the reason modal is open just ignore any change to the current line
+    // This a hack to solve https://github.com/openmsupply/mobile/issues/1011
+    // Underlying issue requires data table rewrite
+    if (isReasonsModalOpen) return;
 
     stocktakeItem.setCountedTotalQuantity(database, quantity);
     if (reasons.length > 0) this.assignReason(stocktakeItem);
--- a/src/pages/StocktakeEditPage.js
+++ b/src/pages/StocktakeEditPage.js
@@ -97,7 +97,6 @@ export class StocktakeEditPage extends React.Component {
       modalKey: null,
       isModalOpen: false,
       isResetModalOpen: false,
-      stocktakeItem: null,
       isStocktakeEditModalOpen: false,
       reasons: [],
       isReasonsModalOpen: false,
@@ -297,7 +296,10 @@ export class StocktakeEditPage extends React.Component {
         return (
           <TouchableOpacity
             onPress={() => {
-              this.openBatchModal(stocktakeItem);
+              this.setState({
+                isStocktakeEditModalOpen: true,
+                currentStocktakeItem: stocktakeItem,
+              });
             }}
           >
             <View style={localStyles.modalControl}>
@@ -417,10 +419,6 @@ export class StocktakeEditPage extends React.Component {
     this.setState({ isStocktakeEditModalOpen: false });
   };
 
-  openBatchModal = item => {
-    this.setState({ stocktakeItem: item, isStocktakeEditModalOpen: true });
-  };
-
   renderReasonModal = () => {
     const { currentStocktakeItem, isReasonsModalOpen, reasons } = this.state;
     // The below findIndex would fail if title was changed on central server!
@@ -502,7 +500,7 @@ export class StocktakeEditPage extends React.Component {
       data,
       isResetModalOpen,
       isModalOpen,
-      stocktakeItem,
+      currentStocktakeItem,
       isStocktakeEditModalOpen,
       isReasonsModalOpen,
     } = this.state;
@@ -546,7 +544,7 @@ export class StocktakeEditPage extends React.Component {
 
         <StocktakeBatchModal
           isOpen={isStocktakeEditModalOpen}
-          stocktakeItem={stocktakeItem}
+          stocktakeItem={currentStocktakeItem}
           database={database}
           genericTablePageStyles={genericTablePageStyles}
           onConfirm={this.onConfirmBatchModal}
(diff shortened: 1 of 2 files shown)


src/widgets/modals/StocktakeBatchModal.js
```diff
@@ -218,8 +225,26 @@ export default class StocktakeBatchModal extends React.Component {
...
+  renderReasonModal = () => {
+    const { database } = this.props;
+    const { reasonModalOpen, currentBatch, reasons } = this.state;
+    const { option } = currentBatch;
+    const currentReasonIndex = option && reasons.findIndex(reason => reason.id === option.id);
+
+    return (
+      <GenericChooseModal
+        isOpen={reasonModalOpen}
+        data={database.objects('Options')}
```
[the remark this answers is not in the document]
Chris-Petty: Fixed in https://github.com/openmsupply/mobile/pull/1013/commits/94434bdb3b8ab1fdd0cca585933ff489baef5782

src/pages/StocktakeEditPage.js
```diff
@@ -141,7 +140,7 @@ export class StocktakeEditPage extends React.Component {
   assignReason = stocktakeItem => {
     const { database } = this.props;
     if (stocktakeItem.shouldApplyReason) {
-      this.onOpenReasonModal();
+      this.setState({ isReasonsModalOpen: true, currentStocktakeItem: stocktakeItem });
```
wlthomson: I think a comment here would be a good idea, keep a paper trail.

src/pages/StocktakeEditPage.js
```diff
@@ -159,15 +158,18 @@ export class StocktakeEditPage extends React.Component {
...
   onEndEditing = (key, stocktakeItem, newValue) => {
     const { database } = this.props;
-    const { reasons } = this.state;
+    const { reasons, isReasonsModalOpen } = this.state;
 
     if (key !== 'countedTotalQuantity' || newValue === '') return;
     const quantity = parsePositiveInteger(newValue);
     if (quantity === null) return;
+    // If the reason modal is open just ignore any change to the current line
+    // This a hack to solve https://github.com/openmsupply/mobile/issues/1011
```
wlthomson: Don't think we should tie our codebase to our version control:

```
// This a hack to ensure the current stock take item is not unexpectedly
// updated by the user immediately before or while the reasons modal is open.
```

src/pages/StocktakeEditPage.js
Chris-Petty: YOU'RE A COMMENT

src/pages/StocktakeEditPage.js
Chris-Petty: Annoyingly you can't make a multiline suggestion in github

src/pages/StocktakeEditPage.js
```diff
@@ -159,15 +158,18 @@ export class StocktakeEditPage extends React.Component {
...
+    // Underlying issue requires data table rewrite
+    if (isReasonsModalOpen) return;
```
wlthomson: I think logic is clearer if this is at the top of the method straight after deconstructing. No need to parse anything if the data is wrong.

src/pages/StocktakeEditPage.js
Chris-Petty: I actually don't know what you'd have me comment here.

`// Open reason selection modal`?

src/pages/StocktakeEditPage.js
```diff
@@ -141,7 +140,7 @@ export class StocktakeEditPage extends React.Component {
   assignReason = stocktakeItem => {
     const { database } = this.props;
     if (stocktakeItem.shouldApplyReason) {
-      this.onOpenReasonModal();
+      this.setState({ isReasonsModalOpen: true, currentStocktakeItem: stocktakeItem });
     } else {
       stocktakeItem.applyReasonToBatches(database);
```
wlthomson: I think this method call is a bit confusing, defo needs a comment or explicitly pass `null`. It looks like @joshxg removed his previous `removeReason` but I do think there should be a `clearReasonsFromBatches` as it's a lot clearer what's going on.

src/pages/StocktakeEditPage.js
wlthomson: A comment on the `if` statement (as explained more below). The code reads: 

```
IF shouldApplyReason
THEN
  updateState
ELSE
  applyReasonToBatches
```

I think this is confusing without a comment.

```

src/pages/StocktakeEditPage.js
Chris-Petty: I 100% agree, but I wasn't going that far refactoring it.

Chris-Petty:
I mean, I've already merged it. If you want requested changes don't be afraid to hit the more demanding button. Hold me to higher standards too!

PR #1047, opened 2019-07-04
#987 handle incorrectly formatted transact entry dates

### LINKED ISSUE #987: Transaction sync failure on invalid entry date

wlthomson:
Fixes #987.

Hotfix for transact sync bug. 

Not the most elegant, but at this stage just designed to handle buggy transaction records. Could be made a bit more tidy by incorporating similar logic into existing date conversion methods (but only when/if we want this logic to extend to other records).

Sync to desktop is handled by existing date conversion logic (already maps `null` values to ISO date `0000-00-00T00:00:00`).

## Change summary

Update realm schema to make `transact.entry_date` optional. 

## Testing

- [ ] Incoming sync handles default transaction entry date (`00-00-0000`).
- [ ] Outgoing sync handles `null` transaction entry dates.
- [ ] Supplier invoice page correctly displays transactions with `null` entry dates.
- [ ] Supplier invoice page correctly displays transactions with `null` entry dates.
- [ ] Customer invoice page correctly displays transactions with `null` entry dates.
- [ ] Customer invoice page correctly displays transactions with `null` entry dates.

N/A.

### DIFF
--- a/src/database/DataTypes/Transaction.js
+++ b/src/database/DataTypes/Transaction.js
@@ -446,7 +446,7 @@ Transaction.schema = {
     serialNumber: { type: 'string', default: 'placeholderSerialNumber' },
     otherParty: { type: 'Name', optional: true },
     comment: { type: 'string', optional: true },
-    entryDate: { type: 'date', default: new Date() },
+    entryDate: { type: 'date', optional: true },
     type: { type: 'string', default: 'placeholderType' },
     status: { type: 'string', default: 'new' },
     confirmDate: { type: 'date', optional: true },
--- a/src/pages/CustomerInvoicePage.js
+++ b/src/pages/CustomerInvoicePage.js
@@ -161,43 +161,42 @@ export class CustomerInvoicePage extends GenericPage {
   };
 
   renderPageInfo = () => {
+    const { transaction } = this.props;
     const infoColumns = [
       [
         {
           title: `${pageInfoStrings.entry_date}:`,
-          info: formatDate(this.props.transaction.entryDate) || 'N/A',
+          info: formatDate(transaction.entryDate) || 'N/A',
         },
         {
           title: `${pageInfoStrings.confirm_date}:`,
-          info: formatDate(this.props.transaction.confirmDate),
+          info: formatDate(transaction.confirmDate),
         },
         {
           title: `${pageInfoStrings.entered_by}:`,
-          info: this.props.transaction.enteredBy && this.props.transaction.enteredBy.username,
+          info: transaction.enteredBy && transaction.enteredBy.username,
         },
       ],
       [
         {
           title: `${pageInfoStrings.customer}:`,
-          info: this.props.transaction.otherParty && this.props.transaction.otherParty.name,
+          info: transaction.otherParty && transaction.otherParty.name,
         },
         {
           title: `${pageInfoStrings.their_ref}:`,
-          info: this.props.transaction.theirRef,
+          info: transaction.theirRef,
           onPress: this.openTheirRefEditor,
           editableType: 'text',
         },
         {
           title: `${pageInfoStrings.comment}:`,
-          info: this.props.transaction.comment,
+          info: transaction.comment,
           onPress: this.openCommentEditor,
           editableType: 'text',
         },
       ],
     ];
-    return (
-      <PageInfo columns={infoColumns} isEditingDisabled={this.props.transaction.isFinalised} />
-    );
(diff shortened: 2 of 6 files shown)


src/database/DataTypes/Transaction.js
```diff
@@ -446,7 +446,7 @@ Transaction.schema = {
     serialNumber: { type: 'string', default: 'placeholderSerialNumber' },
     otherParty: { type: 'Name', optional: true },
     comment: { type: 'string', optional: true },
-    entryDate: { type: 'date', default: new Date() },
+    entryDate: { type: 'date', optional: true },
```
joshxg: I think the original issue was for `Transactions` but I guess extending this to `Requisition`'s is good also, as per your tests. At the moment, though you've only covered the `Transaction` case, so a `Requisition` incoming would still try to assign `null` to the required `entryDate` field which will throw an error.

joshxg:
Oh this would also require a schema bump

src/database/DataTypes/Transaction.js
wlthomson: Yeah, I mentioned that in the PR. As I understood it, the idea was to hotfix patch the issue to prevent any buggy behaviour in the wild. I'm not sure if this is something we want to actually go with long-term, which is why I haven't extended it to any other cases/objects yet.

wlthomson: As I understand it, this change leaves open the question of whether we want to integrate these changes further in future. From what I gathered, the idea was to hotfix patch the issue to prevent any buggy behaviour in the wild. I'm not sure if this is something we want to actually go with long-term, which is why I haven't extended it to any other cases/objects yet.

joshxg: sorry, I mis-read - I guess it just seemed like with the test cases and adding to requisitions code to handle `null` dates it seemed like it had just been forgotten. 

Without this change in the `Requisition` model, there can't be `null` entry dates. So I think either adding it to the `Requisition` model, or removing the code from the requisition pages seems correct? I would go for adding it to the `Requisition` model also, as `Transactions` shouldn't be coming in hot with `null` `entryDate` fields, but they are, so might happen with `Requisition` also. This does sound like a slippery slope though :)

wlthomson:
> Just need to cover `Requisition` also. I'm also assuming this is the solution that was decided on by you/Chris?

Solution initially proposed by @andreievg and returned to after further discussion (#897). It's probably the tidiest way of solving the problem, and after investigation, there are only a few logical areas touched by this schema change, so it's also *relatively* regression-proof.

src/database/DataTypes/Transaction.js
wlthomson: Currently the changes are only to `Transaction` stuff, no added code to `Requisitions` yet. It's quite easy to hotfix some additional changes to `Requisition` at a later date. Since we're not 100% on how mobile should handle malformed dates and haven't isolated the cause of the issue on the desktop side, it might be better to touch as little of the codebase as possible for now. OK with other suggestions, though.

wlthomson:
> Oh this would also require a schema bump

Bumped!

src/pages/CustomerInvoicePage.js
```diff
@@ -165,7 +165,7 @@ export class CustomerInvoicePage extends GenericPage {
       [
         {
           title: `${pageInfoStrings.entry_date}:`,
-          info: formatDate(this.props.transaction.entryDate),
+          info: formatDate(this.props.transaction.entryDate) || 'N/A',
```
Chris-Petty: Ancient history but you'd think the OG developers would have at least destructured this.props...

wlthomson: 9a5ebf0afdee319f322308af4cfea6829f2c5f5b 😄 .

wlthomson:
Not sure if people are aware, but this looks like the likely culprit: https://github.com/sussol/msupply/issues/3422.

wlthomson:
Rebased onto master for patch fix.

PR #1052, opened 2019-07-05
#945 Setting sync promise rejection

### LINKED ISSUE #945: Adding program list to mobile store causes unhandled promise rejection warning
Build Number: 2.3.0-rc0 develop

Description:
If you change the master lists on the mobile store to include a program-list, it rejects some promise. 

image

Reproducible: yes

Reproduction Steps:
1. Have a mobile device with no program list
2. Add a program list to its store
3. sync
4. In dev you'll get an unhandled promise rejection

joshxg:
Fixes #945 #972 

## Change summary
- Just adds an early return for `Setting` record types in sync post processing.
- I think this is too hacky and I want a better solution. This mainly shows the source of the problem.

## Testing
- [ ] Syncing any data which updates `Setting` records does not throw a rejected promise warning - i.e. changing a store tag for a store

I'm not 100% on this post processing yet, but from what I can tell at a high-level:

- A `recordQueue` is created which holds all records from incoming sync
- For each of the records in the `recordQueue`, create another queue of functions to execute for the post processing.
- all the while taking care of crashes etc.

So from what I can tell at a high level, the `recordQueue` is probably to large, as only `Transaction` records and `Requisition` records have functions which are added to the `functionQueue`. So I think this should be changed to have a lookup table of records which should actually be added to the `recordQueue` - maybe better than defaulting to True for adding and having an exclusion table.

### DIFF
--- a/src/sync/PostSyncProcessor.js
+++ b/src/sync/PostSyncProcessor.js
@@ -92,6 +92,7 @@ export class PostSyncProcessor {
     this.recordQueue.forEach((recordType, recordId) => {
       // Use local database record, not what comes in sync. Ensures that records are
       // integrated information (definitely after sync is done).
+      if (recordType === 'Setting') return;
       const internalRecord = this.database.objects(recordType).filtered('id == $0', recordId)[0];
       this.enqueueFunctionsForRecordType(recordType, internalRecord);
     });


Chris-Petty:
>So from what I can tell at a high level, the recordQueue is probably to large, as only Transaction records and Requisition records have functions which are added to the functionQueue. So I think this should be changed to have a lookup table of records which should actually be added to the recordQueue - maybe better than defaulting to True for adding and having an exclusion table.

@joshxg It should only add functions for records transactions/requisitions that need some handling once synced, so I wouldn't expect the queue to be too significant. One edge case would be if a site doesn't sync for a month and then syncs 500 customer requisitions.

I had been reading about closures and general JS function shenanigans so came up with this wacky system. In hindsight I think the better way would have been similar to what you've suggested. I did consider making a realm table along the lines of `PostSyncQueue`. This would also have been more robust against sync failing or powering off I think.

joshxg:
Closing due to duplciate PR #1069

Chris-Petty:
@joshxg can delete branch?

PR #1061, opened 2019-07-15
#1038 External storage permissions

### LINKED ISSUE #1035: Realm not being opened correctly after exporting
## Error in mSupply Mobile

**Error** in **MainActivity**
Cannot access realm that has been closed.

[View on Bugsnag](https://app.bugsnag.com/sustainable-solutions-nz-ltd/msupply-mobile/errors/5d14b088267ced001965c96d?event_id=5d14b0880045f1c033770000&i=gh&m=ci)

## Stacktrace

    src/database/UIDatabase.js:95 - value
    src/pages/MenuPage.js:43 - exportData

[View full stacktrace](https://app.bugsnag.com/sustainable-solutions-nz-ltd/msupply-mobile/errors/5d14b088267ced001965c96d?event_id=5d14b0880045f1c033770000&i=gh&m=ci)

### LINKED ISSUE #1038: Data export doesn't work/Android storage permission not being given
## Describe the bug
When installing mSupply mobile, it appears that storage access isn't being given to the app.

### To reproduce
1. Install mSupply mobile on a tablet (should prompt with permissions, should include storage)
2. Check app settings for permissions
3. Should have Storage permissions

### Expected behaviour
Should ask for storage permission. Should get storage permission

### Version and device info
 - App version: 2.3.4
 - Tablet model: Lenovo, techno (maybe Samsung?)
 - OS version: *

joshxg:
Fixes #1038 Fixes #1035  Fixes#993

*I believe 993 was being caused by invalid characters in a store name when exporting realm*

## Change summary

- On Android Marshmellow/API Level > 22/Android 6.0+ a new permission group was added for 'dangerous' features - i.e. writing to external storage, location, phone calling etc. This permission group [on this page](https://developer.android.com/training/permissions/requesting) requires permissions to be given at run time. For API level < 23, these permissions can be granted when installing the app (I think for <4, you don't need permissions!).
- As the app is potentially on a range of devices below and above 6.0+ we account for both cases. Permission should be asked for during installation with the permissions required in `AndroidManifest.xml` for android 6.0+ - for anything below, `AndroidPermissions` from `react-native` requests permission from the user.
- Also changed exportData slightly. The code was originally closing the `realm` with a comment/explanation
```
// If the database is not closed, there is a small chance of corrupting the data if currently in
// a transaction.
```
I think this is maybe under-thought, surely closing the realm mid transaction is more prone to corruption than copying the file? I changed this to instead not export the data at all if in a transaction. Also this was throwing an error as the realm wasn't being opened/assigned correctly afterwards.
- Also added a validation method for checking there is enough space on the device and that the path exists - just to be sure
- Have also added a `Toast` for visual feedback after exporting as it was a little annoying not having any

### Note: Should be tested on a device using API Level > 22 AND API Level < 23.
*To export data, touch the mSupply man on the menu page for 5+ seconds to enter admin mode. The export data button should appear. Data is exported to `downloads/msupply_mobile/storename-date.realm`
#### API Level > 22: Will *not* ask for permissions when installing
- [ ] After clicking export data, a user should be prompted for storage permission
    - [ ] The prompt should be clear
- [ ] Granting permissions should export the data, and give visual feedback (via a toast) to the user.
- [ ] Not accepting permissions should export the data and give visual feedback to the user.

#### API Level < 23: *Will* ask for permissions when installing
- [ ] After clicking export data, the data should be exported with visual feedback (via a toast)  to the user

- Difficult to test being in a transaction while exporting data.
- Need a place to upload realm's, so we can have a 'upload database' button. A large Vannuatu site's realm is `6mb`
- I think we're getting close to needing an 'admin' page


src/database/UIDatabase.js
```diff
@@ -35,36 +33,53 @@ export class UIDatabase {
...
+    // the realm file.
+    if (!permissionSuccess) return { success: false, ...this.ERRORS.ERROR_NO_PERMISSION };
+    if (realm.isInTransaction) return { success: false, ...this.ERRORS.ERROR_IN_TRANSACTION };
+
+    // Finally try to create the backup/exported realm
+    try {
+      await RNFS.mkdir(exportFolder);
+      await RNFS.copyFile(realmPath, `${exportFolder}/${copyFileName}.realm`);
+    } catch (error) {
+      return { success: false, ...this.ERRORS.ERROR_UNKNOWN };
```
mark-prins: any value in adding `error` to the message property?

src/database/UIDatabase.js
```diff
@@ -35,36 +33,53 @@ export class UIDatabase {
     this.database = database;
   }
 
+  ERRORS = {
+    ERROR_IN_TRANSACTION: { code: 'ERROR_IN_TRANSACTION', message: 'Database is in a transaction' },
+    ERROR_NO_PERMISSION: { code: 'ERROR_NO_PERMISSION', message: 'Storage permission not granted' },
+    ERROR_UNKNOWN: { code: 'ERROR_UNKOWN', message: 'ERROR_UNKNOWN' },
```
wlthomson: Nit: `ERROR_UNKNOWN` typo. Shouldn't the error message be descriptive, e.g. "Unknown error occurred"?

src/database/UIDatabase.js
```diff
@@ -35,36 +33,53 @@ export class UIDatabase {
...
-  exportData(filename = 'msupply-mobile-data') {
-    // TO DO: add method to react-native-database.
 
+  async exportData(filename = 'msupply-mobile-data') {
     const { realm } = this.database;
-    const realmPath = realm.path;
+    const { path: realmPath } = realm;
     const exportFolder = `${RNFS.ExternalStorageDirectoryPath}/Download/mSupplyMobile_data`;
+    const copyFileName = `${filename}${formatDate(new Date(), 'dashes')}`.replace(
+      /[~\\\\/|?*<:>"+]/g,
```
wlthomson: I think in general good to give regexes comments.

src/utilities/fileSystem.js
```diff
@@ -0,0 +1,34 @@
...
+export async function backupValidation(path) {
+  let returnObject = { success: true };
+  try {
+    // Fetch the size of the realm file.
+    const { size } = await RNFS.stat(path);
+    // Fetch the free space remaining on the device.
+    const { freeSpace } = await RNFS.getFSInfo();
+    // Ensure there is twice the amount of space required for copying the
+    // realm file, to be certain
+    if (freeSpace < size * 2) {
```
wlthomson: Is there a reason for needing to ensure there is double the space available? This might be problematic for larger files.

src/utilities/requestPermission.js
```diff
@@ -0,0 +1,44 @@
...
+ * need to be granted for permissions a part of the dangerous group,
+ * which can't be granted during installation.
+ * See:
+ * https://developer.android.com/training/permissions/requesting
+ * https://facebook.github.io/react-native/docs/permissionsandroid
+ *
+ *
+ * @property {String} permission The permission request, examples below
+ * @property {String} message    A reason for requesting the permission
+ * @return   {Object} {success, permissionType, result, error}
```
wlthomson: Is there a reason for returning `permissionType`? It looks like it isn't touched by the method.

src/utilities/fileSystem.js
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Helper method to validate that a file exists and there is enough
+ * space on disk to create a duplicate.
+ * @param  {String} path  Path of the file to check
+ * @return {Object} Object with the shape:
+ * { success: true/false, error:}
+ */
+export async function backupValidation(path) {
+  let returnObject = { success: true };
```
wlthomson: Nit: `return returnObject;` feels a bit weird. Maybe something like `validationResult`?

src/database/UIDatabase.js
```diff
@@ -35,36 +33,53 @@ export class UIDatabase {
     this.database = database;
   }
 
+  ERRORS = {
```
wlthomson: TODO: wonder if at some point it will be good to manage all `ERRORS` in a single source file and import rather than define in scope of each file?

src/database/UIDatabase.js
```diff
@@ -35,36 +33,53 @@ export class UIDatabase {
...
-   * Closes the database, exports the realm file to '/Download/mSupplyMobile\ data' on device
-   * file system. The app will crash if anything tries to access the database while it is closed.
+   * Exports the realm file to '/Download/mSupplyMobile\ data' on device file system.
+   * Ensures there is enough space, the realm exists and requests storage permission,
+   * if required.
    */
-  exportData(filename = 'msupply-mobile-data') {
-    // TO DO: add method to react-native-database.
 
+  async exportData(filename = 'msupply-mobile-data') {
```
wlthomson: Wonder if worth putting this and directory path in a constant? e.g.:

```
EXPORT_DIR = '/Download/mSupplyMobile_data'
EXPORT_FILE = 'msupply-mobile-data'
```

Or, if you like objects:

```
PATHS: {
   export_dir: '/Download/mSupplyMobile_data'
   export_file: 'msupply-mobile-data'
}
```

Or some variation of this.

src/database/UIDatabase.js
joshxg: definitely, cheers ff2930f8d4232831b656fd516710f2c6025548bd

src/database/UIDatabase.js
joshxg: 2666aae4249731967e787db0861901fce19e276d

src/database/UIDatabase.js
joshxg: definitely, cheers 842c72e1047ef6f3a686a43ee9ee7bdf9763e37b

src/utilities/fileSystem.js
joshxg: Reasoning was that I wanted some buffer than having exactly the number of bytes and `* 2` seemed arbitrarily as good as any other number..!

src/utilities/requestPermission.js
joshxg: Mainly as it's a generic method there might be some use out of it!

src/utilities/fileSystem.js
joshxg: 87d05a28c19de7ca04e603b92a4a94c9a58f34d7 cheers

src/database/UIDatabase.js
joshxg: definitely. I didn't really like doing this but I do like returning more than a boolean

src/database/UIDatabase.js
joshxg: 5c6bf9ca298ce98bf2b615059c7410aa90aebfc5 cheers

src/utilities/requestPermission.js
wlthomson: In this case the parameter is essentially `const` right? As the caller will already have the variable and the method is "pure" with respect to that parameter, is it worth returning it and potentially causing confusion?

src/utilities/fileSystem.js
wlthomson: Fair enough, not sure if a constant valued buffer would be better. Having it as a function of the realm file size means that the bigger the file, the bigger the required buffer, which could be problematic?

joshxg:
Closed due to duplicate PR #1047

PR #1085, opened 2019-08-02
#1080 item batches incorrect supplier

### LINKED ISSUE #1080: Item batches always have supplier as default supplier when synced out
Templates are ded, @joshxg is going to resurrect them.

**Description**

While I was looking there, I think I've spotted a more subtle bug from when only ever having one supplier was also assumed: https://github.com/openmsupply/mobile/blob/master/src/sync/outgoingSyncUtils.js#L92

This would make it look like all stock came from the same supplier (the one set on the mobile store's name record as default I think) regardless of what store/name the supplier invoice came from.

**Expected behaviour**

That the supplying store is saved on the the itemBatch when the records are created, and that in sync out it simply reads the ID from the related `Name` record.

joshxg:
Fixes #1080

## Change summary
Adds code to use the `Transactions` confirmation code. Only run for `supplierInvoices` and will add the supplier invoices `otherParty` field to the `ItemBatch.supplier` field when created.

In mSupply when batches are created in a stock take no `name_ID` field is applied, so no changes need to be made there?

## Testing

- [ ] Create a supplier invoice with an item, using a non-supplying store store see the item line has the correct supplier field. EDIT: finalise invoice.
- [ ] sync the invoice, see the `item_line.name_ID` record is correct
- [ ] Run migration code to see if item lines are set correctly
    - Could do this by removing the line in code which sets the supplier field, and putting the migration code in mSupplyMobile component did mount
    - Upgrade a mSupplyMobile app with a new version

I think this has potential to not be trivial, so some ramblings from a mad man:

I *think* I covered all the use cases of batches being created synced out etc.

- `ItemBatch` records are assigned the correct `supplier` on incoming sync
- `Stocktake` don't set a supplier when creating a new `ItemBatch`
-  `CustomerInvoice`s are always confirmed, so never run the `Transaction.confirm` code

Maybe within the migration code I need to check other status'? I don't think it's possible to be confirmed, but not finalised - is there a case where a `confirm` status supplier invoice could be synced to mobile and the finalization happens on mobile, so it doesn't run the `Transaction.confirm` code? But in that case, the `ItemBatch` record would be synced from desktop also and would have it's `name_ID` field set, which would set the `ItemBatch.supplier` field on incoming sync..?


src/dataMigration.js
```diff
@@ -188,6 +188,46 @@ const dataMigrations = [
...
+  // all outgoing ItemBatches had their supplier field set to the default
+  // supplying store record for its store. 2.3.6 ensured that the supplier
+  // of a ItemBatch was correctly synced out, maintaining it's supplier field.
+  // This migration code
+  {
+    version: '2.3.6',
+    migrate: database => {
+      const supplierInvoiceType = 'supplier_invoice';
+      const finalisedStatus = 'finalised';
+      const itemBatches = database.objects('ItemBatch');
```
andreievg: Hmm, maybe this should only run for ItemBatches with quantity > 0

src/dataMigration.js
```diff
@@ -188,6 +188,46 @@ const dataMigrations = [
...
+        // introduced it and set the otherParty of the related transaction
+        // to the supplier of the related batch.
+        itemBatches.forEach(batch => {
+          const { id, transactionBatches } = batch;
+          // Ensure there are transaction batches before trying to find the transaction.
+          if (!(transactionBatches && transactionBatches.length !== 0)) return;
+          // There should only be one supplier invoice related to the item batch,
+          // in case somethings wrong and there are two, use the last one iterated over.
+          // Also ensure it is finalised. If it isn't, the supplier will be set when confirmed.
+          const supplierInvoice = transactionBatches.find(transactionBatch => {
```
andreievg: I am still keen on using realm 'power' since it's available
```
supplierInvoiceTransactionBatches = transactionBatches.filtered('transaction.type = $0 && transaction.status = $1', supplierInvoiceType, finalisedStatus) 
if(supplierInvoiceTransactionBatches.length < 0) return;
```
 const { otherParty: supplier } = supplierInvoiceTransactionBatches[0].transaction;

src/database/DataTypes/Transaction.js
```diff
@@ -395,6 +395,7 @@ export class Transaction extends Realm.Object {
       itemBatch.batch = this.adjustBatchName(batch);
       itemBatch.costPrice = packedToOneCostPrice;
       itemBatch.sellPrice = packedToOneSellPrice;
+      itemBatch.supplier = this.otherParty;
```
andreievg: should check isIncomingInvoice just in case:
https://github.com/openmsupply/mobile/blob/d44edcda7c7da26b1b58006a01c18d07f7fc31b0/src/pages/CustomerInvoicesPage.js#L85-L91

src/sync/outgoingSyncUtils.js
```diff
@@ -89,7 +89,7 @@ const generateSyncData = (settings, recordType, record) => {
         cost_price: String(record.costPrice),
         sell_price: String(record.sellPrice),
         total_cost: String(record.costPrice * record.numberOfPacks),
-        name_ID: settings.get(SUPPLYING_STORE_NAME_ID),
+        name_ID: String(record.supplier) && String(record.supplier.id),
```
andreievg: Screen Shot 2019-08-13 at 2 28 01 PM

`name_ID: record.supplier && String(record.supplier.id)`

src/dataMigration.js
joshxg: makes sense, but before I do it can I just make sure, because I don't really know them well enough: in `outgoingSyncUtils`, `ItemBatch` records aren't filtered by quantity. Do you think having an `ItemBatch` with potentially the wrong supplier, even though `0` quantity will effect reporting at all?

src/dataMigration.js
joshxg: 5ae6b5cdf48ac889ff2138b7e3e55b43b3318817

src/database/DataTypes/Transaction.js
joshxg: hmm, do you mean the getter `Transaction.isIncoming` - This only checks for the `supplier_invoice` type, which is done in the query, or do you mean something else?

src/sync/outgoingSyncUtils.js
joshxg: oops 54055944ed9ec2ad5d7dd51c97b47d2febc64407

joshxg:
When you're happy I will do the changes I did in the last migration PR - migration folder with version files. Don't want to do alot of changes right now

joshxg:
@andreievg could you check up on requested changes please

joshxg:
@andreievg  :(

joshxg:
♪┏(・o･)┛♪

joshxg:
♪┗ ( ･o･) ┓♪

joshxg:
♪┏(・o･)┛♪

Chris-Petty:
░░┌─┐░░░░░░░░░░┌─┐░░
 ░╔╡▐▐╞╝░┌─┐√░░╔╡▐▐╞╝░
░░└╥╥┘░╚╡▌▌╞╗ ░░└╥╥┘░░ 
   ░░╚╚░░░░└╥╥┘░░░ ╚╚░░░ 
 ░░░░░░░░░░╝╝░░░ DANCE PARTY
 ░░┌─┐░░░░░░░░░░░┌─┐░░ 
 ░╔╡XX╞╝░░░┌─┐░░╔╡XX╞╝░
 ░░└╥╥┘░░╚╡XX╞╗░░└╥╥┘░░
  ░░░╚╚░░░░└╥╥┘░░░░╚╚░░░
 ░░░░░░░░░░╝╝░░░░░░░░░░

joshxg:
♪┗ ( ･o･) ┓♪

joshxg:
♪┏(・o･)┛♪

joshxg:
♪┗ ( ･o･) ┓♪

wlthomson:
♪ ٩( ᐛ ) و  ♪

Chris-Petty:
Maybe we should @andreievg ?

joshxg:
♪┏(・o･)┛♪

joshxg:
I guess this should be updated to 3.0.0 patch migration now?

wlthomson:
## TESTS

- [x] Create a supplier invoice with an item, using a non-supplying store store see the item line has the correct supplier field
- [ ] Sync the invoice, see the item_line.name_ID record is correct. **FAILED (`item_line.name_ID` is empty string, see screenshot below)**. 

#1080-item-batches-incorrect-supplier

- [ ] Run migration code to see if item lines are set correctly. **FAILED (looks like no records were updated, sync not triggered)**.

Looks like @andreievg has already looked over this, so not sure if I'm doing something wrong with my testing here?

joshxg:
@wlthomson 

looks like I missed a step .. were the invoices finalised? suppliers only get set on SI's for finalised SI's 🤣

joshxg:
@wlthomson 

Just ran it through again. This is what I did for migration on hufflepuff harry potter demo data:

- Ensured all invoices/stocktakes were finalised in my data file
- Ran this code:

```
    this.database.write(() => {
      const itemBatches = this.database.objects('ItemBatch');
      itemBatches.forEach(itemBatch => {
        this.database.update('ItemBatch', { ...itemBatch, supplier: null });
      });
    });
```
- exported data and ran `adb pull ./storage ~/testing1` to check I had no item batches with a supplier
- Upgraded my app (I just changed the migration code versions to line up and not have to deal with stuff)
- Exported again - checked suppliers were applied

For updating:

Just finalised a SI with a comment `joshX` and a line, checked on desktop 

```
alert(String(ds.transact.query("comment == :1";"@josh1@").transLines.itemLine.name_ID[0]=ds.store.get(ds.name.query("name == :1";"@huf@")[0].supplying_store_id).name_ID))
```

wlthomson:
> @wlthomson
> 
> Just ran it through again. This is what I did for migration on hufflepuff harry potter demo data:
> 
>     * Ensured all invoices/stocktakes were finalised in my data file
> 
>     * Ran this code:
> 
> 
> ```
>     this.database.write(() => {
>       const itemBatches = this.database.objects('ItemBatch');
>       itemBatches.forEach(itemBatch => {
>         this.database.update('ItemBatch', { ...itemBatch, supplier: null });
>       });
>     });
> ```
> 
>     * exported data and ran `adb pull ./storage ~/testing1` to check I had no item batches with a supplier
> 
>     * Upgraded my app (I just changed the migration code versions to line up and not have to deal with stuff)
> 
>     * Exported again - checked suppliers were applied
> 
> 
> For updating:
> 
> Just finalised a SI with a comment `joshX` and a line, checked on desktop
> 
> ```
> alert(String(ds.transact.query("comment == :1";"@josh1@").transLines.itemLine.name_ID[0]=ds.store.get(ds.name.query("name == :1";"@huf@")[0].supplying_store_id).name_ID))
> ```

Got it. Finalising fixed it. 

Followed these steps and replicated your results, so passing all tests.

## TESTS

- [x] Create a supplier invoice with an item, using a non-supplying store store see the item line has the correct supplier field. EDIT: finalise invoice.
- [x] sync the invoice, see the item_line.name_ID record is correct
- [x] Run migration code to see if item lines are set correctly

joshxg:
@wlthomson have updated from develop and merged.. although there is still the question of the migration version - I dunno if we are doing a `2.3.6` or if this should be migration for `3.0.0` 🤷‍♂

wlthomson:
> @wlthomson have updated from develop and merged.. although there is still the question of the migration version - I dunno if we are doing a `2.3.6` or if this should be migration for `3.0.0` 🤷‍♂

It makes more sense to go into `2.3.6` (since its not data table related), but since `3.0.0` should be released soon, it's not too much of a concern?

TL;DR whatever works.

joshxg:
Have bumped the migration version and fixed conflicts

src/dataMigration.js
andreievg: The reason i mention this was to minimise sync out records.

Maybe it's ok to do it for all batches but ONLY if otherParty has changed

src/dataMigration.js
```diff
@@ -188,6 +188,46 @@ const dataMigrations = [
...
+            finalisedStatus
+          );
+          // If there aren't any transaction batches, exit.
+          if (!(supplierInvoiceTransactionBatches.length > 0)) return;
+          const { transaction } = supplierInvoiceTransactionBatches[0];
+          // If somehow the transaction batch doesn't have a transaction, return
+          if (!transaction) return;
+          const { otherParty: supplier } = transaction;
+          if (!supplier) return;
+          database.update('ItemBatch', { id, supplier });
```
andreievg: can you do this only if otherParty has changed ?

src/database/DataTypes/Transaction.js
andreievg: hmm, i was just looking at the code:
```
    const newNumberOfPacks = isIncomingInvoice
        ? itemBatch.numberOfPacks + packedToOneQuantity
        : itemBatch.numberOfPacks - packedToOneQuantity;
```

it implies this method is used for incoming and outgoing transactions, not sure if that's the case, but if it is then may be setting supplier on the batch to something undesirable. Would inventory adjustments use this method also ?

joshxg:
@andreievg  have added 98bfbfc5400e4e069356dd866a870c81c1a46913 to confirmation code also

src/dataMigration.js
joshxg: 515fdd07f890f9b71d59cf6c5a3f7faf4df96984

PR #1086, opened 2019-08-02
#1062 Rejecting ItemBatches not for the mobile store

### LINKED ISSUE #1062: Incoming sync to integrate records with own storeID only
## Describe the bug
In IncomingSyncUtilities, we don't check that incoming sync record is actually for current store id. This may cause issue (and already has in Ivory Coast) where site A data is transferred to site B, causing site B to integrate site A data as their own

### To reproduce 
Haven't tried, but:
1. Set up two mobile sites from one data file
2. Try making both sites collector of the other
3. Add some transactions that modify stock
4. Sync both mobiles sites
5. Should see item line info from one sites mixed with the other one

### Expected behaviour
Incoming records for foreign site should not be integrated

### Version and device info
 - 2.3.4

Implementer to list which record types need this storeID check, and as @Chris-Petty suggested, there may not be an easy solution to records like trans lines as they don't have store id, only linked transact has store id

Also as @Chris-Petty suggested, bug snag this behaviour when it happens

joshxg:
Fixes #1062 

## Change summary

- Rejects any `ItemBatch` incoming records with a `storeID` different from the app's `storeID`

## Testing

- [ ] Sync an `[item_line]` from desktop which has an incorrect storeID for the apps store. It doesn't get integrated.
    - See: https://github.com/openmsupply/mobile/issues/1062#issuecomment-516368874 for a way to do this from desktop
    - Alternatively, hack `integrateRecord` with (I think, I didn't test this way just might be easier!):
```
  if (internalRecordType === 'ItemBatch') {
    syncRecord.store_ID = 'not this store!';
  }
```

I think this is fine for now, but potentially better solutions for better general defensive sync:

- List of all record types of store data which shouldn't be sent. Check storeID's of everything, ensuring it is for this store (and exists). Bugsnag anything caught.
Seems like this is a basic sanity check, but would need to compile a list of exactly what records shouldn't be getting sent and most likely do this check not in the current `sanityCheckMethod`, (which is only for create/update, not merge/delete) as it most likely needs to be run earlier. Would need to be careful the list is correct, such that any caught bugsnags would indicate a BIG PROBLEM in sync.

There is also a lot of other bugsnaging that could be done all over, so maybe need to put thought in the best way to do this (passing to bugsnag custom, non-error objects) but maybe I am over thinking it

### DIFF
--- a/src/sync/incomingSyncUtils.js
+++ b/src/sync/incomingSyncUtils.js
@@ -322,6 +322,8 @@ export const createOrUpdateRecord = (database, settings, recordType, record) => 
       break;
     }
     case 'ItemBatch': {
+      // Not for this store
+      if (record.store_ID !== settings.get(THIS_STORE_ID)) break;
       const item = database.getOrCreate('Item', record.item_ID);
       const packSize = parseNumber(record.pack_size);
       internalRecord = {


joshxg:
are we good to merge this?

Chris-Petty:
@wlthomson @joshxg shall we agree that if reviewer approves, they should just merge on the bat. Keeping in mind the merger should have tested it of course?

I'll let @wlthomson hit the button on this one.

wlthomson:
Merged 👍 .

PR #1098, opened 2019-08-13
#1095 by program refactor 3

### LINKED ISSUE #1095: ByProgramModal refactor
## Describe

`ByProgramModal` component was quite rushed and is confusing. There is an issue #1039 which requires a small change, but I think it's a good opportunity to rewrite/refactor the component.

### Expected

The component is going to have quite a complex state regardless of implementation. State needs to hold the current data that has been selected, the modal data for at least the current step, but ideally any previous steps (selections like order type are not trivial to compute), as if there is no data to select from for the current step, a user should have feedback. i.e. no periods for the previously selected program+order type. So using a reducer hook should help manage this.

### Version and device info

- App version: 2.3+

### Additional context

Have done some work on this already. To break it up there will be PR's for

- [ ] `Step` component
- [ ] New `byProgramModal`, including a reducer
- [ ] Bringing it together

joshxg:
Fixes #1095 

#### Branched from #1097 

Last PR for refactor

- Hooks up the new `ByPRogramModal` to `SupplierRequisitionsPage` and `StocktakesPage`
- Tried to make minimal changes as possible, even though the logic in these is pretty spaghetti, as it should be getting change in `DataTable` rewrite

## Testing
- [ ] When creating a requisition with a program defined, by program modal opens. If no program is defined, the modal does not open
- [ ] When creating a stocktake with a program defined, by program modal opens. If no program is defined, the modal does not open

**Creating a program requisition:**
- [ ]  must select steps in order - program, supplier, period, order type.
- [ ] Can back pedal steps, where the progress is adjusted correctly
- [ ] Only order types which are defined for a program are displayed
- [ ] Only periods for the order type defined are displayed
- [ ] Order types, when being selected show the correct details (MOS etc.)
- [ ] Periods, when being selected show the correct details (number of requisitions etc.)
- [ ] Switching between general and program orders adjusts the steps correctly
- [ ] OK is enabled when all steps are complete
- [ ] closing the modal, closes the modal
- [ ] programs items are added to the requisition

**Creating a stocktakes requisition:**
- [ ] must select steps in order - program, name or name if toggled
- [ ] Can back pedal steps, where the progress is adjusted correctly
- [ ] Name used is the name applied to the stock take
- [ ] programs items are added to the stocktake

### DIFF
--- a/src/database/utilities/createRecord.js
+++ b/src/database/utilities/createRecord.js
@@ -142,7 +142,7 @@ const createRequisition = (
   user,
   { otherStoreName, program, period, orderType = {}, monthsLeadTime = 0 }
 ) => {
-  const { name: orderTypeName, maxMOS, thresholdMOS } = orderType;
+  const { name: orderTypeName = '', maxMOS = 0, thresholdMOS = 0 } = orderType || {};
   const regimenData =
     program && program.parsedProgramSettings ? program.parsedProgramSettings.regimenData : null;
   const daysToSupply = ((monthsLeadTime || 0) + (maxMOS || 1)) * 30;
--- a/src/pages/StocktakesPage.js
+++ b/src/pages/StocktakesPage.js
@@ -51,10 +51,8 @@ export class StocktakesPage extends React.Component {
 
   createNewStocktake = properties => {
     const { currentUser, database } = this.props;
-    const { steps, program } = properties;
-    const stocktakeName = steps[1].name;
+    const { program, name: stocktakeName } = properties;
     let stocktake;
-
     database.write(() => {
       stocktake = createRecord(database, 'Stocktake', currentUser, stocktakeName, program);
       stocktake.addItemsFromProgram(database);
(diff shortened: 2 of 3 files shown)


src/database/utilities/createRecord.js
```diff
@@ -142,7 +142,7 @@ const createRequisition = (
   user,
   { otherStoreName, program, period, orderType = {}, monthsLeadTime = 0 }
 ) => {
-  const { name: orderTypeName, maxMOS, thresholdMOS } = orderType;
+  const { name: orderTypeName = '', maxMOS = 0, thresholdMOS = 0 } = orderType || {};
```
wlthomson: Nice way of setting defaults. Do you need the or guard with `orderType = {}` in the parameters?

joshxg: unfortuantely defaults only work with undefined - not null - pretty ugly 🤣

wlthomson: 😢 .

Issue #1090, opened 2019-08-05
CustomerInvoicePage new DataTable


Chris-Petty:
## Is your feature request related to a problem? Please describe.

Data table is being rewritten. Context is CustomerInvoicePage is good starting to make sure it's up to par with old.

## Describe the solution you'd like

Rewritten using hooks API rather than class API. Uses new VirtualisedList based data table.
Review GenericPage and usage of GenericTablePage.

## Describe alternatives you've considered

None.

## Additional context

__Add any other context or screenshots about the feature request here.__

joshxg:
I think it's worthwhile, since we're refactoring to:

- Try to consolidate logic in all the 'pages' which are exactly the same, more or less to also give more flexibility for the inevitable customizations. i.e a better way to handle:
    - columns
    - cells
    - modals

Lots of options. E.g. for columns: Each page could be passed their columns through redux/navigator props which are a composition of objects created and determined by the current active modules.

Although I'm not sure how much time we have

Chris-Petty:
> I think it's worthwhile, since we're refactoring to:
> 
> * Try to consolidate logic in all the 'pages' which are exactly the same, more or less to also give more flexibility for the inevitable customizations. i.e a better way to handle:
>   
>   * columns
>   * cells
>   * modals
> 
> Lots of options. E.g. for columns: Each page could be passed their columns through redux/navigator props which are a composition of objects created and determined by the current active modules.
> 
> Although I'm not sure how much time we have

I think we don't have time, but good point. Related to #1092 potentially

Chris-Petty:
More specifically regarding redux/nav props, we should evaluate using redux more extensively and revisit nav -> out of data table scope IMO.

joshxg:
definitely out of scope. I agree using redux more extensively is a good idea though. Primarily for passing data which doesn't change frequently as it can cause quite a lot of performance issues if the data starts to be mutated a lot

joshxg:
Closed by #1108

Chris-Petty:
@joshxg I originally intended to leave this open till customer invoice page is properly done, styles and all. Basically when it's got the final flavour of refactor on it.

Issue #1043, opened 2019-07-01
DataTable refactor/rewrite


joshxg:
## Description

Old table is bad. We're rewriting it. See context (below) or https://github.com/sussol/react-native-data-table/issues/46 

## Plan

We'll operate a local import of code for the data table. At the end we'll take the core parts back to react-native-data-table repo

## Branches

**Feature branch:** #1043-data-table-rewrite

## Tasks
**Each task to have its own issue, branch and PR.**
### Utility
- [x] Styles
    - [x] New Props
        - ISSUE: #1119  
        - PR: #1120 
    - [x] Styles
        - ISSUE: #1090 
        - PR:   #1121 
    - [x] Row onPress
        - ISSUE: #1145 
        - PR:   #1146 
    - [x] Optimized virtualized list props
        - ISSUE: #1140 
        - PR:   #1150 
    - [x] Factoring renderCells
        - ISSUE: #1159 
        - PR: #1163 
    - [x] Factoring renderModal
        - ISSUE: #1160  
        - PR:  #1168 
    - [x] New ExpiryInputCell
        - ISSUE: #1155   
        - PR:  #1161 
    - [x] Replacing SearchBar component
        - ISSUE: #1176 
        - PR: #1178 
    - [x] Handle finalizing records not updating state
        - ISSUE: #1171 
        - PR: #1174 
    - [x] Handle navigation side effects not updating state
        - ISSUE: #1172 
        - PR: #1175 
    - [x] Programmatically scrolling data table.
        - ISSUE: #1181 
        - PR: #1187 
    - [x] Tidy up flux methods
        - ISSUE:  #1169 
        - PR: #1204 
    - [x] Dynamic actions 
        - ISSUE:  #1210
        - PR: #1213 
    - [x] Column update
        - ISSUE:  #1209 
        - PR: #1214 
    - [x] small updates
        - ISSUE:  #1252
        - PR:
    - [x] Column typing
        - ISSUE:  #1250
        - PR: #1250
    - [ ] State Initialisers
        - ISSUE:  #1249 
        - PR:
    - [x] Price and unit columns
        - ISSUE:  #1248 
        - PR: #1248 
    - [x] useCallback calls
        - ISSUE:  #1247 
        - PR: #1247 
    - [x] useSyncListener
        - ISSUE:  #1246 
        - PR: #1246 
        
### Pages
Need to rewrite all pages to use new data table.

- [x] CustomerInvoicePage for reference on all following pages 
[...]

PR #1114, opened 2019-08-18
#1110 New CustomerInvoicePage sorting feature

### LINKED ISSUE #1110: New CustomerInvoicePage sorting feature

The new CustomerInvoicePage needs a sorting feature.

We need to use a plain array filled with realm objects for displaying data, not a realm results object. This means there is a choice

- Check if the field to sort on is a getter or a real field. If it is a real field, sort on the backing data of the page, a realm results, then apply the search filter, then convert back to an array. If it is a getter, sort using array methods. Could potentially not apply the search filter and just return a sorted set of data, clearing the search filter

OR

- Sort using array methods

Should check performance of these two methods.

Two options as described above

## Additional context

Have done simple sorting tests previously and sorting on a plain array tends to be just as fast, if not faster in some cases than using realm.

Could debounce this

joshxg:
Fixes #1110 

#### EPIC: #1043
~~#### BRANCHED FROM #1090 - Should be merged before this~~ done

## Change summary

- Adds a debounce method stolen from `underscore.js`
- Adds sorting functionality
    - I found layouts were a fraction faster/the same as using realm for up to 100 records, with realm starting to pull slightly ahead after that, for non-getter fields. The same for getter fields. Didn't think the performance gain of a few miliseconds was worth the complexity increase. However, very easy to refactor if we do think it is worth it.

## Testing
- [ ] Customer invoice page can be sorted be column

- Rename sorting method when integrating this feature into develop
- Different way to steal `underscore.js` methods?

#### Branches:

- [ ] #1115 - Data table filtering 
- [ ] #1116 - Data table data type synchronization
- [ ] #1117 - Data table refactor

### DIFF
--- a/src/pages/CustomerInvoicePage.js
+++ b/src/pages/CustomerInvoicePage.js
@@ -307,7 +307,7 @@ const deselectAll = () => ({
 });
 
 const sortData = (sortBy, isAscending) => ({
-  type: 'sortData',
+  type: 'sortBy',
   sortBy,
   isAscending,
 });
--- a/src/pages/CustomerInvoicePage.js
+++ b/src/pages/CustomerInvoicePage.js
@@ -3,7 +3,7 @@
  * Sustainable Solutions (NZ) Ltd. 2019
  */
 
-import React, { useState, useReducer, useCallback } from 'react';
+import React, { useState, useReducer, useCallback, useMemo } from 'react';
 import PropTypes from 'prop-types';
 import { View, StyleSheet } from 'react-native';
 import { SearchBar } from 'react-native-ui-components';
@@ -11,7 +11,7 @@ import Icon from 'react-native-vector-icons/Ionicons';
 import FAIcon from 'react-native-vector-icons/FontAwesome';
 
 import { createRecord } from '../database';
-import { formatDate, parsePositiveInteger, sortDataBy } from '../utilities';
+import { formatDate, parsePositiveInteger, debounce } from '../utilities';
 import { buttonStrings, modalStrings, pageInfoStrings, tableStrings } from '../localization';
 import { AutocompleteSelector, PageButton, PageInfo, TextEditor } from '../widgets';
 import { BottomConfirmModal, PageContentModal } from '../widgets/modals';
@@ -27,7 +27,6 @@ import {
 
 import globalStyles, { dataTableColors } from '../globalStyles';
 
-const DATA_TYPES_SYNCHRONISED = ['TransactionItem', 'TransactionBatch', 'Item', 'ItemBatch'];
 const MODAL_KEYS = {
   COMMENT_EDIT: 'commentEdit',
   THEIR_REF_EDIT: 'theirRefEdit',
@@ -310,7 +309,6 @@ export const CustomerInvoicePage = ({
   database,
   genericTablePageStyles: pageStyles,
   runWithLoadingIndicator,
-  topRoute,
 }) => {
   const [tableState, dispatch] = useReducer(reducer, {
     data: transaction.items.slice(),
@@ -324,7 +322,7 @@ export const CustomerInvoicePage = ({
   const [modalKey, setModalKey] = useState(null);
   const [modalIsOpen, setModalIsOpen] = useState(false);
   const { ITEM_SELECT, COMMENT_EDIT, THEIR_REF_EDIT } = MODAL_KEYS;
-  const { data, dataState, searchTerm, sortBy, isAscending } = tableState;
+  const { data, dataState, sortBy, isAscending } = tableState;
   let isSelection = false;
 
   // eslint-disable-next-line no-restricted-syntax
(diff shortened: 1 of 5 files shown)


src/utilities/sortDataBy.js
```diff
@@ -38,4 +38,33 @@ export const sortDataBy = (data, sortBy, sortDataType, isAscending = true) => {
...
+      if (isAscending) return [...data.sort((a, b) => Number(a[sortBy]) - Number(b[sortBy]))];
+      return [...data.sort((a, b) => Number(b[sortBy]) - Number(a[sortBy]))];
+    case 'date':
+      if (isAscending) return [...data.sort((a, b) => new Date(b[sortBy]) - new Date(a[sortBy]))];
+      return [...data.sort((a, b) => new Date(a[sortBy]) - new Date(b[sortBy]))];
+    case 'boolean':
+      if (isAscending) return [...data.sort((a, b) => b[sortBy] - a[sortBy])];
+      return [...data.sort((a, b) => a[sortBy] - b[sortBy])];
+    default:
+      return [...data];
```
Chris-Petty: Potentially triggers unnecessary re-render due to array ref change? Could throw warning/error for invalid sortDataType instead?

src/utilities/sortDataBy.js
```diff
@@ -38,4 +38,33 @@ export const sortDataBy = (data, sortBy, sortDataType, isAscending = true) => {
...
 
+/**
+ * Sorts an array of objects, returning a new array.
+ * Sorts strings, numbers, dates or booleans.
+ * Types: 'string', 'number', 'date', 'boolean'.
+ * @param {Array}   data          Array of objects to sort
+ * @param {String}  sortBy        Key for the field to sort by
+ * @param {String}  sortDataType  The type of data to sort
+ * @param {Boolean} isAscending   True if ascending, false otherwise.
+ */
```
Chris-Petty: return?

src/pages/CustomerInvoicePage.js
```diff
@@ -627,7 +648,7 @@ CustomerInvoicePage.propTypes = {
   database: PropTypes.object.isRequired,
   genericTablePageStyles: PropTypes.object.isRequired,
   runWithLoadingIndicator: PropTypes.func.isRequired,
-  topRoute: PropTypes.bool.isRequired,
+
```
Chris-Petty: Remove blank line.
That said, I only left this here cause it was part of a feature: Refresh a page table if navigating back to it.

src/pages/CustomerInvoicePage.js
```diff
@@ -251,6 +250,19 @@ const reducer = (state, action) => {
...
+      const { sortBy } = action;
+      const columnKeyToDataType = {
+        itemCode: 'string',
+        itemName: 'string',
+        availableQuantity: 'number',
+        totalQuantity: 'number',
+      };
+
+      const newData = newSortDataBy(data, sortBy, columnKeyToDataType[sortBy], isAscending);
+      return { ...state, data: newData, sortBy, isAscending: !isAscending };
```
Chris-Petty: Should `isAscending` be negated here? Should also look at old implementation; If state sortBy was not this action's sortBy, then isAscending should be true. If state sortBy matches action's sortBy, this isAscending is false.

src/pages/CustomerInvoicePage.js
```diff
@@ -251,6 +250,19 @@ const reducer = (state, action) => {
       }
       return { ...state, dataState: newDataState };
     }
+    case 'sortBy': {
+      const { data, isAscending } = state;
+      const { sortBy } = action;
+      const columnKeyToDataType = {
```
Chris-Petty: Just an observation: Probably slightly more optimal to declare this outside the reducer, but I approve the convenience of having it right where it's used.

src/utilities/sortDataBy.js
joshxg: 5d9841e29875921b64d2bccc5e6b61b2819fb48b

src/utilities/sortDataBy.js
joshxg: a697c3b69edd32028590a87610fe50f13db11663

src/pages/CustomerInvoicePage.js
joshxg: 1f070fedd9c8e44b55a3b1d298669f89383e0e89

src/pages/CustomerInvoicePage.js
joshxg: ee4d03b77b3eb9ddf69486e6e1a195e790d649f0  woah oops, funky logic - this should be correct

src/pages/CustomerInvoicePage.js
joshxg: 1000% this should not be here - just not 100% where to put it just yet

joshxg:
Have merged in #1043-data-table-rewrite and rebased

src/pages/CustomerInvoicePage.js
Chris-Petty: I suppose it will be constant for most pages. Deal with it soon(_tm_).

src/pages/CustomerInvoicePage.js
Chris-Petty: I commented "nice" on the commit. Idk why we can comment on commits yet.

PR #1115, opened 2019-08-18
#1111 New CustomerInvoicePage filtering feature

### LINKED ISSUE #1111: New CustomerInvoicePage filtering feature

The new CustomerInvoicePage needs a filtering feature.

The same as sorting - need an array of realm objects to display data, but also have access to a realm results backing data.

options:
- Filter on the backing data using realm queries, apply sorting, convert to an array. I think clearing sorting and simply filtering and converting is a viable option. Is filtering ever needed on getter fields?

OR

- Filter and return the data array using basic array methods

I'd suspect that realm would be a fair bit quicker for this. Especially if we index common filter fields, which is a valid use case. 

Should filtering be done on potentially many fields?

As above, couple options.

## Additional context

Might need to check the performance of both options, highly suspect realm queries to be faster. Is clearing the current sort state terrible after a search?

Should also debounce this

joshxg:
Fixes #1111

**Forgot to rename this branch before publishing, oopsies**
## Change summary

- Adds a simple filtering mechanism
- Went with the simple approach, using realm to query a `backingData` in state, returning a filtered version of that.
- Could add realm sorting on top, but would also have to determine if the field is a getter/setter etc etc - just doing it simple

## Testing
- [ ] `CustomerInvoicePage` filtering functionality works

As above - dealing with sorting while filtering. I think keeping this simple for now is a good option, personally - just removing the sorting. Hard to know if the added complexity is really actually worth it. I don't see adding a filter then pushing a sort button to be odd

### DIFF
--- a/src/pages/CustomerInvoicePage.js
+++ b/src/pages/CustomerInvoicePage.js
@@ -148,12 +148,26 @@ const reducer = (state, action) => {
 
   switch (action.type) {
     case 'filterData': {
-      const { backingData, filterDataKeys } = state;
+      const { backingData, filterDataKeys, sortBy, isAscending } = state;
       const { searchTerm } = action;
+
+      const columnKeyToDataType = {
+        itemCode: 'string',
+        itemName: 'string',
+        availableQuantity: 'number',
+        totalQuantity: 'number',
+      };
+
       const queryString = filterDataKeys
         .map(filterTerm => `${filterTerm} BEGINSWITH[c]  $0`)
         .join(' OR ');
-      const newData = backingData.filtered(queryString, searchTerm).slice();
+
+      const newData = newSortDataBy(
+        backingData.filtered(queryString, searchTerm).slice(),
+        sortBy,
+        columnKeyToDataType[sortBy],
+        isAscending
+      );
       return { ...state, data: newData };
     }
     case 'editCell': {
--- a/src/pages/CustomerInvoicePage.js
+++ b/src/pages/CustomerInvoicePage.js
@@ -147,6 +147,15 @@ const reducer = (state, action) => {
   };
 
   switch (action.type) {
+    case 'filterData': {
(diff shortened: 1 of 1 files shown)


src/pages/CustomerInvoicePage.js
```diff
@@ -390,7 +406,12 @@ export const CustomerInvoicePage = ({
     });
   };
 
-  const onSearchChange = () => {};
+  const onSearchChange = searchTerm => {
+    dispatch(filterData(searchTerm));
+  };
+
+  const instantDebouncedDispatch = useMemo(() => debounce(dispatch, 250, true), []);
```
Chris-Petty: unused?

src/pages/CustomerInvoicePage.js
```diff
@@ -390,7 +406,12 @@ export const CustomerInvoicePage = ({
     });
   };
 
-  const onSearchChange = () => {};
+  const onSearchChange = searchTerm => {
+    dispatch(filterData(searchTerm));
```
Chris-Petty: would be cleaner and make the func given to onChange have this better variable name if you just put debounce in here?

src/pages/CustomerInvoicePage.js
joshxg: Used for a header cell

src/pages/CustomerInvoicePage.js
joshxg: I think I should do a PR for `SearchBar` to accept an optional `action` prop so we can bypass this function altogether. I haven't tried but I *think* putting `debounce` in here will return a new function each call, resulting in all invocations not having the same reference to the timeout variable to cancel consecutive calls?

joshxg:
Have added post filtering sorting : 25538b38b4bb68f42afc6496054e666897f4569f

(Just duplicated the `columnKeyToDataType` mapping object until we find a nice home

PR #1116, opened 2019-08-18
#1112 New CustomerInvoicePage data types shnychronizationh feature

### LINKED ISSUE #1112: New CustomerInvoicePage data type synchronization feature

When sync occurs and a record is updated - for example a new customer invoice or an item name changes, we want our data table pages to re-render to display the data.

A custom hook which is re-usable in any page that wants to subscribe to the database change events.

Hook should handle subscribing and unsubscribing and filter the data based on data types passed and the change being sync related. Should accept a callback which when called will re-render the data table.

Should just use the implemented observer pattern in `react-native-data-table`, since it's there, rather than realms observer pattern.

- Using realms observer pattern
- Not having this feature at all
- Just using a callback and `useEffect` 

## Additional context
N/A

joshxg:
Fixes #1112

## Change summary

- Adds a simple database listener 

**NOTE:** Currently when sync occurs, `mSupplyMobileAppContainer` re-renders, re-rendering the entire app (and for some reason also unmounting like, everything). So, without actually plugging this into the page, the data table will refresh on syncing a synchronized data type .. or, all data types. So it is not actually hooked up. I think it is a reasonably large refactor to fix this problem, so I've not done it yet. I had already written this hook, though, so I figured I'd PR it in. 

Have tested and it was triggering the event on the correct data types. Quite simple to hook up - just need a `useDatabaseListener( .. )` and a `callback` which sends a `dispatch` ....... `refreshData` 🤣 - could use a better action, which finds the rows the sync changes are related to, but doing this rather than just re-rendering seems premature. 

## Testing
- [ ] Hooking this hook up to a component will trigger the callback passed on incoming sync change for a data type provided.

Could add to this hook:
- Toast functionality
- Grouping the debounced calls, such that you can see '3 item names changed' etc.

### DIFF
--- a/src/hooks/useDatabaseListener.js
+++ b/src/hooks/useDatabaseListener.js
@@ -1,0 +1,59 @@
+/* eslint-disable import/prefer-default-export */
+import { useEffect, useRef } from 'react';
+import { debounce } from '../utilities';
+
+/**
+ * Custom hook for subscribing to database changes from sync.
+ * Takes a callback to call when a notified of a particular
+ * subscribed event. An array of data types to subscribe to
+ * and the database to listen to changes.
+ * When a component calling this method unmounts, changes will
+ * be unsubscribed.
+ * Callback is debounced, only being called after 10seconds of the
+ * last sync notification.
+ * Filtering is done in this hook, only notifying of sync changes
+ * for record types passed in dataTypes
+ * @param {Func}   callback  A function to call when notified of changes.
+ * @param {Array}  dataTypes Record types to listen to changes for.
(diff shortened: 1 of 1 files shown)


Chris-Petty:
I wooould approve/review this... But I'd just be merging dead code at this stage.

Issue for the overall rerender problem?

joshxg:
I'm biased because I don't want to have done this for no reason, but if it's done and going to be used *eventually*, why not 🤷‍♀

joshxg:
Pages re-render behaviour is now more tightly controlled, so the whole app re-rendering shouldn't* be causing updates. So this is now needed

PR #1117, opened 2019-08-18
#1113 New CustomerInvoicePage refactor

### LINKED ISSUE #1113: New CustomerInvoicePage factoring

`CustomerInvoicePage` has a lot of the logic within the file - a lot of this is reusable code that can be used in other pages. Should be factored out.

I think factoring out the following is a good idea as most pages use this logic:

- Reducer
    - By factoring all logic from each case statement into it's own method, we can construct composed reducers for individual pages. Each page could then take whichever methods it needs for it's own reducer
    - Could then use a 'factory' like method to create these reducers, such that each page can simply call the factory method and have it's reducer returned, dependent on not only the page, but potentially other factors in `state` or `settings`

- Actions
    - All pages pretty much use very similar/the same actions. Also if they're using basically the same reducer, would be best to factor out actions

- Columns
    - A lot of pages share columns - the only variable between pages would be the width of the columns. I think factoring these out and using the same method as above with the reducer is a good idea for reusable code and also for enabling extensible code which can determine, essentially, which 'object to instantiate' during runtime

- `PageInfo`
    - More questionable? But alot of the `PageInfo` 'columns' are shared - or at least the 'parts' i.e. comment, dates etc. 
    - I think the above method is also a very good idea for managing `PageInfo` columns

Doing this will most likely half the size of each page component, making them far easier to understand/read while also making these pages more robust to changing requirements in the future.

Not doing this and copy pasting code everywhere

## Additional context

Might be quite a bit of code to review, so might be best to split into two parts

joshxg:
Fixes #1113 

## Change summary

- Add `routeName` to `screenProps` for each page.

As per the linked issue, refactored:
- Reducer: Put all methods in one file. Made a `getReducer` method which creates a reducer for a page, given the `routeName` and `initialState`
- Columns: Refactored out the columns array. Made a `getColumns` method which creates a `columns` array for a page given it's `routeName`
- Factored out all `actions`
- Factored out all `Icon` components - resuable elsewhere. These are simple constants. I thought just having constants is fine, rather than methods which you pass parameters too. I think if props on the components need to change, it would be fine to just create a new one - I doubt the props are going to change much, and if we do create functions then rather than now?
- Created a `usePageReducer` custom hook, which will return a reducer and state injected with the columns for a specific page

## Testing

Just refactoring, so no tests required?

Still want to refactor `PageInfo`. Also I think a lot of the current event handling could be changed to be handled by the reducer, as they're most defnitely shared logic - in particular modals, adding items, adding items from master lists, adding comments and filtering


src/pages/dataTableUtilities/reducer/reducerMethods.js
```diff
@@ -0,0 +1,235 @@
...
+  });
+
+  return {
+    ...state,
+    dataState: newDataState,
+  };
+};
+
+/**
+ * Focus' the next cell, after the current cell provided in action
```
Chris-Petty:
```suggestion
 * Focus the next appropriate cell after the current cell provided in action
```

src/pages/dataTableUtilities/reducer/reducerMethods.js
```diff
@@ -0,0 +1,235 @@
...
+  for (const [rowKey, rowState] of newDataState.entries()) {
+    if (rowState.isSelected) {
+      newDataState.set(rowKey, {
+        ...rowState,
+        isSelected: false,
+      });
+    }
+  }
+  return { ...state, dataState: newDataState };
+};
```
Chris-Petty:
```suggestion
};

```

src/pages/dataTableUtilities/reducer/reducerMethods.js
```diff
@@ -0,0 +1,235 @@
...
+ */
+export const sortData = (state, action) => {
+  const { data, isAscending } = state;
+  const { sortBy } = action;
+  const columnKeyToDataType = {
+    itemCode: 'string',
+    itemName: 'string',
+    availableQuantity: 'number',
+    totalQuantity: 'number',
+  };
```
Chris-Petty: can come later, but this needs to be extensible?

src/pages/CustomerInvoicePage.js
```diff
@@ -336,22 +69,25 @@ export const CustomerInvoicePage = ({
...
     backingData: transaction.items,
     data: transaction.items.slice(),
     database,
+    keyExtractor,
     dataState: new Map(),
     currentFocusedRowKey: null,
     searchTerm: '',
     filterDataKeys: ['item.name'],
-    sortBy: 'itemName',
+    sortBy: '',
```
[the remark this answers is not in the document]
joshxg: Essentially that would mean dispatching a function on mount/first render to sort the data. Worth it you reckon?

src/pages/dataTableUtilities/columns.js
```diff
@@ -0,0 +1,57 @@
...
+ * Sustainable Solutions (NZ) Ltd. 2016
+ */
+
+import { tableStrings } from '../../localization';
+
+const PAGE_COLUMN_WIDTHS = {
+  customerInvoice: [2, 4, 2, 2, 1],
+};
+
+const PAGE_COLUMNS = {
```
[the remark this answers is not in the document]
joshxg: I am going for an implementation where the functions called (`getColumns`) may in the future have an additional parameter, say `settings`, or `database`, or some minor logic to determine if a particular pages columns need to change. 

I.E supplier invoice, with vaccine module enabled could have a `VVM Status` column. 

So, passing `settings`, we could check for a particular `module` - the `module` combined with the `routeName` would map to a particular set of columns and widths (& the same with reducer), such that the internal logic of the page doesn't actually have to change.

Do you think this is worthwhile?

src/pages/dataTableUtilities/reducer/getReducer.js
```diff
@@ -0,0 +1,61 @@
...
+  filterData,
+  editTotalQuantity,
+  selectRow,
+  deselectRow,
+  deselectAll,
+};
+
+const PAGE_REDUCERS = {
+  customerInvoice,
+};
```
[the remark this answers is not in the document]
joshxg: as above, I am trying to go for similar logic with both `reducer`, `columns` and ideally `PageInfo` (i.e. program requisitions have different `PageInfo` - I think it will be a common usecase in future customizations)

src/pages/dataTableUtilities/reducer/reducerMethods.js
joshxg: 100% - I am thinking at this point, just a constant in this file, any other suggestions? Potentially larger reducer methods could have their own file.. but that's pretty ugly

Chris-Petty: Later it is, saving my thought power

src/pages/dataTableUtilities/reducer/getReducer.js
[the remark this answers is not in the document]
Chris-Petty: Discuss above https://github.com/openmsupply/mobile/pull/1117#discussion_r315021137

joshxg:
Have rebased to data-table-filtering branch, shortcutting the data-types-sync branch and updated with sorting and filtering logic

Chris-Petty:
idk what you did and probs why rebasing is sketchy to play around with frequently, but #1-data-table-filtering is now #1112-data-table-syhcroniszsgon

Chris-Petty:
Jokes I just got confused by the linking of stuff everywhere lol

Chris-Petty:
> NOTE: Branched from #1112 - should be merged before this

I think this needs updating. Also I think these links would be more useful if they went to the PR rather than the issue

joshxg:
done

PR #1120, opened 2019-08-20
#1119 data table new component props

### LINKED ISSUE #1119: New Data Table components require new props

The new components added for the new `DataTable` - `Row`, `Cell` etc. require new props for styling.

a `prop` added for each styling required i.e. `viewStyle` to style the `View` in a component, or `textStyle` to style the `Text` component.

As it is a very common use case that components in react-native use `flex` to adjust their size, and that this is normally a dynamic property, I think it is a good idea to have components have a `width` optional property, which when used injects the passed style (or default) with a `flex` property with a value of the passed `width prop`.

Another common use case is to use `borderRight` in a row based table for indicating the column seperators. I think it would be beneficial to have a `isLastCell` property, which when true, will inject the style with a `borderRightWidth: 0` property.

Using `context` and `defaultProps` extensively. I think the above solution of using `props` is cleaner and simpler

Not having the additional `props` and having this logic within a `Page` component which uses the components. I think the use-cases are extremely common and if written in a way which can be ignored is a great feature

## Additional context

N/A

joshxg:
Fixes #1119 

## Change summary

- Added `isLastCell` and `width` `props` to data table components as described in the issue.
- Added `style` props to each component
- Added a `getAadjustedStyle` method in the `utilities` folder of `DataTable` to inject `flex` and  `borderRightWidth: 0` properties into a cells style

## Testing

**Internal code not really testable?**

Default props?


joshxg:
Did some hacking to see if there was gonna be a quick fix.

#### Navigation Stack re-rendering
- `ScreenProps` in `ReduxNavigator` is causing the large re-renders. Whenever `mSupplyMobileAppContainer` re-renders, all components in the navigation stack also re-render. Would need to be set as a constant somewhere (`componentDidMount` most likely) rather than a new object.

Current screenProps:
```
    this.screenProps = {
      database: this.database,
      settings: this.settings,
      logOut: this.logOut,
      currentUser: this.state.currentUser,
      database: this.database,
      runWithLoadingIndicator: this.runWithLoadingIndicator,
      isInAdminMode: this.state.isInAdminMode,
    };
```

#### Causes of `mSupplyMobileAppContainer` re-rendering:
- `syncState` also causes `mSupplyMobileAppContainer` to re-render. Propose that `SyncModal` and `NavigationBar` subscribe to the redux store to fetch the `syncState`, rather than being passed by `mSupplyMobileAppContainer`.
- `isLoading` causes `mSupplyMobileAppContainer` to re-render. It needs to in order to display the spinner, but fixing the above problem with `screenProps` fixes this.

#### Causes of `ReduxNavigation` re-render
- `ReduxNavigation` is being passed the current `navigationState` by `mSupplyMobileAppContainer`, which it does not use (but is always unequal) and also through redux in props also, causing every re-rendering to also re-render `ReduxNavigation`.
- Should be/can be `memoized`

Overall there doesn't seem to be a quick fix for the whole problem, but there are some which can be easily improved. I think that it is important to refactor this component. With a few hacks I improved the layout time of navigating to certain components by a lot - from ~900ms layout time to around ~150ms

src/widgets/DataTable/Cell.js
```diff
@@ -1,36 +1,48 @@
...
+import { getAdjustedStyle } from './utilities';
 
 /**
  * Renders a cell that displays a string/number value within
  * a react-native `Text` component.
  *
- * @param {string|number} value The value to render in cell
+ * @param {string|number} value     The value to render in cell
+ * @param {Object}        viewStyle Style object for the containing View
+ * @param {Object}        textStyle Style object for the inner Text
```
Chris-Petty: Add `isLastCell`

src/widgets/DataTable/Cell.js
```diff
@@ -1,36 +1,48 @@
...
+  width: PropTypes.number,
+  isLastCell: PropTypes.bool,
 };
 
 Cell.defaultProps = {
   value: '',
+  textStyle: defaultStyles.textStyle,
+  viewStyle: defaultStyles.viewStyle,
+  width: 0,
+  isLastCell: 0,
```
Chris-Petty:
```suggestion
  isLastCell: false,
```

src/widgets/DataTable/Cell.js
```diff
@@ -1,36 +1,48 @@
...
+    <View style={internalViewStyle}>
+      <Text style={textStyle}>{value}</Text>
     </View>
   );
 });
 
+const defaultStyles = StyleSheet.create({
+  viewStyle: {},
+  textStyle: {},
+});
```
Chris-Petty: Seems kind of redundant?

Chris-Petty: Suppose it makes the old just handing an array of styles with overrides work the same?

src/widgets/DataTable/DataTable.js
```diff
@@ -7,31 +8,41 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import { StyleSheet, VirtualizedList, VirtualizedListPropTypes } from 'react-native';
 
-const DataTable = React.memo(({ renderRow, renderHeader, ...otherProps }) => (
+/**
+ * Base DataTable component. Thin wrapper around VirtualizedList, providing
+ * a header component. All VirtualizedList props can be passed through,
+ * however renderItem is renamed renderRow.
+ *
+ * @param {Func}   renderRow    renaming of virtualizedList renderItem prop.
```
Chris-Petty:
```suggestion
 * @param {Func}   renderRow    Renaming of VirtualizedList renderItem prop.
```

src/widgets/DataTable/HeaderCell.js
```diff
@@ -14,6 +16,8 @@ import { TouchableOpacity, Text, StyleSheet, View, TouchableOpacityPropTypes } f
  * @prop {ReactElement} SortAscComponent      Icon component for ascending sorting
  * @prop {ReactElement} SortDescComponent     Icon component for descending sorting
  * @prop {ReactElement} SortNeutralComponent  Icon component for neutral state, no sort.
+ * @prop {Object}       containerStyle        Style object for the wrapping Touchable or View.
+ * @prop {Object}       textStyle             Style object for the inner text component.
```
Chris-Petty: Hope I'm not too annoying
```suggestion
 * @prop {Object}       textStyle             Style object for the inner Text component.
```

src/widgets/DataTable/HeaderCell.js
```diff
@@ -66,6 +82,10 @@ HeaderCell.propTypes = {
   SortAscComponent: PropTypes.element,
   SortDescComponent: PropTypes.element,
   SortNeutralComponent: PropTypes.element,
+  containerStyle: PropTypes.object,
+  textStyle: PropTypes.object,
+  width: PropTypes.number,
+  sortable: PropTypes.bool,
```
Chris-Petty: `isLastCell` missing

src/widgets/DataTable/HeaderCell.js
```diff
@@ -1,7 +1,9 @@
+/* eslint-disable react/forbid-prop-types */
```
Chris-Petty: I reckon we just turn this off globally

src/widgets/DataTable/TouchableCell.js
```diff
@@ -52,18 +69,16 @@ TouchableCell.propTypes = {
   dispatch: PropTypes.func.isRequired,
   renderChildren: PropTypes.func.isRequired,
   TouchableComponent: PropTypes.func.isRequired,
+  containerStyle: PropTypes.object,
+  textStyle: PropTypes.object,
+  isLastCell: PropTypes.bool,
```
Chris-Petty: missing `width`

src/widgets/DataTable/utilities.js
```diff
@@ -0,0 +1,7 @@
+/* eslint-disable import/prefer-default-export */
+export const getAdjustedStyle = (style, width, isLastCell) => {
```
Chris-Petty: JS doc?

src/widgets/DataTable/Cell.js
joshxg: Oops missed a few added here f51041edef541cdc14ec5d1cfb8ac21dd374992f

src/widgets/DataTable/Cell.js
joshxg: I definitely don't like this, but I'm not sure what to put. I was thinking of adding either our styles, or some basic styles for it to still look like a table before porting them over to `react-native-data-table`?

src/widgets/DataTable/EditableCell.js
```diff
@@ -27,6 +27,9 @@ import { getAdjustedStyle } from './utilities';
  * @param {Object} touchableStyle Style object for the wrapping Touchable component
  * @param {Object} viewStyle Style object for the wrapping View component
  * @param {Object} textStyle Style object for the inner Text component
+ * @param {Number}        width Optional flex property to inject into styles.
+ * @param {Bool}          isLastCell Indicator for if this cell is the last
```
Chris-Petty: This why I don't bother aligning white space generally. I'll generally start the comment on tab after the param name and thats it.

src/widgets/DataTable/HeaderCell.js
```diff
@@ -9,15 +9,18 @@ import { getAdjustedStyle } from './utilities';
...
+ * @param {String}       columnKey             The key for the column the cell heads.
+ * @param {Func}         onPressAction         Action for dispatching on press
+ * @param {Func}         dispatch              Dispatcher to backing reducer
+ * @param {ReactElement} SortAscComponent      Icon component for ascending sorting
+ * @param {ReactElement} SortDescComponent     Icon component for descending sorting
+ * @param {ReactElement} SortNeutralComponent  Icon component for neutral state, no sort.
+ * @param {Object}       containerStyle        Style object for the wrapping Touchable or View.
+ * @param {Object}       textStyle             Style object for the inner text component.
+ * @param {Number}        width               Optional flex property to inject into styles.
+ * @param {Bool}          isLastCell          Indicator for if this cell is the last
```
Chris-Petty: Once space out 😛

src/widgets/DataTable/HeaderCell.js
joshxg: 587bfcd3594d21a1a3a206d05cdcc7c1d1499773

Note to self: commit all the suggestions first :(

src/widgets/DataTable/HeaderCell.js
joshxg: dbf837f7e9dcd5894189d2b535575dd2daa9a3a5

src/widgets/DataTable/HeaderCell.js
joshxg: I'll do a PR into develop, I reckon?

src/widgets/DataTable/TouchableCell.js
joshxg: 57ef234e0b7c00a885b67abf201ed88a8aedf4e9

src/widgets/DataTable/utilities.js
joshxg: 5030ce17dc5cc74477a90aa295569ee00ea0f881

src/widgets/DataTable/HeaderCell.js
joshxg: I do not have the required skills to fix this :(

joshxg:
Have removed default styles from components

PR #1121, opened 2019-08-20
#1099 Data table styles

### LINKED ISSUE #1099: Data table rewrite, styles

The data table rewrite requires a styling.

Just using props seems like the most efficient and flexible solution

Using `Context` API to pass styling - seems less flexible and potentially an over engineered solution

## Additional context

Will be going into the Data table rewrite feature branch

joshxg:
Fixes #1099 

## Change summary

- Adds `newDataTableStyles` and `newPageStyles` files instead of the current `pageStyles` and `dataTableStyles` files - I thought to keep these separate - when we have finished changing each table after this feature is complete, we can more easily see if there is anything else that can be tidied up
- Does not use `StyleSheet` - `StyleSheet` has done more or less nothing for a long time. It's main use case was for refering to style sheets over the bridge through IDs, but this was removed a long time ago and doesn't look like it's on the radar to re-implement https://facebook.github.io/react-native/docs/stylesheet.html
- Adds a container component `DataTablePageView` which is simple two `Views` with styles to wrap each table page

## Testing

- [ ] Use either the mobile docs, or if you have a fantastic memory, check that the `CustomerInvoice` page looks the same as before any changes!

- `defaultStyles` might need some work, or could add a few styles to them? Wasn't 100% what to do with them.


src/widgets/DataTable/TouchableCell.js
```diff
@@ -68,7 +68,7 @@ TouchableCell.propTypes = {
   onPressAction: PropTypes.func.isRequired,
   dispatch: PropTypes.func.isRequired,
   renderChildren: PropTypes.func.isRequired,
-  TouchableComponent: PropTypes.func.isRequired,
+  TouchableComponent: PropTypes.node,
```
Chris-Petty: node or element? A 30sec research tells me that a `number` would be valid here, which would be poop.

src/widgets/icons.js
```diff
@@ -2,16 +2,17 @@
...
 import FAIcon from 'react-native-vector-icons/FontAwesome';
 
 import { dataTableColors } from '../globalStyles';
 
-export const SortAscIcon = <FAIcon name="sort-asc" size={15} color="black" />;
-export const SortNeutralIcon = <FAIcon name="sort" size={15} colour="black" />;
-export const SortDescIcon = <FAIcon name="sort-desc" size={15} colour="black" />;
+export const SortAscIcon = <FAIcon name="sort-asc" size={15} style={{ marginRight: 10 }} />;
+export const SortNeutralIcon = <FAIcon name="sort" size={15} style={{ marginRight: 10 }} />;
+export const SortDescIcon = <FAIcon name="sort-desc" size={15} style={{ marginRight: 10 }} />;
```
Chris-Petty: Margin left? Or does the title have margin right too?

Chris-Petty: There are many ways to do this and IDK what's the best.

src/globalStyles/newDataTableStyles.js
```diff
@@ -0,0 +1,128 @@
...
+      justifyContent: 'center',
+      textAlign: 'center',
+    },
+  },
+  cellText: {
+    left: {
+      marginLeft: 20,
+      textAlign: 'left',
+      fontFamily: APP_FONT_FAMILY,
+      fontSize: Dimensions.get('window').width / 100,
```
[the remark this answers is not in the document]
joshxg: Not that I could find so I made one 9f24d4f1ac209282c4de44d91ec3c0585d1f5905

src/pages/CustomerInvoicePage.js
```diff
@@ -67,7 +71,6 @@ const keyExtractor = item => item.id;
 export const CustomerInvoicePage = ({
   transaction,
   database,
-  genericTablePageStyles: pageStyles,
```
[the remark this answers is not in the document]
joshxg: me too - just a couple more things to take out of `screenProps` 🤣

src/widgets/DataTable/TouchableCell.js
joshxg: Yeah, so I went a little further down the rabbit hole. All kinds of answers and I can't find a solid standard. I've had the most 'success' with `node` but happy to change to anything

src/widgets/icons.js
joshxg: argh this was a gooden. Took me awhile to figure out how `generic..` was getting the right margin on the icons hehe. I was really just taking the styles and applying them, trying to 'tidy' up where I could. 

There's certainly a lot of redrawing, but maybe one area of performance enhancing at a time? 🤣 

image

Issue #1123, opened 2019-08-20
mSupplyMobileAppContainer causing excessive re-renders


joshxg:
## Describe the bug

The component `mSupplyMobileAppContainer` causes excessive re-renders due to it's state changing.

The main culprits are the current `syncState` and also `isLoading` (from opening the `Spinner` component).

Although it would be more efficient to have `mSupplyMobileAppContainer` not deal with `syncState` (I think `isLoading` is potentially unavoidable?) I believe the problem is to do with passing `ReduxNavigator` props which are not equal, causing the entire stack to re-render. 

`mSupplyMobileAppContainer` subscribes to redux and fetches data which could also be a source of re-renders 

The problem here is in particular having large re-renders of components when opening the spinner and when sync occurs - this causes the entire app to be unresponsive.

### To reproduce

1. Set the `syncInterval` in `mSupplyMobileAppContainer` to a smaller number (1 sec, or 10 if you're not a nutter)
2. Add logs to `mSupplyMobileAppContainer`'s `render` method
3. See the re-rendering
**Bonus**
4. If you put the sync interval to 1 second, the app will be unresponsive

### Expected behaviour

re-rendering `mSupplyMobileAppContainer` should not causes the entire app to re-render

### Version and device info

- App version: All?
- Tablet model: All
- OS version: All

### Additional context

I think in general `mSupplyMobileAppContainer` should be refactored as there are a few areas that could be improved

joshxg:
@andreievg @Chris-Petty 

Did some hacking to see if there was gonna be a quick fix.

#### Navigation Stack re-rendering
- `ScreenProps` in `ReduxNavigator` is causing the large re-renders. Whenever `mSupplyMobileAppContainer` re-renders, all components in the navigation stack also re-render. Would need to be set as a constant somewhere (`componentDidMount` most likely) rather than a new object.

Current screenProps:
```
    this.screenProps = {
      database: this.database,
      settings: this.settings,
      logOut: this.logOut,
      currentUser: this.state.currentUser,
      database: this.database,
      runWithLoadingIndicator: this.runWithLoadingIndicator,
      isInAdminMode: this.state.isInAdminMode,
    };
```

#### Causes of `mSupplyMobileAppContainer` re-rendering:
- `syncState` also causes `mSupplyMobileAppContainer` to re-render. Propose that `SyncModal` and `NavigationBar` subscribe to the redux store to fetch the `syncState`, rather than being passed by `mSupplyMobileAppContainer`.
[...]

PR #1124, opened 2019-08-21
#1122 Navigating unmounts components

### LINKED ISSUE #1122: Navigating unmounts components
## Describe the bug

When navigating in the app, each page on the `NavigatigationStack` is `unmounted` every time navigation occurs.  This is a significant preformance degradation, as when a user is navigating to a page 3 deep, there are three large unmounting, then remounting process' occuring, including an initial render. These are extremely slow process' compared to a simple re-render.

### To reproduce
1. Put a log in `componentDidUnmount` of some pages you will navigate to
2. Start navigating

1. Use a function which uses `runWithLoadingIndicator`
2. See pages in the stack (a part from mSupplyAppContainer) unmount

I.e - Add a log in `CustomerInvoices` page and navigate to a `CustomerInvoice`, see `CustomerInvoicesPage` unmount.

### Expected behaviour

Components should not be unmounting.

### Version and device info

- App version: I would say all, but possibly 2.3+
- Tablet model:  All
- OS version:  All

### Additional context

**This is about the unmounting of components in the navigation stack rather than the many re-renders of mSupplyMobileApp**

I believe there are two issues here. The large re-rendering is because of `props` being passed. Simple re-rendering of pages should not cause unmounting of components that are already mounted.

joshxg:
Fixes #1122 

## Change summary

- Whenever `mSupplyMobileAppContainer` re-renders, `ReduxNavigator` re-renders. When `ReduxNavigator` re-renders, a new `ReduxContainer` is created and generates a new navigation stack.
- When the new navigation stack is created, all components currently in that stack are mounted, and when the old navigation stack gets garbage collected, the components in the stack are unmounted.

Example of navigating from `CustomerInvoicePage` back to `CustomerInvoicesPage`

image

1. Render root component (Creating a new navigation stack, re-renders due to prop changes from being subscribed to redux which passes new navigation state in props)
2. Mount a new `MenuPage` (From the new navigation stack)
3. Render the new `MenuPage`  
4. Mount a new `CustomerInvoicesPage` (From the new navigation stack)
5. Render a new `CustomerInvoicesPage`
6. Unmount the old `MenuPage` (Garbage collection)
7. Unmount the old `CustomerInvoicesPage` (Garbage collection)
8. Unmount the `CustomerInvoicePage` (Garbage collection)
9. Render the new `CustomerInvoicesPage` (New top of the stack)

#### With this change

image

1. Render root component
2. Render `MenuPage`
3. Render `CustomerInvoicesPage`
4. Render `CustomerInvoicePage`
5. Render `CustomerInvoicesPage`
6. Unmount `CustomerInvoicePage`
7. Render root component

## Testing

- [ ] Navigating does not throw any errors
- [ ] Navigating does not unmount components currently in the navigation stack

**Note:** this does not fix the rendering problems described in #1123 - that would require more changes, this issue was solely for the expensive unmounting process. 

This does significantly improve navigation performance, but not as much as fixing the rendering issues would.

### DIFF
--- a/src/navigation/ReduxNavigation.js
+++ b/src/navigation/ReduxNavigation.js
@@ -4,9 +4,11 @@ import { connect } from 'react-redux';
 import { createReduxContainer } from 'react-navigation-redux-helpers';
 import AppNavigator from './Navigator';
 
+const ReduxNavigator = createReduxContainer(AppNavigator, 'root');
+
 const ReduxNavigation = props => {
   const { nav, dispatch, screenProps } = props;
-  const ReduxNavigator = createReduxContainer(AppNavigator, 'root');
+
   return <ReduxNavigator state={nav} dispatch={dispatch} screenProps={screenProps} />;
 };
 


src/navigation/ReduxNavigation.js
```diff
@@ -4,9 +4,11 @@ import { connect } from 'react-redux';
 import { createReduxContainer } from 'react-navigation-redux-helpers';
 import AppNavigator from './Navigator';
 
+const ReduxNavigator = createReduxContainer(AppNavigator, 'root');
+
 const ReduxNavigation = props => {
   const { nav, dispatch, screenProps } = props;
-  const ReduxNavigator = createReduxContainer(AppNavigator, 'root');
+
```
wlthomson: Can make this a nice one-liner:

```
const ReduxNavigation = ({ nav, dispatch, screenprops }) => <ReduxNavigator state={nav} dispatch={dispatch} screenProps={screenProps} />;
```

joshxg:
I can't replicate this at all - is this simply navigating between the pages or is there another step? It looks as though the update is occurring on a modal rather than the `CustomerInvoicePage` also. There's an issue somewhere to replace the modal component as it is antiquated and has problems with updating it's own state

wlthomson:
> I can't replicate this at all - is this simply navigating between the pages or is there another step? It looks as though the update is occurring on a modal rather than the `CustomerInvoicePage` also. There's an issue somewhere to replace the modal component as it is antiquated and has problems with updating it's own state

Sorry, I forgot to specify that the error case occurs when navigating back from the new invoice modal, I've edited accordingly. I assumed this was still an issue, as the page is getting unmounted while navigating.

joshxg:
Oh right. At the moment modals are not closed when clicking the back button, rather the normal navigation behaviour occurs (which I hope to change soon). So it does make sense that the component unmounts in that case as it is being popped off the navigation stack. I am not getting the unmounted component noop, but I am quite certain that it is because of the modal component that we use. Which manipulates state directly rather than through `setState`, so often gets into an inconsistent state and doesn't know if it's Arthur or Martha.

Wrapping the component with this conditional would fix this problem: 

```
{isCreatingInvoice && (
          <SelectModal
            isOpen={isCreatingInvoice}
            options={database.objects('Customer')}
            placeholderText={modalStrings.start_typing_to_select_customer}
            queryString="name BEGINSWITH[c] $0"
            sortByString="name"
            onSelect={name => {
              this.onNewInvoice(name);
              this.setState({ isCreatingInvoice: false });
            }}
            onClose={() => this.setState({ isCreatingInvoice: false })}
            title={modalStrings.search_for_the_customer}
          />
        )}
```

as we would ensure the mounting and unmounting are controlled by our state rather than the modals internal state. However this page will be refactored shortly, with all the others #1125

wlthomson:
> Oh right. At the moment modals are not closed when clicking the back button, rather the normal navigation behaviour occurs (which I hope to change soon). So it does make sense that the component unmounts in that case as it is being popped off the navigation stack. I am not getting the unmounted component noop, but I am quite certain that it is because of the modal component that we use. Which manipulates state directly rather than through `setState`, so often gets into an inconsistent state and doesn't know if it's Arthur or Martha.
> 
> Wrapping the component with this conditional would fix this problem:
> 
> ```
> {isCreatingInvoice && (
>           <SelectModal
>             isOpen={isCreatingInvoice}
>             options={database.objects('Customer')}
>             placeholderText={modalStrings.start_typing_to_select_customer}
>             queryString="name BEGINSWITH[c] $0"
>             sortByString="name"
>             onSelect={name => {
>               this.onNewInvoice(name);
>               this.setState({ isCreatingInvoice: false });
>             }}
>             onClose={() => this.setState({ isCreatingInvoice: false })}
>             title={modalStrings.search_for_the_customer}
>           />
>         )}
> ```
> 
> as we would ensure the mounting and unmounting are controlled by our state rather than the modals internal state. However this page will be refactored shortly, with all the others #1125

Got it, makes sense. Updated to pass tests.

PR #1142, opened 2019-08-22
#1141 utility require cycles

### LINKED ISSUE #1141: Require cycles in settings and other utility methods
## Describe the bug

There are many require cycles in settings and other various utility methods i.e. merge records/create records.

I don't know if this is 100% my problem, but it seems so.

Removing: `import { createRecord } from '../database';` from `CustomerInvoicePage` makes the app unusable

image

### To reproduce

1. `CustomerInvoicePage`
2. Remove the above line
3. Restart the app

### Expected behaviour

No require cycles

### Version and device info

- App version: 2.3.5+
- Tablet model: all
- OS version: all

### Additional context

Have spent a couple hours trying to fix this without being able to. The actual error I am getting is `Module AppRegistry is not a registered callable module (calling runApplication)`. Standard fixes seem to suggest clearing caches etc. However I just backtracked through commits and found the one commit that had the above line in it is where the app stops being runnable. 

Removing and adding this line is all it takes to bypass the error.

Also there are heaps of warnings when loading the app which are annoying

joshxg:
Fixes #1141 

## Change summary

Removes remaining require cycles a part from one from `react-native-data-table` https://github.com/sussol/react-native-database/issues/4

Don't appreciate how this has been done but 1) it has to be (keeping random import lines in line just to keep the cycles running is not great) and 2) can make issues for the real issues which are non trivial

#### `createRecord` & `numberSequenceUtilities`
I don't really think there's a problem here. The number sequence utilities do seem nicer in their own file, but obviously that creates a cycle. Some major refactoring could happen but I don't think there's much to gain and to be honest I'd rather see all this common logic in the one file.

#### `incomingSyncUtils` -> `createInternalRecord` -> `incomingSyncUtils`
I think this was a good idea implemented poorly. `createOrUpdateRecord` is the target for refactoring here, I think. 

#### `settings/index` -> `mSupplyMobileApp` -> `settings/index`
Basic change

#### `mergeRecords` -> `incomingSyncUtils` -> `mergeRecords`
The terrible one - I think `createOrUpdateRecord` being the problem makes sense here though. If it wasn't so massive it wouldn't look as disastrous. 

#### `utilities/index` -> `utilities/utilities` -> `utilities/index`
Another simple one, extracted constants to a constants folder

#### `database/index` -> `database/schema` -> `database/datatypes` -> `datatypes/transaction`
Another simple one, just imported directly from the file

## Testing

- [ ] Do something


src/sync/incomingSyncUtils.js
```diff
@@ -744,3 +757,138 @@ export const integrateRecord = (database, settings, syncRecord) => {
...
+ * Merge two existing records. One record is retained and the other is merged. After the objects
+ * are merged, the merged object is deleted from the database. The merge operation updates the
+ * fields of the retained record to reference the same data as the merged object.
+ *
+ * @param  {Realm}   database            Local database.
+ * @param  {Object}  settings            Local settings.
+ * @param  {string}  internalRecordType  Internal record type for merge operation.
+ * @param  {Object}  syncRecord          Data representing the sync record.
+ */
+export const mergeRecords = (database, settings, internalRecordType, syncRecord) => {
```
wlthomson: I think this guy is probably the next on the chopping block 😆!

src/sync/incomingSyncUtils.js
```diff
@@ -684,22 +678,41 @@ export const createOrUpdateRecord = (database, settings, recordType, record) =>
       break;
     }
     case 'Period': {
-      const period = database.update(recordType, createPeriodInternalRecord(record, database));
+      const period = database.update(recordType, {
```
wlthomson: Nice, much easier to follow with these functions unwinded.

src/main.js
```diff
@@ -25,4 +25,5 @@ function App() {
   );
 }
 
+YellowBox.ignoreWarnings(['Setting a timer']);
```
wlthomson: Temporary?

src/database/utilities/constants.js
```diff
@@ -0,0 +1,16 @@
+/* eslint-disable import/prefer-default-export */
+/**
+ * mSupply Mobile
+ * Sustainable Solutions (NZ) Ltd. 2019
+ */
+
+export const MILLISECONDS_PER_DAY = 24 * 60 * 60 * 1000;
```
wlthomson: Going to make a potentially controversial suggestion here:

```
const HOURS_PER_DAY = 24;
const MINUTES_PER_HOUR = 60;
const SECONDS_PER_MINUTE = 60;
const MILLISECONDS_PER_MINUTE = 1000;

const MILLISECONDS_PER_MINUTE = MILLISECONDS_PER_SECOND * SECONDS_PER_MINUTE
const MILLISECONDS_PER_HOUR = MILLISECONDS_PER_MINUTE * MINUTES_PER_HOUR
const MILLISECONDS_PER_DAY = MILLISECONDS_PER_HOUR * HOURS_PER_DAY
```

Chris-Petty: Thanks for starting WW3

src/main.js
joshxg: Nah, I was sneaking it in here hehe. We set a timer for longer than `react-native` likes. Could make a wrapper which would start another timer after the last one had finished. But that's just as bad as this line here. I don't think it'll be fixed soon and saves me having to put `console.disableYellowBox = true` every time I make a new branch 🤣

joshxg: oh I'm also following the `react-native` repo about the issue.. never going to be changed

src/sync/incomingSyncUtils.js
joshxg: maybe, but I can't see a good separation of the logic where it would make it easier than just having it all sequential, a part from the switch at the end. Sometimes a method just gotta be big 🤷‍

src/database/utilities/constants.js
joshxg: 23a081c2ad1a430e53100d870fe5ee12f248c634 thanks good idea

Chris-Petty:
@wlthomson said testing was fine, so here goes!

Issue #1092, opened 2019-08-08
Consistent React hooks code


joshxg:
@Chris-Petty @sworup @Stanzzer @DhanyaHerath @mark-prins  @andreievg @wlthomson 

## Description
With React hooks being usable now in the code base and with them being awesome and how they are quite a big feature that does change how regular components are written quite a bit, I think it would be good to just have a discussion on a consistent way to use them before we end up with a mess

### Context
Hooks are designed so that code can be more easily shared between components. To share the code a consistent style/way might make this easier.

For example: `useReducer` hook is awesome for managing certain components with complex/multiple state manipulations, but comes with a bit of extras (potentially) i.e. action creations, actions and the reducer itself.

Also `CustomHooks`

Even if the code isn't going to be shared, it's nice to be able to extract out a reducer/actions from a component, but that can be annoying for checking out the code if there are many different styles throughout the code base for action creators etc.

## Design discussion

Questions to start:

- File structure
    - does it need to change overall? i.e. ducks/feature first
    - tactics for handling components written using hooks, where do reducers/actions/creators/selectors/custom hooks go

- Hooks
    - Consistent 'style' for writing hooks - i.e. should you always use action creators for actions (so that switching files becomes less needed when checking out code)

I'm not suggesting strict guidelines, but I think a discussion about how to try to make this somewhat consistent would be helpful, especially for me

Chris-Petty:
@joshxg I think we should go over this a bit when doing broader refactoring for new data table in pages

PR #1143, opened 2019-08-22
#1136 New CustomerInvoicePage refactor two

### LINKED ISSUE #1136: New DataTable CustomerInvoicePage refactor part two

the new `CustomerInvoicePage` requires further refactoring.

Looking to refactor:

- The 'base object' in state renamed to a common/consistent name - `object`.. `baseObject`, `pageObject`? suggestions welcome! looking at the last one I think.
- `PageInfo` - similar/the same as how columns are currently done. Injecting `infoColumns` into a pages state through `usePageReducer`
- Modal logic in state: Putting opening/closing modal logic in state. I think having a `basicModalOpen` action/reducer case for `customerInvoicePage`, where nothing is required to be set before opening is the way to go. For more complex modal interactions - i.e. opening the `StocktakeBatchModal`, some property needs to be set in state. I think making individual modal opening actions and cases should be the way to go. As long as there is a consistent `isModalOpen` property and `modalKey` in state, the logic can be easily shared.
- Adding items: haven't looked to much at this, but am looking at using a common action/reducer case for all pages to add an item. If the logic between `Transaction`/`Requisition`/`Stocktake` are to different, having individual actions for these is still a solid way to go, I think. They would at least be shared between a couple of pages?
- Adding master list items: Same as above
- Wrapper around `HeaderRow` accepting `columns` and sorting logic will be a simple way of creating headers for each page.
- Searching fully in state
- deleting logic
- `checkForFinaliseError` - not sure what to do with this hehe but seems smelly

Other possibilities
- `renderButtons` - A wrapper around this could also be possible?
- `renderModalContent` - not sure but would be cool!

Leaving it how it is

## Additional context

@Chris-Petty - seems a bit big for one PR, might be a couple. Anything you think I should just leave along or any additional things you think I should do?

joshxg:
Fixes #1136

## Change summary

- Put modal logic in the reducer
- Put selection logic in the reducer
- Extracted page info columns out, similar to columns/reducer methods. This will inject the `tableState` with a *method* (Require dispatch) to call which will return the page info columns for a page.
- Put `pageObject` in state (transaction for customerInvoice page)
- Adding items logic as reducer methods (including from master list. Requisition add from masterlist might need to change this logic a bit as it needs a store, simple change though.)
- Deleting items logic as reducer methods
- Added a wrapper around `HeaderRow` and `HeaderCell`
- Fixed `topLeftSection` styles

## Testing

*Internal changes, not really testable?*

Note that the reducer can sometimes be called a couple of times for *any* dispatch (from what I have gathered). This is an optimization. Calling the reducer method is supposed to be non-blocking, which means that after a dispatch a render cycle can occur, which may call the reducer again. As the reducer *should* be pure it *shouldn't* matter. So, any reducer method should take care of any side effects it is executing*. I believe reducer being called can be cancelled and also if the diff is the same the result is not used.

*Added the line in `Transaction.addItemsFromMasterList* to for an early exit for this case

This is pretty significant, since we use database side effects in the reducer. I think it will be fine, we could add some middleware to handle the side effects, but could be over kill?

### DIFF
--- a/src/hooks/usePageReducer.js
+++ b/src/hooks/usePageReducer.js
@@ -6,6 +6,7 @@
 import { useReducer, useCallback } from 'react';
 import getReducer from '../pages/dataTableUtilities/reducer/getReducer';
 import getColumns from '../pages/dataTableUtilities/columns';
+import getPageInfo from '../pages/dataTableUtilities/pageInfo';
 import { debounce } from '../utilities/index';
 
 /**
@@ -31,7 +32,8 @@ const usePageReducer = (
   instantDebounceTimeout = 250
 ) => {
   const columns = getColumns(page);
-  const [state, dispatch] = useReducer(getReducer(page), { ...initialState, columns });
+  const pageInfo = getPageInfo(page);
+  const [state, dispatch] = useReducer(getReducer(page), { ...initialState, columns, pageInfo });
 
   const debouncedDispatch = useCallback(debounce(dispatch, debounceTimeout), []);
   const instantDebouncedDispatch = useCallback(
--- a/src/hooks/usePageReducer.js
+++ b/src/hooks/usePageReducer.js
@@ -41,13 +41,18 @@ const usePageReducer = (
     pageInfo,
   });
 
-  const debouncedDispatch = useCallback(debounce(dispatch, debounceTimeout), []);
+  const thunkDispatcher = action => {
+    if (typeof action === 'function') action(dispatch, state);
+    dispatch(action);
+  };
+
+  const debouncedDispatch = useCallback(debounce(thunkDispatcher, debounceTimeout), []);
   const instantDebouncedDispatch = useCallback(
-    debounce(dispatch, instantDebounceTimeout, true),
+    debounce(thunkDispatcher, instantDebounceTimeout, true),
     []
   );
 
-  return [state, dispatch, instantDebouncedDispatch, debouncedDispatch];
+  return [state, thunkDispatcher, instantDebouncedDispatch, debouncedDispatch];
 };
 
 export default usePageReducer;
--- a/src/hooks/usePageReducer.js
+++ b/src/hooks/usePageReducer.js
@@ -3,7 +3,7 @@
  * Sustainable Solutions (NZ) Ltd. 2019
  */
 
-import { useReducer, useCallback } from 'react';
+import { useReducer, useCallback, useMemo } from 'react';
 import getReducer from '../pages/dataTableUtilities/reducer/getReducer';
 import getColumns from '../pages/dataTableUtilities/columns';
 import getPageInfo from '../pages/dataTableUtilities/pageInfo';
@@ -12,8 +12,9 @@ import { debounce } from '../utilities/index';
 /**
  * useReducer wrapper for pages within the app. Creates a
  * composed reducer through getReducer for a particular
- * page as well as fetching the required columns and inserting
- * them into the initial state of the component.
+ * page as well as fetching the required data table columns
+ * and page info columns and inserting them into the initial
+ * state of the component.
  *
  * Returns the current state as well as three dispatchers for
  * actions to the reducer - a regular dispatch and two debounced
@@ -33,7 +34,7 @@ const usePageReducer = (
 ) => {
   const columns = getColumns(page);
   const pageInfo = getPageInfo(page);
-  const memoizedReducer = useCallback(getReducer(page), []);
+  const memoizedReducer = useMemo(() => getReducer(page), []);
   const [state, dispatch] = useReducer(memoizedReducer, {
     ...initialState,
     columns,
--- a/src/hooks/usePageReducer.js
+++ b/src/hooks/usePageReducer.js
@@ -48,7 +48,7 @@ const usePageReducer = (
 
   const thunkDispatcher = action => {
     if (typeof action === 'function') action(dispatch, state);
-    dispatch(action);
+    else dispatch(action);
   };
 
   const debouncedDispatch = useCallback(debounce(thunkDispatcher, debounceTimeout), []);
--- a/src/hooks/usePageReducer.js
+++ b/src/hooks/usePageReducer.js
@@ -16,6 +16,11 @@ import { debounce } from '../utilities/index';
  * and page info columns and inserting them into the initial
  * state of the component.
  *
+ * Dispatch returned from useReducer is wrapped allowing the use
+ * of thunks. Actions can return either a plain object or a function.
+ * If a function is returned, it is called, rather than dispatched,
+ * allowing actions to perform side-effects.
+ *
  * Returns the current state as well as three dispatchers for
  * actions to the reducer - a regular dispatch and two debounced
  * dispatchers - which group sequential calls within the timeout
@@ -32,8 +37,8 @@ const usePageReducer = (
   debounceTimeout = 250,
   instantDebounceTimeout = 250
 ) => {
-  const columns = getColumns(page);
-  const pageInfo = getPageInfo(page);
+  const columns = useMemo(() => getColumns(page), [page]);
+  const pageInfo = useMemo(() => getPageInfo(page), [page]);
   const memoizedReducer = useMemo(() => getReducer(page), []);
   const [state, dispatch] = useReducer(memoizedReducer, {
     ...initialState,
--- a/src/hooks/usePageReducer.js
+++ b/src/hooks/usePageReducer.js
@@ -33,7 +33,12 @@ const usePageReducer = (
 ) => {
   const columns = getColumns(page);
   const pageInfo = getPageInfo(page);
-  const [state, dispatch] = useReducer(getReducer(page), { ...initialState, columns, pageInfo });
+  const memoizedReducer = useCallback(getReducer(page), []);
+  const [state, dispatch] = useReducer(memoizedReducer, {
+    ...initialState,
+    columns,
+    pageInfo,
+  });
 
   const debouncedDispatch = useCallback(debounce(dispatch, debounceTimeout), []);
   const instantDebouncedDispatch = useCallback(
--- a/src/pages/CustomerInvoicePage.js
+++ b/src/pages/CustomerInvoicePage.js
@@ -8,7 +8,6 @@ import PropTypes from 'prop-types';
 import { View } from 'react-native';
 import { SearchBar } from 'react-native-ui-components';
 
-import { createRecord } from '../database';
 import { debounce, MODAL_KEYS, getModalTitle } from '../utilities';
 import { buttonStrings, modalStrings } from '../localization';
 
@@ -48,6 +47,7 @@ import {
   openBasicModal,
   closeBasicModal,
   addMasterListItems,
+  addItem,
 } from './dataTableUtilities/actions';
 
 import globalStyles, { SUSSOL_ORANGE, newDataTableStyles, newPageStyles } from '../globalStyles';
@@ -227,14 +227,7 @@ export const CustomerInvoicePage = ({
             queryString="name BEGINSWITH[c] $0 OR code BEGINSWITH[c] $0"
             queryStringSecondary="name CONTAINS[c] $0"
             sortByString="name"
-            onSelect={item => {
-              database.write(() => {
-                if (!transaction.hasItem(item)) {
-                  createRecord(database, 'TransactionItem', transaction, item);
-                }
-              });
-              dispatch(closeBasicModal());
-            }}
+            onSelect={item => dispatch(addItem(item, 'TransactionItem'))}
             renderLeftText={item => `${item.name}`}
             renderRightText={item => `${item.totalQuantity}`}
           />
--- a/src/pages/CustomerInvoicePage.js
+++ b/src/pages/CustomerInvoicePage.js
@@ -75,8 +75,8 @@ export const CustomerInvoicePage = ({
     routeName,
     {
       pageObject: transaction,
-      backingData: transaction.items,
-      data: transaction.items.sorted('item.name').slice(),
+      backingData: transaction.items.sorted('item.name'),
+      data: transaction.items.slice(),
       database,
       keyExtractor,
       dataState: new Map(),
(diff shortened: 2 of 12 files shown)


src/pages/dataTableUtilities/reducer/reducerMethods.js
```diff
@@ -252,3 +262,124 @@ export const sortData = (state, action) => {
...
+};
+
+/**
+ * Sets the modal open state to false, closing any
+ * modal that is open.
+ *
+ * @param {Object} state  The current state
+ * Action: { type: 'closeBasicModal' }
+ */
+export const closeBasicModal = state => ({ ...state, modalIsOpen: false, modalKey: '' });
```
Chris-Petty: Not sure if it's a bad practice, but could use the empty string for modalKey as equivalent of `false` and remove modalIsOpen all together (would have to use a boolean evaluation on the empty string on the modal `open` prop though).

src/pages/dataTableUtilities/reducer/reducerMethods.js
```diff
@@ -252,3 +262,124 @@ export const sortData = (state, action) => {
...
+    pageObject.removeItemsById(database, itemsById);
+    database.save(pageObjectType, pageObject);
+  });
+
+  const newDataState = new Map();
+  const newData = backingData.slice();
+
+  return {
+    ...state,
+    data: newData,
```
Chris-Petty: While it is a bit janky, I think it's ok so long as 
A: The database.write() is synchronous and runs after the newData slice
B: Isn't too slow (though it's only on deleting so not too problematic if there is some delay)

joshxg: I think it's ok for now. Though some thunkers or something would work well. Possibly have a think about implementing our own simple database middleware, like I think you suggested. Would be an easy migration if we decide to go that route, I think

src/pages/dataTableUtilities/reducer/reducerMethods.js
```diff
@@ -252,3 +262,124 @@ export const sortData = (state, action) => {
...
+
+  const itemsById = Array.from(dataState.keys()).filter(rowKey => dataState.get(rowKey).isSelected);
+
+  database.write(() => {
+    pageObject.removeItemsById(database, itemsById);
+    database.save(pageObjectType, pageObject);
+  });
+
+  const newDataState = new Map();
+  const newData = backingData.slice();
```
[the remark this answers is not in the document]
joshxg: yup, live update

src/pages/dataTableUtilities/reducer/reducerMethods.js
```diff
@@ -252,3 +262,124 @@ export const sortData = (state, action) => {
...
+ * @param {Object} state  The current state
+ * @param {Object} action The action to act upon
+ */
+export const deleteItemsById = (state, action) => {
+  const { database, pageObject, dataState, hasSelection, backingData } = state;
+  const { pageObjectType } = action;
+
+  if (!hasSelection) return state;
+
+  const itemsById = Array.from(dataState.keys()).filter(rowKey => dataState.get(rowKey).isSelected);
```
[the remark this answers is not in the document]
joshxg: 6e52630ddf7469b54ecbe432dede4f857ed59512

src/pages/dataTableUtilities/reducer/reducerMethods.js
```diff
@@ -252,3 +262,124 @@ export const sortData = (state, action) => {
...
+ * @param {Object} action The action to act upon
+ * Action: { type: 'editPageObject', value, field, pageObjectType }
+ */
+export const editPageObject = (state, action) => {
+  const { database, pageObject } = state;
+  const { value, pageObjectType, field } = action;
+
+  if (pageObject[field] !== value) {
+    database.write(() => {
+      pageObject[field] = value;
```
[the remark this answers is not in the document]
joshxg: Just checked it out. It is possible! Do you think more explicit is better? I'm happy to do that. I figured it would come at some point when the logic between different fields starts to differ

src/pages/dataTableUtilities/reducer/reducerMethods.js
joshxg: well I don't think so, `''` is falsey and it's javascript!

a530db1774b92255e7f02aca0051ada655fbc9fb

joshxg: forgot one a87c3f9d7da9cfea9a6590536c926f58a39f520c

src/pages/dataTableUtilities/reducer/reducerMethods.js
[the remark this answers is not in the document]
Chris-Petty: I think explicit is better, deleting this method would be mean less invitation to skip being explicit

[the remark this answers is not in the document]
joshxg: b6575d58661aed26dfe5faca59243fbf1d02adf6
1ad3a1b8d3c3a9e0d03a4cb515cccb1a2ffe342a
3caf7f99901b574ee0e6aec3c89de5ac5169258d
0d04575dbdfd38473fb938025763b700251020e2

added `editTheirRef` and `editComment` reducer methods and actions and removed export of `editPageObject`. I think a `editPageObjectSetter` might be a better idea, rather than check if the `pageObject[field]` is a function, and to call rather than assign, but maybe not. Either way I didn't do the change for `editTotalQuantity`, I think it's fine to figure out a bit later, maybe after a couple of pages to make sure we have a broader scope that: two methods for editting fields/using setters or one and how to run the file system for all these reducers methods/actions (and to consider for not just this, but other hooks that might be used throughout the app, aaaand possibly selectors 🤣 ) in #1092 ?

src/pages/CustomerInvoicePage.js
```diff
@@ -74,120 +71,47 @@ export const CustomerInvoicePage = ({
...
-        return modalStrings.search_for_an_item_to_add;
-      case COMMENT_EDIT:
-        return modalStrings.edit_the_invoice_comment;
-      case THEIR_REF_EDIT:
-        return modalStrings.edit_their_reference;
+  const [tableState, dispatch, instantDebouncedDispatch, debouncedDispatch] = usePageReducer(
+    routeName,
+    {
+      pageObject: transaction,
+      backingData: transaction.items,
```
Chris-Petty: This also needs to be sorted, as the when deleting items this is used again. Can either be lazy and add sorted here, or you can pass a method for initial state so you can do the sort once then just slice the result for `data`.

src/pages/dataTableUtilities/reducer/reducerMethods.js
```diff
@@ -252,3 +262,160 @@ export const sortData = (state, action) => {
...
+
+  const itemsIds = Array.from(dataState.keys()).filter(rowKey => dataState.get(rowKey).isSelected);
+
+  database.write(() => {
+    pageObject.removeItemsById(database, itemsIds);
+    database.save(pageObjectType, pageObject);
+  });
+
+  const newDataState = new Map();
+  const newData = backingData.slice();
```
Chris-Petty: Testing I found a problem with this:
image
Basically need to reapply the sorting? Or splice the the same rows out of `data` instead of slicing again but that's pretty sketchy (e.g. the delete failed on DB, but rows disappear in UI)

src/pages/CustomerInvoicePage.js
joshxg: hmm, after using `sorted` so much recently for vaccines stuff and knowing it doesn't sort in-place, I some how switched to think it was in place..! - just kept `backingData` as unsorted as it doesn't really need to be, but sorted the actual display data

src/pages/dataTableUtilities/reducer/reducerMethods.js
joshxg: 14a209667c787a312149128358aca51bb6213f12

improved the deletion speed a loooot!

joshxg:
Fixed the finalization behaviour with these commits here: 

c6386789961b07b851fdf72dca622c4f7e5be665
ac2089061b525a8cb710ed2862005e00757221d1
17c110ad96a0892fe4ef7b0b20dccd3869affe6a
55da0752305c42848ae4bc67f964eda4061c778a

Also added a change to `usePageReducer` in order to fix the non-pure reducer problem:

623294cf89ed8aae5a24f795fc4efc0ef11d92eb
891155535fdb7e836bd9ce4ee6130cf8b7455d34
7bb388d8fd2e271e7880fa6d52dbcc3dacb55b82
854c917bb27afc010eccd6cf4a0f0489ac6aaa6d
ebb52285fcbf62517d4f7f47bc40625aba7dd395
b75009b378558b844a007c5748fde1be63732017
2dde2be1a03493549a574c4705848e5692affff3

src/hooks/usePageReducer.js
```diff
@@ -32,22 +37,27 @@ const usePageReducer = (
...
+
+  const debouncedDispatch = useCallback(debounce(thunkDispatcher, debounceTimeout), []);
   const instantDebouncedDispatch = useCallback(
-    debounce(dispatch, instantDebounceTimeout, true),
+    debounce(thunkDispatcher, instantDebounceTimeout, true),
     []
   );
 
-  return [state, dispatch, instantDebouncedDispatch, debouncedDispatch];
+  return [state, thunkDispatcher, instantDebouncedDispatch, debouncedDispatch];
```
Chris-Petty: I want to keep with the trend and call this `thunker` but don't haha

src/pages/dataTableUtilities/actions.js
```diff
@@ -60,31 +61,75 @@ export const closeBasicModal = () => ({
...
-  type: 'editComment',
-  pageObjectType,
-  value,
-});
-
-export const deleteItemsById = (pageObjectType, object) => ({
-  type: 'deleteItemsById',
-  pageObjectType,
-  object,
+export const addMasterListItems = objectType => (dispatch, state) => {
```
Chris-Petty: At this point is it worth while js docing the actions too?

src/hooks/usePageReducer.js
```diff
@@ -32,22 +37,27 @@ const usePageReducer = (
...
   const [state, dispatch] = useReducer(memoizedReducer, {
     ...initialState,
     columns,
     pageInfo,
   });
 
-  const debouncedDispatch = useCallback(debounce(dispatch, debounceTimeout), []);
+  const thunkDispatcher = action => {
+    if (typeof action === 'function') action(dispatch, state);
+    dispatch(action);
```
Chris-Petty: I'd have thought this line would have to be in the `else` clause of the above `if`? Dispatch is receiving a function as an arg otherwise.

joshxg: oopsies d9b0921ca870399557e490400cb446f387f8a742

src/pages/dataTableUtilities/actions.js
joshxg: Yeah definitely - I think the file structure starts to matter more also (`thunk.js`/`action.js`? or `module1Actions.js` etc? Dunno!) .. I think all of these actions/reducers need to be tidied when we have a solid understanding of the groupings etc?

src/pages/dataTableUtilities/actions.js
```diff
@@ -60,31 +61,75 @@ export const closeBasicModal = () => ({
...
+  let addedItem;
+
+  database.write(() => {
+    if (pageObject.hasItem(item)) return;
+    addedItem = createRecord(database, addedItemType, pageObject, item);
+  });
+
+  if (addedItem) dispatch({ type: 'addItem', item: addedItem });
+  else dispatch(closeBasicModal());
+};
```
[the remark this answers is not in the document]
joshxg: Ideally, I think you would mock the database calls in unit tests, and then everything else is essentially pure. One of the advantages of using sagas over thunks. Thunks are quite a lot easier/quicker to implement though!

I *think* you should be able to hook `useReducer` up with `react-native-debugger` and `redux-dev-tools`, but maybe not while we are not using `linkingObjects` with realm

src/pages/CustomerInvoicePage.js
```diff
@@ -101,15 +102,18 @@ export const CustomerInvoicePage = ({
...
     hasSelection,
+    backingData,
   } = tableState;
 
+  // Transaction is impure - finalization logic prunes items, deleting them from the transaction.
+  // Since this does not manipulate the state through the reducer, data object does not get
+  // updated.
+  if (transaction.isFinalised && data.length !== backingData.length) dispatch(refreshData());
+
+  const pageInfoColumns = useMemo(() => pageInfo(pageObject, dispatch), []);
```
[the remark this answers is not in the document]
joshxg: 🤷‍♀

src/pages/dataTableUtilities/actions.js
Chris-Petty: Yea, let it mature a bit first 👍

src/pages/CustomerInvoicePage.js
```diff
@@ -74,120 +72,50 @@ export const CustomerInvoicePage = ({
...
-  const getModalTitle = () => {
-    switch (modalKey) {
-      default:
-      case ITEM_SELECT:
-        return modalStrings.search_for_an_item_to_add;
-      case COMMENT_EDIT:
-        return modalStrings.edit_the_invoice_comment;
-      case THEIR_REF_EDIT:
-        return modalStrings.edit_their_reference;
+  const [tableState, dispatch, instantDebouncedDispatch, debouncedDispatch] = usePageReducer(
```
Chris-Petty: `tableState` should really be `state` now that it has everything and the kitchen sync. Or `pageState`, but state is fine as it usually implies it holds all state for this parent.

joshxg: 0e371798027b366c7f70ae9b1271b341d7fb418f

src/pages/CustomerInvoicePage.js
```diff
@@ -105,16 +105,16 @@ export const CustomerInvoicePage = ({
     backingData,
   } = tableState;
 
+  const { isFinalised } = transaction;
```
Chris-Petty: Get halfway?

joshxg: 0475969e5f99e9a8c161a742cd4206e8cc030e1b kind of related to this comment

PR #1146, opened 2019-08-26
#1145 onPress prop to Row component

### LINKED ISSUE #1145: New DataTable clickable rows

Rows need to be clickable in order to navigate

`Row` component should have an `onPress` prop. 

I think from a UX perspective it is potentially a source of confusing when rows look more or less exactly the same, but some are clickable and some aren't (current implementation). Some pages now use a column with an icon for clicking (stock take batches).

I think it's best to be consistent with this and add an icon to every row that is clickable. Potentially have the whole row be clickable, but also have an icon to indicate that it is clickable.

Just use the current implementation

## Additional context

N/A

joshxg:
#### Note: Branched from #1136 

Fixes #1145

## Change summary

- Adds a simple `onPress` props to `Row`

As in the docstring, the behaviour is odd when you have a cell which is `Touchable` and the `Row`, as both events will be fired. I cannot find a way to cancel the propogation of a native event being fired on a child component. I think the solution is to wrap each non-touchable cell in a `Touchable` which would trigger the `Row` `onPress`. This would require a bit more refactoring and the use-case isn't needed right now.

Note that this is a bit weird because `EditableCell` uses an `onPress`

## Testing
- [ ] Clicking a `Row` component with an `onPress` prop triggers the `onPress`.
- [ ] Clicking a `Row` without an `onPress` prop does nothing

I wouldn't be opposed to having a `mSupplyRow` component which would do a lot of the rendering of cells as well as handling logic related to having clickable rows and adding an icon indicating that a row is clickable!

### DIFF
--- a/src/widgets/DataTable/Row.js
+++ b/src/widgets/DataTable/Row.js
@@ -4,6 +4,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 import { View } from 'react-native';
+import { TouchableOpacity } from 'react-native-gesture-handler';
 
 /**
  * Renders a row of children as outputted by renderCells render prop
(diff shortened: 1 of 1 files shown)


src/widgets/DataTable/Row.js
```diff
@@ -4,21 +4,30 @@ import React from 'react';
...
  *
  * @param {object} rowData Data to pass to renderCells callback
  * @param {string|number} rowKey Unique key associated to row
  * @param {object} rowState State to pass to renderCells callBack
+ * @param {func} onPress function to call on pressing the row. See below.
+ * @param {object} viewStyle Style object for the wrapping View component
  * @param {func} renderCells renderProp callBack for rendering cells based on rowData and rowState
  *                          `(rowKey, columnKey) => {...}`
- * @param {object} viewStyle Style object for the wrapping View component
+ * Tap gesture events will be captured in this component for any taps on cells within
```
Chris-Petty: I think a comment like this is usually put in the section before @param list

joshxg: a327c609106a22dc650f14ef549d47c96f01fe8f

PR #1149, opened 2019-08-27
#1144 database settings singletons

### LINKED ISSUE #1144: UIDatabase/Settings should be singletons
@Chris-Petty @mark-prins @andreievg @wlthomson 

## Description

In the app we have many interfaces to persistant storage, with two 'main' ones, `UIDatabase` and `MobileAppSettings`. Both of these are classes which are instantiated in `mSupplyMobileAppContainer` and passed around in the app through props through the `ReduxNavigator` component. Specifically using `screenProps`, which is an old/deprecated method that needs to be removed (see #1123), alternatively replaced with using redux to pass the database access object. 

Since these two classes are only ever instantiated once and are passed around to just about every part of the app - to every page/container, a lot of display components and almost every `Realm Class` has at least one method which requires the database to be passed, I think these should be singletons that are required in files that need them.

i.e. in `UIDatabase.js`, export like: `export default new UIDatabase(database)`

I think there's an argument that coding to a concrete class without any dependency injection can cause problems when the implementation changes. I think the likelihood of the implementation changing, while maintaining the same API is extremely unlikely (i.e. if the database changes, the API will not stay the same, anyway, unless we had *another* database access layer on top).

joshxg:
Fixes #1144 

~~## NOTE: Branched from #1142~~

## Change summary

- Created a file `BaseDatabase.js` which exports an instance of `react-native-database`. Should be the single instance of this in the entire app.
- Created a method `getUIDatabaseInstance` in `UIDatabase.js` and exported that - removed the export of `UIDatabase`. Because of require cycles when importing `UIDatabase` into a realm data type (i.e. `Transaction` -> `schema` -> `BaseDatabase` -> `UIDatabase` -> `Transaction`), the database being wrapped needed to be injected. 
- Exported a single instance of `UIDatabase` from `database/index.js`, and trying to import `UIDatabase` will pass the same instance when the method is called (basically using a *slightly more* standard singleton pattern here).
- Added an import of `UIDatabase` to `Settings` and exported a single instance.
- Added imports of `UIDatabase`, `BaseDatabase` (for the synchroniser) and `Settings` to `mSupplyMobileAppContainer`
- Left the passing of database and settings in props, for now. Data table refactor will take care of a lot of that, I think, and the rest can be done in new issues related to fixing navigation/redux. New code can definitely import as needed.

**Note:** ECMAScript Imports are asynchronously loaded - the order cannot be guaranteed *see: [exploring JS](https://exploringjs.com/es6/ch_modules.html#sec_modules-in-browsers) or the [ECMA script specs](http://www.ecma-international.org/ecma-262/6.0/#sec-modules)*
However, scripts being imported *are* guaranteed to be *executed*, prior to importing. 

## Testing

- [ ] Terrible test but: Should run as it did prior to this change..!

Singletons are definitely a pattern that should be used [sparingly](https://www.google.com/search?client=firefox-b-d&q=why+singlestons+are+bad+js) but I think this is a good use case for getting into a better position to fix some more immediate problems. Possibly using a similar pattern for `Synchroniser` and the `Authenticators` would/could be a good idea also

### DIFF
--- a/src/database/BaseDatabase.js
+++ b/src/database/BaseDatabase.js
@@ -1,0 +1,7 @@
+import { Database } from 'react-native-database';
+
+import { schema } from './schema';
+
+const BaseDatabaseInstance = new Database(schema);
+
+export default BaseDatabaseInstance;
--- a/src/database/BaseDatabase.js
+++ b/src/database/BaseDatabase.js
@@ -1,3 +1,16 @@
+/**
+ * mSupply Mobile
+ * Sustainable Solutions (NZ) Ltd. 2019
+ */
+
+/**
+ * Interface to a Realm, to be wrapped by
+ * SyncDatabase or UIDatabase.
+ *
+ * Only one instance will be active at any point
+ * in the applications life-cycle.
+ */
+
 import { Database } from 'react-native-database';
 
 import { schema } from './schema';
--- a/src/database/UIDatabase.js
+++ b/src/database/UIDatabase.js
@@ -1,3 +1,4 @@
+/* eslint-disable import/no-mutable-exports */
 /**
  * mSupply Mobile
  * Sustainable Solutions (NZ) Ltd. 2019
@@ -28,7 +29,7 @@ const translateToCoreDatabaseType = type => {
   }
 };
 
-export class UIDatabase {
+class UIDatabase {
   constructor(database) {
     this.database = database;
     this.EXPORT_DIRECTORY = '/Download/mSupplyMobile_data';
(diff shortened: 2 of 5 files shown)


joshxg:
I checked with adding bundled realm that @andreievg had added, worked fine (maybe should make an issue to add that back)

Chris-Petty:
Couldn't test on tablet with existing database (I was on a data table branch, then moved to this branch), got following on app load:
Screenshot_20190828-123037_mSupply Mobile

joshxg:
Can't replicate it - I'll try build a couple of apk's and upgrade. The variable `database` would be either the props passed to a couple of components sync/finalize modals, the variable passed in `screenProps` or something to do with the removed code which instantiated the database instances previously (`database` and `this.database`)

joshxg:
Built two release APK's - one on develop (2.3.5) and one from this branch (2.3.6). Installed the 2.3.5 APK - initialized. Installed 2.3.6 APK - upgraded. Couldn't reproduce the error. I am gonna say it's most likely to do with the development environment bundler/caching. Mainly due to the variable that isn't used anymore (and not being able to reproduce)

PR #1153, opened 2019-08-28
#1148 final customer invoice page

### LINKED ISSUE #1148: Final new CustomerInvoicePage changes

The final `CustomerInvoicePage` changes to complete the first page with the new `DataTable`

- No editing when finalised
- Remove `checkForFinaliseError` (?)
- Doc strings/comments to clarify anything
- "nice it up" - destructuring where needed, rename any variables  etc

@Chris-Petty anything else you can think of?

Leaving it as is

## Additional context

Hopefully the last issue to finish off `CustomerInvoicePage`!

joshxg:
Fixes #1148 

## Change summary

- Added disabled when finalized behavior
- Added doc strings/comments to `CustomerInvoicePage`
- Moved `checkForFinaliseError` to it's own file**
- Added `KeyboardAwareScrollView`
- `keyboardType` prop to `EditableCell` and a couple other props

## Testing

- [ ] Can finalize a customer invoice
- [ ] Can edit rows which are under the keyboard
- [ ] Customer invoices are disabled when finalized

Initially I gave suggestions for the finalization code, but I resisted refactoring as I think overall it needs to change slightly in order for the 're-render' problem to be fixed. I've moved it now and have an idea of how to change the finalization code (happy to talk about it), might make an issue when being able to refactor a bunch of this other code looks like it will happen. Would understand if it should just go back, but I think just moving it is a small improvement


joshxg:
Added placeholder logic for editable cell.. not happy about it though!

Also not sure if we want to handle these cases
image


[ cells exceeding size by stupid, unrealistic amounts - ellipsismode maybe]

joshxg:
actually it's a bit nastier than that 

image

Chris-Petty:
@joshxg Yea, old pages/tables in mobile ellipsis. I think we should we do it in Cell/EditableCell rather than state (be much jank in the latter). Alternatively could be done in the renderCells method.

src/utilities/finalisation.js
```diff
@@ -0,0 +1,25 @@
...
+export function checkForCustomerInvoiceError(customerInvoice) {
+  if (customerInvoice.items.length === 0) {
+    return modalStrings.add_at_least_one_item_before_finalising;
+  }
+
+  if (customerInvoice.totalQuantity === 0) {
+    return modalStrings.record_stock_to_issue_before_finalising;
+  }
+  return null;
+}
```
Chris-Petty: Thought you were going to make this a class method on transaction? Though in terms of what is callable where I can understand keeping it out (or you'd have to check this isn't being called on transaction with type supplier invoice)

src/widgets/DataTable/DataTable.js
```diff
@@ -19,7 +20,14 @@ import { StyleSheet, VirtualizedList, VirtualizedListPropTypes } from 'react-nat
...
     {renderHeader()}
-    <VirtualizedList style={style} renderItem={renderRow} {...otherProps} />
+    <KeyboardAwareScrollView style={style} enabled behaviour="padding" keyboardVerticalOffset={500}>
+      <VirtualizedList
+        keyboardShouldPersistTaps="always"
+        style={style}
+        renderItem={renderRow}
+        {...otherProps}
+      />
+    </KeyboardAwareScrollView>
```
Chris-Petty: What was this fixing?

src/utilities/finalisation.js
joshxg: yeah - I want to really badly :( I didn't want to put the method 'as is', at that would mean returning strings, so that would mean changing `FinaliseModal`.  As above I thought it would be more suitable to properly refactor/rewrite this when we need to fix the overall render issues (which when you look deeper into it will include changing up `NavigationBar`, `FinalizeModal` and `SyncModal`). Happy to put it back where it was if you think that's better!

src/widgets/DataTable/DataTable.js
joshxg: without it, when I went to edit cells that would be underneath the keyboard, the cell would lose focus and the keyboard would dismiss automatically and not let me edit

src/hooks/usePageReducer.js
```diff
@@ -40,24 +41,34 @@ const usePageReducer = (
...
+  // Reference to the current state object, indepdent of closures.
+  const stateRef = useRef(pageState);
+  const getState = () => stateRef.current;
+
+  // Basic dispatch function.
+  const dispatch = action => {
+    const newState = memoizedReducer(getState(), action);
+    setPageState(newState);
+    stateRef.current = newState;
   };
```
Chris-Petty: If we do figure a better way to do this, then it's easy enough to refactor this anyway.

src/hooks/usePageReducer.js
```diff
@@ -40,24 +41,34 @@ const usePageReducer = (
...
+  const [pageState, setPageState] = useState({
     ...initialState,
     columns,
     pageInfo,
   });
 
-  const thunkDispatcher = action => {
-    if (typeof action === 'function') action(dispatch, state);
-    else dispatch(action);
+  // Reference to the current state object, indepdent of closures.
```
Chris-Petty:
```suggestion
  // Reference to the current state object, independent of closures.
```

src/utilities/finalisation.js
Chris-Petty: Nah I'm still happy with this!

src/pages/CustomerInvoicePage.js
```diff
@@ -133,6 +134,7 @@ export const CustomerInvoicePage = ({
     } = newDataTableStyles;
     return columns.map(({ key: colKey, type, width, alignText }, index) => {
       const isLastCell = index === columns.length - 1;
+      const isDisabled = isFinalised || (rowState && rowState.disabled);
```
Chris-Petty: `rowState.isDisabled`?
Could use `rowData.transaction.isFinalised?

src/globalStyles/newDataTableStyles.js
```diff
@@ -52,11 +52,18 @@ const newDataTableStyles = {
       color: DARK_GREY,
     },
   },
+  editableCellUnfocused: {
+    textAlign: 'right',
```
Chris-Petty: Doesn't this need to be handled by above styles for cellText alignment?

src/pages/CustomerInvoicePage.js
joshxg: hmm, is `rowState.isDisabled` ever even set anywhere? What would be the use case for having a particular row disabled? Would also be very difficult to set each `rowState` field to `disabled` when finalising. Maybe should just remove `rowState` from this condition? Can use `rowData.transaction.isFinalised`, dunno if there's much to gain, pretty much have to use `pageObject.isFinalised` in the dependencies anyway?

src/globalStyles/newDataTableStyles.js
joshxg: ah yes of course it does :(

src/pages/CustomerInvoicePage.js
Chris-Petty: Never mind, I assume this is actually separated for handling stuff the the customer invoice list where only some rows are disabled ?

joshxg: ahh that's the use case! I guess I'll keep it in then 🤣

Chris-Petty: Well, maybe that'd make sense if we're going to factor out renderCells

PR #1164, opened 2019-09-02
#1125 customer invoices page new data table

### LINKED ISSUE #1125: Refactor CustomerInvoicesPage to use new DataTable

Need to use the new `DataTable` in CustomerInvoicesPage.

- A more or less complete refactor, using the first refactored page `CustomerInvoicePage` as a guide.
- This will require using hooks and the new DataTable `component`.

N/A

## Additional context

This is a placeholder issue. Please investigate and compare the page with `CustomerInvoicePage`. Each page being refactored may have their own quirks. The tissue is encouraged to describe any additional problems and differences with the particular page here, to find a solution which encourages further extensibility and code-reuse with other pages, if possible.

Chris-Petty:
Fixes #1125 

## Change summary

Refactor of customer invoices to use new data table and page reducer

**Problems that @joshxg might have done in other branches that I don't want to rework:**
- "STATUS" column needs localisation
- "ENTRY DATE" column omitted cause new code useful to it in another PR 

## Testing

- [ ] Every page feature works

### DIFF
--- a/src/pages/CustomerInvoicesPage.js
+++ b/src/pages/CustomerInvoicesPage.js
@@ -29,7 +29,7 @@ import {
   closeBasicModal,
   deleteTransactionsById,
 } from './dataTableUtilities/actions';
-import { newSortDataBy } from '../utilities';
+import { MODAL_KEYS, newSortDataBy } from '../utilities';
 import usePageReducer from '../hooks/usePageReducer';
 import DataTablePageView from './containers/DataTablePageView';
 
@@ -63,7 +63,7 @@ export const CustomerInvoicesPage = ({ currentUser, navigateTo, routeName }) => 
   const renderNewInvoiceButton = () => (
     <PageButton
       text={buttonStrings.new_invoice}
-      onPress={() => dispatch(openBasicModal('Anything will do'))}
+      onPress={() => dispatch(openBasicModal(MODAL_KEYS.SELECT_CUSTOMER))}
     />
   );
 
--- a/src/pages/CustomerInvoicesPage.js
+++ b/src/pages/CustomerInvoicesPage.js
@@ -3,82 +3,70 @@
  * Sustainable Solutions (NZ) Ltd. 2019
  */
 
-import React from 'react';
+import React, { useCallback } from 'react';
 import PropTypes from 'prop-types';
+import { View } from 'react-native';
+import { SearchBar } from 'react-native-ui-components';
 
-import { GenericPage } from './GenericPage';
-import { PageButton } from '../widgets';
+import { UIDatabase, createRecord } from '../database';
+import { buttonStrings, modalStrings, navStrings } from '../localization';
 import { BottomConfirmModal, SelectModal } from '../widgets/modals';
+import {
+  PageButton,
+  CheckedComponent,
+  UncheckedComponent,
+  DisabledCheckedComponent,
+  DisabledUncheckedComponent,
+} from '../widgets';
+import { DataTable, Row, Cell, CheckableCell, DataTableHeaderRow } from '../widgets/DataTable';
+import {
+  selectRow,
+  deselectRow,
+  deselectAll,
+  sortData,
+  filterData,
+  openBasicModal,
+  closeBasicModal,
+  deleteTransactionsById,
+} from './dataTableUtilities/actions';
+import usePageReducer from '../hooks/usePageReducer';
+import DataTablePageView from './containers/DataTablePageView';
 
-import { createRecord } from '../database';
-import { formatStatus, sortDataBy } from '../utilities';
-import { buttonStrings, modalStrings, navStrings, tableStrings } from '../localization';
+import { SUSSOL_ORANGE, newDataTableStyles, newPageStyles } from '../globalStyles';
 
-const DATA_TYPES_SYNCHRONISED = ['Transaction'];
+const keyExtractor = item => item.id;
 
-/**
- * Renders the page for displaying customer invoices.
- *
- * @prop   {Realm}          database      App database.
- * @prop   {func}           navigateTo    Callback for navigation stack.
- * @state  {Realm.Results}  transactions  Transactions filtered by |customer_invoice|.
- */
-export class CustomerInvoicesPage extends React.Component {
-  constructor(props) {
-    super(props);
-    this.state = {
-      transactions: props.database.objects('CustomerInvoice'),
-      selection: [],
-      isCreatingInvoice: false,
-    };
-    this.dataFilters = {
-      searchTerm: '',
-      sortBy: 'serialNumber',
-      isAscending: false,
-    };
-  }
-
-  onNewInvoice = otherParty => {
-    const { database, currentUser } = this.props;
-    let invoice;
-    database.write(() => {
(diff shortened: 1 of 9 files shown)


src/pages/CustomerInvoicesPage.js
```diff
@@ -3,252 +3,231 @@
...
+    initialState()
+  );
+  const { data, dataState, sortBy, isAscending, columns, modalKey, hasSelection } = state;
 
-  navigateToInvoice = invoice => {
-    const { database, navigateTo } = this.props;
+  const renderNewInvoiceButton = () => (
+    <PageButton
+      text={buttonStrings.new_invoice}
+      onPress={() => dispatch(openBasicModal('Anything will do'))}
```
joshxg: accidental param?

src/pages/CustomerInvoicesPage.js
```diff
@@ -3,252 +3,231 @@
...
+    />
+  );
 
+  const navigateToInvoice = invoice => {
```
joshxg: Just checking in on using redux navigation - are we to do that after?

src/pages/CustomerInvoicesPage.js
```diff
@@ -3,252 +3,231 @@
...
-  }
-}
+        extraData={dataState}
+        renderRow={renderRow}
+        renderHeader={renderHeader}
+        keyExtractor={keyExtractor}
+        getItemLayout={getItemLayout}
+      />
+      <BottomConfirmModal
+        isOpen={hasSelection} // TODO
```
joshxg: `//TODO` what

src/pages/dataTableUtilities/columns.js
```diff
@@ -7,13 +7,42 @@ import { tableStrings } from '../../localization';
...
+    title: tableStrings.customer,
+    sortable: true,
+  },
+  status: {
+    key: 'status',
+    title: tableStrings.status,
+    sortable: true,
+  },
+  entryDate: {
+    // TODO: unbork it
```
joshxg: tbh I hate `//TODO`'s a lot - they never get do'd and just sit there forever if there's no issue made (and even with an issue hehe)

src/utilities/sortDataBy.js
```diff
@@ -41,15 +41,25 @@ export const sortDataBy = (data, sortBy, sortDataType, isAscending = true) => {
...
+    sortDataType ||
+    {
+      otherPartyName: 'string',
+      itemCode: 'string',
+      itemName: 'string',
+      status: 'string',
+      serialNumber: 'number',
+      availableQuantity: 'number',
+      totalQuantity: 'number',
+    }[sortBy];
```
joshxg: this is going to end up massive (one for every column we use) - I think just defining a `const sortDataTypes` or something outside of the method might be better?

joshxg:
Actually forgot to mention a workflow: `CustomerInvoicesPage` -> `CustomerInvoicePage` with a non-finalised invoice (might have to make before) -> finalise -> `CustomerInvoicesPage` - list doesn't update - this happens in a couple other places I'm yet to find an elegant solution (if we can't think of something nice right now I think happy to merge and think about it)

src/pages/CustomerInvoicesPage.js
Chris-Petty: Lol that was a joke, but technically the statement in that string is true.

Should use constant you reckon?

src/pages/CustomerInvoicesPage.js
Chris-Petty: Yea, I'd rather you show me the light on that one after.

src/pages/CustomerInvoicesPage.js
Chris-Petty: Whoops, thanks (it's already done)

src/pages/dataTableUtilities/columns.js
Chris-Petty: That's fair. This one is a bit different cause I know that your branches potentially have the code to do this, so me doing it is a bit derp. Could make an issue "CIs page needs entryDate column" if you like?

src/utilities/sortDataBy.js
Chris-Petty: Sure, although that doesn't fix it being huge. I was wondering if I should put it in the column definitions?

Chris-Petty:
> Actually forgot to mention a workflow: `CustomerInvoicesPage` -> `CustomerInvoicePage` with a non-finalised invoice (might have to make before) -> finalise -> `CustomerInvoicesPage` - list doesn't update - this happens in a couple other places I'm yet to find an elegant solution (if we can't think of something nice right now I think happy to merge and think about it)

This used to be addressed by the `topRoute` prop. Basically, it'd refresh the results object the page had if it became the top route again.

joshxg:
ah cool - sounds like a sufficient solution! - could check `topRoute === routeName` and `refreshData` 🤣

src/pages/CustomerInvoicesPage.js
joshxg: I guess it's fine - there is a PR soon for factoring modal logic. If you like that it might change how this is done, but since this page only uses one modal, maybe not. Maybe just confusing for the next person who looks at this code, and a little annoying for the person who has to add an extra modal to this page

src/pages/CustomerInvoicesPage.js
joshxg: cool~

src/pages/dataTableUtilities/columns.js
joshxg: I think there will be a couple things - navigation & status column(?) maybe just a new issue to revisit the page?

src/utilities/sortDataBy.js
joshxg: That means the columns lookup table has to use the `column key` as the lookup table `key`, which means you can only ever have one column definition for a certain key. I don't think we use multiple column definitions at the moment, but it's pretty restrictive. I don't think it'd be tooooo big as an object in the module scope?

Chris-Petty:
That or an customEffect listening to DB changes 😉

src/pages/CustomerInvoicesPage.js
Chris-Petty: https://github.com/openmsupply/mobile/pull/1164/commits/07bf3098cfa0aabd16e0fa37ef9005ad4cc004d2

src/pages/CustomerInvoicesPage.js
Chris-Petty: https://github.com/openmsupply/mobile/pull/1164/commits/6b2db3009da30e1e1183e8f8e47a9cdfe436637a

src/utilities/sortDataBy.js
Chris-Petty: Fair, I agree. https://github.com/openmsupply/mobile/pull/1164/commits/24f2a1ed1038dce3312805a32ec7966bc0e778c5

src/pages/dataTableUtilities/columns.js
Chris-Petty: https://github.com/openmsupply/mobile/issues/1167

joshxg:
awesome cheers

Issue #1159, opened 2019-08-31
Factor out renderCells from page components


joshxg:
## Is your feature request related to a problem? Please describe.

`renderCells` method is more or less exactly the same in each `Page`. This could be factored out to ensure there are less duplication and therefore less prone to mistakes, errors and work if changes need to be made.

## Describe the solution you'd like

Each Cell has very similar props which need to be passed, so keeping all the cells in a largish method seems to be the most efficient idea. Also the use of `JSX` for components is definitely the best idea, in my opinion. So rather than using some lookup table or something of the like, I think having a simple `Row` wrapper, which contains the `renderCells` method used in almost every page, but potentially slightly larger, given it will contain each type of cell.

## Describe alternatives you've considered

Using the same method as `getColumns`, `getPageInfo` etc.

I think this is not as an effective solution to this problem, give `getColumns` and `getPageInfo` are providing configurations for views, rather than an actual view itself. Potentially there could be a `getCells` method which returns objects of props for `Cells`, but I think these problems are different to warrant different solutions.

## Additional context

I think the same should be done for `Modal`s

joshxg:
Closed by #1163

PR #1168, opened 2019-09-03
#1160 Factor out modals

### LINKED ISSUE #1160: Factor out renderModalContent from page components

`renderModalComponent` method is more or less exactly the same in each `Page`. This could be factored out to ensure there are less duplication and therefore less prone to mistakes, errors and work if changes need to be made.

Currently all the Modals are quite messy. In general they need a larger tidy up. Props for `AutoCompleteSelector` are almost exactly the same for all, except for the options needed and `onSelect` function. I think a similar solution to #1159 is needed, a wrapper component which contains a case statement which, when passed a `modalKey`, renders the correct modal.

Doing nothing!

## Additional context

We have quite a lot of mess with modals, from other wrappers `SelectModal`, `PageContentModal` wrapping `BottomConfirmModal` wrapping `Modal` from `react-native-modal-box` etc. I can see the argument that adding another wrapping modal is just adding more rubbish on - I think it is warranted though!

joshxg:
Fixes #1160 

## Change summary

- Very similar to #1159 - factors out modals to be commonly used in multiple places

Added a `ModalContainer` component to be used instead of `PageContentModal` - basically the same thing but re-written with `react-native` base `Modal` component (it is significantly/noticeably faster, imo!).

## Testing
**Internal changes not testable..?**

`getModalOnSelect` -- this method would need to be defined in every `page`. So far it has worked well up to `StocktakeEditPage` which I would say is the most complex page in regard to modals. Everything else.. just works! But definitely open to suggestions on a better option for this.

I hope to be able to use `ModalContainer` and have each 'modal' widget be able to be wrapped by `ModalContainer`, while also having this wrapper, and that's it. 

Would understand deleting `ModalContainer` and replacing with `PageContentModal` if it was better!


src/widgets/modals/DataTablePageModal.js
```diff
@@ -0,0 +1,112 @@
...
+            options={UIDatabase.objects('Customer')}
+            isOpen={isOpen}
+            placeholderText={modalStrings.start_typing_to_select_customer}
+            queryString="name BEGINSWITH[c] $0"
+            sortByString="name"
+            onSelect={onSelect}
+          />
+        );
+
+      case MODAL_KEYS.SELECT_SUPPLIER:
```
Chris-Petty: I think it's worth thinking about the direction here, soon every modal throughout the app's pages will be in this one file. As we continue with this approach eventually the various pages will be virtually redundant and to get an idea of what any one page looks like requires looking at several files.

Thoughts?

src/pages/CustomerInvoicePage.js
```diff
@@ -205,6 +171,19 @@ export const CustomerInvoicePage = ({ transaction, runWithLoadingIndicator, rout
     />
   );
 
+  const getModalOnSelect = () => {
+    switch (modalKey) {
+      case ITEM_SELECT:
+        return value => dispatch(editComment(value, 'Transaction'));
```
Chris-Petty: There miiiight be a chance you didn't test adding items aye 😛

src/widgets/modals/ModalContainer.js
```diff
@@ -0,0 +1,158 @@
...
+    </View>
+  );
+
+  return (
+    <View style={modalContainer}>
+      <Modal
+        visible={isVisible}
+        animationType="slide"
+        fullScreen={fullScreen}
+        transparent={true}
```
Chris-Petty: I'm ok with doing boolean shortcut:
```suggestion
        transparent
```

src/widgets/modals/ModalContainer.js
```diff
@@ -0,0 +1,158 @@
...
+      {onClose && !noCancel && (
+        <View style={closeButtonContainer}>
+          <CloseButton />
+        </View>
+      )}
+    </View>
+  );
+
+  return (
+    <View style={modalContainer}>
```
Chris-Petty: Kinda curious, would have expected this inside the modal?

src/widgets/modals/ModalContainer.js
```diff
@@ -0,0 +1,158 @@
+/**
+ * mSupply Mobile
+ * Sustainable Solutions (NZ) Ltd. 2019
+ */
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { StyleSheet, Text, TouchableOpacity, View, Modal } from 'react-native';
```
Chris-Petty: Glad to be on the RN modal.

src/widgets/modals/DataTablePageModal.js
joshxg: I think you will have to look at several files to fully understand a page. But I think that's ok. Essentially very similar/exact code is duplicated in 8+ pages. I think this consolidation will reduce bugs/inconsistencies and actually make it easier, after a small learning curve - you understand how this component works - it will be easier to understand the next page a little easier. Changes also become easier - one place to make a change effecting a large chunk/all of the app.

Chris-Petty: Fair, just apprehension from mSupply desktop; done adhoc and with process variables the same idea is a nightmare.

Chris-Petty: Lets see how it pans out 😄 currently it's fine though.

src/widgets/modals/ModalContainer.js
joshxg: Me too .. ! (it is very strange..)

src/pages/CustomerInvoicePage.js
joshxg: daf85c90cb54d5abd7a278bf437d0b2324da4418

joshxg:
NOTE this is into #1163

PR #1173, opened 2019-09-03
#1170 Extract key extractor & item layout functions

### LINKED ISSUE #1170: Extract keyExtractor and getItemLayout from pages

Each page uses a `keyExtractor` and `getItemLayout` function. They are exact duplicates everywhere. They should be extracted to their own file.

Extract each to their own file. Not sure exactly which file. `DataTableUtilities/utilities`...?

Not doing this

## Additional context

@Chris-Petty anything else that should be extracted?

joshxg:
Fixes #1170 

## Change summary

- Adds `getItemLayout` and `keyExtractor` to a utilities file.

Not the greatest factoring, but might prevent a mistake in the future

## Testing

*internal change can't test?*

N/A

### DIFF
--- a/src/pages/CustomerInvoicePage.js
+++ b/src/pages/CustomerInvoicePage.js
@@ -12,7 +12,7 @@ import { SearchBar } from 'react-native-ui-components';
 
 import { MODAL_KEYS, getModalTitle } from '../utilities';
 import { buttonStrings, modalStrings } from '../localization';
-import { DEFAULT_KEY_EXTRACTOR, DEFAULT_GET_ITEM_LAYOUT } from './dataTableUtilities/utilities';
+import { recordKeyExtractor, getItemLayout } from './dataTableUtilities/utilities';
 import { UIDatabase } from '../database';
 import { BottomConfirmModal, PageContentModal } from '../widgets/modals';
 import {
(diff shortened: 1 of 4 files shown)


src/pages/CustomerInvoicePage.js
```diff
@@ -12,6 +12,7 @@ import { SearchBar } from 'react-native-ui-components';
 
 import { MODAL_KEYS, getModalTitle } from '../utilities';
 import { buttonStrings, modalStrings } from '../localization';
+import { DEFAULT_KEY_EXTRACTOR, DEFAULT_GET_ITEM_LAYOUT } from './dataTableUtilities/utilities';
```
Chris-Petty: Looks weird to me to name function variable exports in UPPER_SNAKE_CASE, I'd normally reserve that for sources of string constants. What do you reckon?

joshxg: yup I agree - any suggestion though? Obviously landed here cause I was out of ideas. I'll update SI if/when I commit new names 🤣

Chris-Petty: Just regular `keyExtractor` or maybe `recordKeyExtractor`/`idExtractor`.

Basically camelCase is fine?

joshxg:
Have updated from #1043, added to SI and renamed

PR #1178, opened 2019-09-04
#1176 New SearchBar component

### LINKED ISSUE #1176: SearchBar shouldn't have internal state
## The suggestion
`SearchBar` component from `react-native-ui-components` uses internal state. This state cannot be effected outside of the component. 

**Why?**
You can't, for example, programmatically reset the search text.

Not resetting the search text?

## Additional context

I think we should just create a `SearchBar` component in our own repo. It doesn't seem the project is used by anyone?

joshxg:
Fixes #1176 

## Change summary

- Essentially a copy of the `SearchBar` component in `react-native-ui-components`, just as a stateless functional component.

## Testing
The component is in use in `SupplierInvoice`, `CustomerInvoice` and `CustomerInvoices` at the time of this PR.
- [ ] Search in one of the above pages, should filter the list by the filter keys for that page.
- [ ] Adding an item removes the search filter, if there is one
- [ ] Cross on RHS of component clears the current selection

N/A


src/pages/dataTableUtilities/reducer/reducerMethods.js
```diff
@@ -281,10 +282,16 @@ export const addMasterListItems = state => {
...
  */
 export const addItem = (state, action) => {
-  const { data } = state;
+  const { backingData } = state;
   const { item } = action;
 
-  return { ...state, data: [item, ...data], modalKey: '', sortBy: '', searchTerm: '' };
+  return {
+    ...state,
+    data: [item, ...backingData.slice(0, backingData.length - 1)],
```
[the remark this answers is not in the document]
joshxg: oops I forgot about this hehe.

When adding an item - it should show in the list. If you're currently filtering, it would be weird that it showed in the filtered list. It would be weird if it did not show in the list. I think it's nice to add it to the top of the list.

So: clear the current filtering selection - make `data` have the new item, with a new array of all items in `backingData` except for the last element, which was shifted to the start of the array.

Strange bit of logic - I'll should have commented it. Let me know if this behaviour is correct in your head too - if you decide it is, I'll comment it hehe

joshxg:
Have updated this with correct debouncing logic. Sorry!

I have added back internal state! This is to accomadate for debouncing logic. (if the callback isn't called, value wouldn't update)

src/widgets/SearchBar.js
```diff
@@ -37,30 +40,49 @@ export const SearchBar = ({
...
+    if (value !== textValue) setTextValue(value);
+  }, [value]);
+
   const internalViewStyle = { ...viewStyle, borderColor: color };
   const internalTextStyle = { ...textInputStyle, color };
 
-  const debouncedCallback = useMemo(
-    () => newValue => debounce(onChangeText(newValue), debounceTimeout),
-    [onChangeText]
+  // For Debouncing to funtion, need to use the same function as the
```
Chris-Petty: `To debounce the function...`?

src/widgets/SearchBar.js
```diff
@@ -37,30 +40,49 @@ export const SearchBar = ({
...
+  // last call. Memoize this to keep the reference semi-stable. Keeping
+  // this reference is an optimization, not a requirement so can change.
```
Chris-Petty: I mean, there are no dependancies given to useCallBack so it will be stable and won't change if you try right?

src/widgets/SearchBar.js
```diff
@@ -17,14 +17,17 @@ import { debounce } from '../utilities/index';
...
+ * is only invoked once every `debounceTimeout`
+ *
+ * @param {String} color           Color of the entire component (monochrome only).
+ * @param {String} onChangeText    Callback for changing text (debounced).
+ * @param {String} value           Text value of the search input.
+ * @param {String} placeholder     Placeholder value.
+ * @param {String} autoFocus       Autofocus this component on mounting.
+ * @param {String} textInputStyle  Style of the TextInput component.
+ * @param {String} viewStyle       Style of the wrapping View component.
+ * @param {String} debounceTimeout Time in milliseconds to debounce the onChangeText callback.
```
Chris-Petty: Could be dumb, but maybe a HOC like:

```javascript
withDebounce = (component, timeout, [callbackPropsToDebounce]) => {...}
```

Not sure how HOCs like that might play with hooks.

src/widgets/SearchBar.js
```diff
@@ -37,30 +40,49 @@ export const SearchBar = ({
   debounceTimeout,
   ...textInputProps
 }) => {
+  const [textValue, setTextValue] = useState('');
```
Chris-Petty: I thought original problem was that the old search bar had state? useEffect and other tweaks fix the prob?

src/pages/dataTableUtilities/reducer/reducerMethods.js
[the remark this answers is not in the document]
Chris-Petty: Nah I like. Do comment though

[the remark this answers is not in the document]
joshxg: 1f812359cdb14877e869bbce78195f35ee54e65a

src/widgets/SearchBar.js
joshxg: 3a42e4d98fc377162c598fa5d0a806304d370cb4

src/widgets/SearchBar.js
joshxg: there's always a chance memoized data can be kicked out of the cache

src/widgets/SearchBar.js
joshxg: hm it's quite cool - maybe I'll have a play with something

src/widgets/SearchBar.js
joshxg: I  guess I named it incorrectly. It was more that the component was fully uncontrolled

joshxg:
Have updated and fixed conflicts

joshxg:
Change requests done and updated

Issue #1137, opened 2019-08-21
New DataTable post refactor porting


joshxg:
## Is your feature request related to a problem? Please describe.

Currently we are developing the new `DataTable` on a feature branch with the `DataTable` components within. These need to be shifted to `react-native-data-table` once complete.

## Describe the solution you'd like

1. PR into `react-native-data-table` the completed `DataTable` and related components
2. Remove the related components from the `DataTable` feature branch (see epic)
3. Bump the `react-native-data-table` version in this repo

## Describe alternatives you've considered

N/A

## Additional context

Should link to the `react-native-data-table` PR here. This will be a reasonable sized PR which will need to be merged into master in that repo before this can be completed.

joshxg:
I don't think this should block 3.0.0 - can easily go into 3.1.0?

PR #1187, opened 2019-09-08
#1181 New DataTable auto-scroll and focus logic

### LINKED ISSUE #1181: Programmatically scrolling new DataTable

Can't use `KeyboardAwareScrollView` - need to programmatically scroll

- Put a reference to the `DataTable` in state ..
- `Row` pass `index` to cells.
- `focusNextCell` - add `ROW_HEIGHT` to scroll position
- `focusCell` passes `rowIndex`, scrolls to `ROW_HEIGHT * index`

???

## Additional context

N/A

joshxg:
#### EPIC: #1043 

Fixes #1181 

## Change summary

- Adds scrolling to top logic for a row when focused
- Adds focus next logic for cells

## Testing

See #1043 

- At the time of this PR, only two pages `SupplierInvoice` and `CustomerInvoice` had been completed. Possibly breaking change on any new pages after that.
- This decreases the size of the tables viewport. I think this is a good tradeoff as the currently edited area is far more responsive while trading for scrolling behaviour (hiding rows which aren't in the viewport for a second or so when scrolling). This also means we don't need to use the hack for react-native rendering 100+ text inputs and crashing.
- Remove hack-editable cell for above area and focus cell/focus next reducer logic? Keep it because why not?

### DIFF
--- a/src/widgets/DataTable/DataTable.js
+++ b/src/widgets/DataTable/DataTable.js
@@ -126,7 +126,7 @@ DataTable.defaultProps = {
   getItemCount: items => items.length,
   initialNumToRender: 20,
   removeClippedSubviews: true,
-  windowSize: 2,
+  windowSize: 4,
   columns: [],
 };
 
--- a/src/widgets/DataTable/DataTable.js
+++ b/src/widgets/DataTable/DataTable.js
@@ -38,9 +38,11 @@ const DataTable = React.memo(({ renderRow, renderHeader, style, data, columns, .
   // Array of column keys for determining ref indicies.
   const editableColumnKeys = useMemo(
     () =>
-      columns.reduce((acc, column) => {
-        if (column.type === 'editable' || column.type === 'date') return [...acc, column.key];
-        return acc;
+      columns.reduce((columnKeys, column) => {
+        if (column.type === 'editable' || column.type === 'date') {
+          return [...columnKeys, column.key];
+        }
+        return columnKeys;
       }, []),
     [columns]
   );
--- a/src/widgets/DataTable/DataTable.js
+++ b/src/widgets/DataTable/DataTable.js
@@ -5,72 +5,100 @@
  */
 
 import PropTypes from 'prop-types';
-import React from 'react';
+import React, { useMemo, useRef, useCallback } from 'react';
 import { StyleSheet, VirtualizedList, VirtualizedListPropTypes } from 'react-native';
+import RefContext from './RefContext';
 
 /**
- * Base DataTable component. Thin wrapper around VirtualizedList, providing
- * a header component. All VirtualizedList props can be passed through,
- * however renderItem is renamed renderRow.
+ * Base DataTable component. Wrapper around VirtualizedList, providing
+ * a header component, scroll to top and focus features.
+ * All VirtualizedList props can be passed through, however renderItem
+ * is renamed renderRow.
  *
  * Managing focus and scrolling:
- * Can manage focusing and auto-scrolling for editable cells.
- * Three parameters are passed in `renderRow`:
- * - `getCellRef`  : lazily creates a ref for a cell.
- * - `focusNext`   : Focus' the next editable cell. Call during onEditingSubmit.
- * - `adjustToTop` : Scrolls so the focused row is at the top of the list.
+ * Can manage focusing and auto-scrolling for editable cells through react context API.
+ *
+ * Four parameters are passed in through the refContext:
+ *
+ * - `getRefIndex`   : Gets the ref index for an editable cell given the columnkey and row index.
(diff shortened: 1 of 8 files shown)


joshxg:
This PR uses `React.Context` to pass ref logic around, rather than through props. I think this is a good use case - as well as a good use case for `refs`. I think the responsiveness is a lot smoother and I personally think this UI logic is better here than in actions/reducers which are handling business logic. 

I think this is *possibly* more complex - however if you took a developer who knew nothing about either implementation - I'd probably argue that having cells randomly turn into textinputs and manage focus through a map is a little bit more obscure, than using some straight forward refs to manage focus with a little bit of a confusing way to manage the array? 100% bias here though 🤣

src/widgets/DataTable/TextInputCell.js
```diff
@@ -0,0 +1,140 @@
...
+      <View style={internalViewStyle}>
+        <TextInput
+          ref={getCellRef(refIndex)}
+          placeholder={placeholder}
+          style={internalTextStyle}
+          value={usingPlaceholder ? '' : String(value)}
+          placeholderTextColor={placeholderColour}
+          onChangeText={onEdit}
+          onSubmitEditing={focusNext}
+          underlineColorAndroid="#CDCDCD"
```
mark-prins: Shouldn't the colours come from `colors.js`? ( this and in line 137 )
or does that not matter for some things?

src/widgets/NewExpiryDateInput.js
```diff
@@ -29,45 +35,44 @@ import { parseExpiryDate, formatExpiryDate } from '../utilities';
...
     width,
     debug,
     placeholder,
+    rowIndex,
   }) => {
     if (debug) console.log(`- ExpiryTextInputCell: ${value}`);
 
+    const { focusNextCell, getRefIndex, getCellRef } = React.useContext(RefContext);
+
+    console.log('#################################');
```
mark-prins: is this debug code? (i.e. behind the `debug` flag) or temp code (i.e. shouldn't be here)

src/widgets/NewExpiryDateInput.js
```diff
@@ -106,40 +111,21 @@ export const NewExpiryDateInput = React.memo(
...
           placeholder={placeholder}
           style={textStyle}
           value={expiryDate}
+          placeholderTextColor={placeholderColour}
           onChangeText={setExpiryDate}
-          autoFocus={isFocused}
           onSubmitEditing={onSubmit}
           onEndEditing={finishEditingExpiryDate}
-          underlineColorAndroid={dataTableColors.editableCellUnderline}
+          underlineColorAndroid="#CDCDCD"
```
mark-prins: could the colour const be moved to a shared location (maybe `colors.js`)?
( and line 157 )

src/widgets/DataTable/DataTable.js
```diff
@@ -61,7 +126,8 @@ DataTable.defaultProps = {
   getItemCount: items => items.length,
   initialNumToRender: 20,
   removeClippedSubviews: true,
-  windowSize: 11,
+  windowSize: 2,
```
mark-prins: I'm all for it - just wondering the impact of the reduced windowSize and am currently unable to build and therefore view for myself.

src/widgets/DataTable/DataTable.js
```diff
@@ -5,36 +5,100 @@
...
-    >
+const DataTable = React.memo(({ renderRow, renderHeader, style, data, columns, ...otherProps }) => {
+  // Reference to the virtualized list for scroll operations.
+  const virtualizedListRef = useRef();
+
+  // Array of column keys for determining ref indicies.
+  const editableColumnKeys = useMemo(
+    () =>
+      columns.reduce((acc, column) => {
+        if (column.type === 'editable' || column.type === 'date') return [...acc, column.key];
```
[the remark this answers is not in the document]
joshxg: 5b3df16b800279c7512bd6b9be68cbef90ee8196

src/widgets/DataTable/TextInputCell.js
joshxg: all of this folder is being shifted to `react-native-data-table` repo when it's done , which it won't have access to. I've added the colour as a prop instead, with that colour as a default de60b21bd620ee9e1d5ce9d35fc617fffec50bd9. Maybe when we shift this to the repo we can see if a `styles` file is needed here #1137

src/widgets/NewExpiryDateInput.js
joshxg: 38a9f10ddde43bf9e87908dcd76be4d3d6e04179

joshxg:
Going to merge this - I have kept all the code for the previous focusing method and avoidance of react-native textinput bug, if it turns out not to be a good idea!

PR #1202, opened 2019-09-10
#1167 CustomerInvoicesPage fixes

### LINKED ISSUE #1167: DataTable CustomerInvoicesPage fixes

Putting off the title request, as other refactor branches cover the code for handling this well.

Revisit CIs page when date cells are available in base refactor branch

- Needs an EntryDate field added
- Status column needs correct formatting
- New Invoice button doesn't get closed after navigating away
- Navigation done through redux

Redoing the work. Which is a [waste of my time](https://www.youtube.com/watch?v=WTr1M7DwLkc)

joshxg:
Fixes #1167 

## Change summary

- Added columns: expiry date, status
- Added use of `DataTableRow` and `DataTablePageModal`.
- Added use of redux for navigation.
- Added `useNavigationFocusListener` with a couple of changes to make it safer.

## Testing

See #1043 

Memoizing on press handlers: Memoizing all of these is basically memoizing everything and defeats the purpose. The press handlers which aren't memoized ARE causing re-renders, but only on smaller components - `SearchBar`, buttons etc. [ row onPress is handled by `DataTableRow` and doesn't cause re-renders - using a binding or closure means no memoizing].

This is a pretty good case to use class based components for pages, rather than functional. That means no hooks - but I don't know if that's a good enough reason to use functional components for these larger stateful pages.

### DIFF
--- a/src/pages/CustomerInvoicesPage.js
+++ b/src/pages/CustomerInvoicesPage.js
@@ -72,6 +72,7 @@ export const CustomerInvoicesPage = ({
   useNavigationFocusRefresh(dispatch, navigation);
 
   // On Press Handlers
+  const closeModal = () => dispatch(closeBasicModal());
   const onFilterData = value => dispatch(filterData(value));
   const onNewInvoice = () => dispatch(openBasicModal(MODAL_KEYS.SELECT_CUSTOMER));
   const onRemoveInvoices = () => dispatch(deleteTransactionsById());
@@ -94,7 +95,7 @@ export const CustomerInvoicesPage = ({
       case MODAL_KEYS.SELECT_CUSTOMER:
         return otherParty => {
           reduxDispatch(createCustomerInvoice(otherParty, currentUser));
-          dispatch(closeBasicModal());
+          closeModal();
         };
       default:
         return null;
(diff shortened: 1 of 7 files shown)


src/pages/CustomerInvoicesPage.js
```diff
@@ -189,49 +151,49 @@ export const CustomerInvoicesPage = ({ currentUser, navigateTo, routeName, navig
...
+      <DataTablePageModal
+        fullScreen={false}
         isOpen={!!modalKey}
-        options={UIDatabase.objects('Customer')}
-        placeholderText={modalStrings.start_typing_to_select_customer}
-        queryString="name BEGINSWITH[c] $0"
-        sortByString="name"
-        onSelect={name => onNewInvoice(name)}
+        modalKey={modalKey}
         onClose={() => dispatch(closeBasicModal())}
```
mark-prins: `dispatch(closeBasicModal())` is used twice in the file; it could be wrapped in a local method in which case this line could be

```suggestion
        onClose={handleClose}
```

joshxg: 03220bb2f1aebdaea38615cd3cf16a5095859591

PR #1213, opened 2019-09-13
#1210 DataTable Dynamic actions -- 1

### LINKED ISSUE #1210: DataTable dynamic actions

- Thunks shift a good chunk of logic away from the reducer
- My opinion is that an action creator is the best place to do a lot of conditional logic i.e. deciding which action to actually dispatch
- I think it’s best to keep as much logic out of components not related to JSX/Styling
- I think controlling which action creator is executed offers more control over controlling which reducer case is executed

With these thoughts, I think it makes less sense to compose reducers for each page - it may as well just be one PageReducer.

But we still need a way to dynamically change logic, in my opinion. i.e. changing a stocktake quantity cell to pop up a reason modal when reasons are enabled. I think putting this in the page component mixing it with all the JSX is a bit yuck.

I think doing what we did with reducers, but with actions now is better 😂

I.E usePageReducer returns a PageActions and each page can just call PageAction.openModal(x) or PageAction.setCountedQuantity(), and the ‘correct’ action will be called.

I think this makes it simpler as we only need to change PageAction for the pages which actually have overrides

Not doing anything

## Additional context

N/A

joshxg:
#### EPIC: #1043 

Fixes #1210 

## Change summary

- Created `getPageAction` file aand deleted `getReducer`. The data table reducer is now one composed method, `DataTablePageReducer` and actions are now changed during runtime with `getPageActions`. `usePageReducer` now injects a pages state with a `PageAction` object - contains all the actions for `DataTable` pages which can be overriden if there is a use case for behaviour changing under a certian condition (i.e. reasons being in the data base changes stocktake `editCountedQuantity` to also open a reason modal, as well as setting the `countedQuantity` value.
- Renamed some custom hooks. Think they're a lot more descriptive of their functionality.
- custom hooks except a callback now, rather than a dispatch method and calling `refreshData`
- Updated `SupplierRequisitionPage` and `SupplierRequisitionsPage` to use the new `PageAction` object.

## Testing

See #1043 

I think this is simpler and more extensible/modifiable for the future. It could be taken further:

Change the generic `openModal` action - use specific actions for modals for specific modals, so you could change code in for example `SupplierRequisitionsPage` 

```
const usingPrograms = useMemo(() => getAllPrograms(Settings, UIDatabase).length > 0, []);
const { SELECT_SUPPLIER, PROGRAM_REQUISITION } = MODAL_KEYS;
const NEW_REQUISITON = usingPrograms ? PROGRAM_REQUISITION : SELECT_SUPPLIER;
const onNewRequisition = () => dispatch(PageActions.openModal(NEW_REQUISITON));
```

Instead, you could have in `getPageActions`:

```
export const getPageActions = routeName => {
  switch (routeName) {
    case 'programStocktake': {
        const const usingPrograms = getAllPrograms(Settings, UIDatabase).length > 0;
        return usingPrograms ? programRequisitionReducer : requisitionReducer
    } 
```

[ you could also do this check in redux when navigating, aand have a route `programSupplierRequisitionsPage`

Which would allow you to simply use 

```
const onNewRequisition = () => dispatch(PageActions.newRequisition());
```

I think there is still some thought needing to go into how best to handle having multiple changes. I.E if dispensing and programs some how change the same page and both need some action overrides


src/hooks/useNavigationFocus.js
```diff
@@ -5,29 +5,27 @@
...
+export const useNavigationFocus = (callback, navigation) => {
   const subscription = useRef(null);
 
   const subscribe = () => {
     if (subscription.current) return;
-    subscription.current = navigation.addListener('willFocus', () => dispatch(refreshData()));
+    subscription.current = navigation.addListener('willFocus', callback);
   };
 
   const unSubscribe = () => {
```
wlthomson: Nb: is there a use-case for `subscribe` and `unsubscribe` outside of `useEffect`? If not, should they be left out to indicate they should only be run (automatically) when component is mounted/unmounted and are not intended to be used by user code?

src/pages/SupplierRequisitionPage.js
```diff
@@ -96,56 +76,63 @@ export const SupplierRequisitionPage = ({ requisition, runWithLoadingIndicator,
...
-  useDatabaseChangeListener(dispatch, requisition, 'Requisition');
+  useRecordListener(() => dispatch(PageActions.refreshData()), requisition, 'Requisition');
 
   const { isFinalised, comment, theirRef, program } = pageObject;
 
   // On click handlers
-  const onCloseModal = () => dispatch(closeModal());
-  const onSelectItem = () => dispatch(openModal(MODAL_KEYS.SELECT_ITEM));
-  const onViewRegimenData = () => dispatch(openModal(MODAL_KEYS.VIEW_REGIMEN_DATA));
+  const onCloseModal = () => dispatch(PageActions.closeModal());
```
wlthomson: For the functions which are passed down as props, should probably memoise?

src/pages/SupplierRequisitionsPage.js
```diff
@@ -83,28 +71,30 @@ export const SupplierRequisitionsPage = ({
...
   const { SELECT_SUPPLIER, PROGRAM_REQUISITION } = MODAL_KEYS;
   const NEW_REQUISITON = usingPrograms ? PROGRAM_REQUISITION : SELECT_SUPPLIER;
 
   const onPressRow = rowData => () => reduxDispatch(gotoSupplierRequisition(rowData));
-  const onConfirmDelete = () => dispatch(deleteRequisitions());
-  const onCancelDelete = () => dispatch(deselectAll());
-  const onSearchFiltering = value => dispatch(filterData(value));
-  const onNewRequisition = () => dispatch(openModal(NEW_REQUISITON));
-  const onCloseModal = () => dispatch(closeModal());
+  const onConfirmDelete = () => dispatch(PageActions.deleteRequisitions());
```
wlthomson: Same as before, think some of these could be memoised to save a few render cycles?

src/pages/SupplierRequisitionPage.js
joshxg: These are only passed to really basic components - i.e. buttons or the modal, which without a modal key is rendering `null` essentially, and doesn't re-render the page causing a new function to be created and propogate, so I don't really see the point (I'm trying to practise restraint, I really want to memo them :( ) - if it does need performance gains, we can add them but I think at that point we may aswell make the components class based. Also remember that there is stil the entire navigation stack mounted, and therefore all of the components memo'd data is still in the cach - at some point it's basically redundant to do so

src/hooks/useNavigationFocus.js
joshxg: I could see the potential for switching between sync modes (i.e. with sensors) or during a process like dispensing where you wouldnt want updates to occur during issuing a prescription but would if you were just on a list, maybe. I don't see the harm in having the option? If one day it is added, there would be potential for regressions in existing code (unlikely, but ya know), but not the other way round?

PR #1214, opened 2019-09-14
#1209 DataTable columns update -- 2

### LINKED ISSUE #1209: New DataTable Columns
## Describe the bug

Column keys/types etc at the moment are a bit weird. The key in general is used for looking up on an object.

### Additional context

Type at the moment can be used for determining which cell. But, determining if columns etc are editable is annoying. I think changing some keys and types and setting. 

This also effects focusing - as counting the editable columns is only checking for the type 'editable' or 'date' where there are other editable cells

joshxg:
#### BRANCHED FROM: #1210 

Fixes #1209 

## Change summary

- Updated columns to have an `editable` prop
- Updated some keys in columns to be a bit more consistent
- Updated order of columns in `COLUMNS` to be a little clearer
- Added columns which will be used in pages which aren't done yet.

- Updated logic in `DataTable` and `DataTableRow` for correct focusing logic. Was not re-calculating numbers correctly after adding rows.
- Removed `DataTableRow` stlye prop
- Tried to make `DataTableRow` simpler for the type of cell used given a column type.
- Memoized some things to remove a couple of re-renders

## Testing

See #1043 

- `getColumns` folder is getting pretty unwieldy - not sure if there is really a good grouping to separate into files. Possibly the current groups - `string`, `numeric` etc. But I prefer it being one large file - cost of navigating between files > cost of navigating this file, at this point? Even with another possible dozen columns

### DIFF
--- a/src/hooks/usePageReducer.js
+++ b/src/hooks/usePageReducer.js
@@ -50,9 +50,9 @@ export const usePageReducer = (
   debounceTimeout = 250,
   instantDebounceTimeout = 250
 ) => {
-  const columns = useMemo(() => getColumns(page), [page]);
-  const pageInfoColumns = useMemo(() => getPageInfoColumns(page), [page]);
-  const PageActions = useMemo(() => getPageActions(page), [page]);
+  const columns = useMemo(() => getColumns(page), []);
+  const pageInfoColumns = useMemo(() => getPageInfoColumns(page), []);
+  const PageActions = useMemo(() => getPageActions(page), []);
 
   const [pageState, setPageState] = useState({
     ...(initializer ? initializer(pageObject) : initialState),
@@ -66,14 +66,16 @@ export const usePageReducer = (
   const getState = () => stateRef.current;
 
   // Basic dispatch function.
-  const dispatch = action => {
+  const dispatch = useCallback(action => {
     const newState = DataTablePageReducer(getState(), action);
     setPageState(newState);
     stateRef.current = newState;
-  };
+  });
 
-  const thunkDispatcher = action =>
-    typeof action === 'function' ? action(thunkDispatcher, getState) : dispatch(action);
+  const thunkDispatcher = useCallback(
+    action => (typeof action === 'function' ? action(thunkDispatcher, getState) : dispatch(action)),
+    []
+  );
 
   const debouncedDispatch = useCallback(debounce(thunkDispatcher, debounceTimeout), []);
   const instantDebouncedDispatch = useCallback(
--- a/src/pages/SupplierRequisitionsPage.js
+++ b/src/pages/SupplierRequisitionsPage.js
@@ -20,7 +20,7 @@ import { usePageReducer, useNavigationFocus, useSyncListener } from '../hooks';
 import { createSupplierRequisition, gotoSupplierRequisition } from '../navigation/actions';
 import { getItemLayout, recordKeyExtractor } from './dataTableUtilities';
 
-import globalStyles, { SUSSOL_ORANGE, newDataTableStyles, newPageStyles } from '../globalStyles';
+import globalStyles, { SUSSOL_ORANGE, newPageStyles } from '../globalStyles';
 import { buttonStrings, modalStrings } from '../localization';
 
 const initialiseState = () => {
@@ -130,13 +130,11 @@ export const SupplierRequisitionsPage = ({
     listItem => {
       const { item, index } = listItem;
       const rowKey = recordKeyExtractor(item);
-      const { row, alternateRow } = newDataTableStyles;
       return (
         <DataTableRow
           rowData={data[index]}
           rowState={dataState.get(rowKey)}
           rowKey={rowKey}
-          style={index % 2 === 0 ? alternateRow : row}
           columns={columns}
           dispatch={dispatch}
           getAction={getAction}
--- a/src/pages/SupplierRequisitionsPage.js
+++ b/src/pages/SupplierRequisitionsPage.js
@@ -16,7 +16,7 @@ import { DataTable, DataTableHeaderRow, DataTableRow } from '../widgets/DataTabl
 import { UIDatabase } from '../database';
 import Settings from '../settings/MobileAppSettings';
 import { MODAL_KEYS, getAllPrograms, newSortDataBy } from '../utilities';
-import { usePageReducer, useNavigationFocus, useDatabaseListener } from '../hooks';
+import { usePageReducer, useNavigationFocus, useSyncListener } from '../hooks';
 import { createSupplierRequisition, gotoSupplierRequisition } from '../navigation/actions';
 import { getItemLayout, recordKeyExtractor } from './dataTableUtilities';
 
@@ -79,17 +79,16 @@ export const SupplierRequisitionsPage = ({
   } = state;
 
   // Custom hook to refresh data on this page when becoming the head of the stack again.
-  const callback = () => dispatch(PageActions.refreshData());
+  const callback = () => dispatch(PageActions.refreshData(), []);
   useNavigationFocus(callback, navigation);
-
   // Custom hook to listen to sync changes - refreshing data when requisitions are synced.
-  useDatabaseListener(callback, 'Requisition');
+  useSyncListener(callback, 'Requisition');
 
   const usingPrograms = useMemo(() => getAllPrograms(Settings, UIDatabase).length > 0, []);
   const { SELECT_SUPPLIER, PROGRAM_REQUISITION } = MODAL_KEYS;
   const NEW_REQUISITON = usingPrograms ? PROGRAM_REQUISITION : SELECT_SUPPLIER;
 
-  const onPressRow = rowData => () => reduxDispatch(gotoSupplierRequisition(rowData));
+  const onPressRow = useCallback(rowData => reduxDispatch(gotoSupplierRequisition(rowData)), []);
   const onConfirmDelete = () => dispatch(PageActions.deleteRequisitions());
   const onCancelDelete = () => dispatch(PageActions.deselectAll());
   const onSearchFiltering = value => dispatch(PageActions.filterData(value));
@@ -97,13 +96,13 @@ export const SupplierRequisitionsPage = ({
   const onCloseModal = () => dispatch(PageActions.closeModal());
 
   const onCreateRequisition = otherStoreName => {
-    reduxDispatch(createSupplierRequisition({ otherStoreName, currentUser }));
     onCloseModal();
+    reduxDispatch(createSupplierRequisition({ otherStoreName, currentUser }));
   };
 
   const onCreateProgramRequisition = requisitionParameters => {
-    reduxDispatch(createSupplierRequisition({ ...requisitionParameters, currentUser }));
     onCloseModal();
+    reduxDispatch(createSupplierRequisition({ ...requisitionParameters, currentUser }));
   };
 
   const getAction = (colKey, propName) => {
@@ -141,7 +140,7 @@ export const SupplierRequisitionsPage = ({
           columns={columns}
           dispatch={dispatch}
           getAction={getAction}
-          onPress={onPressRow(item)}
+          onPress={onPressRow}
           rowIndex={index}
         />
       );
--- a/src/pages/dataTableUtilities/getColumns.js
+++ b/src/pages/dataTableUtilities/getColumns.js
@@ -21,7 +21,7 @@ const PAGE_COLUMN_WIDTHS = {
 
 const PAGE_COLUMNS = {
   customerInvoice: ['itemCode', 'itemName', 'availableQuantity', 'totalQuantity', 'remove'],
-  supplierInvoice: ['itemCode', 'itemName', 'totalQuantity', 'editableExpiryDate', 'remove'],
+  supplierInvoice: ['itemCode', 'itemName', 'totalQuantity', 'expiryDate', 'remove'],
   customerInvoices: ['serialNumber', 'customer', 'status', 'entryDate', 'comment', 'remove'],
   supplierRequisitions: [
     'serialNumber',
@@ -59,7 +59,7 @@ const PAGE_COLUMNS = {
     'snapshotTotalQuantity',
     'countedTotalQuantity',
     'difference',
-    'modalControl',
+    'batches',
   ],
   stocktakeEditorReasons: [
     'itemCode',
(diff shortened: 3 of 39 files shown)


src/hooks/usePageReducer.js
```diff
@@ -50,9 +50,9 @@ export const usePageReducer = (
   debounceTimeout = 250,
   instantDebounceTimeout = 250
 ) => {
-  const columns = useMemo(() => getColumns(page), [page]);
-  const pageInfoColumns = useMemo(() => getPageInfoColumns(page), [page]);
-  const PageActions = useMemo(() => getPageActions(page), [page]);
+  const columns = useMemo(() => getColumns(page), []);
```
mark-prins: Confused by the change: the API ref says *every value referenced inside the function should also appear in the dependencies array*. It would appear that the `getColumns` method would be re-run for every render? That's the intention? If so, I think I understand and it's all good.

src/hooks/usePageReducer.js
```diff
@@ -66,14 +66,16 @@ export const usePageReducer = (
   const getState = () => stateRef.current;
 
   // Basic dispatch function.
-  const dispatch = action => {
+  const dispatch = useCallback(action => {
     const newState = DataTablePageReducer(getState(), action);
     setPageState(newState);
     stateRef.current = newState;
-  };
+  });
```
mark-prins: should this have an array?

```suggestion
  }, []);
```

src/widgets/DataTable/DataTableRow.js
```diff
@@ -42,187 +42,197 @@ import { formatStatus } from '../../utilities/index';
...
-                value={rowData[columnKey] && rowData[columnKey].toDateString()}
-                width={width}
-                viewStyle={cellContainer[cellAlignment]}
-                textStyle={cellText[cellAlignment]}
-                isLastCell={isLastCell}
-              />
-            );
+            case 'editableDate':
+              return (
+                <NewExpiryDateInput
```
mark-prins: Everything up to this point is generic / flexible, which is why I'm querying: It is an *editableDate* type and the control is `NewExpiryDateInput`. Perhaps the control should be simply `DateInput`?

src/widgets/DataTable/Row.js
```diff
@@ -47,12 +47,12 @@ const Row = React.memo(
     }
 
     const { adjustToTop } = useContext(RefContext);
-
+    const onPressRow = useCallback(() => onPress(rowData));
```
mark-prins: does this need an array? With `rowData` in it? ( I'm not sure, hence the question.. )

src/widgets/DataTable/DataTable.js
```diff
@@ -39,34 +39,40 @@ const DataTable = React.memo(({ renderRow, renderHeader, style, data, columns, .
...
+  // Passes a cell it's ref index.
+  const getRefIndex = (rowIndex, columnKey) => {
+    const columnIndex = editableColumnKeys.findIndex(key => columnKey === key);
+
+    return rowIndex * numberOfEditableColumns + columnIndex;
+  };
 
   // Callback for an editable cell. Lazily creating refs.
   const getCellRef = refIndex => {
     if (cellRefs.current[refIndex]) return cellRefs.current[refIndex];
```
wlthomson: Could memoise this?

```
const getCellRef = useCallback((refIndex) => { ... });
```

src/widgets/DataTable/DataTable.js
```diff
@@ -39,34 +39,40 @@ const DataTable = React.memo(({ renderRow, renderHeader, style, data, columns, .
...
 
     const newRef = React.createRef();
     cellRefs.current[refIndex] = newRef;
+
     return newRef;
   };
 
   // Focuses the next editable cell in the list. Back to the top on last row.
   const focusNextCell = refIndex => {
```
wlthomson: Could memoise this?

```
const focusNextCell = useCallback((refIndex) => { ... });
```

src/widgets/DataTable/DataTableRow.js
```diff
@@ -42,187 +42,197 @@ import { formatStatus } from '../../utilities/index';
...
-                rowIndex={rowIndex}
-              />
-            );
+    // Callback for rendering a row of cells.
+    const renderCells = useCallback(
+      () =>
+        // Map each column to an appropriate cell for a given row.
+        columns.map(({ key: columnKey, type, width, alignText }, index) => {
+          // Indicator if the right hand border should be removed from styles for this cell.
+          const isLastCell = index === columns.length - 1;
```
wlthomson: Nit: could destructure here just for consistency.

```
const { length: columnsLength } = columns;
const isLastCell = index === columnsLength - 1;
```

src/widgets/DataTable/DataTableRow.js
```diff
@@ -42,187 +42,197 @@ import { formatStatus } from '../../utilities/index';
...
-                onUncheckAction={getAction(columnKey, 'onUncheckAction')}
-                dispatch={dispatch}
-                containerStyle={touchableCellContainer}
-                width={width}
-                isLastCell={isLastCell}
-              />
-            );
+          // This cell is disabled if the pageObject is finalised, the row has been explicitly set
+          // as disabled, or the rowData is disabled (i.e. data is an invoice),
+          const isDisabled =
```
wlthomson: Coudl destructure here just for consistency:

```
const { isDisabled: isRowStateDisabled } = rowState;
const { isDisabled: isRowDataDisabled } = rowData;
const isDisabled = isFinalised || isRowStateDisabled || isRowDataDisabled
```

Not sure if should be `rowData.isDisabled` or `rowData.isFinalised`? Looks like comment says the former, but existing code says the latter?
```

File: src/widgets/DataTable/DataTableRow.js [hunk 9]
```diff
@@ -42,187 +42,197 @@ import { formatStatus } from '../../utilities/index';
...
-                value={formatStatus(rowData[columnKey])}
-                width={width}
-                viewStyle={cellContainer[cellAlignment]}
-                textStyle={cellText[cellAlignment]}
-                isLastCell={isLastCell}
-              />
-            );
+          switch (type) {
+            case 'editableString':
+            case 'editableNumeric':
```
wlthomson: I think the types have been updated to not use `'editable'`? Should be switching on `editable`?

src/widgets/DataTable/DataTableRow.js
```diff
@@ -42,187 +42,197 @@ import { formatStatus } from '../../utilities/index';
...
+                <TouchableCell
+                  key={columnKey}
+                  renderChildren={OpenModal}
+                  rowKey={rowKey}
+                  columnKey={columnKey}
+                  onPressAction={getAction(columnKey)}
+                  dispatch={dispatch}
+                  width={width}
+                  isLastCell={isLastCell}
+                  containerStyle={{
```
wlthomson: Maybe should be in component `StyleSheet`?

src/widgets/DataTable/Row.js
wlthomson: I think this is also missing `deps`?

src/hooks/usePageReducer.js
joshxg: From the real docs (stack overflow)

> For both `useMemo` and `useCallback` (which is essentially just a special case of `useMemo`), if the second argument is an empty array, the value will be memoized once and always returned.

src/hooks/usePageReducer.js
joshxg: yup

src/widgets/DataTable/DataTableRow.js
joshxg: mm - I suppose I was copying the previous component, but the component essentially allows for dates in the form `mm/yyyy` and will prepend the last of the previous month (apparently all expiries are the last of the month?)

src/widgets/DataTable/Row.js
joshxg: ya I guess I missed a lot ..

src/hooks/usePageReducer.js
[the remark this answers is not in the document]
joshxg: ya I missed hheaaps

src/pages/SupplierRequisitionsPage.js
```diff
@@ -79,31 +79,30 @@ export const SupplierRequisitionsPage = ({
   } = state;
 
   // Custom hook to refresh data on this page when becoming the head of the stack again.
-  const callback = () => dispatch(PageActions.refreshData());
+  const callback = () => dispatch(PageActions.refreshData(), []);
```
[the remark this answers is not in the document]
joshxg: lol

src/pages/SupplierRequisitionsPage.js
```diff
@@ -79,31 +79,30 @@ export const SupplierRequisitionsPage = ({
...
   const onConfirmDelete = () => dispatch(PageActions.deleteRequisitions());
   const onCancelDelete = () => dispatch(PageActions.deselectAll());
   const onSearchFiltering = value => dispatch(PageActions.filterData(value));
   const onNewRequisition = () => dispatch(PageActions.openModal(NEW_REQUISITON));
   const onCloseModal = () => dispatch(PageActions.closeModal());
 
   const onCreateRequisition = otherStoreName => {
-    reduxDispatch(createSupplierRequisition({ otherStoreName, currentUser }));
     onCloseModal();
+    reduxDispatch(createSupplierRequisition({ otherStoreName, currentUser }));
```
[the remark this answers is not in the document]
joshxg: `src/navigation/actions` is all the navigation actions - this is a redux action, could be named better - it will create and then navigate to a SI

src/pages/dataTableUtilities/getColumns.js
```diff
@@ -68,175 +68,264 @@ const PAGE_COLUMNS = {
...
+    'batches',
   ],
 };
 
 const COLUMNS = () => ({
+  // CODE COLUMNS
   serialNumber: {
+    type: 'string',
     key: 'serialNumber',
     title: tableStrings.invoice_number,
```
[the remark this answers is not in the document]
joshxg: I don't like destructuring non-camel case

src/pages/dataTableUtilities/getColumns.js
```diff
@@ -68,175 +68,264 @@ const PAGE_COLUMNS = {
...
     'reason',
-    'modalControl',
+    'batches',
   ],
 };
 
 const COLUMNS = () => ({
+  // CODE COLUMNS
   serialNumber: {
+    type: 'string',
```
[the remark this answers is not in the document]
joshxg: yeah I was wondering if this is over the top but I agree (with `object.freeze`! 🤣 )

src/hooks/usePageReducer.js
mark-prins: ah. thanks

src/pages/dataTableUtilities/getColumns.js
```diff
@@ -68,175 +68,264 @@ const PAGE_COLUMNS = {
...
+    key: 'status',
+    title: tableStrings.status,
     sortable: true,
+    editable: false,
   },
+
+  // EDITABLE STRING COLUMNS
+
+  batchName: {
+    type: 'editableString',
```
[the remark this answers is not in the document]
joshxg: you're right

src/widgets/DataTable/DataTable.js
```diff
@@ -39,34 +39,40 @@ const DataTable = React.memo(({ renderRow, renderHeader, style, data, columns, .
...
-          return [...columnKeys, column.key];
-        }
+        const { editable } = column;
+        if (editable) return [...columnKeys, column.key];
         return columnKeys;
       }, []),
     [columns]
   );
+  const numberOfEditableColumns = editableColumnKeys.length;
+  const numberOfRows = data.length;
```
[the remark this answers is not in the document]
joshxg: people don't seem to like destructuring lengths 🤷‍♀

src/widgets/DataTable/Row.js
mark-prins: yeah, as many as 4 small things in 100s of lines of code changes. 🤔

src/widgets/DataTable/DataTableRow.js
joshxg: comment wrong

src/widgets/DataTable/DataTableRow.js
joshxg: bit annoying - the `editable` field is mainly to be able to count the number of `editable` columns for easier focusing. The `type` is for easier switching for `cells` ...

src/widgets/DataTable/DataTableRow.js
joshxg: yeah

src/pages/SupplierRequisitionsPage.js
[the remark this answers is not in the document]
wlthomson: Maybe missign something but isn't this a thunk? Looks like `createSupplierRequisition` returns a thunk, doesn't it need to be passed `dispatch` as a param?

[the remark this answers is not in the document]
joshxg: any function returned by an action creator is passed `dispatch` and a function `getState` when it is called by the dispatcher - don't need to pass it here (one of the reasons why thunks are 'too powerful'! see `usePageReducer` for my implementation of the thunk middleware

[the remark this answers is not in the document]
joshxg: oh - and also, we alreayd have thunk middleware for redux! (I didn't even need to add it -- it was already a dependency just never used)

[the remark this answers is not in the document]
wlthomson: > oh - and also, we alreayd have thunk middleware for redux! (I didn't even need to add it -- it was already a dependency just never used)

Oh, is that possibly that's why it works? The thunk dispatcher does what you described, but you're calling it with the normal `dispatch` function passed down by react-redux `Provider`?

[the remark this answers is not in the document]
joshxg: yup - our redux has thunk middleware, so it works basically exactly the same as `usePageReducer`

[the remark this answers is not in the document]
wlthomson: I get that, I'm just a bit confused as to the different use-cases? Shouldn't we be using the `usePageReducer` implementation of `dispatch` for any thunks?

src/widgets/DataTable/DataTable.js
joshxg: Forgot to reply to these - these are contained in an object within a react context which is memoized, so no need

PR #1215, opened 2019-09-14
#1197 StocktakesPage fixes -- 3

### LINKED ISSUE #1197: DataTable StocktakePage fixes
## Describe the bug

Please see PR comments in #1190 
### To reproduce
N/A
### Expected behaviour
N/A

### Additional context

Issue is to make the changes from requested changes in #1190

joshxg:
### NOTE: Branch not runnable due to many files being shifted - branch at #1224 is when the app is runnable again.

Fixes #1197 

## Change summary

- Defined callbacks explicitly
- Using `PageActions` for actions
- Uses new `SearchBar` component
- Added a `stateInitialiser` function

## Testing

See #1043 

- Need to add a `TouchableWithoutFeedback` to `TouchableCell` when it is disabled. Clicking on a finalised rows remove icon navigates to it as the event isn't being captured in the child

### DIFF
--- a/src/pages/StocktakesPage.js
+++ b/src/pages/StocktakesPage.js
@@ -33,7 +33,7 @@ const stateInitialiser = () => {
   const backingData = UIDatabase.objects('Stocktake');
   return {
     backingData,
-    data: backingData.sorted('createdDate', false).slice(),
+    data: backingData.sorted('createdDate', true).slice(),
     keyExtractor: recordKeyExtractor,
     dataState: new Map(),
     searchTerm: '',
@@ -119,6 +119,7 @@ export const StocktakesPage = ({ routeName, currentUser, dispatch: reduxDispatch
           dispatch={dispatch}
           getAction={getAction}
           onPress={onRowPress}
+          rowIndex={index}
         />
       );
     },
--- a/src/pages/StocktakesPage.js
+++ b/src/pages/StocktakesPage.js
@@ -21,7 +21,7 @@ import { BottomConfirmModal, DataTablePageModal } from '../widgets/modals';
 import { DataTable, DataTableHeaderRow, DataTableRow } from '../widgets/DataTable';
 
 import { buttonStrings, modalStrings } from '../localization';
-import globalStyles, { SUSSOL_ORANGE, newDataTableStyles, newPageStyles } from '../globalStyles';
+import globalStyles, { SUSSOL_ORANGE, newPageStyles } from '../globalStyles';
 
 import {
   gotoStocktakeManagePage,
@@ -105,14 +105,11 @@ export const StocktakesPage = ({ routeName, currentUser, dispatch: reduxDispatch
     listItem => {
       const { item, index } = listItem;
       const rowKey = keyExtractor(item);
-      const { row, alternateRow } = newDataTableStyles;
-
       return (
         <DataTableRow
           rowData={data[index]}
           rowState={dataState.get(rowKey)}
           rowKey={rowKey}
-          style={index % 2 === 0 ? alternateRow : row}
           columns={columns}
           dispatch={dispatch}
           getAction={getAction}
--- a/src/pages/StocktakesPage.js
+++ b/src/pages/StocktakesPage.js
@@ -18,16 +18,6 @@ import { BottomConfirmModal, DataTablePageModal } from '../widgets/modals';
 import { PageButton, DataTablePageView } from '../widgets';
 import { DataTable, DataTableHeaderRow, DataTableRow } from '../widgets/DataTable';
 
-import {
-  selectRow,
-  deselectRow,
-  deselectAll,
-  deleteStocktakes,
-} from './dataTableUtilities/actions/rowActions';
-
-import { filterData, sortData } from './dataTableUtilities/actions/tableActions';
-import { openModal, closeModal } from './dataTableUtilities/actions/pageActions';
-
 import { getItemLayout, recordKeyExtractor } from './dataTableUtilities/utilities';
 import globalStyles, { SUSSOL_ORANGE, newDataTableStyles, newPageStyles } from '../globalStyles';
 import { usePageReducer } from '../hooks/usePageReducer';
@@ -62,11 +52,12 @@ export const StocktakesPage = ({ routeName, currentUser, dispatch: reduxDispatch
     dataState,
     sortBy,
     isAscending,
-    columns,
     modalKey,
     hasSelection,
     usingPrograms,
     keyExtractor,
+    columns,
+    PageActions,
   } = state;
 
   const { PROGRAM_STOCKTAKE } = MODAL_KEYS;
@@ -74,8 +65,8 @@ export const StocktakesPage = ({ routeName, currentUser, dispatch: reduxDispatch
   const getAction = useCallback((colKey, propName) => {
     switch (colKey) {
       case 'remove':
-        if (propName === 'onCheckAction') return selectRow;
-        return deselectRow;
+        if (propName === 'onCheckAction') return PageActions.selectRow;
+        return PageActions.deselectRow;
       default:
         return null;
     }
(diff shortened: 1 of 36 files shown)


src/pages/StocktakesPage.js
```diff
@@ -8,44 +8,32 @@
...
 
-const keyExtractor = item => item.id;
-
-export const StocktakesPage = ({ routeName, currentUser, dispatch: reduxDispatch }) => {
-  const [state, dispatch, instantDebouncedDispatch, debouncedDispatch] = usePageReducer(routeName, {
-    backingData: UIDatabase.objects('Stocktake'),
-    data: UIDatabase.objects('Stocktake')
-      .sorted('createdDate', false)
-      .slice(),
+const stateInitialiser = () => {
```
joshxg: Potentially put state initialisers for all pages in a file which can be used by each pages reducer for an easy way to 'reset' state

src/pages/StocktakesPage.js
```diff
@@ -54,95 +42,111 @@ export const StocktakesPage = ({ routeName, currentUser, dispatch: reduxDispatch
...
     modalKey,
     hasSelection,
     usingPrograms,
+    keyExtractor,
+    columns,
+    PageActions,
   } = state;
 
-  const { PROGRAM_STOCKTAKE } = MODAL_KEYS;
+  const callback = useCallback(() => dispatch(PageActions.refreshData()), []);
```
joshxg: `refreshCallback`?

src/pages/StocktakesPage.js
```diff
@@ -54,95 +42,111 @@ export const StocktakesPage = ({ routeName, currentUser, dispatch: reduxDispatch
...
+
+  const onRowPress = useCallback(stocktake => reduxDispatch(gotoStocktakeEditPage(stocktake)), []);
+  const onFilterData = value => dispatch(PageActions.filterData(value));
+  const onCancelDelete = () => dispatch(PageActions.deselectAll());
+  const onConfirmDelete = () => dispatch(PageActions.deleteStocktakes());
+  const onCloseModal = () => dispatch(PageActions.closeModal());
+
+  const onNewStocktake = () => {
+    if (usingPrograms) return dispatch(PageActions.openModal(MODAL_KEYS.PROGRAM_STOCKTAKE));
+    return reduxDispatch(gotoStocktakeManagePage({ stocktakeName: '' }));
```
joshxg: use `export const gotoStocktakeManagePage = ({stocktake, stocktakeName = ""} = {}) => {...` in `navigation/actions`

then just `return reduxDispatch(gotoStocktakeManagePage());`

src/pages/StocktakesPage.js
```diff
@@ -8,44 +8,32 @@
...
-import DataTablePageView from './containers/DataTablePageView';
+
+import { buttonStrings, modalStrings } from '../localization';
+import globalStyles, { SUSSOL_ORANGE, newPageStyles } from '../globalStyles';
 
 import {
   gotoStocktakeManagePage,
   createStocktake,
   gotoStocktakeEditPage,
 } from '../navigation/actions';
```
joshxg: use `navigation/index`

src/pages/StocktakesPage.js
```diff
@@ -54,95 +42,111 @@ export const StocktakesPage = ({ routeName, currentUser, dispatch: reduxDispatch
...
+    return reduxDispatch(gotoStocktakeManagePage({ stocktakeName: '' }));
+  };
 
-  const getAction = useCallback((colKey, propName) => {
+  const getAction = (colKey, propName) => {
     switch (colKey) {
       case 'remove':
-        if (propName === 'onCheckAction') return selectRow;
-        return deselectRow;
+        if (propName === 'onCheckAction') return PageActions.selectRow;
```
joshxg: is having a `COLUMN_KEYS` constant overkill, to remove the `case 'remove':` in place of `case COLUMN_KEYS.REMOVE:`

maybe `return propName === 'onCheckAction' ? PageActions.selectRow : PageActions.deselectRow` or maybe thats ugly

src/pages/StocktakesPage.js
```diff
@@ -155,13 +159,16 @@ export const StocktakesPage = ({ routeName, currentUser, dispatch: reduxDispatch
       <View style={newPageTopSectionContainer}>
         <View style={newPageTopLeftSectionContainer}>
           <SearchBar
-            onChange={value => debouncedDispatch(filterData(value))}
+            onChangeText={onFilterData}
             style={searchBar}
             color={SUSSOL_ORANGE}
             placeholder=""
+            value={searchTerm}
```
joshxg: passing `style`, `color` and `placeholder` values which are equal to defaults

joshxg:
Also forgot to shift finalize error check method again

src/pages/StocktakesPage.js
mark-prins: that's an interesting idea. it would make components smaller and easier to read which I like; just whether it means when changing a component you have to refer to multiple files and whether that is a bother. I think overall, good idea

src/pages/StocktakesPage.js
mark-prins: darn. yes. missed that. it's in a few files isn't it?

src/pages/StocktakesPage.js
mark-prins: probably is overkill. but I like the idea all the same

src/pages/StocktakesPage.js
joshxg: ya will just make an issue to fix em

src/pages/StocktakesPage.js
```diff
@@ -54,95 +42,111 @@ export const StocktakesPage = ({ routeName, currentUser, dispatch: reduxDispatch
...
 
-  const { PROGRAM_STOCKTAKE } = MODAL_KEYS;
+  const callback = useCallback(() => dispatch(PageActions.refreshData()), []);
+  // Listen to sync changing stocktake data - refresh if there are any.
+  useSyncListener(callback, ['Stocktake']);
+  // Listen to navigation focusing this page - fresh if so.
+  useNavigationFocus(callback, navigation);
+
+  const onRowPress = useCallback(stocktake => reduxDispatch(gotoStocktakeEditPage(stocktake)), []);
+  const onFilterData = value => dispatch(PageActions.filterData(value));
```
wlthomson: Should these functions be memoised, too?

src/pages/StocktakesPage.js
wlthomson: In general, I think all string literals should have constants. It reduces errors and indicates clearly the "interface" that the field implements. I like this because it also can be used for run-time type checking:

```
if (Object.values(COLUMN_KEYS).includes(key)) {
  // Do stuff.
} else {
  // Handle error.
}
```

src/pages/StocktakesPage.js
```diff
@@ -54,95 +42,111 @@ export const StocktakesPage = ({ routeName, currentUser, dispatch: reduxDispatch
...
+  const PageButtons = useCallback(() => {
     const { verticalContainer, topButton } = globalStyles;
     return (
       <View style={verticalContainer}>
-        <PageButton
-          style={topButton}
-          text={buttonStrings.new_stocktake}
-          onPress={() => newStocktake()}
-        />
+        <PageButton style={topButton} text={buttonStrings.new_stocktake} onPress={onNewStocktake} />
```
wlthomson: I know you don't like destructuring underscore variables, but suggestion anyway:

```
const { new_stocktake: newStocktakeText } = buttonStrings
<PageButton style={topButton} text={newStocktakeText} onPress={onNewStocktake} />
```

src/pages/StocktakesPage.js
joshxg: You're right this is the way I was thinking also - The problem being at this point is that the `initialState` is sometimes derived from the props passed to the page, so there needs to be some parameter passing here. 

The impurity is the main reason I haven't done it. Yet to think of a good solution which enables the use of the initialiser function *as well* as having it usable within the reducer

src/pages/StocktakesPage.js
joshxg: in an ideal world they all would be - but that kinda defeats the purpose. If we feel the performance gains are neccessarry we can add them (I think these pages mightv'e been better as class based components, so memoizing all of this wouldn't be neccessary)

PR #1223, opened 2019-09-14
#1220 CustomerInvoicesPage fixes -- 8

joshxg:
### NOTE: Branch not runnable due to many files being shifted - branch at #1224 is when the app is runnable again.

Fixes #1220 

## Change summary

- Added use of custom hooks: `useNavigationFocus` and `useSyncListener` to refresh on incoming sync (if a traansaction comes in) as well as on re-focusing the screen
- Added use of `PageActions` for actions
- Explictly defined functions for callbacks

## Testing

See #1043 

### Related areas to think about

N/A

PR #1224, opened 2019-09-14
#1218 CustomerInvoicePage  fixes -- 9

### LINKED ISSUE #1218: CustomerInvoicePage to use PageActions
## The suggestion
`CustomerInvoicePage` was refactored before added `PageActions`.

**Why?**
It's broken

N/A

## Additional context

_Add any other context or screenshots about the refactor suggestion here._

joshxg:
#### BRANCHED FROM: #1223 

Fixes #1218 

## Change summary

- Added a `stateInitialiser` function
- Added use of `PageActions` for actions
- explicitly defined all callbacks
- Fixed a minor issue with toggle bar/page button alignment
- Fixed an issue with `DropDownCell` placeholder prop
- Added the use of `useRecordListener` custom hook to listen for being finalised

## Testing

See #1043 

- I think a container component which would do all alignment in a consistent way for buttons would be good. Similar to `DataTableRow`, could also be good. Passing some array of button keys as a selection and it just rendered them correctly in the right positions. i.e.

pass: `[[ADD_ITEM, ADD_FROM_MASTER_LIST], [USE_SUGGESTED_QUANTITES, CREATE_AUTOMATIC_ORDER']]`

And the component would render a 2x2 of the buttons in that order. Really should just be able to pass `PageActions`, and each button could call an action, which could be overriden for a particular page. Very flexible and consistent 🤷‍♀ 

Would then have basically full control over every page - can easily create a custom page with custom page info, custom columns, custom buttons where all actions (basically all logic) could be custom - all column logic/cell editing/buttons and even finalisation

### DIFF
--- a/src/database/DataTypes/Requisition.js
+++ b/src/database/DataTypes/Requisition.js
@@ -7,7 +7,7 @@ import Realm from 'realm';
 import { complement } from 'set-manipulator';
 
 import { createRecord, getTotal } from '../utilities';
-import UIDatabase from '../UIDatabase';
+import { UIDatabase } from '..';
 
 /**
  * A requisition.
--- a/src/database/DataTypes/StocktakeBatch.js
+++ b/src/database/DataTypes/StocktakeBatch.js
@@ -1,4 +1,5 @@
 import Realm from 'realm';
+
 import { createRecord } from '../utilities';
 
 /**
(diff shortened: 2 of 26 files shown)


joshxg:
Filtering should be on `item.name` AND `item.code` for this page

PR #1256, opened 2019-09-17
#1250 column types constant

### LINKED ISSUE #1250: Column types should be constant

in `getColumns` there are a bunch of `type` strings.

All these `types` should be defined as constants.

Not doing anything

## Additional context

https://github.com/openmsupply/mobile/pull/1215#discussion_r324482261

wlthomson:
Fixes #1250.

## Change summary

Replaces column type strings with an enum.

Not sure about the ideal file to export this from or the best identifiers to use. Not completely sure about the domain of these constants (where to be used, where not), but I think I got the main use cases. Feedback welcomed on all these.

## Testing

- DO TO.

Might be part of a bigger convo to have later re. how we organise "typing" in mobile in general (i.e. if we want to make it part of our design architecture that we have different type definitions for different domains, e.g. incoming 4D types, internal realm types, internal component types etc.)


src/pages/dataTableUtilities/getColumns.js
```diff
@@ -5,6 +5,8 @@
 
 import { tableStrings } from '../../localization';
 
+import COLUMN_TYPES from '../../widgets/DataTable/types';
```
joshxg: Need to do this for the `keys` also

Also I think `pages/datatableUtilities` is more suitable place

wlthomson: Yeah, agreed. Done.

src/pages/dataTableUtilities/constants.js
```diff
@@ -0,0 +1,63 @@
...
+  NUMERIC_EDITABLE: 'editableNumeric',
+  NUMERIC: 'numeric',
+  STRING_EDITABLE: 'editableString',
+  STRING: 'string',
+};
+
+export const COLUMN_KEYS = {
+  AVAILABLE_QUANTITY: 'availableQuantity',
+  BATCHES: 'batches',
+  BATCH: 'batch',
```
joshxg: I don't see this being used

joshxg: im stupid - it's batches that isn't used. I think this seems a little confused? There are keys in here mixed with fields of the `COLUMNS` object? I think either have just the keys, or separate them into two different objects? Keys are fields/getters on realm models/objects - it seems like this could actually introduce more errors - using a key which is actually a field in `COLUMNS`?

src/pages/dataTableUtilities/getColumns.js
```diff
@@ -25,112 +27,148 @@ const PAGE_COLUMN_WIDTHS = {
...
+    COLUMN_KEYS.QUESTION,
+    COLUMN_KEYS.EDITABLE_VALUE,
+    COLUMN_KEYS.EDITABLE_COMMENT,
   ],
-  regimenDataModal: ['question', 'editableValue', 'editableComment'],
 };
 
 const COLUMNS = () => ({
   // CODE COLUMNS
   invoiceNumber: {
```
joshxg: can do 

```
const field = "field";

const obj = {
  [field]: "value"
};

console.log(Object.keys(obj));
```

logs "field"

src/widgets/DataTable/DataTableRow.js
```diff
@@ -29,6 +29,8 @@ import TextInputCell from './TextInputCell';
 
 import { formatStatus } from '../../utilities/index';
 
+import { COLUMN_TYPES } from '../../pages/dataTableUtilities/constants';
```
joshxg: forgot this one: export from `dataTableUtilities/index`

wlthomson: 4d1161b32797fb3e027fcae76df281c1063ceec1.

src/pages/dataTableUtilities/getColumns.js
wlthomson: Done.

src/pages/dataTableUtilities/constants.js
wlthomson: Yeah, it's a bit confusing to differentiate how different strings as used here because they get "post-processed" at other places in the code base. Updated now, let me know if I'm still not understanding it.

PR #1257, opened 2019-09-17
#1128 stock page refactor

### LINKED ISSUE #1128: Refactor StockPage to use new DataTable
Need to use the new `DataTable` in `StockPage`.

- A more or less complete refactor, using the first refactored page `CustomerInvoicePage` as a guide.
- This will require using hooks and the new DataTable `component`.

N/A

## Additional context

This is a placeholder issue. Please investigate and compare the page with `CustomerInvoicePage`. Each page being refactored may have their own quirks. The tissue is encouraged to describe any additional problems and differences with the particular page here, to find a solution which encourages further extensibility and code-reuse with other pages, if possible.

joshxg:
Fixes #1128 

## Change summary

- Refactors `StockPage` to use the new `DataTable`
- Added a `ItemDetails` component which is really just a hack putting the expansion from https://github.com/openmsupply/mobile/blob/master/src/pages/StockPage.js the old stock page wrapped in a scroll view and modal
- Added colour props to `PageInfo`

Not great, can iterate if theres other ideas to make it nicer
stockpage

## Testing

See #1043 

- A better replacement for expansions is needed - more generic
- Need to refactor `PageInfo` it's a bit crazy!


src/widgets/DataTable/DataTableRow.js
```diff
@@ -56,11 +56,14 @@ const DataTableRow = React.memo(
       editableCellText,
       editableCellTextView,
       editableCellUnfocused,
+      selectedRow,
       alternateRow,
       row,
     } = newDataTableStyles;
 
-    const rowStyle = rowIndex % 2 === 0 ? alternateRow : row;
+    const { isSelected = false } = rowState || {};
```
mark-prins: nifty! bonus points for a concise one-liner

src/widgets/DataTable/DataTableRow.js
```diff
@@ -56,11 +56,14 @@ const DataTableRow = React.memo(
...
+    // If the row is selected, use selectedRow style, otherwise alternate row style on index.
+    const rowStyle = isSelected ? selectedRow : (rowIndex % 2 === 0 && alternateRow) || row;
```
mark-prins: a rare case where the destructuring is unhelpful - `row` by itself didn't tell me that it was a style object, whereas `newDataTableStyles.row` would

PR #1260, opened 2019-09-18
#1248 Price/Unit Columns

### LINKED ISSUE #1248: SupplierRequisition price unit columns not implemented
## Describe the bug

The price and unit columns in `SupplierRequisitionPage` (for program requisitions) are not implemented correctly (at all actually)

This is because I wanted to change the way `price` was done for an item.

In master it's done like [this](https://github.com/openmsupply/mobile/blob/55057105f3da721188cc8761a65cf253aa74cca9/src/pages/SupplierRequisitionPage.js#L155-L179) - takes the maximum price of all master list name joins for an item, or the price on an item itself. I think price is quite a highly customizable feature 

I think it would be best to change this to be more dynamic, I propose:

- Add a `get priceZ` (??) field to `Item`
- Within this method, call a utility function `getItemPrice(this, masterListNameJoins)` which returns the max, or this.price.

I am not sure if calling a function here is the right move - but the idea is that it would, *if* price customizations come in it would look more like:

```
get priceZ(){
    const priceStrategy = Settings.get(THIS_STORE_PRICING_STRATEGY);
    switch (priceStrategy){
        type(CAMBODIA_SPECIAL) return cambodiaSpecialPrice();
        type(IVORY_COAST) return ivoryCoastPrice();
    }
}
```

Actually, this is probably over engineered. I think just have a `getter` which returns the current pricing algorithm..

### To reproduce

Development bug

### Expected behaviour

To have working price and unit columns

### Version and device info

`DataTable` refactor branch

### Additional context

N/A

joshxg:
Fixes #1248 

## Change summary

- Added pass through of `get` from `UIDatabase` so we can use it now
- Added new call `getSetting` to `UIDatabase` as because all of the many layers on the database there are many dependency cycles - easier to just import `UIDatabase`
- Added pricing algorith mfor requisition items - max price of all master list items for this store

## Testing

See #1043 

N/A

### DIFF
--- a/src/database/UIDatabase.js
+++ b/src/database/UIDatabase.js
@@ -136,6 +136,10 @@ class UIDatabase {
     }
   }
 
+  get(...args) {
+    return this.database.get(...args);
+  }
+
   addListener(...args) {
     return this.database.addListener(...args);
   }
@@ -174,6 +178,11 @@ class UIDatabase {
 
   write(...args) {
     return this.database.write(...args);
+  }
+
+  getSetting(key) {
+    const setting = this.database.get('Setting', key, 'key');
+    return setting && setting.value;
   }
 }
 
--- a/src/database/UIDatabase.js
+++ b/src/database/UIDatabase.js
@@ -182,7 +182,7 @@ class UIDatabase {
 
   getSetting(key) {
     const setting = this.database.get('Setting', key, 'key');
-    return setting && setting.value;
+    return (setting && setting.value) || '';
   }
 }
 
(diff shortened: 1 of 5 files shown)


src/database/UIDatabase.js
```diff
@@ -175,6 +179,11 @@ class UIDatabase {
   write(...args) {
     return this.database.write(...args);
   }
+
+  getSetting(key) {
+    const setting = this.database.get('Setting', key, 'key');
+    return setting && setting.value;
```
joshxg: mmm maybe this should be `(setting && setting.value || '')`

mark-prins: I'm not sure about usage and the expectations of callers.. so will defer to whatever you think best. I like having a default myself

Issue #845, opened 2019-04-03
Update ESLint to disable forbidden prop types


joshxg:
ESLint rules to forbid certain prop types (e.g objects, arrays) is solid but not necessary for this project. Nearly every file has this disabled.

If anyone has any other updates they have found would be good I can add them with this?

joshxg:
```
onPress={
              stocktakeItem.shouldHaveReason
                ? () =>
                    this.setState({ currentStocktakeItem: stocktakeItem }, this.onOpenReasonModal)
                : () => {}
            }
```

This is the indentation problem - prettier adds two extra spaces on line 4 as prettier formats ternaries differently than the linter (Usually this isn't a prioblem, e.g this code is being changed, but found it while writing it initially)

Chris-Petty:
Seems a bit of an edge case and indentation consistency everywhere else isn't worth giving up for I think.

But yea disabling the forbid prop types would be good I think.

wlthomson:
I've been having generic indentation issues with these two rules conflicting:

`.eslintrc`

```
{
  ...
  "indent": ["error", 2, { "SwitchCase": 1 }]
}
```

`.prettierrc`

```
{
  "tabWidth": 2
}
```

This problem only occurred recently, but it's interfered to the extent that I've had to disable applying prettier on save. It could be something else externally interferring, but I spent a good amount of time trying to figure out before giving up. If nobody else is having problems, it's probably just my setup doing something weird.

I mea culpa that much of the current setup was my doing, because I thought it would be cool for us to have our own "sussol style". In hindsight, the flexibility of our config has ended up becoming a bit of a hassle to wrangle and has ended up becoming a bit counterproductive. In the future, it might be worth trialing something like [eslint-config-prettier](https://github.com/prettier/eslint-config-prettier) to the end of keeping our config as automated and small as possible.

joshxg:
Going to do this today. I would also like to disable the preference for default exports. Objections?

wlthomson:
> Going to do this today. I would also like to disable the preference for default exports. Objections?

I'm OK with disabling it, but in the case of forbidden prop types, have you looked into using `arrayOf`, `objectOf` and  `Shape`? I like type checking, so I lean towards this before resorting to disabling it.

joshxg:
I definitely like prop types also. The problem at the moment is that we pass around the `Realm` database instance & `settings`. `PropTypes` don't have a `Realm` `PropType` afaik

wlthomson:
[...]

PR #1266, opened 2019-09-18
#1249 extract state initialisers

### LINKED ISSUE #1249: State initialisers should be extracted
## Describe the bug

Each page uses a `stateInitialiser` function. This function is within the `XXXPage` component. It should be extracted for the benefits:

- Simple method to use to 'reset' state of a component by the components reducer
- Component becomes a bit cleaner.

I think this would potentially be a good method to call after pushing to the navigation stack - so that poping (and returning to the screen) would result in a clean page, rather than a potentially dirty one with still selected items etc. Of course you can just add into the code to unselect all items etc etc, but it is a lot simpler to just re-init the state. Would ensure nothing was missed (i.e. forgetting to reset `modalKey: ''` to ensure the modal is closed.

Couple of options as pointed out here: https://github.com/openmsupply/mobile/pull/1215#discussion_r324481839

I'm going to make a claim that reading from the DB is an OK thing to do in the reducer. Essentially we are already doing this with `backingData` and any `pageObject`. Reading a requisitions `items` field is essentially equivalent to reading `UIDatabase.objects("X")`

I think having these functions in `pageReducers`?

Also https://github.com/openmsupply/mobile/pull/1216#discussion_r324486286

Need to also ensure `usePageReducer` can be used without having one of these intiialiser functions defined. Also each initialiser may need to be derived from props passed to the component (i.e. a requisition)

### To reproduce

Development

### Expected behaviour

N/A

### Version and device info

- App version:
- Tablet model:
- OS version:

wlthomson:
Fixes #1249.

## Change summary

Simplifies page logic by moving initialisation logic into a dedicated file and updating the `usePageReducer` hook signature to more closely resemble the standard React `useReducer` hook. 

## Testing

- [ ] All pages still work and don't crash!

Updated initialisation naming to UK English 🇬🇧 style (both US and UK were being used in different contexts). If people prefer US 🇺🇸 naming, can update with a quick find/replace.

### DIFF
--- a/src/hooks/usePageReducer.js
+++ b/src/hooks/usePageReducer.js
@@ -45,6 +45,7 @@ import { debounce } from '../utilities/index';
  */
 export const usePageReducer = (
   initialState,
+  initialiser,
   debounceTimeout = 250,
   instantDebounceTimeout = 250
 ) => {
@@ -52,10 +53,10 @@ export const usePageReducer = (
   const columns = useMemo(() => getColumns(page), []);
   const pageInfoColumns = useMemo(() => getPageInfoColumns(page), []);
   const PageActions = useMemo(() => getPageActions(page), []);
-  const initialiser = useMemo(() => getPageInitialiser(page), []);
+  const pageInitialiser = useMemo(() => initialiser || getPageInitialiser(page), []);
 
   const [pageState, setPageState] = useState({
-    ...(initialiser ? initialiser(pageObject) : initialState),
+    ...(pageInitialiser ? pageInitialiser(pageObject) : initialState),
     columns,
     getPageInfoColumns: pageInfoColumns,
     PageActions,
--- a/src/hooks/usePageReducer.js
+++ b/src/hooks/usePageReducer.js
@@ -43,13 +43,12 @@ import { debounce } from '../utilities/index';
  * @param {Number} instantDebounceTimeout Timeout period for an instant debounce
  */
 export const usePageReducer = (
-  page,
   initialState,
   initializer,
-  pageObject,
   debounceTimeout = 250,
   instantDebounceTimeout = 250
 ) => {
+  const { page, pageObject } = initialState;
   const columns = useMemo(() => getColumns(page), []);
   const pageInfoColumns = useMemo(() => getPageInfoColumns(page), []);
   const PageActions = useMemo(() => getPageActions(page), []);
--- a/src/hooks/usePageReducer.js
+++ b/src/hooks/usePageReducer.js
@@ -10,6 +10,7 @@ import {
   getPageActions,
   getColumns,
   getPageInfoColumns,
+  getPageInitialiser,
 } from '../pages/dataTableUtilities';
 
 import { debounce } from '../utilities/index';
@@ -37,14 +38,13 @@ import { debounce } from '../utilities/index';
  *
  * @param {String} page                   routeName for the current page.
  * @param {Object} initialState           Initial state of the reducer
- * @param {Func}   initializer            Function to generate the initial state (optional)
+ * @param {Func}   initialiser            Function to generate the initial state (optional)
  * @param {Object} pageObject             base PageObject for the page i.e. a Requisition.
  * @param {Number} debounceTimeout        Timeout period for a regular debounce
  * @param {Number} instantDebounceTimeout Timeout period for an instant debounce
  */
 export const usePageReducer = (
   initialState,
-  initializer,
   debounceTimeout = 250,
   instantDebounceTimeout = 250
 ) => {
@@ -52,9 +52,10 @@ export const usePageReducer = (
   const columns = useMemo(() => getColumns(page), []);
   const pageInfoColumns = useMemo(() => getPageInfoColumns(page), []);
   const PageActions = useMemo(() => getPageActions(page), []);
+  const initialiser = useMemo(() => getPageInitialiser(page), []);
 
   const [pageState, setPageState] = useState({
-    ...(initializer ? initializer(pageObject) : initialState),
+    ...(initialiser ? initialiser(pageObject) : initialState),
     columns,
     getPageInfoColumns: pageInfoColumns,
     PageActions,
--- a/src/pages/dataTableUtilities/getPageInitialiser.js
+++ b/src/pages/dataTableUtilities/getPageInitialiser.js
@@ -1,0 +1,216 @@
+/**
+ * mSupply Mobile
+ * Sustainable Solutions (NZ) Ltd. 2019
+ */
+
+import { UIDatabase } from '../../database';
+import Settings from '../../settings/MobileAppSettings';
+
+import { newSortDataBy, getAllPrograms } from '../../utilities';
+import { recordKeyExtractor } from './utilities';
+
+const customerInvoiceInitialiser = pageObject => ({
+  pageObject,
+  backingData: pageObject.items,
+  data: pageObject.items.sorted('item.name').slice(),
+  keyExtractor: recordKeyExtractor,
+  dataState: new Map(),
+  searchTerm: '',
+  filterDataKeys: ['item.name'],
+  sortBy: 'itemName',
+  isAscending: true,
+  modalKey: '',
+  modalValue: null,
+  hasSelection: false,
+});
+
+const customerInvoicesInitialiser = () => {
+  const backingData = UIDatabase.objects('CustomerInvoice');
+  return {
+    backingData,
+    data: newSortDataBy(backingData.slice(), 'serialNumber', false),
+    keyExtractor: recordKeyExtractor,
+    dataState: new Map(),
+    searchTerm: '',
+    filterDataKeys: ['otherParty.name'],
+    sortBy: 'serialNumber',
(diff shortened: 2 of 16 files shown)


src/pages/dataTableUtilities/getPageInitialiser.js
```diff
@@ -0,0 +1,213 @@
...
+ * mSupply Mobile
+ * Sustainable Solutions (NZ) Ltd. 2019
+ */
+
+import { UIDatabase } from '../../database';
+import Settings from '../../settings/MobileAppSettings';
+
+import { newSortDataBy, getAllPrograms } from '../../utilities';
+import { recordKeyExtractor } from './utilities';
+
```
joshxg: There are two other components which use `NewStocktakeBatchModal` and `RegimenDataModal` -- They're very weird components that I hate, but they'll be broken. Not 100% sure the best way to handle. I don't think bother moving their initialisers into the file but keeping it usable in `usePageReducer`?

src/pages/dataTableUtilities/getPageInitialiser.js
```diff
@@ -0,0 +1,213 @@
...
+  searchTerm: '',
+  filterDataKeys: ['item.name'],
+  sortBy: 'itemName',
+  isAscending: true,
+  modalKey: '',
+  modalValue: null,
+  hasSelection: false,
+});
+
+const customerInvoicesInitialiser = () => {
```
joshxg: do you think these need doc strings or ott

src/pages/dataTableUtilities/getPageInitialiser.js
```diff
@@ -0,0 +1,213 @@
...
+    supplierInvoice: supplierInvoiceInitialiser,
+    supplierInvoices: supplierInvoicesInitialiser,
+    supplierRequisition: supplierRequisitionInitialiser,
+    supplierRequisitions: supplierRequisitionsInitialiser,
+  };
+
+  return pageToInitialiser[page];
+};
+
+export default getPageInitialiser;
```
joshxg: https://blog.neufund.org/why-we-have-banned-default-exports-and-you-should-do-the-same-d51fdc2cf2ad

joshxg:
I'll also note that after this is merged we can use it in a few pages where `useNavigationFocus` is used. Instead of `refreshData`, we can call an action `reInit`? which will set the state as the initial state again (I think this should be a new issue, not done here)

src/pages/dataTableUtilities/getPageInitialiser.js
wlthomson: Yeah, I'm not a big fan of `default`. I think it's best to address this in a separate issue, for consistencies sake with other methods (e.g. `getColumns`, `getPageInfoColumns` also `default export`, so my though would be to keep `getPageInitialiser` consistent). 

If we want to go with this pattern, will also need to update the eslint config as we currently use air-bnb's `prefer-default-export`: https://github.com/airbnb/javascript#modules--prefer-default-export.

joshxg: you're right! Was going to do it here.. oopsies https://github.com/openmsupply/mobile/issues/845#issuecomment-524062278 (more just bringing this to attention than requiring some sort of change right now :))

joshxg:
I've also had a change of heart 🤣  re my last comment - after playing around I like how the state stays the exact same after navigating !

src/pages/dataTableUtilities/getPageInitialiser.js
```diff
@@ -0,0 +1,213 @@
...
+    filterDataKeys: ['serialNumber', 'otherStoreName.name'],
+    sortBy: 'serialNumber',
+    isAscending: true,
+    modalKey: '',
+    hasSelection: false,
+  };
+};
+
+const getPageInitialiser = page => {
+  const pageToInitialiser = {
```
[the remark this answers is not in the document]
wlthomson: Oops, I meant to finish this off with a closure. I like the idea of modules only exporting functions without any "implicit" closures. See what you think: 5eecb6f87c1eda83782ba1b3f50424d8346860a2.

src/hooks/usePageReducer.js
```diff
@@ -37,25 +38,24 @@ import { debounce } from '../utilities/index';
  *
  * @param {String} page                   routeName for the current page.
  * @param {Object} initialState           Initial state of the reducer
- * @param {Func}   initializer            Function to generate the initial state (optional)
+ * @param {Func}   initialiser            Function to generate the initial state (optional)
```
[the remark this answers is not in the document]
wlthomson: 8dc05cd9142146d6a373e0177c940c551157e1e8.

My opinion if anyone interested: I think that having an initialiser parameter leaves the usage a bit ambiguous, as it implies the caller has to have knowledge of what pages the reducer knows about or always provide a backup initialiser. I think it'd be nice to extract out all initialisation from the caller by essentially restricting it to callers which have hardcoded initialisers (can think of these as the set of pages the function is defined over). It seems counter-intuitive, but I think that adding the optional parameter actually makes the reducer more as opposed to less tightly coupled.

src/pages/dataTableUtilities/getPageInitialiser.js
wlthomson: 7fe5aed1a656a7173223ed4d6169891948aadd4c.

Updated `NewStocktakeBatchModal` to use new `usePageReducer`. I think `RegimenDataModal` should be fixed in its own issue (it still uses `useReducer` with its own actions and reducer).

src/pages/dataTableUtilities/getPageInitialiser.js
wlthomson: 🤣 . 98bc3c103448b3fcef6587e2ff3c600f89eeb246.

src/pages/dataTableUtilities/getPageInitialiser.js
```diff
@@ -0,0 +1,322 @@
...
+
+  return {
+    pageObject: requisition,
+    backingData,
+    data: showAll
+      ? backingData.sorted('item.name').slice()
+      : backingData.filter(item => item.isLessThanThresholdMOS),
+    keyExtractor: recordKeyExtractor,
+    dataState: new Map(),
+    currentFocusedRowKey: null,
```
joshxg: dont need this

src/pages/dataTableUtilities/getPageInitialiser.js
```diff
@@ -0,0 +1,322 @@
...
+  backingData: requisition.items,
+  data: requisition.items.sorted('item.name').slice(),
+  keyExtractor: recordKeyExtractor,
+  dataState: new Map(),
+  searchTerm: '',
+  filterDataKeys: ['item.name', 'item.code'],
+  sortBy: 'itemName',
+  isAscending: true,
+  modalKey: '',
+});
```
joshxg: missing `modalValue`

src/pages/dataTableUtilities/getPageInitialiser.js
```diff
@@ -0,0 +1,322 @@
...
+};
+
+/**
+ * Gets data for initialising a customer requisition page from an associated requisition.
+ *
+ * @param    {Requisition}  requisition
+ * @returns  {object}
+ */
+const customerRequisitionInitialiser = requisition => ({
+  requisition,
```
joshxg: should be `pageObject`

src/pages/dataTableUtilities/getPageInitialiser.js
```diff
@@ -0,0 +1,322 @@
...
+import { recordKeyExtractor } from './utilities';
+
+/**
+ * Gets data for initialising a customer invoice page from an associated transaction.
+ *
+ * @param    {Transaction}  transaction
+ * @returns  {object}
+ */
+const customerInvoiceInitialiser = transaction => ({
+  transaction,
```
joshxg: pageObject

src/hooks/usePageReducer.js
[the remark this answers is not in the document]
joshxg: more tightly coupled with what?

The intention with this was essentially to create a middleware for `useReducer`.. it slowly expanded into injecting props for a bit of DRYness (all of them could just be called in a page). 

I think possibly the best option would be to make a `useThunkReducer` hook - which is the middleware hook. Then, the `usePageReducer` hook could just inject properties. However, I also wonder now if we should just put the field injections straight into the initial state - maybe that would be less confusing heh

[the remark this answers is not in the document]
wlthomson: Oops, just saw this. I agree with that.

My anti-initialisation parameter thoughts clarified:

I like the idea of `usePageReducer` only being usable by pages which are known to the application, as it allows all the state to be abstracted away from the page.  It feels a bit wrong for the caller to be initialising state which it doesn't maintain.

Thinking of DRY, it also makes sense that all pages should have the source of truth for their initialisers in a single place, rather than having a division between pages with and without hardcoded initialisers.

In that context, I think your suggestion of separating out `usePageReducer` into two reducer is a good one, it allows a clearer separation between the more static "page" state and the more dynamic/side-effecty "data" state.

PR #1269, opened 2019-09-18
#1267 Missing StocktakesPage toggle

### LINKED ISSUE #1267: Missing StocktakesPage toggle
## Describe the bug

Missing Old/Current toggle on `StocktakesPage`

### To reproduce
Go to page no see toggle

### Expected behaviour

The toggle should be there

### Version and device info

2.3.6 development

joshxg:
Fixes #1267 

## Change summary

- Adds back the toggle I forgot about

## Testing

See #1043 

Conflicts with #1266 

Note that this messes up the back navigation a little bit. When navigating back `refreshData` is called which will just display everything. This needs to change with changes made in #1266 such that the state is re-init, rather than `refreshData`

### DIFF
--- a/src/pages/dataTableUtilities/actions/tableActions.js
+++ b/src/pages/dataTableUtilities/actions/tableActions.js
@@ -73,6 +73,16 @@ export const showFinalised = () => ({ type: ACTIONS.SHOW_FINALISED });
 export const showNotFinalised = () => ({ type: ACTIONS.SHOW_NOT_FINALISED });
 
 /**
+ * Wrapper around showFinalised/showNotFinalised to toggle between. Determines the
+ * correct action to dispatch.
+ *
+ * @param {Bool} showFinalised Indicator wheter finalised rows are currently displayed.
+ */
+export const toggleShowFinalised = showingFinalised => {
+  if (showingFinalised) return showNotFinalised();
+  return showFinalised();
+};
+/**
  * Shows all items, regardless of current stock on hand, toggles
  * showAll to true and removes the current search filtering. Sort is
  * kept stable.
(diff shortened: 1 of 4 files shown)


src/pages/dataTableUtilities/actions/tableActions.js
```diff
@@ -62,6 +62,26 @@ export const hideOverStocked = () => ({ type: ACTIONS.HIDE_OVER_STOCKED });
  */
 export const hideStockOut = () => ({ type: ACTIONS.HIDE_STOCK_OUT });
 
+/**
+ * Hides all rows which have a status of finalised.
+ */
+export const showFinalised = () => ({ type: ACTIONS.SHOW_FINALISED });
```
mark-prins: :nit this one `showFinalised` and that one `showNotFinalised` are only called by an internal method, so don't need to be exported. There's a really minor overhead in the export, but it also declares intention - and I wondered if those guys aren't supposed to be called externally?

joshxg: I wonder this myself - I landed on "ahhh, why not" .. not sure what the best thing is. No harm in being called directly, but the intention is that the wrapper is called, for ease of use

joshxg: oh - I also hate these names if you have any other suggestions 🤣 - I didn't want to use `past` and `current`, I felt they were less explicit ?

mark-prins: might as well leave them then - if ok to call directly.

I stumbled on the Past and Current names later on (as in, didn't know what they meant at first) - so prefer the `showFinalised` and `showNotFinalised` My only other suggestion would make the naming worse, so let's not go there ( including 'only' to make it clear what is being shown `showOnlyFinalised` or `showFinalisedOnly` - but I like that idea less than what you have ). also, elsewhere in mSupply we have unfinalised rather than not. but not is more logical.

so, in summary. all good as it is. thanks.

PR #1271, opened 2019-09-18
#1270 Checkable cell icon

### LINKED ISSUE #1270: Checkable cell icon
## Describe the bug

The checkable cell icon being used is different from master

### To reproduce

1. Go to a page
2. Click on a checkable cell

### Expected behaviour

I think we should be using `md-remove-circle` for the checked component icon

### Version and device info

data table refactor

### Additional context

N/A

joshxg:
Fixes #1270 

## Change summary

- Adds the correct icon name
- Slight change in colour to unchecked icons to make it more obvious

Not exactly like the old cell, but I like this :)
image

## Testing

See #1043 

N/A


joshxg:
Closed by #1271

PR #1272, opened 2019-09-18
#1268 ItemDetails modal animation

### LINKED ISSUE #1268: ItemDetails exit animation
## Describe the bug

`ItemDetails` modal unmounting does not correctly animate

### To reproduce

1. Go to Stock page
2. Click on a row
4. close it
3. see error

### Expected behaviour

Should show the exit enimation

### Version and device info

Data table refactor 

### Additional context

Just need to fix the conditional logic on an item being passed in props

joshxg:
Fixes #1268 

## Change summary

- Adds a condition on the content of the modal rather than the modal itself so that the animation runs when unmounting
- Also fixed a style oopsies

## Testing

See #1043 

N/A

### DIFF
--- a/src/widgets/modals/ItemDetails.js
+++ b/src/widgets/modals/ItemDetails.js
@@ -29,8 +29,6 @@ import { DARKER_GREY, SUSSOL_ORANGE } from '../../globalStyles';
  * @param {Any}    modalProps Any additional props for the modal component.
  */
 export const ItemDetails = ({ isOpen, item, onClose, ...modalProps }) => {
-  if (!item) return null;
-
   const headers = {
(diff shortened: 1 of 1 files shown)


joshxg:
(if you're interested 🤣 )

- When the 'close' action is invoked, it re-renders the component with a null item and `isOpen = false`
- When a null item is passed, the component just gets removed from the DOM (as per the condition, if !item, return null)
- By moving this condition, the component still re-renders but the modal content is null
- The `Modal` is passed the prop `isOpen = false`, which triggers the close animation

mark-prins:
I thought as much. looking forward to testing now and seeing these things in action 😀

PR #1291, opened 2019-09-19
#1288 Keyboard dismissal on last cell

### LINKED ISSUE #1288: Keyboard should dismiss when submitting the last cell

When editing, the keyboard should dismiss after submitting when it is the last row.

`if (refIndex+1 === numberOfEditableCells) Keyboard.dismiss()` 

in `focusNextCell` in `DataTable.js`

None

## Additional context

It's annoying. Especially with one/few row(s)

joshxg:
Fixes #1288 

## Change summary

- When you get to the end of the list of items in a table, dismiss the keyboard on submitting

## Testing

See #1043 

N/A

### DIFF
--- a/src/widgets/DataTable/DataTable.js
+++ b/src/widgets/DataTable/DataTable.js
@@ -6,7 +6,7 @@
 
 import PropTypes from 'prop-types';
 import React, { useMemo, useRef, useCallback } from 'react';
-import { StyleSheet, VirtualizedList, VirtualizedListPropTypes } from 'react-native';
+import { StyleSheet, VirtualizedList, VirtualizedListPropTypes, Keyboard } from 'react-native';
 import RefContext from './RefContext';
 
 /**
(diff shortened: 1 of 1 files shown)


src/widgets/DataTable/DataTable.js
```diff
@@ -69,11 +69,13 @@ const DataTable = React.memo(({ renderRow, renderHeader, style, data, columns, .
     return newRef;
   };
 
-  // Focuses the next editable cell in the list. Back to the top on last row.
+  // Focuses the next editable cell in the list. On the last row, dismiss the keyboard.
   const focusNextCell = refIndex => {
+    if (refIndex + 1 === numberOfEditableCells) return Keyboard.dismiss();
```
wlthomson: I always get scared by `===` in these contexts because of possible off-by-one errors.

Maybe:

```
const maxIndex = numberOfEditableCells - 1;
if (refIndex >= maxIndex) return Keyboard.dismiss();
```

joshxg: 0ae68dcedb4cbaafc17fe7ec415ad690259577ff

I've actually thought about this a bit - I prefer the off-by-one error actually get thrown rather than silently ignored, when values should be specific. Used variables though

PR #1303, opened 2019-09-21
#1299 Memoize components

### LINKED ISSUE #1299: Components should be Memoized
## The suggestion

Quite a few components should be memoized, possibly with a handler for `propsEqual` on a minimum set of props which cause re-renders.

Examples
- Buttons,
- SearchBar
- Modals

etc

**Why?**
Faster and don't have to ensure all props are memoized if they aren't in the comparison 

Not do anything

## Additional context

N/a

joshxg:
Fixes #1299 

## Change summary

- Memoized `getActions` command in each page. Reduces some running of `DataTableRow` which was resulting in nothing actually changing
- Changed how `PageInfo` was being used. Was re-rendering after any page render, now only on changing a value
- Memoized `SearchBar`, only re-render when value changes
- Memoized `BottomConfirmModal` - re-render when opening/closing only.
- Memoized `DataTableModal` - re-render when onpening/closing only.
- Memoized `Icon`s a little bit 

## Testing

See #1043 

N/A


src/pages/CustomerInvoicePage.js
```diff
@@ -76,17 +76,13 @@ export const CustomerInvoicePage = ({ transaction, runWithLoadingIndicator, rout
...
-  const renderPageInfo = useCallback(
-    () => (
-      <PageInfo
-        columns={getPageInfoColumns(pageObject, dispatch, PageActions)}
-        isEditingDisabled={isFinalised}
-      />
-    ),
-    [comment, theirRef, isFinalised]
-  );
+  const pageInfoColumns = useCallback(getPageInfoColumns(pageObject, dispatch, PageActions), [
```
wlthomson: Wondering if it might be worth adding some comments just indicating what method does and the purpose of the memoisation, e.g. 

```
 // Initialise column getter. Memoised to only run component mount.
 const pageInfoColumns = useCallback(getPageInfoColumns(pageObject, dispatch, PageActions), [
    comment,
    theirRef,
    isFinalised,
  ]);
```

src/pages/CustomerInvoicePage.js
```diff
@@ -155,7 +152,7 @@ export const CustomerInvoicePage = ({ transaction, runWithLoadingIndicator, rout
     <DataTablePageView>
       <View style={newPageTopSectionContainer}>
         <View style={newPageTopLeftSectionContainer}>
-          {renderPageInfo()}
+          <PageInfo columns={pageInfoColumns} isEditingDisabled={isFinalised} />
```
wlthomson: 👍 .

src/widgets/DataTable/DataTable.js
```diff
@@ -96,6 +96,11 @@ const DataTable = React.memo(({ renderRow, renderHeader, style, data, columns, .
     [numberOfEditableCells]
   );
 
+  const renderItem = useCallback(
```
wlthomson: Might be worth indicating when the importance of this memoisation in terms of the render cycle.

src/widgets/SearchBar.js
```diff
@@ -18,7 +18,10 @@ import { debounce } from '../utilities/index';
  * with a magnifying glass icon and clear button.
  *
  * Debounces input by the user, such that the onChangeText callback
- * is only invoked once every `debounceTimeout`
+ * is only invoked once every `debounceTimeout`.
+ *
+ * NOTE: This component is exported MEMOIZED. See below for propsAreEqual
```
wlthomson: Thinking back to when I was new to hooks, might be more helpful here to explain the why rather than the what, e.g.:

```
This component only re-renders when the |value| prop is updated.
```

src/widgets/modals/DataTablePageModal.js
```diff
@@ -31,6 +32,9 @@ import NewSocktakeBatchModal from './NewStocktakeBatchModal';
 /**
  * Wrapper around ModalContainer, containing common modals used in various
  * DataTable pages.
+ *
+ * NOTE: Exported component is MEMOIZED - see below for propsAreEqual implementation.
```
wlthomson: See above comment 🆙 .

src/widgets/SearchBar.js
joshxg: It's only here to try to minimize a mistake someone in the future might have wondering why their component might not be rendering as per normal. The comment you suggest is on the `propsAreEqual` function itself

src/pages/CustomerInvoicePage.js
joshxg: I think this sort of comment will get outdated quite quickly so i'm not 100% sold on the need? essentially it is: 

```
// This component is memoized and updated when this occurs
memoizeThisFunction(someFunction, [when this occurs])
```

wlthomson: Arguably all comments will get out of date, but I still think they're worth doing.

If you're introducing features to a shared codebase, I think it makes sense to add minimal documentation indicating the fundamental life cycle of a component.

I wouldn't suggest commenting all `useCallback(args, ...)` because I think you're right that that could be a bit too much. But, I think `useCallback(args, [])` has a unique enough use-case (i.e. run only on mount) such that adding a comment could significantly reduce the time it would take a newbie to understand the code.

I agree that commenting too much isn't great, because of maintenance, and that ideally you want self-documenting code, but I'd still lean towards documenting key life cycle events of a component (i.e. ones which have clear analogues to life cycle methods).

src/widgets/SearchBar.js
wlthomson: Sure, but it seems like you could just do the one comment? You could prob get away with it just in the docstring as you'd think `propsAreEqual` is fairly self-documenting in this instance.

PR #1307, opened 2019-09-23
#1065 remove old code

### LINKED ISSUE #1065: TypeError in MainActivity
## Error in mSupply Mobile

null is not an object (evaluating 'this.itemBatch.totalQuantity')

[View on Bugsnag](https://app.bugsnag.com/sustainable-solutions-nz-ltd/msupply-mobile/errors/5d31e78d248f24001a1f59d4?event_id=5d31e78d004aeabb08bc0000&i=gh&m=ci)

## Stacktrace

    /Users/joshuagriffin/projects/mobile/src/database/DataTypes/TransactionBatch.js:142 - value
    /Users/joshuagriffin/projects/mobile/src/database/DataTypes/TransactionItem.js:251 - value
    /Users/joshuagriffin/projects/mobile/src/database/DataTypes/TransactionItem.js:200 - value
    /Users/joshuagriffin/projects/mobile/src/database/DataTypes/TransactionItem.js:178 - value
    /Users/joshuagriffin/projects/mobile/src/pages/CustomerInvoicePage.js:106 - 
    /Users/joshuagriffin/projects/mobile/src/database/UIDatabase.js:157 - readFileGeneric
    /Users/joshuagriffin/projects/mobile/src/pages/CustomerInvoicePage.js:105 - onEndEditing

[View full stacktrace](https://app.bugsnag.com/sustainable-solutions-nz-ltd/msupply-mobile/errors/5d31e78d248f24001a1f59d4?event_id=5d31e78d004aeabb08bc0000&i=gh&m=ci)

### LINKED ISSUE #1165: Remove unused code after DataTable refactor

Code is being replaced with new code in #1043 - the new code is being tagged with `NewX...`, so that existing code will continue to work while refatoring. 

- Remove all old code.
- All new code prefixed with `New...` should be renamed.

Just doing this from the start

#### Current files:
- `newDataTableStyles`
- `newSortDataBy`
- `NewExpiryTextInput`

Potentiaally could be done when vaccines is merged..eventually? - if so, this issue should be marking the old versions as deprecated

joshxg:
Fixes #1165 

## Change summary

- `newDataTableStyles` -> `dataTableStyles` -- removed OLD `dataTableStyles`
- `newPageStyles` -> `pageStyles` -- removed OLD `pageStyles`
- `NewExpiryTextInput` -> `ExpiryTextInput` -- removed OLD `ExpiryTextInput`
- `NewStocktakeBatchModal` -> `StocktakeBatchModal` -- removed OLD `StocktakeBatchModal`
- `NewRegimenDataModal` -> `RegimenDataModal` -- removed OLD `RegimenDataModal`
- `NewExpiryTextInput` -> `ExpiryTextInput` -- removed OLD `ExpiryTextInput` 
- Removed `SelectModal` - not used
- Removed some random old tests - not used
- Removed `react-native-data-table` dependency
- Removed `react-native-generic-data-table-page-for-react-native-table-apps` dependency
- Removed code from `RealmExplorer` page so I could remove dependencies

## Testing

See #1043 

This is just renaming things and deleting the old stuff. Can just do stuff in the app and see everything working fine then would be all good to approve, imo


src/widgets/modals/StocktakeBatchModal.js
```diff
@@ -1,413 +1,175 @@
...
+    dataState,
+    sortBy,
+    isAscending,
+    modalKey,
+    modalValue,
+    keyExtractor,
+    PageActions,
+    columns,
+    getPageInfoColumns,
+  } = state;
```
wlthomson: This throws an error which can be traced to `pageObject` not being defined. Not sure, but I don't think it's related to this PR, just seems to have been revealed by migrating this modal to the reducer.

Fixed by adding `pageObject` to the initialisation logic in `usePageReducer`:

```
 const [pageState, setPageState] = useState({
    pageObject,
    ...(pageInitialiser ? pageInitialiser(pageObject) : initialState),
    columns,
    getPageInfoColumns: pageInfoColumns,
    PageActions,
  });
```

Not sure how this got introduced, maybe when I updated the initialisation logic?

src/widgets/modals/StocktakeBatchModal.js
```diff
@@ -1,413 +1,175 @@
...
-            style={localStyles.reasonCell}
-          >
-            {editable && <Icon name="external-link" size={14} color={SUSSOL_ORANGE} />}
-            <Text style={{ width: '80%' }} numberOfLines={1} ellipsizeMode="tail">
-              {stocktakeBatch.option ? stocktakeBatch.option.title : programStrings.not_applicable}
-            </Text>
-          </TouchableOpacity>
-        );
-      }
+        return PageActions.editStocktakeBatchName;
```
wlthomson: Nit:

Could turn this into a mapper quite nicely, e.g.:

```
return pageActions[mapColumnAction[colKey]]
```

src/widgets/modals/StocktakeBatchModal.js
```diff
@@ -1,413 +1,175 @@
...
+      />
+    ),
+    [sortBy, isAscending]
+  );
+
+  const PageButtons = () => (
+    <PageButton
+      text={buttonStrings.add_batch}
+      onPress={() => dispatch(PageActions.addStocktakeBatch())}
+      isDisabled={stocktakeItem.stocktake.isFinalised}
```
wlthomson: Even though it's in the schema, might want to add a guard here to prevent undefined object error.

src/pages/CustomerRequisitionPage.js
```diff
@@ -168,18 +168,18 @@ export const CustomerRequisitionPage = ({ requisition, runWithLoadingIndicator,
...
+    pageTopLeftSectionContainer,
+    pageTopRightSectionContainer,
+  } = globalStyles;
   return (
     <DataTablePageView>
-      <View style={newPageTopSectionContainer}>
-        <View style={newPageTopLeftSectionContainer}>
+      <View style={pageTopSectionContainer}>
+        <View style={pageTopLeftSectionContainer}>
           {renderPageInfo()}
```
wlthomson: Nit:

Returning method calls inside JSX looks a bit weird to me. Might be nicer to assign it to a component variable:

```
const PageInfo = renderPageInfo()
return (
  <DataTablePageView> 
  ...
   <View style=PageTopLeftSectionContainer>
     <PageInfo />
   </View>
  ...
...
```
```

File: src/pages/StocktakeEditPage.js [hunk 5]
```diff
@@ -178,18 +178,18 @@ export const StocktakeEditPage = ({
       />
     );
 
-    return <View style={newPageTopRightSectionContainer}>{program ? null : ManageStocktake}</View>;
+    return <View style={pageTopRightSectionContainer}>{program ? null : ManageStocktake}</View>;
```
wlthomson: Not entirely convinced of this myself, but going to suggest it anyway:

```
const ManageStocktakeOrNull = program ? null : ManageStocktake;
return (
  <View style={...}>
    <ManageStocktakeOrNull/>
  </View>;
```

src/widgets/DataTable/DataTableHeaderRow.js
```diff
@@ -22,37 +22,40 @@ import { newDataTableStyles } from '../../globalStyles';
...
-  />
-));
+const DataTableHeaderRow = React.memo(({ columns, sortBy, isAscending, dispatch, sortAction }) => {
+  const { headerRow, headerCells, cellText } = dataTableStyles;
+  return (
+    <HeaderRow
+      style={headerRow}
+      renderCells={() =>
+        columns.map(({ key, title, sortable, width, alignText }, index) => {
+          const sortDirection = isAscending ? 'ASC' : 'DESC';
```
wlthomson: The usual nit: can these be made into constants?

src/widgets/DataTable/DataTableHeaderRow.js
```diff
@@ -22,37 +22,40 @@ import { newDataTableStyles } from '../../globalStyles';
...
+          const directionForThisColumn = key === sortBy ? sortDirection : null;
+          const isLastCell = index === columns.length - 1;
```
wlthomson: Here I do think you want to use `>=` rather than `===`. Acknowledging what you said about preferring errors last time, I think here it's not guaranteed that you'll throw an error if you're off by one (e.g. if somehow you skip the last cell and ran over), so it's probably a good idea to play defensively.

joshxg: we're mapping over columns, if one was skipped and run over, we'd be in a different array and probably be seg faulting or something..!

src/widgets/DataTable/DataTableHeaderRow.js
joshxg: this PR is for moving files, not making code changes

src/pages/StocktakeEditPage.js
joshxg: same as above

src/pages/CustomerRequisitionPage.js
joshxg: Unfortuantely creating components like this will actually start to re-mount components, rather than re-render them. Sucks cause it is way nicer

src/widgets/modals/StocktakeBatchModal.js
joshxg: I think we need to trust our databases constraints, otherwise whats the point? Although I agree defensive code is great, this seems a little far

src/widgets/modals/StocktakeBatchModal.js
joshxg: not 100% sure but I don't think this will work for every case - although I hate having `getActions` in every page and would love a nice solution

src/widgets/modals/StocktakeBatchModal.js
joshxg: sorry! I should have mentioned this gets fixed here https://github.com/openmsupply/mobile/pull/1305/files

src/widgets/DataTable/DataTable.js
```diff
@@ -6,7 +6,7 @@
 
 import PropTypes from 'prop-types';
 import React, { useMemo, useRef, useCallback } from 'react';
-import { StyleSheet, VirtualizedList, VirtualizedListPropTypes } from 'react-native';
+import { StyleSheet, VirtualizedList, VirtualizedListPropTypes, Keyboard } from 'react-native';
```
mark-prins: have these changes snuck in from another branch?
I remember reviewing a PR with this in it last week

src/globalStyles/dataTableStyles.js
```diff
@@ -23,56 +21,201 @@ export const dataTableColors = {
 };
 
 export const dataTableStyles = {
-  text: {
+  cellContainer: {
```
mark-prins: if this PR is just removing old code and moving stuff around.. why the new styles?

src/widgets/DataTable/DataTable.js
joshxg: hmm yeah https://github.com/openmsupply/mobile/pull/1291/files -- I'm not sure why that is coming up in the diff. I will merge this after the other PR's and ensure I have based correctly

src/globalStyles/dataTableStyles.js
joshxg: sorry maybe I shouldv'e been more clear. I created a new file `newDataTableStyles` and made new styles in there for the new data table to ensure everything was still working throughout the refactor. I then shifted the code from `newDataTableStyles` to `dataTableStyles` and removed `newDataTableStyles` - maybe this was a poor decision

mark-prins: no - it's all good. just checking that nothing has snuck in that might cause a conflict later on

src/widgets/modals/StocktakeBatchModal.js
wlthomson: > I think we need to trust our databases constraints, otherwise whats the point? Although I agree defensive code is great, this seems a little far

I'm a big believer in defensive programming in the strict sense (i.e. defend against everything, even if you don't think it could ever happen).

I'd say the point of a schema is to tell the DB what constraints to enforce internally and to give a contract to your code to implement. JS isn't typed, so there's no compile-time constraints on `stocktakeItem`. IMO you trust the DB to enforce it's constraints internally, but you shouldn't assume that also applied to user code, e.g.:


```
  const PageButtons = () => {
    stocktakeItem = undefined;
    return (
      <PageButton
        text={buttonStrings.add_batch}
        onPress={() => dispatch(PageActions.addStocktakeBatch())}
        isDisabled={stocktakeItem.stocktake.isFinalised}
      />
    );
  };
```

Doing this throws an `undefined is not an object` error. Sure, it's a dumb thing to do, but point being that the DB doesn't do anything to prevent this happening.

There's also the fact that there's no guarantee that the schema will remain the same over time, e.g. what if someone renamed `stocktake` to `linkedStocktake` and forgot to update the code?

You could argue that throwing an error is the best thing to do when user code breaks a DB constraint, but for a user, the RN error (red screen of death) is pretty horrific, better to handle it gracefully with defensive code where possible IMO.

src/pages/CustomerRequisitionPage.js
wlthomson: > Unfortuantely creating components like this will actually start to re-mount components, rather than re-render them. Sucks cause it is way nicer

Even if you memoise?

src/widgets/DataTable/DataTableHeaderRow.js
wlthomson: Similar to the defensive programming point above. There's no harm in not protecting against it, so IMO it makes sense to. Could be a hangover from C++ days, but in general it seems like a bad idea to rely on assumptions (IMO most bugs are the result of unforeseen results of assumptions).

src/widgets/DataTable/DataTableHeaderRow.js
wlthomson: Doesn't have to be done here? I think a good practice (what we usually do?) is to suggest making a new issue for changes that don't fit within the remit of a PR (or dismiss if you don't think worth doing).

src/widgets/DataTable/DataTableHeaderRow.js
wlthomson: defensive

joshxg:
Since this is moving a lot of files this is going to cause any new PRs to be heavily conflicted. Was meant to be just shifting of files - happy to make new issues for anything that has come up but would like to merge this and not have to branch off it ?

src/widgets/DataTable/DataTableHeaderRow.js
joshxg: sure no problem - I will probably not make a new PR and just sneak it into some other PR, or you can !

src/widgets/modals/StocktakeBatchModal.js
wlthomson: My point is you can have some guarantees on the schema, but you can't guarantee that a field will be present on a js object, even if it instantiates a realm class. So, I wouldn't suggest defensively coding against everything, just against referencing fields on an object without a guard. I think the whole DB aspect in this instance is a red-herring as it seems like Realm doesn't enforce as much as it maybe should?

wlthomson: Made a mistake above, turns out Realm does enforce more than I thought. Realm prevents `Stocktake.stocktakeItem = undefined` (can't write outside a transaction, and even in a transaction it seems to ignore it?), however you can still do `Stocktake = undefined`, which is pretty much a variation on the same deal.

I think we agree it's a balance, I just think making any assumptions about "this won't be a bug" is dangerous and the small investment up-front is worth it in terms of being able to guarantee stable software to the client.

wlthomson:
> Since this is moving a lot of files this is going to cause any new PRs to be heavily conflicted. Was meant to be just shifting of files - happy to make new issues for anything that has come up but would like to merge this and not have to branch off it ?

Merge it, choo choo.

PR #1308, opened 2019-09-23
#1263 migrate RealmExplorer

### LINKED ISSUE #1263: RealmExplorer page
## The suggestion

`RealmExplorer` page is using the old `DataTable` and should be refactored with all the other pages. This is a page which can be accessed in admin mode to browse the currently opened relam.

**Why?**
Consistency in the app, faster and that would mean being able to remove a large number of old dependencies that aren't used anywhere else.

Not bothering

## Additional context

Not sure if should bother? @mark-prins @Chris-Petty @wlthomson @andreievg

wlthomson:
Fixes #1263.

## Change summary

Refactor of `RealmExplorer` to remove all dependencies on old data table. 

- Replaced all data table components with RN components (exception of `SearchBar`, see below).
- Updated `RealmExplorer` to pure component using hooks.
- Refactored data parsing logic.

Seems a bit slow, probably worth doing some memoisation to prevent unnecessary re-renders? Feedback from those more experienced with `useCallback` etc would be appreciated.

## Testing

- [ ] All tables display correctly.
- [ ] Filter strings work.
- [ ] Table state is reset when the filter string is cleared.
- [ ] No errors thrown when the search string is cleared.

- Replace `FlatList` and `VirtualizedList` components with `DataTable`. 
  - @joshxg Is this worth doing? AFAIK, `DataTable` consists mostly of a wrapper around `VirtualizedList`, so assuming this would be relatively straightforward?
- Replace `SearchBar` (`react-native-ui-components`) with `SearchBar` (`widgets/SearchBar.js`).
- Use proper Sussol styles.


src/pages/RealmExplorer.js
```diff
@@ -3,201 +3,214 @@
...
+      const fields = Object.entries(properties)
+        .map(([field, type]) => ({
+          [field]: parseType(type),
+        }))
+        .reduce((acc, field) => ({ ...acc, ...field }), {});
+      return { [name]: fields };
+    })
+    .reduce((acc, object) => ({ ...acc, ...object }));
+
+const OBJECT_TYPES = getObjectTypes(schema);
```
joshxg: a little more differentiation between realm object types and primitives would be appreciated took a sec to realize this was the realm object types and not related to the primitive mapping

src/pages/RealmExplorer.js
```diff
@@ -3,201 +3,214 @@
...
+ * @prop   {UIDatabase}     database       App wide database.
+ * @state  {string}         objectString   Current database object.
+ * @state  {string}         searchString   Current search string. Used to update current object.
+ * @state  {string}         filterString   Current filter string. Used to update filtered data.
+ * @state  {Realm.Results}  objectData     Reference to current database object results. Used to
+ *                                         roll back filter state when filter is reset.
+ * @state  {Realm.Results}  filteredData   Reference to current database object results after filter
+ *                                         has been applied. Displayed to the user.
+ */
+export const RealmExplorer = ({ database }) => {
```
joshxg: just import `UIDatabase` so we can remove passing of database object through props

src/pages/RealmExplorer.js
```diff
@@ -3,201 +3,214 @@
...
-            textStyle={globalStyles.text}
-            width={1}
-            text={key}
-          />
-        );
-      }
-    }
-    return <Header style={globalStyles.header}>{headerCells}</Header>;
-  }
+  const onSearchChange = searchString => {
```
joshxg: the way this is done there's two renders per character type. (when you use internal `SearchBar` it will auto debounce) - but will still cause two renders per update

src/pages/RealmExplorer.js
```diff
@@ -3,201 +3,214 @@
...
+      newState.filteredData = newState.objectData.filtered(filterString);
+    } catch (err) {
+      newState.filteredData = updateObject ? newState.objectData : filteredData;
     }
   }
 
-  onSearchChange(searchTerm) {
-    const { database } = this.props;
-    const { dataSource } = this.state;
+  return newState;
```
joshxg: SRP - personally I think this is doing to much

```
const filterData = (newFilterString, state) => {
    const { objectData } = state;
    try...
    const newData = objectData.filtered(filterString)

    catch..
    const newData = objectData

    return {...state, objectData: newData }
}
```

and a similar method for `searchString`


Then, in the component

```
const onFilterData = useCallback((value) => setState((state) => filterData(value)), []);
```

can pass a function to `setState` and it will inject the current `state`

src/pages/RealmExplorer.js
```diff
@@ -3,201 +3,214 @@
...
+  return (
+    <View style={[globalStyles.container]}>
+      <SearchBar onChange={onSearchChange} placeholder="Table name" />
+      <SearchBar onChange={onFilterChange} placeholder="Filter string" />
+      <FlatList
+        data={headerData}
+        keyExtractor={headerRowKeyExtractor}
+        renderItem={renderHeaderRow}
+      />
+      <VirtualizedList
```
joshxg: flat list is a wrapper around virtualized list - they do the same thing, except `FlatList` has a simpler API and is simpler to use in general. Just use one or the other https://facebook.github.io/react-native/docs/flatlist.html#listheadercomponent

src/pages/RealmExplorer.js
```diff
@@ -3,201 +3,214 @@
...
+        data={data}
+        getItem={(d, i) => d[i]}
+        getItemCount={d => d.length}
```
joshxg: general rule of never use annonymous functions serves well - unless it's some inconsequential tiny component, which this isn't.

Also the amount of time saved typing `d` is huge compared to something a little more descriptive like `data` but possibly not worth it

src/pages/RealmExplorer.js
```diff
@@ -3,201 +3,214 @@
...
+    return <View style={styles.row}>{cells}</View>;
+  };
+
+  const headerData = [objectFields];
+  const data = state.filteredData;
+
+  return (
+    <View style={[globalStyles.container]}>
+      <SearchBar onChange={onSearchChange} placeholder="Table name" />
+      <SearchBar onChange={onFilterChange} placeholder="Filter string" />
```
joshxg: I believe you just need to use `value={filterString}` and change `onChange` to `onChangeText` to use the new `SearchBar`

src/pages/RealmExplorer.js
```diff
@@ -3,201 +3,214 @@
...
+        <Text style={styles.cellText}>{columnKey}</Text>
       </View>
-    );
-  }
-}
+    ));
+    return <View style={styles.row}>{cells}</View>;
+  };
+
+  const renderRow = row => {
```
joshxg: I probably wouldn't bother but this is a good candidate for `useCallback` will memoize and existing rows when you filter

src/pages/RealmExplorer.js
```diff
@@ -3,201 +3,214 @@
...
-          style={globalStyles.container}
-          listViewStyle={globalStyles.container}
-          dataSource={dataSource}
-          renderRow={this.renderRow}
-          renderHeader={this.renderHeader}
-        />
+  const objectFields = OBJECT_FIELDS[state.objectString];
+
+  const headerRowKeyExtractor = (_, index) => index.toString();
+  const rowKeyExtractor = ({ id }) => id;
```
joshxg: these are types of things that will create a new reference every render and trigger a re-render on pretty much everything - if you define outside of scope of the component don't need to memoize

src/pages/RealmExplorer.js
```diff
@@ -3,201 +3,214 @@
...
+
+  const renderHeaderRow = headerRow => {
```
joshxg: `useCallback` candidate

src/pages/RealmExplorer.js
```diff
@@ -3,201 +3,214 @@
...
+    return <View style={styles.row}>{cells}</View>;
+  };
+
+  const renderRow = row => {
+    const { item } = row;
+    const cells = Object.entries(objectFields).map(([columnKey, columnType]) => {
+      const cell = item[columnKey];
+      const cellValue = parseCell(cell, columnType);
+      return (
+        <View style={styles.cell}>
```
joshxg: need to give a key here

src/pages/RealmExplorer.js
wlthomson: That's much nicer. Just to clarify, did you mean to put `[state.filterString]` as the second arg of `useCallback`? Also think that `filterData(value)` should be `filterData(value, state)`?

joshxg: Note: `setState` here would be index 1 return from your `useState` call cant remember what it was hehe

if you pass a function to the `setState` return value, the current state gets injected for you https://reactjs.org/docs/hooks-reference.html#usestate

so:

- Function being passed to your `setState`: `oldState => filterData(value)`
- Function being passed into `useCallback`: `value => setState(oldState => filterData(value))`

The dependent here, is only `setState` - but I *think* you can use the initial value of this from the closure, not 100% but if  so, it can be fully memoized. 

You could technically do some funny things with functions to keep a constant reference, with a closure to `setState`, but more understandable to just memoize I think.

src/pages/RealmExplorer.js
```diff
@@ -3,201 +3,214 @@
...
-  'SyncOut',
-  'Stocktake',
-  'StocktakeItem',
-  'StocktakeBatch',
-  'User',
-];
+const typeMapper = new Map(
+  Object.entries(REALM_TYPES)
+    .map(([type, realmTypes]) => realmTypes.map(realmType => [realmType, type]))
+    .flat()
```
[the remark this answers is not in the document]
wlthomson: Checked against `v10.15` (version specified in `.nvmrc`) and no issues. If issues pop up, definitely can remove it, but as we are already babelling, it's probably OK?

src/pages/RealmExplorer.js
wlthomson: c9d5f59d202b2c892b933fefee9017dad994524e.

src/pages/RealmExplorer.js
wlthomson: Agree 100%. a7496e193ca5ad1349e01923bbc2387fba3c02d3.

src/pages/RealmExplorer.js
wlthomson: Went for `VirtualizedList`, just for consistency with `DataTable`. `FlatList` header replaced with `ListHeaderComponent` prop. b56a04041e336171188728ccea36b0be8db36eaf.

src/pages/RealmExplorer.js
wlthomson: Good point, I think initially the choice was temporary to stop eslint from shouting. 39d6839a865cd541b1fd8f28e9675cd8355a55ba.

src/pages/RealmExplorer.js
wlthomson: Good point. 589030767bf1963dcbb6c3ef10a11973f2e48ec0.

src/pages/RealmExplorer.js
wlthomson: Nice. debe6b09d03cf8a21b0f19b5db721bb1fc358dd9.

src/pages/RealmExplorer.js
wlthomson: Very good point. 43364eb68f575859a3987b0a25620e79be34c120.

src/pages/RealmExplorer.js
wlthomson: d5156601ed255ae7a0bdbcd7992104d21a7b5424.

src/pages/RealmExplorer.js
```diff
@@ -3,201 +3,278 @@
...
-  database: PropTypes.object.isRequired,
+  const RealmSearchBar = useCallback(getSearchBarRenderer(onSearchChange), [onSearchChange]);
+  const RealmFilterBar = useCallback(getFilterBarRenderer(onFilterChange), [onFilterChange]);
+  const RealmTable = useCallback(getTableRenderer(realmObjectFields), [realmObjectFields]);
+
+  return (
+    <View style={[globalStyles.container]}>
+      <RealmSearchBar searchString={searchString} />
+      <RealmFilterBar filterString={filterString} />
+      <RealmTable data={filteredData} />
```
joshxg: my opinion is that this looks pretty but makes it harder to actually know whats going on at a glance - takes more cognitive load to actually understand the components and requires more thought into memoization and other performance enhancing techniques. I like to use this style when there are conditions on what and when to render something, but not to wrap to many basic components in other components 

```
<RealmSearchBar searchString={searchString}/>
```

with 2? wrapping functions isn't easier to read than:

```
<SearchBar value={filterString} onChangeText={onFilterChange} placeholder="Filter string" />
```

And provides no additional benefit that I can see given that `SearchBar` is already memoized and re-renders only when `value` changes

joshxg:
Need to resolve conflicts and I will merge

wlthomson:
> Need to resolve conflicts and I will merge

Resolved 👍 .

PR #1331, opened 2019-10-06
#1327 OnPress Crash

### LINKED ISSUE #1327: Clicking on a customer requisition row throws an error
## Describe the bug

When a customer requisition row is clicked, the following error is thrown (sometimes it takes a few seconds for it to appear):

#1327-customer-requisition-row-error

Note: this bug seems to sometimes also occur when the "Use Suggested Quantities" button is pressed.

### To reproduce

1. Go to 'Customer Requisitions`.
2. Click on any row.
3. Wait a few seconds...
4. See error.

### Expected behaviour

No error is thrown.

### Version and device info

- App version: 2.3.7.
- Tablet model: TB-X304L (replicated with emulator).
- OS version: 8.1.0.

### Additional context

See https://github.com/openmsupply/mobile/issues/1244#issuecomment-537723537.

joshxg:
Fixes #1327, #1340.

## Change summary

- Correctly overrides `onPress` for `TouchableNoFeedbacl` to be a no op onPress

## Testing

- [ ] Clicking a row in for example `CustomerRequisitionPage` doesn't throw an error

Not getting a crash on pressing `Use Suggested Quantities` button

### DIFF
--- a/src/widgets/DataTable/TouchableNoFeedback.js
+++ b/src/widgets/DataTable/TouchableNoFeedback.js
@@ -17,7 +17,7 @@ import PropTypes from 'prop-types';
 const onPressNoOp = () => {};
 
 const TouchableNoFeedback = ({ children, style, ...touchableProps }) => (
-  <TouchableWithoutFeedback onPress={onPressNoOp} {...touchableProps}>
+  <TouchableWithoutFeedback {...touchableProps} onPress={onPressNoOp}>
     <View style={style}>{children}</View>
   </TouchableWithoutFeedback>
 );


joshxg:
note that this also fixes the problem with being able to select finalised rows and delete them

src/widgets/DataTable/TouchableNoFeedback.js
```diff
@@ -17,7 +17,7 @@ import PropTypes from 'prop-types';
 const onPressNoOp = () => {};
 
 const TouchableNoFeedback = ({ children, style, ...touchableProps }) => (
-  <TouchableWithoutFeedback onPress={onPressNoOp} {...touchableProps}>
+  <TouchableWithoutFeedback {...touchableProps} onPress={onPressNoOp}>
```
wlthomson: What does this fix involve? Were the unpacked parameters overriding `onPress`? If so, might be worth making the intended override explicit, maybe something like:

```
const TouchableNoFeedback = ({ children, style, ...touchableProps }) => {
  const touchableWithoutFeedbackProps = { ...touchableProps, onPress: onPressNoOp };
  return <TouchableWithoutFeedback {...touchableWithoutFeedbackProps}>;
}
```

joshxg: yeah, overriding.

I think your suggestion will cause a new object and unneccessary re-renders, I can try it out

wlthomson: No worries. Will merge.

Issue #1339, opened 2019-10-10
Stocktake not finalising without focussing on quantity


wlthomson:
## Describe the bug

When a new stocktake is created via `StocktakeManagePage`, the actual quantities of the chosen items are automatically populated to the snapshot quantities. When attempting to finalise the resulting stocktake, the user is prompted with the following error:

![Screen Shot 2019-10-11 at 10 26 07 AM](https://user-images.githubusercontent.com/43223637/66607842-9df21d00-ec11-11e9-9534-32c82d52166d.png)

### To reproduce

1. Go to "Stocktakes".
2. Click on "New Stocktake".
3. Select some items and create a new stocktake.
4. Attempt to finalise the new stocktake.
5. See error.

### Expected behaviour

Should be able to finalise a stocktake with the default actual quantities?

### Version and device info

- App version: 2.3.7.
- Tablet model: TB-X304L (replicated with emulator).
- OS version: 8.1.0.

### Additional context

N/A.

Chris-Petty:
Snapshot quantities shouldn't automatically fill counted quantity. If #1343 has done that, then that change needs to be undone.

joshxg:
Closed by #1346

PR #1346, opened 2019-10-11
#1343 stocktake updates

### LINKED ISSUE #1343: StocktakeEditPage updates
## Describe the bug

Few bugs in `StocktakeEditPage` highlighted in testing: https://github.com/openmsupply/mobile/issues/1242#issuecomment-540822222

- name `PageInfo` is disabled
- Items are set to the snapshot quantity automatically
- Placeholder does not show

additional:

- can't remove items  from the stocktake
- Bottom confirm modal goes away when nothing is selected, but should always stay

### To reproduce

1. 
- try to edit a name in a stocktake
- see error

2.
- Add an item to a stocktake
- see error

3. 
- try to remove an item from a stocktake
- see error

### Expected behaviour

The opposite of the bugs

### Version and device info

`DataTable` feature branch in development

### Additional context

N/A

joshxg:
Fixes #1343 

## Change summary

- Correctly using placeholders and not counted state for rows in a stocktake
- Made the bottom modal thing in `StocktakeManagePage`  always open
- Corrected logic for remove items from a `Stocktake`
- Corrected logic for the keyboard type for editable cells

## Testing

- [ ] When editing a cell that should be numeric, the numeric keyboard is opened
- [ ] When adding an item to a stocktake, initial state is to be NOT COUNTED
- [ ] When changing a value in a stocktake row, the state is the value entered
- [ ] Can remove items from a `Stocktake`
- [ ] When finalising a `Stocktake`, `Not Counted` rows are pruned

`DataTableRow` getting pretty messy at this point. Could refactor soon!

### DIFF
--- a/src/widgets/DataTable/DataTableRow.js
+++ b/src/widgets/DataTable/DataTableRow.js
@@ -29,7 +29,8 @@ import TextInputCell from './TextInputCell';
 
 import { formatStatus } from '../../utilities';
 
-import { COLUMN_TYPES } from '../../pages/dataTableUtilities';
+import { COLUMN_TYPES, COLUMN_NAMES } from '../../pages/dataTableUtilities';
+import { tableStrings } from '../../localization/index';
 
 /**
  * Wrapper component for a mSupply DataTable page row.
@@ -88,7 +89,15 @@ const DataTableRow = React.memo(
 
           switch (type) {
             case COLUMN_TYPES.EDITABLE_STRING:
-            case COLUMN_TYPES.EDITABLE_NUMERIC:
+            case COLUMN_TYPES.EDITABLE_NUMERIC: {
+              // Special condition for stocktake counted total quantity cells.
+              // Use the placeholder 'Not counted' when a stocktake item or batch
+              // has not been counted yet.
+              let placeholder = '';
+              if (columnKey === COLUMN_NAMES.COUNTED_TOTAL_QUANTITY) {
+                placeholder = rowData.hasBeenCounted ? '' : tableStrings.not_counted;
+              }
+
               return (
                 <TextInputCell
                   key={columnKey}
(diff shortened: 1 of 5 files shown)


joshxg:
I got a little confused - this also basically contains #1339

src/widgets/DataTable/DataTableRow.js
```diff
@@ -88,7 +89,15 @@ const DataTableRow = React.memo(
...
             case COLUMN_TYPES.EDITABLE_STRING:
-            case COLUMN_TYPES.EDITABLE_NUMERIC:
+            case COLUMN_TYPES.EDITABLE_NUMERIC: {
+              // Special condition for stocktake counted total quantity cells.
+              // Use the placeholder 'Not counted' when a stocktake item or batch
+              // has not been counted yet.
+              let placeholder = '';
+              if (columnKey === COLUMN_NAMES.COUNTED_TOTAL_QUANTITY) {
+                placeholder = rowData.hasBeenCounted ? '' : tableStrings.not_counted;
+              }
```
wlthomson: Maybe can remove the mutable state here, something like:

```
const defaultPlaceholder = '';
const notCountedPlaceholder = tableStrings.not_counted;
const isNotCounted = columnKey === COLUMN_NAMES.COUNTED_TOTAL_QUANTITY && !rowData.hasBeenCounted;
const placeholder = isNotCounted ? notCountedPlaceholder : defaultPlaceholder;

```

joshxg: as much as I don't like mutating - if it makes it easier to read I just do it, and it seems to be easier to read than those lines? (and also more open to other special conditions if we need to use placeholders for other column types?) :(

wlthomson: Personally, I find `let` hard to read, because it's harder to reason about the variable state (because it can be changed). Looking at it again, I think I went over the top with variables. You could make the suggestion a bit cleaner:

```
const isNotCounted = columnKey === COLUMN_NAMES.COUNTED_TOTAL_QUANTITY && !rowData.hasBeenCounted;
const placeholder = isNotCounted ? tableStrings.not_counted : '';
```

Let me know either way and I'll merge 👍 .

joshxg: i think no, simply cause the next cell that needs a placeholder will be screwed!

Issue #484, opened 2017-08-19
A script for converting localisation files into csv and back to js


andreievg:
It would be very handy to have a script that parses all files under /localisation and converts them to csv files, one per language. 
Csv file can then be converted to excel, and sent for translation, once translated a script should be able to do the reverse.

Chris-Petty:
It'd be very worthwhile making sure that whatever format exported/imported conforms to popular localisation editing tool format(s)

joshxg:
I think this should be done soon! 

Idea: Keep strings in `.json` files, import them to use as the parameter for the localization singletons.

Then use a package such as https://www.npmjs.com/package/json2csv (first one I googled) to convert. simple!

diegoevangelisti:
I agree with the idea of keeping the data in `.json` files to can easily export it to csv and import it later. Based on Andrei’s thoughts of having one file per language, this can be a probable flow:

New `.json` files, one for each language:
gb.json, fr.json, gil.json, fl.json, la.json

Example of file structure for gb.json

```
{
	“authStrings": {
		“loggin_in”: “Loggin in ...”,
		“username”: ”Login”,
		…
		},
	“buttonStrings”: {
		…
		}
	…
}
```
So each file contains all the information related with one language and the data is organized by localization files.
(authStrings, buttonStrings, etc)

All files can be imported in localisation/index.js and exported like this:

`const gb = JSON.parse(fs.readFileSync(‘./gb.json’)`
`const fr = JSON.parse(fs.readFileSync(‘./fr.json’)`
And so on for the rest.

`export { gb, fr, gil, fl, la }`

After that all files in /localization can have access to corresponding string:

Example for authStrings.js

```
Import { gb, fr, gil, tl, la } from ‘./index.js'

export const authStrings = new LocalizedStrings({

gb: {
    logging_in: gb.authStrings.loggin_in,
    login: gb.authStrings.login,
    password: gb.authStrings.password,
    repeat_password: gb.authStrings.repeat_password,
    user_name: gb.authStrings.user_name,
    email: gb.authStrings.email,
    warning_sync_edit: gb.authStrings.warning_sync_edit,
},
fr: {
    logging_in: fr.authStrings.loggin_in
    login: gb.authStrings.login,
    password: gb.authStrings.password,
    user_name: gb.authStrings.user_name,
},
gil: {
    logging_in:  gil.authStrings.loggin_in,
    login: gil.authStrings.login,
    password: gil.authStrings.password
[...]

PR #1412, opened 2019-10-19
#1400 column header spacing

### LINKED ISSUE #1400: Column headers not great when not english
## Describe the bug

Column headers aren't looking great when not in english

### To reproduce

1. Go to supplier/customer requisitions or stocktake edit page in french/laos/kiribati

### Expected behaviour

Should look good

### Version and device info

2.3.6 development

### Additional context

image

image

image

Tetum/English are fine

joshxg:
Fixes #1400 

## Change summary

- Added some explicit new lines in a few strings
- Changed margin on right aligned cells to give a bit more space
- Added a few french translations

Kiribati/laos customer requisitions still doesn't look good. The strings are long, I think it would require increasing the size of the header row by about 50%. I think we could look into making sure the translations are correct and are as short as possible before doing that. Especially considering a couple strings have `?` at the end, as if the translations weren't actually complete and a bit of guessing was going on?

## Testing

- [ ] Customer requisition page in particular looks fine in english/french (Was the focus of the PR in the end)
- [ ] Hide/show over stock toggle in french looks fine

Need to update a lot of translations. There is a veyr old issue here #484 which I think is worth doing. Wouldn't take long and would be super helpful!

image

image

image


src/localization/buttonStrings.js
```diff
@@ -27,9 +27,9 @@ export const buttonStrings = new LocalizedStrings({
     view_regimen_data: 'View Regimen Data',
   },
   fr: {
-    // TODO: view_regimen_data
+    view_regimen_data: 'Données de régimes',
     add_batch: 'Ajouter un lot',
-    add_master_list_items: "Ajouter une liste-Maître d'articles ?",
+    add_master_list_items: 'Ajouter à\npartir des\nlistes maîtres',
```
wlthomson: Noticed this translation has changed. Just checking this is the correct translation?

wlthomson: I think can get away with a single newline here:

```suggestion
    add_master_list_items: 'Ajouter à partir\ndes listes maîtres',
```

Before:

#1400-column-header-spacing-1

After:

#1400-column-header-spacing-2

joshxg: oops sorry I meant to put that in the top comment. I took this translation for desktop explicitly. The `?` tipped me off it may not have been translated correctly:

Original: Ajouter une liste-Maître d'articles -- Add a master list of articles
New: Ajouter à partir des listes maîtres -- Add from master lists

@alainsussol this is correct? :)

joshxg: Ah nice thanks - f83afcc

wlthomson: Google translate says you're right 😄 . For sake of getting `v3.0.0` out, will merge (can always fix later, if needed 👍 ).

joshxg:
have fixed conflicts

src/localization/buttonStrings.js
alainsussol: Yes this is correct...

PR #1416, opened 2019-10-19
#1407 add correct query parameter

### LINKED ISSUE #1407: Program requisition pricing not working
## Describe the bug

The price column in a program supplier requisition is not working

### To reproduce

1. Create a Program where an item has a price on desktop, relate it to the mobile store
2. Create a program requisition on mobile
4. See the price does not show!

### Expected behaviour

The price should show!

### Version and device info

2.3.6 development

### Additional context

Current code:

```
    // Query string: Query for all MasterList.IDs related to this store and this item.
    const queryString = `${masterListIds
      .map(id => `masterList.id == '${id}'`)
      .join(' OR ')} AND item = $0`;

    // Return the maximum price of all MasterListItems.
    return UIDatabase.objects('MasterListItem')
      .filtered(queryString, this)
      .max('price');
  }
```

Need to change the line
```
.filtered(queryString, this)
```

To

```
.filtered(queryString, this.item)
```

joshxg:
Fixes #1407 

## Change summary

- Just adding the correct parameter for the query

## Testing

- [ ] Create two masters lists for the mobile store. At least one should be a program. Add an item with different prices. Create a program requisition with the item. Price should be the max of those two prices.

This pricing doesn't seem right but was how I was told to do it. Can change it later when we get told it's incorrect!

### DIFF
--- a/src/database/DataTypes/RequisitionItem.js
+++ b/src/database/DataTypes/RequisitionItem.js
@@ -148,7 +148,7 @@ export class RequisitionItem extends Realm.Object {
 
     // Return the maximum price of all MasterListItems.
     return UIDatabase.objects('MasterListItem')
-      .filtered(queryString, this)
+      .filtered(queryString, this.item)
       .max('price');
   }
 


joshxg:
Just keep as general behavior for now. 

If there are new pricing requirements I envisage having each pricing algorithm separated completely, and the `get price()` method will just be a switch dependent on the 'customization'. Maybe that makes sense 🤷‍♂ 

I wanted to do this originally but was trying not to overengineer and code for non-existent requirements? heh

wlthomson:
> Just keep as general behavior for now.
> 
> If there are new pricing requirements I envisage having each pricing algorithm separated completely, and the `get price()` method will just be a switch dependent on the 'customization'. Maybe that makes sense 🤷‍♂
> 
> I wanted to do this originally but was trying not to overengineer and code for non-existent requirements? heh

I agree. Is a nice use-case for templating. I was more just double-checking that this is intended behaviour for all clients (including all IC clients) and won't cause any unexpectedness?

PR #1417, opened 2019-10-19
#1405 Multiline Cell component

### LINKED ISSUE #1405: Cell component should be multi line by default
## Describe the bug

`Cell` is a single line. This cna be a bit problematic when item names are long. Especially given that their units and sizes need to be in the name.

### To reproduce

1. On desktop edit an item name to be super long
2. Add this item to any page - SI/CI/Stocktake etc
4. See error [It's ellipsized up and cut off even though theres room for a couple lines at least]

### Expected behaviour

Should use two lines

### Version and device info

All, forever I think

### Additional context

- Would need to update PR here: https://github.com/sussol/react-native-data-table/pull/57
- Maaybe need to adjust row height maybe, but worth it most likely

joshxg:
Fixes #1405 

## Change summary

- Adds a `maxLines` prop to `Cell`, default of 2

image

image

## Testing

- [ ] Have an `ItemLine`/`Supplier`/`Customer` with a really long name and look at the name in the `DataTable`, should look all good over two lines

N/A

have updated PR for `react-native-data-table`: https://github.com/sussol/react-native-data-table/pull/57

### DIFF
--- a/src/widgets/DataTable/Cell.js
+++ b/src/widgets/DataTable/Cell.js
@@ -16,18 +16,20 @@ import { getAdjustedStyle } from './utilities';
  * @param {Number}        maxLiens   Maximum number of lines for the Text component
  *                                   in a row. Removing the borderRight if true.
  */
-const Cell = React.memo(({ value, textStyle, viewStyle, width, isLastCell, maxLines, debug }) => {
-  if (debug) console.log(`- Cell: ${value}`);
-  const internalViewStyle = getAdjustedStyle(viewStyle, width, isLastCell);
+const Cell = React.memo(
+  ({ value, textStyle, viewStyle, width, isLastCell, numberOfLines, debug }) => {
+    if (debug) console.log(`- Cell: ${value}`);
+    const internalViewStyle = getAdjustedStyle(viewStyle, width, isLastCell);
 
-  return (
-    <View style={internalViewStyle}>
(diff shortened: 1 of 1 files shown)


src/widgets/DataTable/Cell.js
```diff
@@ -13,15 +13,16 @@ import { getAdjustedStyle } from './utilities';
...
  */
-const Cell = React.memo(({ value, textStyle, viewStyle, width, isLastCell, debug }) => {
+const Cell = React.memo(({ value, textStyle, viewStyle, width, isLastCell, maxLines, debug }) => {
   if (debug) console.log(`- Cell: ${value}`);
   const internalViewStyle = getAdjustedStyle(viewStyle, width, isLastCell);
 
   return (
     <View style={internalViewStyle}>
-      <Text ellipsizeMode="tail" numberOfLines={1} style={textStyle}>
+      <Text ellipsizeMode="tail" numberOfLines={maxLines} style={textStyle}>
```
wlthomson: Just my opinion, think its nice to use consistent props with the underlying component. E.g. `numberOfLines` instead of `maxLines`. I think unless the underlying component is being used in a very different way, it makes sense to model components consistently with the conventions used by the built-in components?

joshxg: 352447b432b5fde3924d06546c5fbe067c662eea

PR #1428, opened 2019-10-21
#1425 Stocktake navigation

### LINKED ISSUE #1425: User is not navigated to StocktakeManagePage when creating a new stocktake
## Describe the bug

Expected behaviour should be as described by the following workflow:

Go to stocktakes -> Click button to create a new stocktake -> Enter stocktake parameters -> Enter stocktake items.

After entering the stocktake parameters via `ByProgramModal`, the user should be presented with the `StocktakeManagePage` to specify items. This works as expected when programs are not enabled, but does not when programs are enabled. 

### To reproduce

1. Disable programs if they are enabled.
2. Create a new stocktake.
3. Note that you are navigated from `ByProgramModal` to `StocktakeEditPage` (instead of to `StocktakeManagePage`).

### Expected behaviour

User should be navigated to `StocktakeManagePage` after creating a new stocktake (invariant of programs enabled or not).

### Version and device info

- App version: 2.3.6.
- Tablet model: Pixel C (emulated).
- OS version: Android API 28

### Additional context

See https://github.com/openmsupply/mobile/issues/1241#issuecomment-544342739.

joshxg:
Fixes #1425 

## Change summary

- Changes a couple of navigation actions to fix the bug as per the issue
    - When creating a `Stocktake`, 3 invocations
        - Creating a program stocktake: create and go to `StocktakeEditPage`
        - Creating a normal stocktake with programs defined - create and go to StocktakeManagePage
        - Creating a normal stocktake with no programs defined: Already on `StocktakeManagePage`, with a list of items selected - go to `StocktakeEditPage`
    - When navigating to `StocktakeEditPage`, two invocations
        - Navigating from `StocktakesPage` [When creating a program stocktake]
        - Navigating from `StocktakeManagePage` [all others]. Don't want to navigate back to `StocktakeManagePage` from `StocktakeEditPage`, so pop and push the stack

## Testing

- [ ] Creating a program stocktake
    - [ ] After creating a program stocktake - navigates to `StocktakeEditPage`
    - [ ] From `StocktakeEditPage`, pressing back goes to `StocktakesPage`
- [ ] Creating a general stocktake [from `ByProgramModal`]
    - [ ] Navigates to `StocktakeManagePage` on creation [name should be set]
    - [ ] Selecting items and confirming navigates to `StocktakeEditPage`
    - [ ] Navigating back from `StocktakeEditPage` goes to `StocktakesPage`
- [ ] Creating a stocktake [with no programs defined]
    - [ ] Navigates to `StocktakeManagePage`  [no stocktake created yet]
    - [ ] Adding items and confirming navigates to `StocktakeEditPage`
    - [ ] Navigating back from `StocktakeEditPage` navigates to `StocktakesPage`

N/A

### DIFF
--- a/src/navigation/actions.js
+++ b/src/navigation/actions.js
@@ -9,6 +9,7 @@ import Settings from '../settings/MobileAppSettings';
 import { createRecord } from '../database/utilities/index';
 import { navStrings } from '../localization/index';
 import { SETTINGS_KEYS } from '../settings/index';
+import { getCurrentRouteName } from './selectors';
 
 /**
  * Navigation Action Creators.
@@ -48,18 +49,24 @@ export const gotoStocktakeManagePage = ({ stocktake, stocktakeName }) =>
  *
  * @param {Object} stocktake  The requisition to navigate to.
  */
-export const gotoStocktakeEditPage = stocktake => {
+export const gotoStocktakeEditPage = stocktake => (dispatch, getState) => {
+  const { nav } = getState();
+
+  const currentRouteName = getCurrentRouteName(nav);
+
   const hasNegativeAdjustmentReasons = UIDatabase.objects('NegativeAdjustmentReason').length > 0;
   const hasPositiveAdjustmentReasons = UIDatabase.objects('PositiveAdjustmentReason').length > 0;
   const usesReasons = hasNegativeAdjustmentReasons && hasPositiveAdjustmentReasons;
 
-  return NavigationActions.navigate({
+  const navigationActionCreator =
+    currentRouteName === 'stocktakes' ? NavigationActions.navigate : StackActions.replace;
+
+  const navigationParameters = {
     routeName: usesReasons ? 'stocktakeEditorWithReasons' : 'stocktakeEditor',
-    params: {
-      title: navStrings.stocktake,
-      stocktake,
-    },
-  });
+    params: { title: navStrings.stocktake, stocktake },
+  };
+
+  dispatch(navigationActionCreator(navigationParameters));
 };
 
 /**
@@ -209,12 +216,9 @@ export const createStocktake = ({ currentUser, stocktakeName, program, itemIds }
     else if (itemIds) stocktake.setItemsByID(UIDatabase, itemIds);
   });
 
-  const stackReplacementAction = StackActions.replace({
-    routeName: 'stocktakeEditor',
-    params: { stocktake, title: navStrings.stocktake },
-  });
+  const nextAction = program || itemIds ? gotoStocktakeEditPage : gotoStocktakeManagePage;
 
-  dispatch(stackReplacementAction);
+  dispatch(nextAction(stocktake));
 };
 
 /**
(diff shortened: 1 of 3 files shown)


src/navigation/actions.js
```diff
@@ -48,18 +49,25 @@ export const gotoStocktakeManagePage = ({ stocktake, stocktakeName }) =>
...
+
   const hasNegativeAdjustmentReasons = UIDatabase.objects('NegativeAdjustmentReason').length > 0;
   const hasPositiveAdjustmentReasons = UIDatabase.objects('PositiveAdjustmentReason').length > 0;
   const usesReasons = hasNegativeAdjustmentReasons && hasPositiveAdjustmentReasons;
 
-  return NavigationActions.navigate({
+  // If navigating from the stocktakesPage, go straight to the StocktakeEditPage. Otherwise,
+  // replace the current page as the user is coming from StocktakeManagePage.
+  const navigationActionCreator =
+    currentRouteName === 'stocktakes' ? NavigationActions.navigate : StackActions.replace;
```
wlthomson: Just to make sure I follow, we push to the stack if navigating from `StocktakesPage`, but replace if `StocktakeManagePage`? Assuming this is so that the previous page on the stack when viewing `StocktakeEditPage` is always `StocktakesPage`?

If this is correct, suggest the comment only needs to explain this (and the rest can be seen in the code), e.g. something simple like

```
// Maintain stockTakes page at the top of the navigation stack.
// If not navigating from stocktakesPage, assume navigating from StocktakeManagePage. 
```

If incorrect, then ignore 😆 .

src/navigation/actions.js
```diff
@@ -209,12 +217,11 @@ export const createStocktake = ({ currentUser, stocktakeName, program, itemIds }
...
 
-  const stackReplacementAction = StackActions.replace({
-    routeName: 'stocktakeEditor',
-    params: { stocktake, title: navStrings.stocktake },
-  });
-
-  dispatch(stackReplacementAction);
+  // If creating a Stocktake with a program or with a list of itemIds, navigate to
+  // the StocktakeEditPage. Otherwise, navigate to the StocktakeManagePage.
+  return program || itemIds
```
wlthomson: Maybe a constant here instead of the first line of the comment? E.g.

```
// If stocktake has items, navigate to StocktakeEditPage, else navigate to StocktakeManagePage.
const stocktakeHasItems = program || itemIds
return dispatch(stocktakeHasItems 
  ? gotoStocktakeEditPage(stocktake) 
  : gotoStocktakeManagePage(stocktakeName, stocktake))

File: src/navigation/actions.js [hunk 3]
```diff
@@ -209,12 +217,11 @@ export const createStocktake = ({ currentUser, stocktakeName, program, itemIds }
...
+    ? dispatch(gotoStocktakeEditPage(stocktake))
+    : dispatch(gotoStocktakeManagePage(stocktakeName, stocktake));
```
wlthomson: Just personal preference, but I think is nicer to put the ternary inside the dispatch (see above)?

PR #1448, opened 2019-10-29
#1439 update pages prop types

### LINKED ISSUE #1439: Add PropTypes to pages

We currently disable `react/forbid-prop-types` for all pages. Using proper PropTypes is a good idea for helping catch errors at run-time.

Update PropTypes using `instanceOf` and `shape` instead of `object`. We can then remove the disabling of `react/forbid-prop-types`.

Leaving as is, or disabling for all files. Although this is a quick solution, I don't think it's a good idea as enforcing types only takes a small amount of work and is a good way to reduce errors and likely more than pay for itself in the long run.

## Additional context

See #845.

wlthomson:
Fixes #1439.

## Change summary

Add PropTypes to pages.

## Testing

- [ ] Navigating to customer invoices in development mode does not show any props errors.
- [ ] Opening a customer invoice in development mode does not show any props errors.
- [ ] Navigating to customer requisitions in development mode does not show any props errors.
- [ ] Opening a customer requisition in development mode does not show any props errors.
- [ ] Navigating to supplier requisitions in development mode does not show any props errors.
- [ ] Opening a supplier requisition in development mode does not show any props errors.
- [ ] Navigating to stock in development mode does not show any props errors.
- [ ] Navigating to stocktakes in development mode does not show any props errors.
- [ ] Editing a stocktake in development mode does not show any props errors.

N/A.


src/database/UIDatabase.js
```diff
@@ -198,4 +198,22 @@ export const getUIDatabaseInstance = database => {
...
+    return null;
+  }
+
+  if (prop instanceof UIDatabase) return null;
+  return new TypeError(`Invalid database prop in ${componentName}`);
+};
+
+export const UIDatabaseType = createUIDatabasePropType(false);
+UIDatabaseType.isRequired = createUIDatabasePropType(true);
+
```
joshxg: this is cool is it a common pattern I've never seen it before

src/pages/CustomerRequisitionsPage.js
```diff
@@ -7,6 +7,8 @@ import React, { useCallback } from 'react';
 import PropTypes from 'prop-types';
 import { View } from 'react-native';
 
+import { navigation } from 'react-navigation';
```
joshxg: this is undefined, I *think* we want https://github.com/react-navigation/core/blob/master/src/routers/StackRouter.js here

src/pages/MenuPage.js
```diff
@@ -155,13 +157,16 @@ export class MenuPage extends React.Component {
...
 
-/* eslint-disable react/require-default-props, react/forbid-prop-types */
 MenuPage.propTypes = {
-  database: PropTypes.object.isRequired,
+  database: UIDatabaseType.isRequired,
   isInAdminMode: PropTypes.bool,
   logOut: PropTypes.func.isRequired,
   navigateTo: PropTypes.func.isRequired,
-  settings: PropTypes.object.isRequired,
+  settings: PropTypes.shape(SETTINGS_TYPES).isRequired,
```
joshxg: this basically does nothing and the `SETTINGS_TYPES` is a bit confusing. `settings` should be an instance of `MobileAppSettings` which is a singleton?

src/settings/index.js
```diff
@@ -21,6 +22,27 @@ export const SETTINGS_KEYS = {
   HARDWARE_UUID: 'Hardware_UUID',
 };
 
+export const SETTINGS_TYPES = {
```
joshxg: as above I dunno whats going on with this?

src/database/UIDatabase.js
wlthomson: I couldn't find any examples with a similar "singleton" design to ours, so this is mostly just a pattern I came up with that I thought went well with our singleton design. But the `.isRequired` is just a property like any other (actually just a convention), so I saw most common prop types using some sort of closure or nested function call in a similar way.

src/pages/MenuPage.js
wlthomson: Yeah, it should just be an `instanceOf`. Not sure what I was going for here!

a236a5c46c1873fbe97aabfae62f2a9433903e99.

src/pages/CustomerRequisitionsPage.js
wlthomson: Had a look through `react-navigation` and their typescript definitions. I think the best way to do it (without ts) is to just make sure the methods we use are defined on the object. 

086a2cfbe6d5bf0e7158195afd9bc3ed7e48ed7a.

joshxg: Hate to hold this up but using `shape` seems to just do nothing? It seems equivalent to using `object` except the linter doesn't complain - if i'm not wrong, why not just use `object`, disable the rule and https://github.com/openmsupply/mobile/issues/845 knock this one away with this

wlthomson: > Hate to hold this up but using `shape` seems to just do nothing? It seems equivalent to using `object` except the linter doesn't complain - if i'm not wrong, why not just use `object`, disable the rule and #845 knock this one away with this

No worry about hold up. Sorry for late reply :).

As I understand it, it works like this:

```
Blah.propTypes = {
  anyObject: PropTypes.object,
  numberObject: PropTypes.objectOf(PropTypes.number), 
  fooBarObject: PropTypes.shape({ foo: PropTypes.string, bar: PropTypes.number }),
  ....
```

`Blah` takes props:

- `anyObject`, can be any object with any keys, values.
- `numberObject`, can have any keys, but all values must be numbers.
- `fooBarObject`, must have keys `foo` and `bar` and respective values of correct type.

So, AFAIK, it goes `PropTypes.object -> PropTypes.objectOf -> PropTypes.shape`, from most general to most specific. `object` doesn't do much type checking at all, so guessing that's probably why eslint doesn't like it.

joshxg: It's been a bit since I looked at this - but when I changed shape to be an arbitrary object to check it was actually doing anything, it didnt complain. It would seem obvious that specifying a shape it would check the fields, but it didnt actually seem to! Maybe i'm wrong though

wlthomson: I _think_ it only complains if the fields are missing, but it doesn't prevent you adding arbitrary ones. Maybe you added fields to the object but didn't remove any?

joshxg: Ahh. I get no warnings at all with an empty object. All it does it ensure that any fields that are defined are the correct type, huh? Not great heh, sorry for not understanding

wlthomson: Yeah I think so, you might be able to make them required. Finishing up prop types is top on my list after the indicator stuff, so hopefully will get around to this today/tomorrow.

PR #1458, opened 2019-10-30
#1457 Internal & External suppliers

### LINKED ISSUE #1457: Supplier requisitions using external rather than internal suppliers
## Describe the bug

The modal for selecting a supplier is using external suppliers and is used for both supplier requisition and supplier invoices. Supplier requisitions need to use internal

### To reproduce

1. Go to `SupplierRequisitionsPage`
2. Click on new `SupplierRequisition`

### Expected behaviour

Should be for internal not external suppliers

### Version and device info

- App version:
- Tablet model:
- OS version:

joshxg:
Fixes #1457 

## Change summary

- Separates `SELECT_SUPPLIER` key to `SELECT_EXTERNAL_SUPPLIER` and `SELECT_INTERNAL_SUPPLIER`
- Uses a the key to switch between `ExternalSupplier` and `InternalSupplier` in `DataTablePageModal`

## Testing

- [ ] When creating a `SupplierRequisition` - the choice of supplier are INTERNAL suppliers
- [ ] When creating a `SupplierInvoice` - the choice of suppliers are EXTERNAL suppliers

Unit tests would have fixed this


joshxg:
it is already using `InternalSupplier` https://github.com/openmsupply/mobile/blob/0dfd251e104205e841fa7817993b0761032764d9/src/widgets/modals/ByProgramModal.js#L114

PR #1473, opened 2019-10-31
#1471 stocktake line editing

### LINKED ISSUE #1471: Editing line on stocktake wipes batch counts
## Describe the bug

When editing a stocktake, you can go item by item and/or edit batches of an item. When editing by item, it does some dark magic system related to the fabled "FEFO" prophecy.

3.0.0 rewrite included saving changes to a cell on every keystroke. If you delete a number before entering a new one, it'll apply changing stock to 0 across all batches. Then as you type in a new number, it only adds to the first expiring (I think). This is bad, as it ruins all your batch stock counts and movements if you then finalised the stocktake.

### To reproduce

  1. Go to a Stocktake that has some lines with existing counts
  2. Inspect batch quantities for a line with stock. Add some extra batches if there is only 1, and add some stock to them
  3. Go back to the item list and edit the item stock, say from 100 to 103:
    - Tap the cell (which focuses the number) and start typing a new one. This will wipe all stock then add the count to the first expiring batch (or last? I'm tired)
    - Tap the cell and set the cursor to the end. backspace once to __10__, removing 90 stock from batches according to reverse FEFO (??? (the order doesn't actually matter regarding this bug)), then adding 93 to the first expiring batch*
  4. Check batches again and see all the stock has shifted violently to the nearest expiry.

_*This method will work even on 100->10->100. Probably not expecting much to change, but it do_

### Expected behaviour

Batch quantities are preserved if you are trying to edit an item counted quantity. 

Basically I think cells in stocktake items should only callback on losing focus rather than on every keystroke. Not sure how else you'd fix this. Also whoops, didn't see this one coming.

### Version and device info

- App version: 3.0.1
- Tablet model: TabA SM-T580
- OS version: YEETDROID

### Additional context

#625 can probably be closed

joshxg:
Fixes #1471 

## Change summary

- Adds a new stateful Cell for editing stocktakes

## Testing

- [ ] Editing any `Stocktake` `Counted Quantity` cell does not change any values until losing focus [i.e. difference column does not change as you type]

This isn't a perfect fix - there are still inconsistencies when opening/closing the modal, clicking finalised etc etc. I don't like this fix but it will do for now, I think

### DIFF
--- a/src/widgets/DataTable/TextInputCellWithState.js
+++ b/src/widgets/DataTable/TextInputCellWithState.js
@@ -1,0 +1,135 @@
+/* eslint-disable react/forbid-prop-types */
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import { View, TextInput } from 'react-native';
+
+import Cell from './Cell';
+import RefContext from './RefContext';
+
+import { getAdjustedStyle } from './utilities';
+
+const TextInputCellWithState = React.memo(
+  ({
+    value,
+    rowKey,
+    columnKey,
+    isDisabled,
+    placeholderColour,
+    editAction,
+    dispatch,
+    isLastCell,
+    width,
+    debug,
+    keyboardType,
+    placeholder,
+    viewStyle,
+    rowIndex,
+    textInputStyle,
+    cellTextStyle,
+    underlineColor,
+  }) => {
+    if (debug) console.log(`- TextInputCellWithState: ${value}`);
+
+    const [internalValue, setInternalValue] = useState({ initialValue: value, newValue: value });
+
+    const { initialValue, newValue } = internalValue;
+
+    if (initialValue !== value) setInternalValue({ initialValue: value, newValue: value });
+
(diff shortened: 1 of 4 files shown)


src/widgets/DataTable/TextInputCellWithState.js
```diff
@@ -0,0 +1,135 @@
...
+
+    const focusNext = () => focusNextCell(refIndex);
+
+    // On editing, set the internal state of this component.
+    const onEdit = updatedValue => setInternalValue({ ...internalValue, newValue: updatedValue });
+
+    // When editing has been completed, submit the final result.
+    const onEnd = ({ nativeEvent }) => {
+      const { text = '' } = nativeEvent || {};
+      dispatch(editAction(text, rowKey, columnKey));
```
wlthomson: I think `columnKey` is redundant here?

joshxg:
Closing this PR until implementation/solution is agreed upon

PR #1478, opened 2019-10-31
#1475 Focus crash

### LINKED ISSUE #1475: Focusing a null pointer
## Error in mSupply Mobile

null is not an object (evaluating 'I((t+1)%h).current.focus')

[View on Bugsnag](https://app.bugsnag.com/sustainable-solutions-nz-ltd/msupply-mobile/errors/5dba542ca6e17e001a903c14?event_id=5dba542c00525113778d0000&i=gh&m=ci)

## Stacktrace

    index.android.bundle:942 - T

[View full stacktrace](https://app.bugsnag.com/sustainable-solutions-nz-ltd/msupply-mobile/errors/5dba542ca6e17e001a903c14?event_id=5dba542c00525113778d0000&i=gh&m=ci)

joshxg:
Fixes #1475 

## Change summary

- Adds a check that a ref exists before trying to focus

Essentially this occurs in the edge case in a stocktake batch modal window where the table is smaller, and the keyboard covers some rows. When the keyboard covers rows, rows aren't rendered, and therefore dont have a ref to focus. 

I'm not sure of a nother work around. Doesn't occur on a normal table, only on stocktake batch which is smaller as explained above

## Testing
* This is very edge case, I think I would need to show someone how to have it occur*
- [ ] Focusing the next cell never crashes the app

N/A

### DIFF
--- a/src/widgets/DataTable/DataTable.js
+++ b/src/widgets/DataTable/DataTable.js
@@ -84,7 +84,7 @@ const DataTable = React.memo(
       const nextCellRef = (refIndex + 1) % numberOfEditableCells;
       const cellRef = getCellRef(nextCellRef);
 
-      return cellRef.current.focus();
+      return cellRef.current && cellRef.current.focus();
     };
 
     // Adjusts the passed row to the top of the list.


joshxg:
Merging due to the app *never* crashing for @wlthomson

wlthomson:
> Merging due to the app _never_ crashing for @wlthomson

:).

PR #1488, opened 2019-11-04
#1486 menu page modules

### LINKED ISSUE #1207: Exporting data throws an error when not enough space
## Describe the bug

When a device has insufficient memory to export, an error is thrown due to trying to check for a non-existent property.

Pretty rare this would occur, I'd imagine. Randomly found exporting data on an emulator that I haven't cleared the data on in a looong time!

### To reproduce

**Setup: Have a device with less free memory than a realm file (10mb or something) - can reduce memory size on emulator easily**

1. Go to `MenuPage`, turn on `AdminMode`
2. Click on `Export Data`

### Expected behaviour

Should warn with the toast that is setup, rather than error. Simple two line change - calling method in `MenuPage` checks for an `error` object, and a `message` within that - should just check for a `message` field.

### Version and device info

- App version: 2.3+
- Tablet model: All
- OS version: All

### LINKED ISSUE #1486: Changes to MenuPage for modules

For additional modules - i.e. vaccines, dispensing we need an entry point

Turn the 3 sections of the main menu page into 4. One for modules

Refactoring the entire app to use bottom tab navigator :)

## Additional context

Vaccines did this. Has an icon to use also.

joshxg:
Fixes #1486 Fixes #1207 

## Change summary

- Refactors `MenuPage` to a function rather than a class
- Fixes a bug when a tablet doesn't have enough space to export throwing an error
- Uses the modules image from VM branch. Can change this/recolour later if we don't like it
- Didn't use the exact code from VM branch as I don't think there's a need to really make the code that DRY as the potential extensions would really just be to add buttons to the `ModulesSection` for VM, dashboard etc. Anything else and we would need to make a new page, it wouldn't work.
- Added `ModuleReducer` -- I think this could be `CustomisationReducer` which would hold things like `usingPrograms`, `usingReasons`. Not 100%, feedback be nice!

Without modules (Badges are being iterated on):

image

With modules:
image

admin mode:

image

## Testing

**NOTE: Badges are new and will be tested in a different issue**

- [ ] Can navigate to `CustomerInvoices`
- [ ] Can navigate to `SupplierInvoices`
- [ ] Can navigate to `CustomerRequisition`
- [ ] Can navigate to `CustomerInvoices`
- [ ] Can navigate to `Stocktakes`
- [ ] Can navigate to `Stock`

Mainly the `ModuleReducer` - in order to dynamicly update the UI, will need to have a case for when sync ends, where the state can be refreshed. Quite easy to do but want to lock in what the state should hold before I implement anything


src/pages/MenuPage.js
```diff
@@ -28,183 +27,208 @@ import {
...
+    backgroundColor: 'white',
+    alignItems: 'center',
+    borderColor: SHADOW_BORDER,
+    borderWidth: 1,
+    justifyContent: 'flex-start',
+    marginTop: 30,
+    paddingTop: 50,
+    marginHorizontal: 30,
+    flex: 1,
+  },
```
mark-prins: There's a lot of similarity between the container styles. I was wondering why it isn't possible to have the same style instead of creating the two and that lead me to think about bringing together the common styles e.g.

```
  container: {
    alignItems: 'center',
    backgroundColor: 'white',
    borderColor: SHADOW_BORDER,
    borderWidth: 1,
    flex: 1,
    marginTop: 30,
    paddingTop: 50,
  },
  moduleContainer: {
    flexDirection: 'row',
    justifyContent: 'center',
    marginHorizontal: 15,
  },
  originalContainer: {
    justifyContent: 'flex-start',
    marginHorizontal: 30,
  },
```

and then something like this, which makes things a bit messy again - so may not be worth it.

```
  const { image, container, originalContainer, moduleContainer, bottomIcon, moduleRow } = styles;

  const containerStyle = Object.assign(
    {},
    container,
    usingModules ? moduleContainer : originalContainer
  );
```

joshxg: In general for styles I've found that creating multiple,  similar styles to just be easier to work with. Taken a bit far here, thanks 56b6dae 

Also what did you use to update the image? (thanks btw!)

mark-prins: I used gimp for the image - there's something odd with the colour palette that I didn't figure out. worked around it by copying one of the other images and pasting in a new layer

PR #1500, opened 2019-11-05
#1492 basic dispensing module page

### LINKED ISSUE #1492: Basic Dispensary page

Can't get a started on dispensing

A basic dispensary page.. something to work with

Not doing this

## Additional context

This is a simple, basic page to get started and start iterating on. By having a basic UI, other core/backend logic can be worked on which needs less consensus/discussion than UI

joshxg:
Fixes #1492 

## Change summary

- Adds a very basic Dispensing module page.

image

- To enable `MenuPage` to turn into a 4 grid over a 3, add `usingDispensary : true` to your store custom data (and will need to reopen the app)

- Should dispensing even be a 'module' 🤷‍♂


Chris-Petty:
This is fine for the basic, but regarding:

> Should dispensing even be a 'module'

In implementation, yes. Could be arguments that it should be in the "customer" section of the app though... but could just look ugly.

src/reducers/PagesReducer.js
```diff
@@ -21,7 +21,7 @@ export const PagesReducer = (state = {}, action) => {
       const { pageObject } = params;
 
       const pageInitialiser = getPageInitialiser(routeName);
-      const pageInitialState = pageInitialiser(pageObject);
+      const pageInitialState = pageInitialiser ? pageInitialiser(pageObject) : {};
```
Chris-Petty: This will conflict with [#1517 changes](https://github.com/openmsupply/mobile/pull/1517/files#diff-f4800056faa9ea0c03eda6581a4e7969R24) but that's fine, though this here assigns an empty object where as that PR assigns null in the falsy case.

PR #1509, opened 2019-11-06
#1489 badges to small

### LINKED ISSUE #1487: Remove badges at the top of pages
## Describe the bug

The badges at the top of pages i.e. `SupplierInvoicePage` do not update correctly and are unneccesary.

### To reproduce

1. Go to `SupplierInvoicePage` with unfinalised invoices
2. Go to a in progress invoice, finalise it.
4. See error - not updated badge

### Expected behaviour

The badge shouldn't be there since there is a toggle for in progress/finalised

### Version and device info

- App version: 3.1 development
- Tablet model: All
- OS version:  All

### Additional context

Solution for this issue is to remove this badge

### LINKED ISSUE #1489: Badges are too small

Need to increase the size of the newly introduced badges.

Should probably increase the size by at least double, maybe three times.

I'm not sure about the shape - I personally think a circle is nice, but not required to change!

N/A

## Additional context

N/A

joshxg:
Fixes #1489 Fixes #1487 

On physical tablets, the badges were hard to read and hard to press.

## Change summary

image

- Boldens text, increases the size
- Increases the size of each badge
- Simplifies code, use element ref rather than absolute values
- Use a custom hook so that we can reuse the logic easily

## Testing

- [ ] Do something

As per comment here https://github.com/openmsupply/mobile/issues/1489#issuecomment-549730303 - I did not change the appearance other than increasing the size


src/widgets/InfoBadge.js
```diff
@@ -1,129 +1,103 @@
...
-    const unfinalisedCountText = pendingCount > 99 ? '99+' : pendingCount;
+  const popoverText = useCallback(
+    (item, key) =>
+      (item.count > 0 || item.text) && (
+        <Text style={badgeTextStyle} key={key}>
+          {`${item.title}: ${item.count || item.text}`}
+        </Text>
+      ),
+    []
+  );
```
sworup: @joshxg I just ran the code in emulator and I got this error.

Screenshot 2019-11-07 at 11 52 45

I had encountered this issue before when I used `view` as wripper. You might notice the `text` wrapper rather then `view` wrapper in the previous code. 

You did not get this error at your end?

src/widgets/InfoBadge.js
```diff
@@ -1,129 +1,103 @@
...
-  }
-}
+  return (
+    <View>
+      {children}
+      {pendingCount && (
+        <View style={mainWrapperStyle}>
+          <TouchableHighlight ref={ref} style={touchableContainerStyle} onPress={show}>
+            <Badge value={unfinalisedCountText} />
+          </TouchableHighlight>
```
sworup: @joshxg I think I have a solution for the error above. 

The conditional above excepts the left side expression to be a boolean, simply checking `pendingCount` would give this error. 

```
pendingCount !== 0
```
...fixes this.

joshxg: oops! 4a7447f81dbdf2e2c73011b10e6f036ee3ce7e33

sworup:
👍 for `usePopover` hook.

sworup:
I just extracted the `Badge` component from `react-native-elements`. Thanks for refactoring it and removing `renderNode` @joshxg.

GaryWilletts:
@sworup your screenshot shows a very different screen size/resolution to Josh's. Which one is the standard for 10" tablets?

joshxg:
As spoken about, but to keep the conversation central:

- I installed an apk with the badges on physical tablets and had people who had never seen the badges try them out [by no means a scientific experiment, but extremely useful).
- The consensus was that the badges were too small to read and click. I've adjusted the sizes here: a5d3bb7da03b608f5debbbc023fb344a01a29bb1 - happy if you want to commit different sizes or give me some numbers to try out 

This gives: 

image

I think having a shape with equal width and height would help

Also of note:

- 66% of the users didn't realise the badges were clickable at all. And they also thought that the popover was pretty redundant and not helpful. Others thought that the popover should be clickable to 'navigate' to the actionable. I think we could think about either removing the popover or adjusting the popover slightly to make it more obvious it is touchable.

joshxg:
Also added this ee602865bbcab6021a7969a7b86c21285963d12f - more code extracted in the custom hook

sworup:
@joshxg Your screenshot looks right actually, I would have been satisfied if I got the same interface.

To make it seem the badge is clickable, an option could be to use animation. Make it glow or size up and then become small. I would not recommend adding text on the badge.

Actionable interface sounds like the best next evolution to badges.

This PR is good to merge from my side.

sworup:
> @sworup your screenshot shows a very different screen size/resolution to Josh's. Which one is the standard for 10" tablets?

@GaryWilletts I was using an older Nexus 9 spec. @joshxg I think is using Nexus 10. His device's screen is definitely bigger.

PR #1514, opened 2019-11-07
#1479 Strange ordering when adding rows

### LINKED ISSUE #1479: Adding items to tables causes strange ordering
## Describe the bug

Adding items to tables causes strange ordering.

Items are always added to the head:

`data: [record, ...backingData.slice(0,backingData.length-1)]`

The ordering of `backingData` may change. After adding one item it is at the head of the list. Adding the next item may not have the exact same ordering. For example:

List:

- A
- B

Add C:

- C
- A
- B

Add D:

- D
- A
- B
- C

This is because adding an item removes ALL user defined sorting, and what is left is the implicit database sort which is a standard sorting on the primary key. If it gets sorted before/after adding by whatever the current sort is it just starts to add complexity. I think just adding to the bottom and stop trying to be cool is the right solution?

### To reproduce

As Above

### Expected behaviour

As Above

### Version and device info

- App version: 3.0.0+
- Tablet model: All
- OS version: All

### Additional context

N/A

joshxg:
Fixes #1479 

## Change summary

- Just changes the way state is mutated. I'm 90%, but not 100% there is no edge case where this would not cause a problem, so putting it into develop rather than master as a patch so we can have time to play

## Testing

Not really a check list of testing, more of a story sorry 🤷‍♂ 

*This PR just changes the way state is held, such that when adding a new row to a table, the row is added to the top of the table, and the remaining rows do not change order, but the current filtering and sorting is removed (as it is possibly not correct anymore).*

Create a list of a few items in a supplier invoice. Easiest to number the quantities `1,2,3` etc. Add an item. It should be added to the head and the order of the remaining rows should be stable when adding more items - whether there is a filtering term or sorting.

This applies to adding of all rows `supplier invoice`, `customer invoice` and `supplier requisition`. All the logic is shared, so testing one page is sufficient.

Try to break the app, by using some funky ordering or workflow of sorting/filtering/adding/editing/removing etc

- [ ] App doesn't break

N/A

### DIFF
--- a/src/pages/dataTableUtilities/reducer/tableReducers.js
+++ b/src/pages/dataTableUtilities/reducer/tableReducers.js
@@ -174,20 +174,12 @@ export const toggleStockOut = state => {
  *
  * Also removes the current sorting and filter.
  */
-export const addRecord = (state, action) => {
-  const { data, searchTerm, backingData } = state;
-  const { payload } = action;
-  const { record } = payload;
-
-  // New records are added to index 0 of the table and all filtering and
-  // sorting are removed. To keep the table consistent while adding items,
-  // if there is a search term, the table is currently being filtered so
-  // fetch all the rows again. Otherwise, just use the current data state.
-  const newData = searchTerm ? backingData.slice(1, backingData.length - 1) : data;
+export const addRecord = state => {
+  const { backingData } = state;
 
   return {
     ...state,
-    data: [record, ...newData],
+    data: backingData.sorted('id', true).slice(),
     modalKey: '',
     sortBy: '',
     searchTerm: '',
--- a/src/pages/dataTableUtilities/reducer/tableReducers.js
+++ b/src/pages/dataTableUtilities/reducer/tableReducers.js
@@ -175,13 +175,13 @@ export const toggleStockOut = state => {
  * Also removes the current sorting and filter.
  */
 export const addRecord = (state, action) => {
-  const { backingData } = state;
+  const { data } = state;
   const { payload } = action;
   const { record } = payload;
 
   return {
     ...state,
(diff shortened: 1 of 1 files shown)


src/pages/dataTableUtilities/reducer/tableReducers.js
```diff
@@ -175,13 +175,19 @@ export const toggleStockOut = state => {
...
-  const { backingData } = state;
+  const { data, searchTerm, backingData } = state;
   const { payload } = action;
   const { record } = payload;
 
+  // New records are added to index 0 of the table and all filtering and
+  // sorting are removed. To keep the table consistent while adding items,
+  // if there is a search term, the table is currently being filtered so
+  // fetch all the rows again. Otherwise, just use the current data state.
+  const newData = searchTerm ? backingData.slice(1, backingData.length - 1) : data;
```
Chris-Petty: From me trying to understand, you're slicing from index 1 because `backingData[0]` should already be the new record and otherwise you'd show it twice when adding it on line 190 (note the old implementation didn't do this)? If that is true... couldn't just use `backingData` directly without any logic? If it's not true, then a line is getting lost?

The second parameter to [slice](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice) is optional and should automatically slice the rest of the array without giving an argument. 
```suggestion
  const newData = searchTerm ? backingData.slice(1) : data;
```
Depending on first comments, the first parameter is also optional too... But then not sure why you need slicing at all when you're spreading?

joshxg: hm you're right that was stupid. I did this instead. Maybe also stupid 611df7586450b6271c90060a5deab019dc396d7c -

Issue #1519, opened 2019-11-08
Usage calculations wrong if there is a SI in progress


joshxg:
## Describe the bug

Usage bug -- relatively obscure -- premise is that null is always less than any date:

Usage starts here, just setting up parameters: https://github.com/openmsupply/mobile/blob/ca52751126a6bdfde01459e76a07b6ed33d45bbb/src/database/DataTypes/Item.js#L73-L78

Calls this to calculate usage over the period: https://github.com/openmsupply/mobile/blob/ca52751126a6bdfde01459e76a07b6ed33d45bbb/src/database/DataTypes/Item.js#L200-L212

Which gets the date an item was added here: https://github.com/openmsupply/mobile/blob/ca52751126a6bdfde01459e76a07b6ed33d45bbb/src/database/DataTypes/Item.js#L86-L94

Which is where the problem is (commented): 

```
  get addedDate() {
    if (this.batches.length === 0) return undefined;
    let itemAddedDate = new Date();
    this.batches.forEach(batch => {
      // If a batches added date is null (which is if there is any transaction
       // which has is not confirmed yet), then itemAddedDate is always null
      const batchAddedDate = batch.addedDate;
      itemAddedDate = batchAddedDate < itemAddedDate ? batchAddedDate : itemAddedDate;
    });
    return itemAddedDate;
  }
```

Method to get a transaction batches added date:
https://github.com/openmsupply/mobile/blob/ca52751126a6bdfde01459e76a07b6ed33d45bbb/src/database/DataTypes/ItemBatch.js#L41-L48

### Expected behaviour

When `itemAddedDate` is newer than the AMC lookback time (90 days by default, editable by store custom data), then the period to calculate consumption for should be from the item added date.

The fix is very easy, one line. But this has been in production for a long time and this changes, some usage quite drastically - for example:
OLD:
```
{"usage":0},{"usage":1.4111111111111112},{"usage":22.366666666666667},{"usage":2.1666666666666665},{"usage":1.1875},{"usage":0.18888888888888888}]
```
NEW/BUG FIXED:
```
[{"usage":0},{"usage":6.0476190476190474},{"usage":95.85714285714286},{"usage":41},{"usage":1.1875},{"usage":1.0625}] 
```

### Version and device info

- App version: 1.0 +
- Tablet model: All
- OS version: All

### Additional context

N/A

joshxg:
Fixed by #1521

Issue #880, opened 2019-04-17
[Store CustomData] - how to best deal with store Custom Data in mobile


andreievg:
For vaccine module store.custom_data (on current mobile store) can have a field usesVaccineModule, which enables different interface in mobile app with optional modules

https://github.com/openmsupply/mobile/pull/868 introduces this change where custom_data for current store is saved in settings and is parsed and access later when needed. 

Some comment links
https://github.com/openmsupply/mobile/pull/868#discussion_r275176486
https://github.com/openmsupply/mobile/pull/868#discussion_r275176554

Figure out a better way to store modules.

joshxg:
@andreievg @Chris-Petty @wlthomson @mark-prins 

- `Settings` is a table in realm, where each object is a `key-value` pair.
- Each `value` is a string.

Currently, on incoming `[Store]` records, if the store is the mobile store, we store the `custom_data` in `Settings`, under the key `THIS_STORE_CUSTOM_DATA` with the value as a stringified JSON object.

In order to utilize some of this `customData` during ivory coast, we had to use node `global`s in order to customize logic, instead we can just use msupply `global` `Settings` 🤣 

Instead of storing all of `[Store]custom_data`, lets extract what we need and put them as their own key:value pairs in `Settings`. 

I propose simply adding some `Setting`s key value pairs (note that value is a string atm :():

```
THIS_STORE_MODULES: {
    *all modules listed here with true/false*
}
```
alternatively single key value pairs:

```
THIS_STORE_VACCINED_UP: false
THIS_STORE_DISPENSED_UP: false
```

We could also have a..

```
THIS_STORE_CUSTOMIZATIONS: {
    lookbackForSomeAMCThing: xx
}
```

When methods are being extended to utilize some different functionality, we can extract the current logic, and make different functions quite easily. Making sure there would be no regressions with current logic, but extending functionality with new logic.

I think this idea needs fleshing out and some polish, but I think it is a good option. Would require a few things, migration code, documenting settings well and changing some (small) existing code. (Also can remove `Store.customData`

joshxg:
Also will just add that described above is currently on the Vaccines branch, NOT in develop.

joshxg:
Also with the new `DataTable`, we can customise routes, without effecting other pages

i.e.

```
const gotoMenuPage(){
    const menuPageCustomizations = Settings.get(THIS_STORE_MENU_PAGE_CUSOMIZATIONS);
    switch (menuPageCustomizations){
[...]

PR #1520, opened 2019-11-08
#1491 remove global usage

### LINKED ISSUE #1491: Remove use of global object

`AMC` calculation uses the `global` object in `Settings`

Not 100% sure of the solution, but certainly removing `global` is the goal. The tissue should look into implementation

N/A

## Additional context

N/A

joshxg:
Fixes #1491 

## Change summary

Main goal of this PR was to remove the usage of the `global` object in `MobileAppSettings`. This meant changing usage logic. Ended up extracting the consumption logic to a method outside  - reasoning
- Having this logic extracted is easier if (when) we need to change usage logic. Will be far easier to create a new algorithm and avoid regressions (i.e. could use an `IF customisation: usage1() else usage2()`). "Pseudo dependency injection" - i'd argue a lot of the business logic should be extracted in such a way
- The logic was a little convoluted. Calling helpers from `Item` -> `ItemBatch` -> `TransactionBatch` was a bit hard to follow, and they've never been reused. Is a lot easier to digest in a small sequential method, in my opinion

Additional changes that sneaked in:
- Saving all fields that are in `[store]custom_data` which effect mobile as their own settings keys [This is an initial step in hopefully, a better way to organise/manage this as it is inevitable this is going to be used more(?)]
- Instead of refresh the `global` object, 'refresh' the settings keys in `PostSyncProcessor`. I wouldv'e liked to use the processor more as it is intended, but we don't have `Store` in schema, so unable to queue up functions for that data type.
- When refreshing the `global` object, also accounts for *removing* keys, which was not being done before
- Also added a sync action for completion of sync, (I shouldn't have done this in this PR, sorry) such that the module reducer can subscribe and update the interface if modules are changed. This was not being done before, you had to restart the app)

## Testing

*Note: Please let me know if further details are needed if this is ever tested*

- [ ] Usage calculation is the 'same as before'!
- [ ] using `monthsLeadTime` custom data is synced (Can use realm browser)
- [ ] using `monthlyConsumptionEnforceLookBackPeriod` custom data is synced (Can use realm browser)
- [ ] using `monthlyConsumptionLookBackPeriod` custom data is synced (Can use realm browser)
- [ ] using `usesDispensaryModule` custom data is synced (Can use realm browser)
- [ ] using `usesDispensaryModule` custom data is synced (Can use realm browser)

- Need to write docs for all this custom data stuff
- Really wanted to rename the monthly consumption settings but would have to write redux persistance migration code etc. Might do in a new issue, but might not worry about it. 
- Found #1519 during this
- In a way this is also starting to address #880 - I would like to see a `DataType` field added to `Settings`.. so it would be nicer to manage different data types


src/mSupplyMobileApp.js
```diff
@@ -151,6 +152,7 @@ class MSupplyMobileAppContainer extends React.Component {
     } else {
       this.postSyncProcessor.processRecordQueue();
     }
+    dispatch(syncCompleteTransaction());
```
joshxg: This naming doesn't really make sense. Want a way to signal 'after post processing sync', but don't want to pass dispatch to post sync processor

src/pages/MenuPage.js
```diff
@@ -56,7 +56,7 @@ const Menu = ({
 
   const MenuButton = useCallback(
     props => <Button style={menuButton} textStyle={buttonText} {...props} />,
-    []
+    [usingDispensary, usingModules]
```
joshxg: This will be messy and not good if there's ever 4+ modules..

src/settings/index.js
```diff
@@ -18,6 +18,19 @@ export const SETTINGS_KEYS = {
   THIS_STORE_NAME_ID: 'ThisStoreNameId',
   THIS_STORE_TAGS: 'ThisStoreTags',
   THIS_STORE_CUSTOM_DATA: 'ThisStoreCustomData',
+  CONSUMPTION_LOOKBACK_PERIOD: 'monthlyConsumptionLookBackPeriod',
+  CONSUMPTION_ENFORCE_LOOKBACK: 'monthlyConsumptionEnforceLookBackPeriod',
+  MONTHS_LEAD_TIME: 'monthsLeadTime',
+  VACCINE_MODULE: 'usesVaccineModule',
+  DISPENSARY_MODULE: 'usesDispensaryModule',
```
joshxg: `uses` `using` and no `prefix` is confusing

src/sync/PostSyncProcessor.js
```diff
@@ -83,6 +86,39 @@ export class PostSyncProcessor {
...
+   * After syncing, recheck and save all of these fields.
+   */
+  processCustomData = () => {
+    // This stores custom data, stored as a JSON stringified object
+    const customData = this.settings.get(SETTINGS_KEYS.THIS_STORE_CUSTOM_DATA);
+
+    try {
+      // If there is no custom data, use an empty object and clear all
+      // customizable settings (set to an empty string)
+      const parsedCustomData = JSON.parse(customData || '{}');
```
joshxg: this is pretty ugly

src/utilities/dailyUsage.js
```diff
@@ -0,0 +1,62 @@
...
+};
+
+/**
+ * Returns if this store is customized such that usage should force a lookback period,
+ * rather than using an items added date.
+ */
+const getAMCEnforcementLookback = () => {
+  const amcEnforcementString = UIDatabase.getSetting('monthlyConsumptionEnforceLookBackPeriod');
+  return amcEnforcementString === 'true';
+};
```
joshxg: this combined with the `checkModule` helper in the `ModuleReducer`.. lots of spread out helpers

src/utilities/dailyUsage.js
```diff
@@ -0,0 +1,62 @@
...
+  const numberOfUsageDays = millisecondsToDays(usagePeriod);
+
+  const usage = UIDatabase.objects('TransactionBatch')
+    .filtered('itemBatch.item.id == $0', itemId)
+    .filtered('transaction.type == $0', 'customer_invoice')
+    .filtered('transaction.confirmDate >= $0 && transaction.confirmDate <= $1', startDate, dateNow)
+    .reduce((sum, { totalQuantity }) => sum + totalQuantity, 0);
+
+  return usage / numberOfUsageDays || 1;
+};
```
joshxg: Wondering if it would be better to split this into two (three) functions - `dailyUsage` which would be a splitter which would call either `normalUsage` or `usageWithLookbackEnforcement` (or `usageWithItemAddedLogic`, `usageJustOverTheLastXDays`..) .. maybe not for now

src/database/DataTypes/ItemBatch.js
```diff
@@ -80,25 +78,7 @@ export class ItemBatch extends Realm.Object {
...
-      startDate,
-      endDate
-    );
-
-    return getTotal(transactionBatches, 'usage');
-  }
-
-  /**
-   * Add a transaction batch to be associated with this batch.
+   * Ad˚d a transaction batch to be associated with this batch.
```
mark-prins:
```suggestion
   * Add a transaction batch to be associated with this batch.
```

src/utilities/dailyUsage.js
mark-prins: yeah, not for now. it's ok as is

PR #1523, opened 2019-11-10
#1518 Corrected select all state

### LINKED ISSUE #1518: StocktakeManagePage all selected is sometimes incorrect
## Describe the bug

The `AllSelected` toggle is sometimes incorrect.

### To reproduce

1. Go to `StocktakeManagePage`
2. Filter by some string
3. Click all selected
4. remove filter
5. See error

### Expected behaviour

N/A

### Version and device info

- App version: 3.0+
- Tablet model: All
- OS version: All

### Additional context

N/A

joshxg:
Fixes #1518 

## Change summary

- When selecting all, all items are actually selected. This is the behaviour I would expect.

## Testing

- [ ] Filter `StocktakeManage` page to a sub set of items and pres all items selected. All items should be selected (after removing the filter)

I think we are looking to change this page anyway (Or at least I would like to!), but probably not before the next version?

### DIFF
--- a/src/pages/dataTableUtilities/reducer/rowReducers.js
+++ b/src/pages/dataTableUtilities/reducer/rowReducers.js
@@ -82,13 +82,13 @@ export const deselectAll = state => {
  * Toggles selection of all rows between selected/false
  */
 export const toggleSelectAll = state => {
-  const { data, keyExtractor, dataState, allSelected } = state;
+  const { keyExtractor, dataState, allSelected, backingData } = state;
 
   const newDataState = new Map(dataState);
 
   const newSelectionState = !allSelected;
 
-  data.forEach(item => {
+  backingData.forEach(item => {
     const rowKey = keyExtractor(item);
     newDataState.set(rowKey, { ...newDataState.get(rowKey), isSelected: newSelectionState });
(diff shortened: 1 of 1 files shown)


Chris-Petty:
Funnily enough, the original implementation was a bit more complex. The state of selecting all items was a reflection of if all the current filtered items was selected. E.g. you could search for all items starting with "para" and hit the toggle, it'd select all those items. When you clear the filter it'd just update the state of the toggle to be off.

Alternatively, if you filtered the table to 1 row, then selected it, the toggle would turn on. Likewise pressing the toggle again would remove selection from only items in the table.

In reality I think no one actually uses it like this lol, it is usually omitted from trainings 'cause it's edgy power user tricks. But it is snazzy if you want to just stocktake specific sets of items rather than literally all items.

Anyway, to maintain that behaviour the code you've changed was fine. You'd instead need to update the toggle state by checking if every row in data is selected, every time you change `data` or `selection`.

joshxg:
I did think about that. I figured that for now, at least, make this change. With discussions regarding a unified UI for item selection and potential changes there wouldn't be too much point investing time, right now.

PR #1524, opened 2019-11-10
#1426 Navigation debouncing

### LINKED ISSUE #1426: Navigation should be debounced
## Describe the bug

Navigating can often lead to the user error of navigating back more than desired

### To reproduce

1. Go to A page nested at least two deep in the navigation stack
2. Click on the back top left button a couple times in quick succession

### Expected behaviour

Should be debounced ~250ms to prevent this user error

### Version and device info

2.3+

### Additional context

N/A

joshxg:
Fixes #1426 

## Change summary

- Just adds debouncing. I've been double clicking a lot and it's super annoying 🤣 

## Testing

- [ ] Spam clicking the top left back button doesn't navigate back twice (should be debounced if press within .4 seconds)

N/A

### DIFF
--- a/src/mSupplyMobileApp.js
+++ b/src/mSupplyMobileApp.js
@@ -1,3 +1,4 @@
+/* eslint-disable react/sort-comp */
 /* eslint-disable no-undef */
 /* eslint-disable no-console */
 /**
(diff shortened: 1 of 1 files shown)


Chris-Petty:
Tested on tablet, seems fine

PR #1525, opened 2019-11-10
#1450 Localization update

### LINKED ISSUE #1450: The french translation for Login or the alert for a wrong password are not right
## Describe the bug

The french translation for Login is displayed as "Se connecter or Ouverture de session". It should be either "Se connecter" OR "Ouverture de session" but not both. You can choose either !!!

Also when you don't have the right password, the alert that is displayed has not been translated

### To reproduce

1. Open msupply mobile
2. Change language to french
3. See error
4. Put a wrong password
5. See Error

### ScreenShot

image

image

### Expected behaviour

Translation should be either "Se connecter" OR "Ouverture de session" but not both

For the alert Translation should be "Utilisateur ou mot de passe incorrect"

### Version and device info

- App version: 3.0.0
- Tablet model: Galaxy Tab A
- OS version: Android 9.0

### Additional context

I remember having put both as translation to choose one or the other but I was probably unclear or even forgot to say that a choice needed to be made...

joshxg:
Fixes #1450 

## Change summary

Updates some incorrect translations

## Testing

- [ ] The translations described in the issue are correct!

N/A

### DIFF
--- a/src/localization/authStrings.js
+++ b/src/localization/authStrings.js
@@ -16,9 +16,9 @@ export const authStrings = new LocalizedStrings({
   },
   fr: {
     logging_in: 'Connexion en cours',
-    login: 'Se connecter or Ouverture de session',
+    login: 'Se connecter',
     password: 'Mot de passe',
-    user_name: "Identifiiant or Nom d'utilisateur",
+    user_name: 'Identifiiant',
   },
   gil: {
     logging_in: 'Taninga bwa e na uki',


src/localization/authStrings.js
```diff
@@ -13,12 +13,13 @@ export const authStrings = new LocalizedStrings({
...
     email: 'E-mail Address',
+    invalid_username_or_password: 'Invalid username or password',
   },
   fr: {
     logging_in: 'Connexion en cours',
-    login: 'Se connecter or Ouverture de session',
+    login: 'Se connecter',
     password: 'Mot de passe',
-    user_name: "Identifiiant or Nom d'utilisateur",
+    user_name: 'Identifiiant',
```
DhanyaHerath: Is this a typo? `Identifiant` maybe? (google translate seems to agree)

alainsussol: @DhanyaHerath Yes it is a typo. Should be as you said ie Identifiant

alainsussol:
@joshxg Identifiant ou mot de passe invalide

src/localization/authStrings.js
joshxg: oops. fe50743

joshxg:
@DhanyaHerath Please re-review

diegoevangelisti:
- [x] The translations described in the issue are correct!

Test done

PR #1581, opened 2019-11-25
#1580 Badges crash on navigating

### LINKED ISSUE #1580: Badges crash when navigating quickly
## Describe the bug

When opening a popover from a badge and quickly navigating, the app crashes

Throwing: `java.lang.Double.doubleValue()' on a null object reference`

### To reproduce

1. Click on a badge to open the popover
2. Click on a button to navigate before the popover opens

badgeCrash

### Expected behaviour

The app shouldn't crash

### Version and device info

- App version: 3.1 development
- Tablet model: all
- OS version: all

### Additional context

Not sure of the cause, exactly

joshxg:
Fixes #1580 

## Change summary

- Turns out it was because of the badge being unmounted and the ref being reset. Which is what the changes in `MenuPage` were from. So the popover was trying to reference a null pointer.
- When this was fixed, there was an issue that the popover would not dissapear after navigating away. So this is where the change to `usePopover` using `useNavigationFocus` and the addition of `willBlur` in `useNavigationFocus` comes from.

## Testing

- [ ] When clicking pressing a badge to open a popover, and tapping a button to navigate, the app does not crash!
- [ ] When clicking pressing a badge to open a popover, and tapping a button to navigate, the popover is not visible
- [ ] When clicking pressing a badge to open a popover, and tapping a button to navigate, when returning to the previous page (menu page) the popover is closed.

You could also fork/pr into this repo to add a simple line here:  https://github.com/SteffeyDev/react-native-popover-view/blob/6b2c0dea3529f5dd821d6dd98394a45a096d79f2/src/Utility.js#L32-L36

`if (!ref) return false` to 'fix' this 

TIL: `defaultProps` and `propTypes` are not compatible with `forwardRef`

### DIFF
--- a/src/hooks/useNavigationFocus.js
+++ b/src/hooks/useNavigationFocus.js
@@ -7,37 +7,43 @@
 import { useEffect, useRef } from 'react';
 
 /**
- * Simple custom hook which subscribes to navigation focus event,
- * (navigating to a page), which will invoke the passed callback
- * when navigating to it.
+ * Simple custom hook which subscribes to navigation focus/blur events,
+ * (navigating to/from a page), which will invoke the passed callback.
  *
  * Main use case is popping a page from the stack and returning to
- * a page whose state has been altered but the UI has not updated.
- *
+ * a page whose state has been altered but the UI has not updated or
+ * triggering a callback when navigating away from a page
  * - Only events for the calling page are subscribed to.
  * - Initial navigation focus event is NOT captured.
  *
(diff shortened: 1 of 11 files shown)


diegoevangelisti:
- [x] When clicking pressing a badge to open a popover, and tapping a button to navigate, the app does not crash!
- [x] When clicking pressing a badge to open a popover, and tapping a button to navigate, the popover is not visible
- [x] When clicking pressing a badge to open a popover, and tapping a button to navigate, when returning to the previous page (menu page) the popover is closed.

Test done

Issue #1585, opened 2019-11-25
Patient helpers for payments


joshxg:
## Is your feature request related to a problem? Please describe.

Need a `Name` helper for a patients available credit

## Describe the solution you'd like

- a helper `availableCredit` which returns the amount of available credit for a patient

## Describe alternatives you've considered

N/A

## Additional context

N/A

joshxg:
Closed by #1587

PR #1588, opened 2019-11-25
#1584 CashIn and CashOut Transactions

### LINKED ISSUE #1584: createCashOut and createCashIn utility methods

- Two utility methods which will sequentially create the required records for `cashOut` and `cashIn` transactions

- `createCashIn(Int amount) : `
    - Create a CC transaction - where the `total` is equal to `-(amount)`
    - Create a RC transaction - Simply related to the patient
    - Create an RC cash_in line - where the `total` is equal to `amount`, and `linkedTransaction` is related to the `CC` transaction

**Note: This will require a `Patient` helper - `availableCredit`, which will return the amount of available credit (`CI.outstanding` - `CC.outstanding`). Possibly require an `off-set CI` createRecord case?

- `createCashOut(Int amount) : `
    - create a PS transaction - simply related to the patient
    - create a CI transaction - where `outstanding` and `total` are equal to `amount`
    - Create a PS Line, where the `total` is equal to `amount` and `linkedTransaction` is related to the `CI`

N/A

## Additional context

N/A

joshxg:
## BRANCHED FROM: #1585 

Fixes #1584 

## Change summary

- Adds create methods for `CashIn`, `CashOut` and `OffsetCI`'s
- Made pretty much all the fields of a `TransactionBatch` optional. This isn't great. Considering making another object type for these cash ledge trans line things. I need to confirm if they are or aren't always `cash_in`/`cash_out` type or if there are another field that can be used to determine if they should be a `TransactionBatch`, or some other object type before I can do that, as if they come in during incoming sync and the `Transaction` hasn't been fetched yet, there is more complicated logic for deleting/creating etc when it's found what kind of object type a `trans_line` should be

## Testing

N/A

N/A


joshxg:
Possibly a Cash Out transaction should create `CustomerCreditLine`s on the `CustomerCredit`s where the amount is coming from

src/database/DataTypes/Name.js
```diff
@@ -56,7 +56,7 @@ export class Name extends Realm.Object {
     const sumOfCredits = this.customerCredits.sum('outstanding');
     const sumOfDebits = this.customerInvoices.sum('outstanding');
 
-    return Math.abs(sumOfDebits - sumOfCredits);
+    return Math.abs(sumOfCredits + sumOfDebits);
```
mark-prins: hm? I'm not following the logic. I would have thought that the available credit is `sumOfCredits - sumOfDebits` though I'm not convinced that mSupply is actually doing that!

src/database/DataTypes/TransactionBatch.js
```diff
@@ -141,18 +141,18 @@ TransactionBatch.schema = {
...
-    packSize: 'double',
-    numberOfPacks: 'double',
+    packSize: 'double?',
+    numberOfPacks: 'double?',
     numberOfPacksSent: { type: 'double', optional: true },
     transaction: 'Transaction',
     note: { type: 'string', optional: true },
-    costPrice: 'double',
-    sellPrice: 'double',
+    costPrice: 'double?',
```
mark-prins: agree that it isn't great having so many optional fields.
*shrugs helplessly*

src/database/DataTypes/Name.js
mark-prins: ah yes. that makes sense, thanks

PR #1591, opened 2019-11-25
#1586 Utility method: pay

### LINKED ISSUE #1586: Payment helper pay()

A utility method which will 'pay' a prescription and correctly handle credit and overpayments

A utility method `pay()`:

- finds the amount being paid and compares against the cost of the prescription resulting in three paths, using credit, even, using overpayment
- if credit is being used: make sure there is enough credit available, and error if not. 
- Create an RC transaction and a RC Transline, which is related to the prescription and has a `total` equal to the amount of cash being paid
- When using credit:
    - Iterate over all the patients CC transactions from oldest to newest, finding the ones which have available credit. For each CC transaction credit will be used from, create a CC Line for that transaction, where the `total` is equal to the amount of credit being used from that transaction and a RC Line, where the `total` is equal to the negative amount being taken, and the `linkedTransaction` is this prescription.
- If using an overpayment, create a `cashIn` transaction equal to the value being overpaid.

## Additional context

joshxg:
## BRANCHED FROM: #1588 

Fixes #1586 

## Change summary

- utility method to pay a prescription off - adding credit for overpayments or using credit for underpayments
- Still not setting exactly all the fields that should be - for example time should probably be set
- Excessive errors is mainly for development - can be cut down possibly a bit later

## Testing

N/A

N/A

### DIFF
--- a/src/utilities/modules/dispensary/pay.js
+++ b/src/utilities/modules/dispensary/pay.js
@@ -6,6 +6,34 @@ import { UIDatabase } from '../../../database/index';
  * Sustainable Solutions (NZ) Ltd. 2019
  */
 
+/**
+ * Helper method to pay off a prescription. Given a cash amount
+ * a script and a patient, all receipts, credits and cash in
+ * transactions will be created and the script will be finalised.
+ *
+ * Paying a script has three paths: underpay, exactpay and overpay.
+ * All payments create a Receipt with a ReceiptLine.
+ *
+ * An underpayment creates a CustomerCreditLine and a ReceiptLine
+ * for each source of credit (CustomerCredit) that credit was used from.
+ * For example if a Patient has overpaid the last two prescriptions, then
+ * there would be two sources of credit. If both sources were used to pay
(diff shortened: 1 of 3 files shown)


src/utilities/modules/dispensary/pay.js
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * @param {Name} patient       A Name that is a patient
+ * @param {Transaction} script A Transaction that is a prescription
+ * @param {Number} cashAmount  The cash amount being paid for the script.
+ */
+
+export const pay = (currentUser, patient, script, cashAmount) => {
+  if (!patient.isPatient) throw new Error('Patient is not a patient');
+
+  if (!script.isPrescription) throw new Error('Script is not a script');
```
mark-prins: 😄 these two errors made me smile. How can a patient not be a patient!

joshxg: this was unintentional 🤣

PR #1595, opened 2019-11-26
#1592 refund utility

### LINKED ISSUE #1592: Payment utility method: refund

Need to be able to 'refund'/'credit'/'return' goods. Initially this will just be for prescriptions, but may extend to supplier credits.

A utility method, `refund(TransactionBatch[] batches)`

- Will take an array of `TransactionBatch`s - creates a `CustomerCredit` 
- Adds a matching but inverse `TransactionBatch` for each passed batch to the `CustomerCredit` created, with a type of `"stock_in"`

N/A

## Additional context

N/A

joshxg:
Fixes #1592 

## Change summary

- Adds refund() utility
- Added a `createRecord` for a `refundLine` - there's really just one more method to write - `cancel`, and that should cover all our use cases. Then it will be very easy to see where the shared/duplicated code is and can revert a few things. Finding it much easier this way

## Testing

N/A

N/A

### DIFF
--- a/src/utilities/modules/dispensary/refund.js
+++ b/src/utilities/modules/dispensary/refund.js
@@ -19,8 +19,8 @@ import { UIDatabase } from '../../../database';
  */
 export const refund = (currentUser, patient, batches) => {
   const forSamePatient = batches.every(({ transaction }) => {
-    const { otherPartyName } = transaction;
-    return patient.id === otherPartyName.id;
+    const { otherParty } = transaction;
+    return patient.id === otherParty?.id;
   });
 
   if (!batches.length) throw new Error('Trying to refund void');
--- a/src/utilities/modules/dispensary/refund.js
+++ b/src/utilities/modules/dispensary/refund.js
@@ -9,7 +9,7 @@ import { UIDatabase } from '../../../database';
 /**
  * Utility to refund a collection of TransactionBatch records. Creates a
  * single CustomerCredit and for each TransactionBatch that should be
- * refund, a RefundLine record, which enters the batch back into stock
+ * refunded, a RefundLine record, which enters the batch back into stock
  * and adds credit for the patient.
  *
  *
--- a/src/utilities/modules/dispensary/refund.js
+++ b/src/utilities/modules/dispensary/refund.js
@@ -6,6 +6,17 @@
 import { createRecord } from '../../../database/utilities';
 import { UIDatabase } from '../../../database';
 
+/**
+ * Utility to refund a collection of TransactionBatch records. Creates a
+ * single CustomerCredit and for each TransactionBatch that should be
+ * refund, a RefundLine record, which enters the batch back into stock
+ * and adds credit for the patient.
+ *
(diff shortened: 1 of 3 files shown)


src/utilities/modules/dispensary/refund.js
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Utility to refund a collection of TransactionBatch records. Creates a
+ * single CustomerCredit and for each TransactionBatch that should be
+ * refund, a RefundLine record, which enters the batch back into stock
+ * and adds credit for the patient.
+ *
+ *
+ * @param {User}               currentUser The currently logged in user.
+ * @param {Name}               patient     The patient to refund to
+ * @param {TransactionBatch[]} batches     Array of batches to be refund
```
mark-prins:
```suggestion
 * @param {TransactionBatch[]} batches     Array of batches to be refunded
```

src/utilities/modules/dispensary/refund.js
```diff
@@ -0,0 +1,44 @@
...
+ */
+export const refund = (currentUser, patient, batches) => {
+  const forSamePatient = batches.every(({ transaction }) => {
+    const { otherPartyName } = transaction;
+    return patient.id === otherPartyName.id;
```
mark-prins: `otherPartyName` couldn't be null, could it?

src/utilities/modules/dispensary/refund.js
```diff
@@ -0,0 +1,44 @@
...
+  });
+
+  if (!batches.length) throw new Error('Trying to refund void');
+  if (!forSamePatient) throw new Error('Batches for different patients!');
+
+  const sumOfCredit = batches.reduce((acc, { total }) => acc + total, 0);
```
mark-prins: are the refunded transactions marked in some way, so they can't be refunded again?

joshxg: No. Need to make another desktop issue to find out different/correct behaviour. A new field may need to be added to `[trans_line]` or something as you can just refund as many times as you like.

Can even refund the same item multiple times in the same credit!

image

src/utilities/modules/dispensary/refund.js
joshxg: cbe90b871c4bdc1e9864c3e184e393690062b99d

src/utilities/modules/dispensary/refund.js
joshxg: oops thats broken completely, cheers - e375cbae25c5ae62762dd3a0d19651c41e921df8

src/utilities/modules/dispensary/refund.js
joshxg: Hope you don't mind me addressing this elsewhere, possibly after a discussion!

mark-prins: yes, I think it needs to be sorted separately - and especially because desktop needs a tidy up first!

PR #1612, opened 2019-11-27
#1603 PopoverDropDown

### LINKED ISSUE #1603: Popover DropDownMenu component

Currently we use large modals for selection actions. This is slow. A faster and less intrusive operation would be good.

A popup-dropdown menu

## Implementation

Probably using `Popover` which was implemented from badges

N/A

## Additional context

N/A

joshxg:
Fixes #1603 

## Change summary

- Adds a basic `PopOverDropDown` component

Bit messy and more hard coded than I'd like. Not very flexible or dynamic but good to have a base component for when the concrete use case comes

dropdownexample1

(The surrounding border seemed more prominent in non-gif format 🤷‍♂ )

## Testing

N/A

N/A

### DIFF
--- a/src/widgets/PopoverDropDown.js
+++ b/src/widgets/PopoverDropDown.js
@@ -92,6 +92,7 @@ const DropDown = ({ anchorRef, isVisible, options, onSelection, onClose, title }
  * left/right/bottom/top position which has the most free space.
  *
  * Usage:
+ * @prop {Object}      navigation The navigation object passed by withNavigation
  * @prop {React.Node}  BaseComponent Any component which will be wrapped in a Touchable.
  * @prop {Array}       options An array of strings, the options to choose.
  * @prop {Func}        onSelection The callback invoked when a selection is made.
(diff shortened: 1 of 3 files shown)


src/widgets/PopoverDropDown.js
```diff
@@ -0,0 +1,159 @@
...
+        <TouchableOpacity onPress={show} ref={ref}>
+          <BaseComponent />
+        </TouchableOpacity>
+        <DropDown
+          anchorRef={ref}
+          options={options}
+          isVisible={visible}
+          onSelection={onSelection}
+          onClose={close}
+          title="Payment Types"
```
mark-prins: would the props for PopoverDropDown be coming later? The title, options and onSelection are set by the caller aren't they?

joshxg: oops - have added the title prop 1409643  - yeah the props you've mentioned would be supplied, as well as a `BaseComponent` to act as an anchor.

Have added a docstring.,. hoping that it's enough even though I think I know it's not

PR #1617, opened 2019-11-27
#1615 deleting doesnt work

### LINKED ISSUE #1615: Removing a stocktake from the Current list doesn't work properly
## Describe the bug

After removing a stocktake from the Current list, it's not deleted until the user goes to Menu Page, comes back to Stocktages page and remove the mentioned stocktake again.

deleted-stocktake-doesnt-go

### To reproduce

1. Go to Stocktakes page.
2. Remove a stocktake from the Current list.
3. See error

### Expected behaviour

Removing a stocktake from the current list must work on the first attempt.

### Version and device info

- App version: v3.0.4
- Tablet model: Lenovo TB-X304L
- OS version: Android 8.1.0

Also tested on Android Emulator (Name: WXGA API 28, OS version: Android 9.0)

joshxg:
Fixes #1615 

## Change summary

Bug was that after navigating back, the `currentRoute` wasn't being set correctly. Solution is to pass the previous route name in the `BACK` navigation, so you can set the `currentRoute` in state when navigating back

## Testing

 Navigate to `CustomerInvoices page - create a new invoice. Go back, try and delete.
- [ ] The invoice is deleted
- [ ] The invoice is removed from the list
- [ ] Try again on another page i.e. `SupplierInvoice`, `StocktakesPage` 

I think `PagesReducer` has gotten to hacky for it to work 'correctly'. Probably should be refactored

### DIFF
--- a/src/mSupplyMobileApp.js
+++ b/src/mSupplyMobileApp.js
@@ -42,6 +42,7 @@ import { UIDatabase } from './database';
 import globalStyles, { textStyles, SUSSOL_ORANGE } from './globalStyles';
 import { UserActions } from './actions';
 import { debounce } from './utilities';
+import { prevRouteNameSelector } from './navigation/selectors';
 
 const SYNC_INTERVAL = 10 * 60 * 1000; // 10 minutes in milliseconds.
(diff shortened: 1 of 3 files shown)


diegoevangelisti:
Great job again @joshxg, It works well for me now!

PR #1620, opened 2019-11-28
#1601 SimpleTable component

### LINKED ISSUE #1601: SimpleTable component

The main `DataTable` component is quite bulky and not suitable for simple lists.

A simple table component which would be a simplified interface over a `FlatList` which would be designed to be non-editable, provide default props and styles

## Implementation

A simple table component which would be a simplified interface over a `FlatList` which would be designed to be non-editable, provide default props and styles

N/A

## Additional context

N/A

joshxg:
Fixes #1601 

## Change summary

-`SimpleTable` Component - have had other uses cases i.e. in vaccines a similar component was made. 
- The regular `DataTable` is a bit heavy for all use cases, especially very simple ones of just showing unchanged, uneditable data

SimpleTable

## Testing

N/A

Not sure about the page export diffs.. eslint/prettier problem I think, will make a new issue


src/widgets/SimpleTable.js
```diff
@@ -0,0 +1,127 @@
...
+      ListHeaderComponent={renderHeaders}
+    />
+  );
+});
+
+SimpleTable.propTypes = {
+  data: PropTypes.array.isRequired,
+  columns: PropTypes.array.isRequired,
+  selectRow: PropTypes.func.isRequired,
+  extraData: PropTypes.object.isRequired,
```
mark-prins: the name of this one didn't reflect the intention for me. it's like `isSelected` more than `extraData`. is there an intention to use it for something else as well?

src/widgets/SimpleTable.js
```diff
@@ -0,0 +1,127 @@
...
+      <Cell
+        value={rowData[key]}
+        viewStyle={cellContainer[alignText]}
+        textStyle={cellText[alignText]}
+        isLastCell={index === columns.length - 1}
+        width={width}
+        key={`${id}${key}`}
+      />
+    ));
+  }, []);
```
mark-prins: ` [rowData]`? or doesn't matter? It's not likely to re-render anyway is it?

src/widgets/SimpleTable.js
joshxg: you're right - it is a bit weird. Mainly because I appear to have forgotten to put this back: 1cd751b7661c040b1fe18e22dfaccd8b133370bd [prop is essentially a trigger to refresh]. Hopefully that makes it clearer, although the name still doesn't help..

src/widgets/SimpleTable.js
joshxg: `rowData` is a parameter not a free variable

src/widgets/SimpleTable.js
mark-prins: the name still seems odd to me, but no biggie. I think merge anyway

src/widgets/SimpleTable.js
mark-prins: will take your word for it ( which means, I dunno. seems to work fine, so let's just leave it )

Issue #1624, opened 2019-12-01
StepsTracker styles, onPress and naming


joshxg:
## Is your feature request related to a problem? Please describe.

`StepsTracker` component styles are a bit nutty. Should have some sort of max height so that having <= 3 steps doesn't have gigantic step numbers. The `onPress` should be disabled on the CURRENT step. The whole component should be renamed `ProgressTracker` (Or something else that isn't `StepsTracker`

## Describe the solution you'd like

Should have some sort of max height so that having <= 3 steps doesn't have gigantic step numbers. The `onPress` should be disabled on the CURRENT step. The whole component should be renamed `ProgressTracker` (Or something else that isn't `StepsTracker`

## Implementation

Should have some sort of max height so that having <= 3 steps doesn't have gigantic step numbers. The `onPress` should be disabled on the CURRENT step. The whole component should be renamed `ProgressTracker` (Or something else that isn't `StepsTracker`

## Describe alternatives you've considered

N/A

## Additional context

N/A

joshxg:
Closed by #1629

PR #1625, opened 2019-12-01
#1599 wizard component

### LINKED ISSUE #1599: Wizard component

We don't have a Wizard component

A component which integrates the two components #1598 and #1597 into a single component for a wizard

N/A

## Additional context

N/A

joshxg:
Fixes #1599 

## Change summary

- Adds a simple layout component `Wizard`.
- Changes `StepsTracker` to use counting from 0 rather than 1 to be consistent with tabs usage. Still displays from 1 - annoying.. - right decision? 🤷‍♂ 
- Changes in `MultiSelectList` and `widgets/index` are to alphabetically sort and remove eslint by-passes - sorry it was an unavoidable situation, I had to!
- Styles of `StepsTracker` and a couple other problems I made an issue for: #1624 

So it'd sort of look like this (with better styles from a non-numpty hopefully!!):
wizard1

And code would look something like:

```
const TabOne = () => <View style={{ flex: 1, backgroundColor: 'red' }}></View>;
const TabTwo = () => <View style={{ flex: 1, backgroundColor: 'blue' }}></View>;
const TabThree = () => <View style={{ flex: 1, backgroundColor: 'yellow' }}></View>;
const tabTitles = ['Title One', 'Title Two', 'Title three'];

export const TestPage = () => {
  const [currentStep, setCurrentStep] = React.useState(0);
  return (
    <Wizard
      tabs={[TabOne, TabTwo, TabThree]}
      titles={tabTitles}
      onPress={x => setCurrentStep(x)}
      currentTabIndex={currentStep}
    />
  );
};
```

## Testing

N/A

Fun facts : Trying to avoid passing props to each inidividual tab. It's possible by using a prop `screenProps` on the navigator container, but that is an object and managing references can be a nightmare and can cause every tab to re-render with simple mistakes. It's also being deprecated! 

I spent a little bit of time making a component to do this:

```
<TabNavigator>
    <TabOne/>
    <TabTwo/>
    <TabThree/>
</TabNavigator>
```

Which is super nice imo! Unfortuantely I realised that react-navigation 5.0 is being completely changed to have navigation work like this: https://blog.expo.io/announcing-react-navigation-5-0-bd9e5d45569e - so I just stopped and will use that when it's released. Only in alpha at the mo : "More adventurous users are already using it in production but it’s not for everyone. " -- lets be adventurous 🤣


joshxg:
me too. I *think* it's going to be easier to manage re-renders keeping them separate, but it might make no difference, in which case using an array of objects is most probably nicer. Will definitely check it out when actually using it in a page

PR #1655, opened 2019-12-06
#1602 icons for dispensing

### LINKED ISSUE #1602: Extra Icons for dispensing

We don't have easy interfaces for icon components which will be used for dispensing

Easy interfaces to icon components which will be used for dispensing

## Implementation

Code which is an easy interface to icon components which will be used for dispensing

N/A

## Additional context

N/A

joshxg:
Fixes #1602 

## Change summary

- Added three icon components..
- Went off road and tried to make all the icons consistent. Making them all functions and giving them all an `Icon` suffix rather than random no suffix/component suffix and some being functions other objects etc

## Testing

N/A

N/A


src/widgets/icons.js
```diff
@@ -57,3 +66,17 @@ ConfirmIcon.propTypes = { style: PropTypes.object };
...
+ChevronDownIcon.defaultProps = { color: WHITE, size: 20 };
+ChevronDownIcon.propTypes = { color: PropTypes.string, size: PropTypes.number };
+
+export const PencilIcon = React.memo(({ color, size }) => (
+  <FAIcon name="pencil" color={color} size={size} />
+));
+PencilIcon.defaultProps = { color: WHITE, size: 20 };
+PencilIcon.propTypes = { color: PropTypes.string, size: PropTypes.number };
+
+export const HistoryIcon = React.memo(() => <FA5Icon name="history" />);
```
mark-prins: I don't see the need for the `memo` on this guy

joshxg: Functional components by default basically are implementing `shouldComponentUpdate` where it will always return true. Wrapping it in `memo` implements a `shouldComponentUpdate` that does a shallow comparison on all props. Since it doesn't have props this is the same as returning false every call, so it will never re-render v always re-rendering, without memo! 

Memo basically has higher effectiveness, the less frequent the props change. Since the props never change, it is at maximum efficiency!

Maybe it's stupid and premature since it's a little component and adding the memo is just additional code that makes pretty small difference. I thought SVGs would have slightly higher re-draw implications than your average native component?

PR #1658, opened 2019-12-08
#1647 Incorrect pricing

### LINKED ISSUE #1647: When creating a program requisition, an item is taking the price of a different Item
## Describe the bug

Related with #1407 
When having two master lists related with a mobile store, it is taking the highest price as expect. But at the same time it is applying that max price to the rest of the products in the ML.

Like in the following example:
Screen Shot 2019-12-05 at 9 39 00 AM

ITEM CODE 050457, Price: 30 -> Ok. Should be 30 because in the other ML it is 15.
ITEM CODE 047285, Price: 30 -> Wrong. Should be 20. This product only belongs to ML New Program and price is 20.

New Program and Test Program are the only two ML programs related with this mobile store.
Screen Shot 2019-12-05 at 9 47 33 AM

In New Program, Price of ITEM CODE 047285 is 20.
Screen Shot 2019-12-05 at 9 38 07 AM

And ITEM CODE 047285 doesn't belong to Test Program. (so it doesn't have a another price rather than 20)
Screen Shot 2019-12-05 at 9 38 46 AM

### To reproduce

- [ ] Create two masters lists for the mobile store. At least one should be a program. Add an item with different prices. Create a program requisition with the item (ITEM 1) and add a different item (ITEM 2).
- [ ] See that ITEM 2 is taking the price of ITEM 1.

### Expected behaviour

ITEM 2 shouldn't take the price of ITEM 1, they are not related.

### Version and device info

- App version: v3.0.4
- OS version: Android 9.0 (AVD)

diegoevangelisti:
Fixes #1647 

## Change summary

When having two master lists related with a mobile store, all items were taking the same max price. This issue was solved adding an extra filter to select the corresponding MasterListItem before applying max('price'). 

## Testing

- [ ] Create two masters lists for the mobile store. (At least one should be a program). 
- [ ] Add two items in both lists and assign different prices to them.
- [ ] Create a new program requisition related with one master list created in step 1.
- [ ] Price for ITEM 1 and ITEM 2 are their maximum among all created ML and they don't take other item's prices.

### DIFF
--- a/src/database/DataTypes/RequisitionItem.js
+++ b/src/database/DataTypes/RequisitionItem.js
@@ -148,6 +148,7 @@ export class RequisitionItem extends Realm.Object {
 
     // Return the maximum price of all MasterListItems.
     return UIDatabase.objects('MasterListItem')
+      .filtered('item.id == $0', this.item.id)
       .filtered(queryString, this.item)
       .max('price');
   }
--- a/src/database/DataTypes/RequisitionItem.js
+++ b/src/database/DataTypes/RequisitionItem.js
@@ -142,13 +142,12 @@ export class RequisitionItem extends Realm.Object {
       .map(({ masterList }) => masterList.id);
 
     // Query string: Query for all MasterList.IDs related to this store and this item.
-    const queryString = `${masterListIds
+    const queryString = `(${masterListIds
       .map(id => `masterList.id == '${id}'`)
-      .join(' OR ')} AND item = $0`;
+      .join(' OR ')}) AND item = $0`;
 
     // Return the maximum price of all MasterListItems.
     return UIDatabase.objects('MasterListItem')
-      .filtered('item.id == $0', this.item.id)
       .filtered(queryString, this.item)
       .max('price');
   }


src/database/DataTypes/RequisitionItem.js
```diff
@@ -148,6 +148,7 @@ export class RequisitionItem extends Realm.Object {
 
     // Return the maximum price of all MasterListItems.
     return UIDatabase.objects('MasterListItem')
+      .filtered('item.id == $0', this.item.id)
```
joshxg: This works, but I think there is a simpler solution. 

here: https://github.com/openmsupply/mobile/blob/de7f6813beb9e2b7874501bc2e28469102c74324/src/database/DataTypes/RequisitionItem.js#L145-L147 

The query already has `Item` - but we're adding it again here?

I *think* that the query is not correct. If you log the query I think it will look something like:

`masterList.id = X1 OR masterList.id = X2 OR, ... OR masterList.id = Xn AND item = Y1`

The evaluation of the conditions is incorrect - it would need to be one of any of the master lists or the last master list AND the item. We would want it to be anyone of the master lists AND the item, so just adding some parenthesis would do the trick:

`(masterList.id = X1 OR masterList.id = X2 OR, ... OR masterList.id = Xn) AND item = Y1`


So we can either add the parenthesis OR do what you have done but remove the `item` check in the above condition - up to you what you think is best, but we don't need it in both places

diegoevangelisti: You are right. It was working not applying a simpler solution as you mentioned (adding parenthesis).
Thanks for your feedback @joshxg!

PR #1660, opened 2019-12-10
#1630 Patients & Prescriber / Dispensary page

### LINKED ISSUE #1630: Patients/Prescriber page

Need a patients/prescriber page component.

A patients/prescriber page component as per wireframe in SUSSOL balsamiq.

A simple data table page with a toggle to toggle between patients and prescribers. Also has a button to add a patient/prescriber 

## Implementation

Maybe need to split this over multiple PRs. Definitely not implementing the dispense/creation functionality at this stage

N/A

## Additional context

N/A

joshxg:
# BRANCHED FROM #1602 

Fixes #1630 

## Change summary

- Adds basic dispensary page as per the wire frames
- Going to be doing small iterations on each of the main pages and cycle through [So for example, haven't added the footer to send the query to get some other patients]

image

- Logic for new patient/dispensing/editing a patient is not there yet - currently just logging the event.
- Filtering/sorting working
- OK performance, not great. Below is with 15,000 patients. Don't really see a use case for sorting on a list of 15,000 but 🤷‍♂ -- (Slowness is caused by using `localeCompare` string prototype method which is evidently very slow)
- This is very similar to like every other page, so not to much to really check. I'm not sure about the `dataSet` toggling and a better way to make that more generic, to be used elsewhere (for example, for `StocktakeItem` and `StocktakeBatch` would be cool

patients

## Testing

N/A

N/A


src/widgets/icons.js
```diff
@@ -80,3 +80,7 @@ PencilIcon.defaultProps = { color: WHITE, size: 20 };
 PencilIcon.propTypes = { color: PropTypes.string, size: PropTypes.number };
 
 export const HistoryIcon = React.memo(() => <FA5Icon name="history" />);
+
```
mark-prins: Just wondering why the `HistoryIcon` gets memo-ised when it has no props?

Issue #1404, opened 2019-10-18
AutoCompleteSelector should use new SearchBar


joshxg:
## Describe the bug

The widget `AutoCompleteSelector` uses the `SearchBar` from `react-native-components`. Using the internal `SearchBar` is better

### To reproduce

1. Go to CI, click create new customer invoice

### Expected behaviour

It would have a button for clearing the search term

### Version and device info

2.3.6 development

### Additional context

N/A

Rebecca367:
@joshxg 
I am not sure about this issue and how to start with this one. I sit possible to shed some light on this one or suggest it to be done by person who has in-depth knowledge about code base?

joshxg:
@Rebecca367 

- `AutocompleteSelector` component uses this component from this library: 
    - https://github.com/openmsupply/mobile/blob/master/src/widgets/AutocompleteSelector.js#L8
    - https://github.com/openmsupply/mobile/blob/c35dafd163f953a54d800421ba76470bb0df1a6f/src/widgets/AutocompleteSelector.js#L93-L102
- This issue is to remove this component and REPLACE it with this component https://github.com/openmsupply/mobile/blob/master/src/widgets/SearchBar.js which is an internal component in our repository

Is that helpful?

joshxg:
Putting this in the backlog and moving out of 3.1.0 release

joshxg:
Closed by #1667

diegoevangelisti:
## Testing

- [x] go to customer invoice
- [x] click on "new invoice"
- [ ] search for customers

During testing in the tablet, I found the filter is not working properly (See that after typing `Baltimor` it doesn't filter other available options).

Also found an orange cancel icon on the right (see picture below) while I guess the expected one is white.

Tablet model Lenovo XB-X304L
Android version 7.1.1
APK: mSupplyMobile-3_2_0-rc1-universal-release.apk

Screenshot:
![Screenshot_2020-01-08-12-03-24-559](https://user-images.githubusercontent.com/32987464/71941952-774b7e80-3220-11ea-9cea-efbb2654a087.jpeg)

@joshxg @Rebecca367 Please let me know in case you consider I'm doing something wrong otherwise I'll create the issue. Thanks!

Chris-Petty:
I think the cancel button is just red for the searchbar by default (and is red in the table pages), seems fine

joshxg:
@diegoevangelisti 

[...]

PR #1667, opened 2019-12-10
#1404 Use of new SearchBar Component in AutoCompleteSelector

Rebecca367:
Fixes issue #1404

## Change summary
Import searchBar from internal file rather than from third part to make it more flexible on changing

## Testing

- [ ] go to customer invoice
- [ ] click on "new invoice"
- [ ] search for customers

Type customer's name on the search bar. check that the filter is working as expected. If it does click on its name or clear that search bar by pressing the 'x' button on the right.

The searchBar is not case sensitive, so don't know it will affect on other area where case sensitivity might be crucial across codebase.

### DIFF
--- a/src/widgets/AutocompleteSelector.js
+++ b/src/widgets/AutocompleteSelector.js
@@ -98,6 +98,7 @@ export class AutocompleteSelector extends React.PureComponent {
           placeholder={placeholderText}
           onChangeText={text => this.setState({ queryText: text })}
           style={[localStyles.text, localStyles.searchBar]}
+          color="white"
         />
         {data.length > 0 && (
           <FlatList
--- a/src/widgets/AutocompleteSelector.js
+++ b/src/widgets/AutocompleteSelector.js
@@ -8,7 +8,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { FlatList, StyleSheet, Text, TouchableOpacity, View } from 'react-native';
 import { complement } from 'set-manipulator';
-import { SearchBar } from '.';
+import { SearchBar } from './SearchBar';
 import { APP_FONT_FAMILY } from '../globalStyles';
 import { generalStrings } from '../localization';
 import { withOnePress } from './withOnePress';
--- a/src/widgets/AutocompleteSelector.js
+++ b/src/widgets/AutocompleteSelector.js
@@ -12,7 +12,6 @@ import { SearchBar } from './SearchBar';
 import { APP_FONT_FAMILY } from '../globalStyles';
 import { generalStrings } from '../localization';
 import { withOnePress } from './withOnePress';
-
 /**
  * A search bar that autocompletes from the options passed in, and allows any of
  * the dropdown options to be selected. Will gracefully handle null values
@@ -100,7 +99,7 @@ export class AutocompleteSelector extends React.PureComponent {
           placeholder={placeholderText}
           onChangeText={this.onChangeText}
           style={[localStyles.text, localStyles.searchBar]}
-          WHITE="#FFFFFF"
+          autocompleteColor="#FFFFFF"
         />
         {data.length > 0 && (
           <FlatList
--- a/src/widgets/AutocompleteSelector.js
+++ b/src/widgets/AutocompleteSelector.js
@@ -32,6 +32,8 @@ export class AutocompleteSelector extends React.PureComponent {
     };
   }
 
+  onChangeText = queryText => this.setState({ queryText });
+
   /**
    * Filters a realm results object. Creates two realm results A, B
    * by two query strings. And concats A to B - A.
@@ -96,7 +98,8 @@ export class AutocompleteSelector extends React.PureComponent {
           autoCorrect={false}
           autoFocus
           placeholder={placeholderText}
-          onChangeText={text => this.setState({ queryText: text })}
+          // onChangeText={text => this.setState({ queryText: text })}
+          onChangeText={this.onChangeText}
           style={[localStyles.text, localStyles.searchBar]}
           color="white"
         />
--- a/src/widgets/AutocompleteSelector.js
+++ b/src/widgets/AutocompleteSelector.js
@@ -101,6 +101,7 @@ export class AutocompleteSelector extends React.PureComponent {
           onChangeText={this.onChangeText}
           style={[localStyles.text, localStyles.searchBar]}
           color={WHITE}
+          placeholderTextColor={WHITE}
         />
         {data.length > 0 && (
           <FlatList
--- a/src/widgets/AutocompleteSelector.js
+++ b/src/widgets/AutocompleteSelector.js
@@ -98,7 +98,6 @@ export class AutocompleteSelector extends React.PureComponent {
           autoCorrect={false}
           autoFocus
           placeholder={placeholderText}
-          // onChangeText={text => this.setState({ queryText: text })}
           onChangeText={this.onChangeText}
           style={[localStyles.text, localStyles.searchBar]}
           color="white"
--- a/src/widgets/AutocompleteSelector.js
+++ b/src/widgets/AutocompleteSelector.js
@@ -93,9 +93,11 @@ export class AutocompleteSelector extends React.PureComponent {
       <View style={localStyles.container}>
         <SearchBar
           autoCapitalize="none"
-          autoCorrect={false}
+          autoCorrect={true}
           autoFocus
           placeholder={placeholderText}
+          debouncedCallback
+          onChangeText={text => this.setState({ queryText: text })}
           style={[localStyles.text, localStyles.searchBar]}
         />
         {data.length > 0 && (
--- a/src/widgets/AutocompleteSelector.js
+++ b/src/widgets/AutocompleteSelector.js
@@ -9,9 +9,10 @@ import PropTypes from 'prop-types';
 import { FlatList, StyleSheet, Text, TouchableOpacity, View } from 'react-native';
 import { complement } from 'set-manipulator';
 import { SearchBar } from './SearchBar';
-import { APP_FONT_FAMILY } from '../globalStyles';
+import { APP_FONT_FAMILY, WHITE } from '../globalStyles';
 import { generalStrings } from '../localization';
 import { withOnePress } from './withOnePress';
+
 /**
  * A search bar that autocompletes from the options passed in, and allows any of
  * the dropdown options to be selected. Will gracefully handle null values
@@ -99,6 +100,7 @@ export class AutocompleteSelector extends React.PureComponent {
           placeholder={placeholderText}
           onChangeText={this.onChangeText}
           style={[localStyles.text, localStyles.searchBar]}
+          color={WHITE}
         />
         {data.length > 0 && (
           <FlatList
--- a/src/widgets/AutocompleteSelector.js
+++ b/src/widgets/AutocompleteSelector.js
@@ -6,9 +6,9 @@
 // eslint-disable-next-line max-classes-per-file
 import React from 'react';
 import PropTypes from 'prop-types';
-import { SearchBar } from 'react-native-ui-components';
 import { FlatList, StyleSheet, Text, TouchableOpacity, View } from 'react-native';
 import { complement } from 'set-manipulator';
+import { SearchBar } from '.';
 import { APP_FONT_FAMILY } from '../globalStyles';
 import { generalStrings } from '../localization';
 import { withOnePress } from './withOnePress';
@@ -95,10 +95,7 @@ export class AutocompleteSelector extends React.PureComponent {
           autoCapitalize="none"
           autoCorrect={false}
           autoFocus
-          color="white"
-          onChange={text => this.setState({ queryText: text })}
           placeholder={placeholderText}
-          placeholderTextColor="white"
           style={[localStyles.text, localStyles.searchBar]}
         />
         {data.length > 0 && (
--- a/src/widgets/AutocompleteSelector.js
+++ b/src/widgets/AutocompleteSelector.js
@@ -99,7 +99,6 @@ export class AutocompleteSelector extends React.PureComponent {
           placeholder={placeholderText}
           onChangeText={this.onChangeText}
           style={[localStyles.text, localStyles.searchBar]}
-          autocompleteColor="#FFFFFF"
         />
         {data.length > 0 && (
           <FlatList
--- a/src/widgets/AutocompleteSelector.js
+++ b/src/widgets/AutocompleteSelector.js
@@ -93,10 +93,9 @@ export class AutocompleteSelector extends React.PureComponent {
       <View style={localStyles.container}>
         <SearchBar
           autoCapitalize="none"
-          autoCorrect={true}
+          autoCorrect={false}
           autoFocus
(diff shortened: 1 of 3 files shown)


src/widgets/AutocompleteSelector.js
```diff
@@ -6,9 +6,9 @@
 // eslint-disable-next-line max-classes-per-file
 import React from 'react';
 import PropTypes from 'prop-types';
-import { SearchBar } from 'react-native-ui-components';
 import { FlatList, StyleSheet, Text, TouchableOpacity, View } from 'react-native';
 import { complement } from 'set-manipulator';
+import { SearchBar } from '.';
```
diegoevangelisti: A simple suggestion here, replacing this import line for: `import { SearchBar } from './SearchBar';`

Rebecca367: Thanks, done!

diegoevangelisti:
@Rebecca367 Just want to add a constructive comment related with this PR. From my point of view, would be useful to have clearer instructions in the testing section. For example, instead of: `do something` it would help having something like: `Type a supplier's name and check that the filter is working as expected` or `Check that the search can be cleared pressing the 'x' button on the right`, etc. More explicit testing to ensure that everything is working well.
Also, clarifying the issue related with this PR (e.g. Fixes issue #1404) will be very handy.
Hope this can help! Again, this is just an opinion :-) Cheers!

src/widgets/AutocompleteSelector.js
```diff
@@ -93,12 +93,11 @@ export class AutocompleteSelector extends React.PureComponent {
...
         <SearchBar
           autoCapitalize="none"
-          autoCorrect={false}
+          autoCorrect={true}
           autoFocus
-          color="white"
-          onChange={text => this.setState({ queryText: text })}
           placeholder={placeholderText}
-          placeholderTextColor="white"
+          debouncedCallback
```
joshxg: There is no prop `debouncedCallback` in the `SearchBar` component! Was this a mistake?

src/widgets/AutocompleteSelector.js
```diff
@@ -93,12 +93,11 @@ export class AutocompleteSelector extends React.PureComponent {
...
+          onChangeText={text => this.setState({ queryText: text })}
```
joshxg: Would be really cool to define this callback as a class method !

for example:

```
export class AutocompleteSelector extends React.PureComponent {
  constructor(props) {
    super(props);
    this.state = {
      queryText: '',
    };
  }

onChangeText = text => ...

...


<SearchBar 
    ....
    onChangeText={this.onChangeText}
    ...
    />
```

src/widgets/AutocompleteSelector.js
```diff
@@ -93,12 +93,11 @@ export class AutocompleteSelector extends React.PureComponent {
       <View style={localStyles.container}>
         <SearchBar
           autoCapitalize="none"
-          autoCorrect={false}
+          autoCorrect={true}
```
joshxg: Why has this been changed? Because we handle a lot of different countries, correcting spelling can be a little bit of a nightmare.. which countries spelling to use etc.. I think it might be best to keep this as `false`?

src/widgets/AutocompleteSelector.js
```diff
@@ -93,12 +93,11 @@ export class AutocompleteSelector extends React.PureComponent {
       <View style={localStyles.container}>
         <SearchBar
```
joshxg: We also need to provide new styles. The old `SearchBar` was white, however the default for this one is orange - we are replacing the search bar, so we should not see any difference here - we need to make it white please!

joshxg:
> @Rebecca367 Just want to add a constructive comment related with this PR. From my point of view, would be useful to have clearer instructions in the testing section. For example, instead of: `do something` it would help having something like: `Type a supplier's name and check that the filter is working as expected` or `Check that the search can be cleared pressing the 'x' button on the right`, etc. More explicit testing to ensure that everything is working well.
> Also, clarifying the issue related with this PR (e.g. Fixes issue #1404) will be very handy.
> Hope this can help! Again, this is just an opinion :-) Cheers!

This is not just good advice or an opinion but what all PRs will require! :)

src/widgets/AutocompleteSelector.js
Rebecca367: yes, I think so. already been removed.

src/widgets/AutocompleteSelector.js
Rebecca367: Is it a necessary class that you would like to implement on this one or we can discuss this later?

src/widgets/AutocompleteSelector.js
Rebecca367: OK, cool. It's logical to set it as "true" isn't it? but anyway I have implemented it in the branch.

src/widgets/AutocompleteSelector.js
Rebecca367: done.

src/widgets/AutocompleteSelector.js
joshxg: This is not a new class, it is defining the callback as a class method - this makes the code easier to understand and also makes the function defined only once, reducing the number of re-renders

src/widgets/SearchBar.js
```diff
@@ -10,8 +10,9 @@ import PropTypes from 'prop-types';
 
 import { StyleSheet, TextInput, View, TouchableOpacity } from 'react-native';
 import { MagnifyingGlass, Cancel } from './icons';
-import { SUSSOL_ORANGE, APP_FONT_FAMILY } from '../globalStyles/index';
+import { APP_FONT_FAMILY } from '../globalStyles/index';
 import { debounce } from '../utilities/index';
+import { WHITE } from '../globalStyles/colors';
```
joshxg: This will change the colour of the component for the whole app, rather than just within this modal. The default being orange is correct, you need to pass through the colour white as a prop

src/widgets/AutocompleteSelector.js
joshxg: no - as explained auto correction with fields that are not standard words is not good

src/widgets/AutocompleteSelector.js
```diff
@@ -95,11 +97,10 @@ export class AutocompleteSelector extends React.PureComponent {
           autoCapitalize="none"
           autoCorrect={false}
           autoFocus
-          color="white"
-          onChange={text => this.setState({ queryText: text })}
           placeholder={placeholderText}
-          placeholderTextColor="white"
+          onChangeText={this.onChangeText}
           style={[localStyles.text, localStyles.searchBar]}
+          color="white"
```
joshxg: Can we use the constant `WHITE` here?

This one defined here:
https://github.com/openmsupply/mobile/blob/78406fdd118175f491b99db6c82cb79eb7006a09/src/globalStyles/colors.js#L21

It would be good to export it in the `index` file - it seems someone forgot!!

https://github.com/openmsupply/mobile/blob/78406fdd118175f491b99db6c82cb79eb7006a09/src/globalStyles/index.js#L26-L41

Rebecca367: @joshxg 
@Chris-Petty 
There are 2 changes regarding to export and import WHITE = '#FFFFFF' to search bar.

src/widgets/AutocompleteSelector.js
```diff
@@ -100,7 +100,7 @@ export class AutocompleteSelector extends React.PureComponent {
           placeholder={placeholderText}
           onChangeText={this.onChangeText}
           style={[localStyles.text, localStyles.searchBar]}
-          color="white"
+          WHITE="#FFFFFF"
```
joshxg: There is no prop `WHITE` on `SearchBar` this doesn't make any sense

src/widgets/SearchBar.js
```diff
@@ -132,7 +133,7 @@ SearchBarComponent.defaultProps = {
   debounceTimeout: 250,
   textInputStyle: defaultStyles.textInput,
   viewStyle: defaultStyles.container,
-  color: SUSSOL_ORANGE,
+  color: WHITE,
```
joshxg: again we've changed the default prop back to `WHITE` .. we only wanted to change the colour of the component to white when it is used in `AutoCompleteSelector`

src/widgets/AutocompleteSelector.js
Rebecca367: check on index.js. Value assigned to WHITE IN index.js.  Imported WHITE to searchBar then use it in Autocompleteselector. That's how it works.

src/widgets/SearchBar.js
Rebecca367: This one is a mistake and confusion.

src/widgets/AutocompleteSelector.js
Chris-Petty: Josh's comment here is correct.

`SearchBar` is a component used by `AutocompleteSelector` (this file that we're commenting on). `WHITE` was not a prop previously, it was `color` and `placeholderTextColor`. You have edited SearchBar, but not in anyway that enables what you're doing here, not that it is necessary. 

All that is required for this PR is to change the imported SearchBar to be the implementation within this repo (as you have done), and to change value provided to `color` and `placeholderTextColor` to be the constant `WHITE` that you added to the exports in globalStyles.

src/widgets/SearchBar.js
Chris-Petty: As Josh pointed out, you have changed the default prop to `WHITE` rather than the usually desired `SUSSOL_ORANGE`. While this does look correct in the AutoCompleteSelector:
Screen Shot 2019-12-16 at 12 26 15

It breaks the design in every other table page in the app which also use the `SearchBar` component:
Screen Shot 2019-12-16 at 12 27 57

Chris-Petty: That said, this file doesn't need any changes.

src/widgets/SearchBar.js
```diff
@@ -125,6 +126,7 @@ const defaultStyles = StyleSheet.create({
     fontFamily: APP_FONT_FAMILY,
     backgroundColor: 'rgba(0, 0, 0, 0)',
     flex: 1,
+    autocompleteColor: '#FFFFFF',
```
Chris-Petty: This won't do anything and also this file needs no changes at all

src/widgets/AutocompleteSelector.js
```diff
@@ -95,11 +96,10 @@ export class AutocompleteSelector extends React.PureComponent {
           autoCapitalize="none"
           autoCorrect={false}
           autoFocus
-          color="white"
-          onChange={text => this.setState({ queryText: text })}
           placeholder={placeholderText}
-          placeholderTextColor="white"
+          onChangeText={this.onChangeText}
           style={[localStyles.text, localStyles.searchBar]}
+          autocompleteColor="#FFFFFF"
```
Chris-Petty: This still won't pass through, as it's not an accepted prop of SearchBar (and shouldn't be!). (I think) the props you need to pass to SearchBar are:
```jsx
<SearchBar
  ...
  color="white"
  placeholderTextColor="white"
  ...
/>
```
Now looking at it, seems a bit over the top making a new constant in global styles for `WHITE: #FFFFFF` when `"white"` is already a CSS constant. If were were going for some sort of off white throughout the app I'd see the see the use case.

I think it'd be good to review how reactjs, props and JSX work together. [This egghead course](https://egghead.io/courses/start-learning-react) helped me a lot a few years ago. In particular, [Write More Reusable React Components with Composable APIs](https://egghead.io/lessons/react-write-more-reusable-react-components-with-composable-apis) is a relevant lesson. I encourage you to watch the whole course though, it's quick and a fairly comprehensive base of react and JSX. The [official docs](https://reactjs.org/docs/components-and-props.html) are of high quality also.

Chris-Petty: Actually, just the prop `color` may be needed if you read the `SearchBar` props.

src/widgets/AutocompleteSelector.js
Rebecca367: I misunderstood the changes that you requested me to do and went on wrong direction instead.
so next time I better ask for clarification on comments to save time.
There are some changes that Chris suggest is added as below
Screen Shot 2019-12-16 at 8 58 53 PM

src/widgets/SearchBar.js
Rebecca367: I was trying to create a const with new name with it which is completely unnecessary and wrong due to misunderstood the whole meaning of it. Thanks for checking on the commits that don't make any sense. Next time, I will try to clarify changes with reviewers what they would like to be changed before spending time and effort on amendment.

src/widgets/AutocompleteSelector.js
Rebecca367: @Chris-Petty You are right. autocompleteColor was not a valid prop name in searchBar. I tried to create a new const as I thought that's what it meant but it's not. Thanks for reviewing the code. I also add placeholderTextColor={WHITE} to searchbar as you mentioned. 
I think I need to learn more and use it more to remember how it works as it's easily forgotten even more after long break or a holiday.

src/widgets/AutocompleteSelector.js
```diff
@@ -95,11 +97,11 @@ export class AutocompleteSelector extends React.PureComponent {
...
           autoCorrect={false}
           autoFocus
-          color="white"
-          onChange={text => this.setState({ queryText: text })}
           placeholder={placeholderText}
-          placeholderTextColor="white"
+          onChangeText={this.onChangeText}
           style={[localStyles.text, localStyles.searchBar]}
+          color={WHITE}
+          placeholderTextColor={WHITE}
```
joshxg: `SearchBar` has no `placeholderTextColor` prop. You do not need this line! :)

src/widgets/SearchBar.js
```diff
@@ -137,6 +138,7 @@ SearchBarComponent.defaultProps = {
   placeholder: '',
   autoFocus: false,
   onFocusOrBlur: null,
+  WHITE: '#FFFFFF',
```
joshxg: Again, this file needs no changes!! :)

What do you think this is doing?

src/widgets/SearchBar.js
```diff
@@ -149,4 +151,5 @@ SearchBarComponent.propTypes = {
   textInputStyle: PropTypes.object,
   viewStyle: PropTypes.object,
   onFocusOrBlur: PropTypes.func,
+  WHITE: PropTypes.string,
```
joshxg: Again, this is not needed!! :)

src/widgets/SearchBar.js
Rebecca367: I think that this might hard-code value to "WHITE" constant as we discussed previously.

src/widgets/SearchBar.js
Rebecca367: Yes, already deleted!

src/widgets/AutocompleteSelector.js
Rebecca367: done. Thanks for your time reviewing.

src/widgets/SearchBar.js
Rebecca367: Sorry, this was a mistake..

PR #1669, opened 2019-12-11
#1666 Corrected usage calculation

### LINKED ISSUE #1666: Monthly usage incorrect
## Describe the bug

Monthly usage seems to have at minimum a monthly usage of 30

### To reproduce

1. Init a new store on mobile
2. Sync some items
3. See monthly usage is at minimum, for all items, 30

### Expected behaviour

There should be no minimum monthly usage.

### Proposed Solution

Not sure as I haven't started the issue yet

### Version and device info

- App version: 3.0.0 development
- Tablet model: All
- OS version: All

### Additional context

At a guess, daily usage is being calculated with a minimum of 1 which is being multiplied by 30 (default days in a month).

joshxg:
Fixes #1666 

## Change summary

- Corrects the usage return value, which was returning `1`, instead of `0`

### This is somewhat confusing - I've tried to explain as simply as I can, I'm sorry

**[ You can check monthly usage by going to `Current Stock` page - and clicking on an item. A popup will open from the bottom of the screen, which will display monthly usage ]**

example:
image

**Usage calculation can be found in `dailyUsage.js`. It is the sum of all usage within the lookback period (defined through store custom data `monthlyConsumptionLookBackPeriod` field, defaults to 30 if not defined in store custom data), over the number of days in the lookback period.**

However, if the item was added (earliest stock in transaction batch confirm date for that item) more recently than the lookback period, the usage is calculated only since the item was added. UNLESS the store custom data `monthlyConsumptionEnforceLookBackPeriod` is set, in which case the lookback period is always used.

**Note: This PRs aim is to only make a small change to ensure the correct usage is being returned when there is no usage**

- [ ] Items which have NO usage, display NO usage

N/A

### DIFF
--- a/src/utilities/dailyUsage.js
+++ b/src/utilities/dailyUsage.js
@@ -58,5 +58,5 @@ export const dailyUsage = item => {
     .filtered('transaction.confirmDate >= $0 && transaction.confirmDate <= $1', startDate, dateNow)
     .reduce((sum, { totalQuantity }) => sum + totalQuantity, 0);
 
-  return usage / numberOfUsageDays || 1;
+  return usage / (numberOfUsageDays || 1);
 };


src/utilities/dailyUsage.js
```diff
@@ -58,5 +58,5 @@ export const dailyUsage = item => {
     .filtered('transaction.confirmDate >= $0 && transaction.confirmDate <= $1', startDate, dateNow)
     .reduce((sum, { totalQuantity }) => sum + totalQuantity, 0);
 
-  return usage / numberOfUsageDays || 1;
+  return usage / (numberOfUsageDays || 1);
```
wlthomson: LGTM.

You could also do `numberOfUsageDays | 1`. Doesn't add anything but it's cool.

joshxg:
@wlthomson 

I want to do that - but I also think, if something has gone wrong, maybe displaying a number that is clearly wrong (a negative) is better than displaying a value that could be valid? what you think?

wlthomson:
> @wlthomson
> 
> I want to do that - but I also think, if something has gone wrong, maybe displaying a number that is clearly wrong (a negative) is better than displaying a value that could be valid? what you think?

I agree, probably better to do that and/or bubble up an exception or log something. As long as negatives don't break any logic down the line then 👍 .

PR #1676, opened 2019-12-12
#1042 Change `Show over stocked` as default in program requisitions

### LINKED ISSUE #1042: Hide/Show overstock button should default to show overstock when finalised
## Describe the bug

UPDATE: This issue should default the show/hide overstock toggle to show overstock when viewing a finalised requisition.

The show/hide overstock button is usable when a requisition is finalized and is a little confusing. This button hides the rows in the requisition which have stock that is over the MOS, defined by the order type chosen for the requisition.

### To reproduce
1. Go to a finalized, program requisition
2. Click on hide/show overstock
3. See error

### Expected behaviour
Shouldn't be usable

### Version and device info
 - App version: 2.3.4
 - Tablet model: All
 - OS version: All

### Additional context

@DhanyaHerath @Chris-Petty @wlthomson @nmadruga

Looking for opinions

Not 100% sure if the expected behaviour should be that this is disabled. I think that it should be disabled, but looking for feedback

### LINKED ISSUE #1402: Filtering should be trimmed
## Describe the bug

Filtering should be trimmed of trailing whitespace

### To reproduce

1. Go to a filterable page, 
2. filter and add trailing white space

### Expected behaviour

Trailing whitespace shouldn't effect the filtering string

### Version and device info

2.3.6 development

### Additional context

N/A

diegoevangelisti:
Fixes #1042 

## Change summary

When opening a finalised program requisition, the toggle was set `Hide over stocked` as default, causing confusions when opening to see its items. Now it is `Show over stocked` as default, and anyway it can by toggle to hide/show as before (option remains enable).

## Testing

- [ ] Create a new program requisition (the program should have over stock items, otherwise modify quantities performing a stocktake).
- [ ] Finalise the requisition
- [ ] Check that the option `Show over stocked items` is set as default and items are visible
- [ ] Press `Hide over stocked` and check toggle works.

Attention: Remember to add items that are over stocked, otherwise the toggle will not make any sense (all items will be shown either hide/show is pressed).

### DIFF
--- a/src/pages/dataTableUtilities/getPageInitialiser.js
+++ b/src/pages/dataTableUtilities/getPageInitialiser.js
@@ -360,11 +360,15 @@ const supplierRequisitionInitialiser = requisition => {
   const route = program ? ROUTES.SUPPLIER_REQUISITION_WITH_PROGRAM : ROUTES.SUPPLIER_REQUISITION;
 
   const sortedData = backingData.sorted('item.name').slice();
+  const isFinalised = requisition.status === 'finalised';
+  const filteredData = !isFinalised
+    ? sortedData.filter(item => item.isLessThanThresholdMOS)
+    : sortedData;
 
   return {
     pageObject: requisition,
     backingData,
-    data: sortedData,
+    data: filteredData,
     keyExtractor: recordKeyExtractor,
     dataState: new Map(),
     searchTerm: '',
@@ -374,7 +378,7 @@ const supplierRequisitionInitialiser = requisition => {
     modalKey: '',
     hasSelection: false,
     modalValue: null,
-    showAll: usingPrograms,
+    showAll: usingPrograms && isFinalised,
     route: ROUTES.SUPPLIER_REQUISITION,
     columns: getColumns(route),
     getPageInfoColumns: getPageInfoColumns(route),
--- a/src/pages/dataTableUtilities/getPageInitialiser.js
+++ b/src/pages/dataTableUtilities/getPageInitialiser.js
@@ -354,16 +354,16 @@ const supplierInvoicesInitialiser = () => {
  * @returns  {object}
  */
 const supplierRequisitionInitialiser = requisition => {
-  const { program, items: backingData } = requisition;
+  const { isFinalised, program, items: backingData } = requisition;
 
   const usingPrograms = !!program;
   const route = program ? ROUTES.SUPPLIER_REQUISITION_WITH_PROGRAM : ROUTES.SUPPLIER_REQUISITION;
 
   const sortedData = backingData.sorted('item.name').slice();
-  const isFinalised = requisition.status === 'finalised';
-  const filteredData = !isFinalised
-    ? sortedData.filter(item => item.isLessThanThresholdMOS)
-    : sortedData;
+  const filteredData =
+    !usingPrograms || isFinalised
+      ? sortedData
+      : sortedData.filter(item => item.isLessThanThresholdMOS);
 
   return {
     pageObject: requisition,
@@ -378,7 +378,7 @@ const supplierRequisitionInitialiser = requisition => {
     modalKey: '',
     hasSelection: false,
     modalValue: null,
-    showAll: usingPrograms && isFinalised,
+    showAll: !usingPrograms || isFinalised,
     route: ROUTES.SUPPLIER_REQUISITION,
     columns: getColumns(route),
     getPageInfoColumns: getPageInfoColumns(route),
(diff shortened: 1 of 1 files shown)


src/pages/dataTableUtilities/getPageInitialiser.js
```diff
@@ -377,7 +374,7 @@ const supplierRequisitionInitialiser = requisition => {
     modalKey: '',
     hasSelection: false,
     modalValue: null,
-    showAll: !usingPrograms,
+    showAll: usingPrograms,
```
joshxg: I think the intention was to still have `showAll` to be false when using a program for a NON-finalised requisition. This makes it easier when for example, you go to first make the requisition, you can see only the items you *should* order.. but when you view an old, finalised, requisition, you display all the items so none are hiding. So I think we should still use `filteredData` - but only when the `status !== "finalised"` -- make sense?

diegoevangelisti: Totally! Yes, I've got it. Now in the filter I am checking the status and `showAll` only will be true when the using a program that is finalised. Let me know what you think about it. Thanks!

src/pages/dataTableUtilities/getPageInitialiser.js
```diff
@@ -360,7 +360,8 @@ const supplierRequisitionInitialiser = requisition => {
   const route = program ? ROUTES.SUPPLIER_REQUISITION_WITH_PROGRAM : ROUTES.SUPPLIER_REQUISITION;
 
   const sortedData = backingData.sorted('item.name').slice();
-  const filteredData = usingPrograms
+  const isFinalised = requisition.status === 'finalised';
+  const filteredData = !isFinalised
```
joshxg: hmm this does work, but only because if it's not a program there are no items yet! It seems a little odd to me, when you think about non-program requisitions. Just because this is basically ignored, I don't think we should not account for that case explicitly.

We filter when the requisition is NOT finalised AND is a program requisition using this one  : https://github.com/openmsupply/mobile/blob/7c64ed28f73f2be998e2fa6f2dafeaa0798f9941/src/pages/dataTableUtilities/getPageInitialiser.js#L359 

Alternatively you can say: we DONT filter when the requisition is finalised OR is NOT a program requisition

Another one is you could use this:
https://github.com/openmsupply/mobile/blob/7c64ed28f73f2be998e2fa6f2dafeaa0798f9941/src/database/DataTypes/Requisition.js#L68-L70

so you can add to this line:
https://github.com/openmsupply/mobile/blob/7c64ed28f73f2be998e2fa6f2dafeaa0798f9941/src/pages/dataTableUtilities/getPageInitialiser.js#L357

`const { isFinalised, program, items:backingData } = requisition;`

and then you have already `isFinalised` to use :)

src/pages/dataTableUtilities/getPageInitialiser.js
```diff
@@ -377,7 +378,7 @@ const supplierRequisitionInitialiser = requisition => {
     modalKey: '',
     hasSelection: false,
     modalValue: null,
-    showAll: !usingPrograms,
+    showAll: usingPrograms && isFinalised,
```
joshxg: I guess this is logical, given we ignore this field when `usingPrograms === false` anyway, but it seems to make more sense in my head to read it as this being true when NOT using programs OR is finalised .. maybe I am just crazy

diegoevangelisti: You're not crazy - at least regarding this topic ;-) - . Now I can see what you mean and makes sense to me as well. `!usingPrograms || isFinalised` is totally explicit and much clearer than the one I was using before. Working on this right now!

src/pages/dataTableUtilities/getPageInitialiser.js
diegoevangelisti: > Alternatively you can say: we DONT filter when the requisition is finalised OR is NOT a program requisition

Totally agree, I am going to use this one instead, makes this point easier to understand.

> `const { isFinalised, program, items:backingData } = requisition;`
> 
> and then you have already `isFinalised` to use :)

Great, I am using it this way now, thanks for the suggestions!

src/pages/dataTableUtilities/getPageInitialiser.js
```diff
@@ -354,15 +354,16 @@ const supplierInvoicesInitialiser = () => {
...
 
   const usingPrograms = !!program;
   const route = program ? ROUTES.SUPPLIER_REQUISITION_WITH_PROGRAM : ROUTES.SUPPLIER_REQUISITION;
 
   const sortedData = backingData.sorted('item.name').slice();
-  const filteredData = usingPrograms
-    ? sortedData.filter(item => item.isLessThanThresholdMOS)
-    : sortedData;
+  const filteredData =
+    !usingPrograms || isFinalised
```
joshxg: I'm being pedantic now - this is not a "must do" - i'm approving anyway. Just a suggestion: Because we're using `!usingPrograms || isFinalised` twice, assigning it to a variable reduces the duplication and also makes this more self documenting:

i.e.: `const shouldShowOverStockItems = !usingPrograms || isFinalised` (too long of a name? 🤣 )

diegoevangelisti: Great suggestion and I agree with you. Regarding the name, maybe removing `should` still makes sense and it is not too long. ( `const showOverStockItems = !usingPrograms || isFinalised`)
Thanks for your feedback, cheers!

PR #1694, opened 2019-12-15
#1632 TextInput with Validation

### LINKED ISSUE #1632: Styled and default proped TextInputs with masks

Need to create forms, which will require `TextInputs` in a form style. Currently use them only really in the `LoginModal` which now they're going to be used more can be abstracted.

`FormTextInput` or something, with default props passed and styling. Also a masking feature would be useful, if only for character limit

## Implementation

As above

N/A

## Additional context

N/A

joshxg:
### BRANCHED FROM #1632 

Fixes #1632 

## Change summary

- Adds `ValidationTextInput` component which should make validation and form control a little easier.
- Primary use will be for creating/editing patients and insurance policies
- If we use patient editing as the example: I am progressing toward the parent having a some sort of `formValid` property (maybe multiple, for each input and some aggregated valid property also), which will control the `SAVE` button on a form. This will be updated by the call back of each input which will call back when the validity changes. When an input is finished being edited, it will callback it's updated values.
- Could potentially use some X and ✓ icons to be a bit more classy 🎩 but can come later I think!
- Could potentially callback to the parent everytime the text changes to save the value. That could be a little easier, but feels nice to abstract the logic into a component

image

## Testing

N/A

N/A

### DIFF
--- a/src/localization/validationStrings.js
+++ b/src/localization/validationStrings.js
@@ -24,7 +24,7 @@ export const validationStrings = new LocalizedStrings({
     repeatPassword: {
       required: 'Enter the repeat password',
     },
-    is_required: 'is required',
+    isRequired: 'is required',
   },
   fr: {},
   gil: {},
--- a/src/localization/validationStrings.js
+++ b/src/localization/validationStrings.js
@@ -24,6 +24,7 @@ export const validationStrings = new LocalizedStrings({
     repeatPassword: {
       required: 'Enter the repeat password',
     },
+    is_required: 'is required',
   },
   fr: {},
   gil: {},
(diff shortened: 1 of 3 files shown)


src/localization/validationStrings.js
```diff
@@ -24,6 +24,7 @@ export const validationStrings = new LocalizedStrings({
     repeatPassword: {
       required: 'Enter the repeat password',
     },
+    is_required: 'is required',
```
mark-prins: I am reluctant to pick this particular scab again... but here goes...
Shouldn't this property be `isRequired`?

joshxg: eba5dc6

Don't be reluctant please! I didn't notice the strings in this file were using camelCase :(

I've made an issue here to try and refactor soon for consistency https://github.com/openmsupply/mobile/issues/1695

PR #1704, opened 2019-12-16
#1540 Fix autocomplete selector should use hooks api

### LINKED ISSUE #1540: AutocompleteSelector should use Hooks API
## The suggestion

AutocompleteSelector should use Hooks API rather than React Class API

**Why?**
Consistency and https://github.com/openmsupply/mobile/issues/1092

Doing nothing.

sworup:
Fixes #1540

## Change summary

- Refactor code 
  - Autocomplete selector should use hooks api
  - Move `ResultRow` component out of `AutocompleteSelector` component
  - User `ResultRow` already present in code, hence sharing the component among `AutocompleteSelector` and `MultiSelectList`.

## Testing

- [ ] Pages using `AutocompleteSelector` should work properly.
- [ ] Pages using `MultiSelectList` (Master-list selection) should work properly.
- [ ] Rest of the interface should be same. This is code refactor.

None

### DIFF
--- a/src/widgets/AutocompleteSelector.js
+++ b/src/widgets/AutocompleteSelector.js
@@ -4,7 +4,7 @@
  */
 
 // eslint-disable-next-line max-classes-per-file
-import React, { useState } from 'react';
+import React, { useState, useCallback } from 'react';
 import PropTypes from 'prop-types';
 import { FlatList, StyleSheet, View } from 'react-native';
 import { complement } from 'set-manipulator';
@@ -77,6 +77,21 @@ export const AutocompleteSelector = props => {
     return [];
   };
 
+  const renderItem = useCallback(
+    item => (
+      <ResultRowWithOnePress
+        data={item}
+        onPress={onSelect}
+        renderLeftText={renderLeftText}
+        renderRightText={renderRightText}
+        showCheckIcon={false}
+      />
+    ),
+    [renderLeftText, renderRightText]
+  );
+
+  const keyExtractor = item => item.id || item.name;
+
   const data = getData();
 
   const ResultRowWithOnePress = withOnePress(ResultRow);
@@ -96,16 +111,8 @@ export const AutocompleteSelector = props => {
       {data.length > 0 && (
         <FlatList
           data={data}
-          keyExtractor={item => item.id || item.name}
-          renderItem={item => (
-            <ResultRowWithOnePress
-              data={item}
-              onPress={onSelect}
-              renderLeftText={renderLeftText}
-              renderRightText={renderRightText}
-              showCheckIcon={false}
-            />
-          )}
+          keyExtractor={keyExtractor}
+          renderItem={renderItem}
           keyboardShouldPersistTaps="always"
           style={localStyles.resultList}
         />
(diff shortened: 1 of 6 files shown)


src/widgets/AutocompleteSelector.js
```diff
@@ -4,42 +4,37 @@
...
+    queryString,
+    queryStringSecondary,
+    primaryFilterProperty,
+    secondaryFilterProperty,
+    options,
+    onSelect,
+    placeholderText,
+    renderLeftText,
+    renderRightText,
+  } = props;
```
joshxg: could just destructure the parameter?

src/widgets/AutocompleteSelector.js
```diff
@@ -77,50 +69,48 @@ export class AutocompleteSelector extends React.PureComponent {
...
+        color="white"
+        onChange={text => setQueryText(text)}
+        placeholder={placeholderText}
+        placeholderTextColor="white"
+        style={[localStyles.text, localStyles.searchBar]}
+      />
+      {data.length > 0 && (
+        <FlatList
+          data={data}
+          keyExtractor={item => item.id || item.name}
```
joshxg: could we define this explicitly so re-renders don't cause a new function to be created?

src/widgets/AutocompleteSelector.js
```diff
@@ -77,50 +69,48 @@ export class AutocompleteSelector extends React.PureComponent {
...
+          renderItem={item => (
```
joshxg: same as above, can we memoize and explicitly define this callback to reduce re-renders?

src/widgets/ResultRow.js
```diff
@@ -9,20 +9,21 @@ import { Text, TouchableOpacity, StyleSheet } from 'react-native';
 import Icon from 'react-native-vector-icons/Ionicons';
 import { APP_FONT_FAMILY, SUSSOL_ORANGE, BACKGROUND_COLOR, GREY } from '../globalStyles';
 
-const ResultRow = React.memo(
+export const ResultRow = React.memo(
   ({ data, renderLeftText, renderRightText, onPress, isSelected, showCheckIcon }) => {
-    const rowPressed = useCallback(() => onPress(data), [data]);
+    const rowPressed = useCallback(() => onPress(data.item), [data.item]);
```
joshxg: nice

src/widgets/index.js
```diff
@@ -4,8 +4,9 @@
  */
 
 export { Button, ProgressBar } from 'react-native-ui-components';
-
+// eslint-disable-next-line import/no-cycle
```
joshxg: In the above files if we export directly from the files rather than from `.` we can remove the cycle, I believe

sworup: I thought about doing that, but did this for consistency. Will revert back.

src/widgets/AutocompleteSelector.js
sworup: Resolved with 81441a9

src/widgets/AutocompleteSelector.js
sworup: Resolved with 81441a9

src/widgets/AutocompleteSelector.js
sworup: Resolved with ae8cb73

src/widgets/index.js
sworup: Resolved with 7ae9c50

PR #1717, opened 2019-12-17
#1706 Add Dashboard dependencies

### LINKED ISSUE #1706: Dashboard dependencies

Need to add dependencies used in the `Dashboard` feature.

Add the dependencies 

- `victory`
- `react-native-svg`
- `victory-native`

## Implementation

Follow instructions: https://github.com/FormidableLabs/victory-native

N/A

## Additional context

N/A

diegoevangelisti:
Fixes #1706

## Change summary

Adding `victory-native` and `react-native-svg` dependencies, needed for creating the charts (pie, bar, etc.)

## Testing

- [ ] Working in this branch, go to the CLI and type `yarn start` (or `react-native run-android`)
- [ ] Check BUILD SUCCESSFUL result.
- [ ] Again in the CLI, type `yarn build`
- [ ] Check BUILD SUCCESSFUL result again.
- [ ] Finally, the app should be running properly.

The proper test for these installations would be taking place once the dependencies are being imported in further development.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -194,7 +194,6 @@ android {
 }
 
 dependencies {
-    implementation project(':react-native-svg')
     implementation fileTree(dir: "libs", include: ["*.jar"])
     implementation "com.android.support:appcompat-v7:${rootProject.ext.supportLibVersion}"
     implementation "com.facebook.react:react-native:+"  // From node_modules
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -194,6 +194,7 @@ android {
 }
 
 dependencies {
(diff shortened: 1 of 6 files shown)


diegoevangelisti:
> I think their instructions are out of date. Since `react-native` 0.60, you shouldn't need to do `react-native link` anymore: https://github.com/react-native-community/cli/blob/master/docs/autolinking.md
> 
> I ran `react-native unlink react-native-svg` and I was still able to build etc, so it seems we don't need to do that step. I think it's best to let `react-native` handle that for us.. because otherwise we will have to make sure this is still working correctly etc when other updates occur

Nice found @joshxg! I just unlinked them

Issue #1693, opened 2019-12-15
Localization for dispensing


joshxg:
## Is your feature request related to a problem? Please describe.

This is an issue for the dispensing feature branch localization, to track all introduced strings. This should be completed toward the end of the feature introduction.

## Describe the solution you'd like

- At minimum, french translations for all strings introduced

### List of issues/cpde where strings are introduced:

- #1694 
- titles in https://github.com/openmsupply/mobile/blob/%23feature-dispensary/src/navigation/actions.js
- #1702 
- #1724 
- #1750 
- #1749 
- #1747 
- #1759
- #1750
- #1734 
- #1810 
- #1784 
- #1785

mark-prins:
[spreadsheet](https://docs.google.com/spreadsheets/d/1r-hdhIHvuZ7XDSiEmL8lLgtLfww5E-XHMrEwoU3BMg0/edit#gid=0) created to track translations

joshxg:
Closed by #2070

PR #1718, opened 2019-12-17
#1702 Add & Edit Patient Modal

### LINKED ISSUE #1702: Add/Edit Patient Modal

Need a way to add and edit patients.

A modal with a form input which allows editing a patient as well as entering details for a new patient

## Implementation

N/A

N/A

## Additional context

N/A

joshxg:
Fixes #1702 

## Change summary

- Adds `PatientEdit` component to be used within a `ModalContainer` component.
- Does not contain any code to actual use it
- Adds `PatientReducer` and `PatientActions` which will be used for adding and editing, as well as most likely a patients history modal

image

Definitely can use some work. Making groupings i.e. "Details", "Address" etc. A date picker for DOB, most likely. But is a good first step. I *think* having the "not saved till finished" approach here is the right way to go, rather than updating instantly on key press/updating. But doing it this way causes all the fields in the state and setters etc, so it is a bit more code 🤷‍♂ 

## Testing
N/A

N/A


src/reducers/PatientReducer.js
```diff
@@ -0,0 +1,60 @@
+/**
+ * mSupply Mobile
+ * Sustainable Solutions (NZ) Ltd. 2019
+ */
+
+import { PATIENT_ACTIONS } from '../actions/PatientActions';
+
+const patientInitialState = () => ({
+  currentPatient: null,
```
mark-prins: I wondered about having a 'validity' property like 
```
const patientInitialState = () => ({
  currentPatient: null,
  firstName: '',
  lastName: '',
  code: '',
  dateOfBirth: '',
  email: '',
  phone: '',
  addressOne: '',
  addressTwo: '',
  country: '',
  isValid: {
    firstName: true,
    lastName: true,
    code: true,
    dateOfBirth: true,
    email: true,
    phone: true,
    country: true,
  },
});
```

which might make the `fieldLookup` below redundant?
`      return { ...state, isValid: { [field]: validity } }; `

( alternative thought was to have the shape `{ value: '', isValid: true }` for each property which I also like )

Just feels like a lot of duplication having properties doubled up with their 'IsValid' version

src/widgets/ValidationTextInput.js
```diff
@@ -56,8 +56,8 @@ export const ValidationTextInput = ({
   // to notify the parent.
   const onCheckValidity = React.useCallback(
     inputToCheck => {
-      const newValidState = onValidate(inputToCheck);
-      if (newValidState !== isValid) onChangeValidState(newValidState);
+      const newValidState = onValidate ? onValidate(inputToCheck) : true;
+      if (newValidState !== isValid && onValidate) onChangeValidState(newValidState);
```
mark-prins:
```suggestion
      if (newValidState !== isValid && onChangeValidState) onChangeValidState(newValidState);
```

src/widgets/modals/PatientEdit.js
```diff
@@ -0,0 +1,265 @@
...
+            <ValidationTextInput
+              value={lastName}
+              isRequired
+              label="Last Name"
+              onSubmitEditing={setLastName}
+            />
+            <ValidationTextInput
+              value={code}
+              onValidate={validateCode}
+              invalidMessage="Must be less than 20 characters"
```
mark-prins: not now: translation string for the validation messages

src/widgets/modals/PatientEdit.js
```diff
@@ -0,0 +1,265 @@
...
+    dateOfBirth,
+    email,
+    phone,
+    addressOne,
+    addressTwo,
+    country,
+  } = patient;
+
+  const canSave =
+    firstNameIsValid &&
```
mark-prins: I'm thinking that if `patient` did have an `isValid` property that contained a bunch of boolean props.. then there'd be a way to make this a little more generic i.e. `Object.values(patient.isValid).every(v=>v);`

src/reducers/PatientReducer.js
joshxg: much better :) (just have the values which have some sort of constraint currently in desktop.. although that's a poor excuse to not have better validation on some of these fields)
fa9b01e

src/widgets/ValidationTextInput.js
joshxg: 4b9f63e

src/widgets/modals/PatientEdit.js
joshxg: Have made this issue to track all the localization changes to be done because I am lazy https://github.com/openmsupply/mobile/issues/1693

src/widgets/modals/PatientEdit.js
joshxg: 079307c

src/reducers/PatientReducer.js
mark-prins: Oh, nice, I like what you've done with pulling out `isValid` and adding in `newValidityState`

Issue #441, opened 2017-07-24
Built in reports/dashboard from synced reports and feedback to the user


Chris-Petty:
Description:  
Mobile users would appreciate feedback on how the app is serving them. This should likely be in the form of a dashboard with some basic reports/charts.

Suggestions:
- Item consumption charts/tables (potentially useful for auditors)
- Item stock level charts
- Number of patients/invoices
- Items with low stock/near expiring stock
- Time since last stock take (and button to navigate to that stocktake)
- Current stock inbound from suppliers
- Current stock requested in requisitions

Perhaps a new DB table storing these calculated reports so that temporal information is readily available. This may be similar to what is already implemented in mSupply, so may be worth investigating that to see what can align and be synced.

Comments:
Talk to clients and consultants about what features in this area would benefit them the most. All of this should have a solid design (Open in the public forum) before implementation.

andreievg:
To summarise our discussion yesterday:

We decided to add a new table to mSupply that will sync to mobile site. 

| Field          |Type                        
|----------------|--------------------------------------------|
|ID              | Text UUID                                  |
|storeID         | Text UUID                                  |
|data            | json                                       |
|reportID        | Text UUID (FK to dashboard_report.ID       |
|type            | Text                                       | 
|title           | Text                                       |
|tabName         | Text (optional)                            |

This should be enough to display synced dashboards report in mobile. json data field will be in the same format as current web dashboard.

TODO add link to mSupply Desktop issue

Issue #1709, opened 2019-12-16
Add Report class getters and setters


joshxg:
## Is your feature request related to a problem? Please describe.

Since the `data` field on the `Report` object will be a stringified JSON object, it would be very helpful to have getters/setters to help save these fields.

## Describe the solution you'd like

I imagine a field `get data()` would be very helpful, which would handle the parsing. The same with a `set data(Object object)` field. Basically getters/setters which make dealing with a JSON Stringified object a lot easier.

## Implementation

Not 100% sure!

## Describe alternatives you've considered

N/A

## Additional context

Maybe this is better done after the `Report` class has been used a bit to understand what would be needed, not sure!

diegoevangelisti:
@joshxg 
We need to add getters for `@property type` and `title` here as well, right?

joshxg:
@diegoevangelisti nope - if you don't provide a getter/setter this will act just the same as a normal object property.

PR #1722, opened 2019-12-17
#1708 Report class added to Realm schema

### LINKED ISSUE #1708: Add Report class to realm schema

Add the `Report` class to the realm schema.

The a report class with the following fields

image

Note that this is slightly different from the previous dashboard PR schema as well as the desktop schema. I don't think the additional fields are needed, but I could be wrong

## Implementation

I think this should be a `Realm` class, as it would be very helpful to have getters and setters later

N/A

## Additional context

N/A

diegoevangelisti:
Fixes #1708

## Change summary

Report class was added to the realm schema with the structure suggested in #1708.

## Testing

No sure. Same as #1706 maybe?

Less fields than suggested as first option here #441 were added to Report schema, so maybe later it needs to by checked and adjusted in consequence.

### DIFF
--- a/src/database/DataTypes/index.js
+++ b/src/database/DataTypes/index.js
@@ -8,7 +8,6 @@ import Realm from 'realm';
 export class Address extends Realm.Object {}
 export class ItemCategory extends Realm.Object {}
 export class ItemDepartment extends Realm.Object {}
-export class Report extends Realm.Object {}
 export class Setting extends Realm.Object {}
 export class SyncOut extends Realm.Object {}
 export class TransactionCategory extends Realm.Object {}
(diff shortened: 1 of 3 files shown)


src/database/DataTypes/index.js
```diff
@@ -8,6 +8,7 @@ import Realm from 'realm';
 export class Address extends Realm.Object {}
 export class ItemCategory extends Realm.Object {}
 export class ItemDepartment extends Realm.Object {}
+export class Report extends Realm.Object {}
```
joshxg: Slightly different this time round, I think, because we would want to use `get` and `set` it would be good to have it's own file.

I think adding a file to `dataTypes/Report.js` and adding that line there. It's a small change, but it will make it easier to add getters and setters later.

diegoevangelisti: Yes, thanks! I was just reading #1709 (related with this) and thinking about the same. I will create that file to later add get() and set() there

PR #1723, opened 2019-12-17
#1709 Add report class getters and setters

### LINKED ISSUE #1709: Add Report class getters and setters

diegoevangelisti:
Fixes #1709?

## Change summary

Getter and setter added to handle data in Report class. Mainly since the data field on the Report object will be a stringified JSON object, no will have getters/setters to help save these fields. 

## Testing

Not sure (Same as #1706 maybe?)
- [ ] Run`yarn start` (or `react-native run-android`)
- [ ] Check BUILD SUCCESSFUL result.
- [ ] Again in the CLI, run `yarn build`
- [ ] Check BUILD SUCCESSFUL result again.
- [ ] Finally, the app should be running properly.

Check #1709 for more details.

### DIFF
--- a/src/database/DataTypes/Report.js
+++ b/src/database/DataTypes/Report.js
@@ -2,6 +2,47 @@
  * mSupply Mobile
  * Sustainable Solutions (NZ) Ltd. 2019
  */
+
 import Realm from 'realm';
 
-export class Report extends Realm.Object {}
+/**
+ * A Report.
+ *
+ * @property  {string}            id
+ * @property  {string}            type
+ * @property  {string}            title
+ * @property  {string}            _data
+ */
+
+export class Report extends Realm.Object {
+  /**
+   * Get data.
+   *
+   * @return  {Object}
+   */
+  get data() {
+    return JSON.parse(this._data);
+  }
+
+  /**
+   * Set data.
+   *
+   * @param  {dataObject}  dataObject
+   */
+  set data(dataObject) {
+    this._data = JSON.stringify(dataObject);
+  }
+}
+
+Report.schema = {
+  name: 'Report',
+  primaryKey: 'id',
+  properties: {
+    id: 'string',
+    type: 'string',
+    title: 'string',
+    _data: 'string',
+  },
+};
+
+export default Report;
--- a/src/database/DataTypes/Report.js
+++ b/src/database/DataTypes/Report.js
@@ -5,6 +5,7 @@
 
 import Realm from 'realm';
 
+import checkIsObject from '../utilities';
 /**
  * A Report.
  *
@@ -30,7 +31,7 @@ export class Report extends Realm.Object {
    * @param  {dataObject}  dataObject
    */
   set data(dataObject) {
-    this._data = JSON.stringify(dataObject);
+    this._data = checkIsObject(dataObject) ? JSON.stringify(dataObject) : {};
   }
 }
 
@@ -46,3 +47,5 @@ Report.schema = {
 };
 
 export default Report;
+
+
--- a/src/database/DataTypes/Report.js
+++ b/src/database/DataTypes/Report.js
@@ -31,7 +31,7 @@ export class Report extends Realm.Object {
    * @param  {dataObject}  dataObject
    */
   set data(dataObject) {
-    this._data = checkIsObject(dataObject) ? JSON.stringify(dataObject) : {};
+    this._data = checkIsObject(dataObject) ? JSON.stringify(dataObject) : JSON.stringify({});
   }
 }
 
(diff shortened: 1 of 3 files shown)


src/database/DataTypes/Report.js
```diff
@@ -2,6 +2,47 @@
...
+  get data() {
+    return JSON.parse(this._data);
+  }
+
+  /**
+   * Set data.
+   *
+   * @param  {dataObject}  dataObject
+   */
+  set data(dataObject) {
```
joshxg: Probably be good to be a bit safer with this one. I think some validation here to be safer when we set the property, and ensure that it is always an object, so when we use the getter, we can always be sure and happy that we will be getting an object, and not an array/string or somethng like that.

Not sure what the best would be, ideas could be checking `typeof === "object"` .. but that will return true for any array as `typeof [] === "object"` is true ..

So, might need a little research into the best way to make sure we are always setting a JS object .. because JS types are very funny 🤣 ... `console.log(typeof null)`, `console.log(typeof new Date())` !

diegoevangelisti: 😅God bless JS. Nice discussion about it [here](https://stackoverflow.com/questions/8511281/check-if-a-value-is-an-object-in-javascript)
Maybe we can go ahead with:
```
function isObject(obj) {
  return obj === Object(obj);
}
``` 
And then test it properly. What do you think?

joshxg: just quickly checking it out I get these results:

```
const isObject = obj => obj === Object(obj);

console.log(isObject({})); // True
console.log(isObject([])); // True
console.log(isObject(null)); // False
console.log(isObject(new Date())); // True
console.log(isObject(new Map())); // True
console.log(isObject("")); // False
console.log(isObject(123)); // False
console.log(isObject(undefined)); // False
```

Ideally we would only want `isObject({})` to return `True` - otherwise if we accidently set a `Map` or something like that, it would not be good!

I think a function similar, but with a few more constraints. I wonder if a larger JS library might have some good ideas - for example underscore.js or lodash 🤣

diegoevangelisti: Yeah, because all of them are Obj for JS. 
Screen Shot 2019-12-18 at 9 52 36 AM
Quite a long definition. I will find a proper way to only evaluate `isObject({})` as true as test it to make sure it works 😄

diegoevangelisti: I _think_ I found the solution using this function. Literally:
> To know if a value is an object that can have properties and be looped through, its constructor can be compared to Object. It doesn't work for objects created from classes, then the instanceof operator can be used instead.

Now the tests are passing:

```
const isObject = dataObject => dataObject && typeof dataObject === 'object' && dataObject.constructor === Object;
```
Screen Shot 2019-12-18 at 10 07 49 AM

joshxg: cool :)

I would put this somewhere in `src/utilities` to be used elsewhere (and also I reckon it should return a boolean always?)

andreievg: Any reason why not just surround with try/catch and if error caught return empty object (or some object that can be visualised as 'no-data' in the chart components) ? Btw incoming sync utility should also guarantee that incoming data is a proper object ?

joshxg: setting `[]` will not throw an error, but it will if we try to use it as an object! If incoming sync utility will guarantee it's an object.. it will be doing exactly this, right?

diegoevangelisti: > Btw incoming sync utility should also guarantee that incoming data is a proper object ?
I believe it should guarantee that, yeah.

diegoevangelisti: >(and also I reckon it should return a boolean always?)
That would be the case, I agree. I'll consider that, thanks.

src/database/DataTypes/Report.js
```diff
@@ -30,7 +31,7 @@ export class Report extends Realm.Object {
    * @param  {dataObject}  dataObject
    */
   set data(dataObject) {
-    this._data = JSON.stringify(dataObject);
+    this._data = checkIsObject(dataObject) ? JSON.stringify(dataObject) : {};
```
joshxg: oops, you'll still need to stringify the `{}`

diegoevangelisti: Thanks! Done here 219774d

joshxg:
nice! 🎉

PR #1725, opened 2019-12-17
#1710 Add Report to incoming sync utils

### LINKED ISSUE #1710: Add Report to incomingSyncUtils

Need to add the new `Report` class to `incomingSyncUtils` to enable sync

Add a `case` for the `Report` class in `incomingSyncUtils`

## Implementation

Dependent on the final schema, need to simply add a case statement which will save a report that has come in hot from the server to the database.

N/A

## Additional context

N/A

diegoevangelisti:
Fixes #1710

## Change summary

Report class was added to list of record types to check in `incomingSyncUtils`

## Testing

Not sure by now. Same as #1706 ?

@joshxg @andreievg 
Should I use `checkIsObject()` here as well? In that case, it is not useful to have it in `Report.js`. Once might be enough (as far as I understand), please correct me if I am wrong. Thanks!

### DIFF
--- a/src/sync/incomingSyncUtils.js
+++ b/src/sync/incomingSyncUtils.js
@@ -252,7 +252,7 @@ export const sanityCheckIncomingRecord = (recordType, record) => {
       canBeBlank: ['units', 'comment', 'order_number'],
     },
     Report: {
-      cannotBeBlank: ['ID', 'title', 'type', '_data'],
+      cannotBeBlank: ['ID', 'title', 'type', 'data'],
       canBeBlank: [],
     },
   };
@@ -262,7 +262,7 @@ export const sanityCheckIncomingRecord = (recordType, record) => {
       containsAllFieldsSoFar &&
       record[fieldName] !== null && // Key must exist.
       record[fieldName] !== undefined && // Field may be undefined.
-      record[fieldName].length > 0, // Fild must not be empty string.
+      record[fieldName].length > 0, // Field must not be empty string.
     true
   );
 
--- a/src/sync/incomingSyncUtils.js
+++ b/src/sync/incomingSyncUtils.js
@@ -1,3 +1,4 @@
+/* eslint-disable no-useless-catch */
 /* eslint-disable camelcase */
 import {
   EXTERNAL_TO_INTERNAL,
@@ -13,6 +14,7 @@ import {
 import { CHANGE_TYPES, generateUUID } from '../database';
 import { deleteRecord } from '../database/utilities';
 import { SETTINGS_KEYS } from '../settings';
+import { checkIsObject } from '../utilities';
 
 const { THIS_STORE_ID, THIS_STORE_TAGS, THIS_STORE_CUSTOM_DATA } = SETTINGS_KEYS;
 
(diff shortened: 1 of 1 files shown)


src/sync/incomingSyncUtils.js
```diff
@@ -493,6 +497,16 @@ export const createOrUpdateRecord = (database, settings, recordType, record) =>
...
       break;
     }
+    case 'Report': {
+      internalRecord = {
+        id: record.ID,
+        title: record.title,
+        type: record.type,
+        _data: record.data,
+      };
+      database.update(recordType, internalRecord);
```
joshxg: yeah, I think you should do a check to make sure `data` is valid object that we want to accept and if not, reject the record (again, we can log to bugsnag later).

if we assign the value like this 
```
const someReport = database.objects("Report")[0]
someReport.data = {some: value}
```

Then the setter will be used.

If we set it with `update`, then the setter is not used.

IMO, we should keep both.. because being safe with your database is a good idea!

diegoevangelisti: Great, thanks for the explanation! Asking before commiting, this one we use checkIsObject  twice (before and after using the setter ).
```
case 'Report': {
      const reportData = database.objects('Report')[0];
      const setData = checkIsObject(record.data)
        ? reportData.data(record.data)
        : reportData.data({}); // Need to add a proper case solution for when record.data is not an Object
      internalRecord = {
        id: record.ID,
        title: record.title,
        type: record.type,
        _data: setData,
      };
      database.update(recordType, internalRecord);
      break;
  }
```

joshxg: Not quite. I think my example confused you, sorry. Hope the below will help!

`const reportData = database.objects('Report')[0];` - this will fetch the first `Report` record in the database. I was just using that as an example of accessing a realm object.

What you have is basically correct:

```
// Destructuring is nice!
const {ID:id, title, type, data} = record;

// We need to parse the data, so be safe and use a try
try{

// Parse the data to let us be able to check if it's correct and valid.
const parsedData = JSON.parse(data);

// Check if we should be setting this data. Even if it's parsed, it could still not be a JSONObject!
const shouldSetData = checkIsObject(JSON.parse(data))

// Now create the record we will put in the database. It's faster and easier to use .update() 
// here - but when we use that, the setter will NOT be used, and _data is a string field. If the
// parsed data is not a valid JSONObject, just set null for now.
const internalRecord = { id, title, type, _data: shouldSetData ? data : null }

// Finally, create the record
database.update(recordType, internalRecord)    
}catch(error){
    // Throw to parent, for now
    throw error
}
break;
```

diegoevangelisti: Thanks, much clearer btw :-) Done here 03ff6d7

src/sync/incomingSyncUtils.js
```diff
@@ -249,14 +251,18 @@ export const sanityCheckIncomingRecord = (recordType, record) => {
       cannotBeBlank: [],
       canBeBlank: ['units', 'comment', 'order_number'],
     },
+    Report: {
+      cannotBeBlank: ['ID', 'title', 'type', '_data'],
```
joshxg: this will be the field coming FROM the server which would be `data`. We are using `_data` on mobile just to be difficult 🤣 but the field in the mSupply database will be `data`

src/sync/incomingSyncUtils.js
```diff
@@ -249,14 +251,18 @@ export const sanityCheckIncomingRecord = (recordType, record) => {
...
+    },
   };
   if (!requiredFields[recordType]) return false; // Unsupported record type
   const hasAllNonBlankFields = requiredFields[recordType].cannotBeBlank.reduce(
     (containsAllFieldsSoFar, fieldName) =>
       containsAllFieldsSoFar &&
       record[fieldName] !== null && // Key must exist.
       record[fieldName] !== undefined && // Field may be undefined.
-      record[fieldName].length > 0, // Fidl must not be empty string.
+      record[fieldName].length > 0, // Fild must not be empty string.
```
joshxg: 🤣

diegoevangelisti: I don't want to change this one... 😆

PR #1727, opened 2019-12-18
#1724 more generic patient edit

### LINKED ISSUE #1724: PatientEdit modal should be more generic

`PatientEdit` was a modal create dfor editing patient in the dispensary branch. It needs to be more generic to cater for other use cases. i.e. prescribers

A more generic `FormControl` component which will take some configuration and manage the inputs

## Implementation

N/A

N/A

## Additional context

N/A

joshxg:
Fixes #1724 

## Change summary

- Removes `PatientEdit`
- Adds `FormControl`
- Adds `form` state to redux with a `FormReducer`. I think this is better and makes more sense.
- Removes a good chunk of logic from `PatientReducer` - still keeping it for history etc later. This is good, I think.
- Adds a bunch of config for defining a form. Will add things like the `Slider` made previously as a different type as well as toggles.
- General idea is that the `FormControl` gets handed an array of what to render, and when the constraints (required/validity) are correct and save is pressed, will pass back an object of just the key/value pairs.

## Testing

N/A

- Need to create a `createRecord` case for `Patient` which will set more values and defaults. That's why the edit/update and create actions are separated.
- `Prescriber`/`InsurancePolicy` will use basically the same thing. Possibly something like abbreviations/directions or the cash register to create a new cash in/cashout basic transaction


src/selectors/form.js
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+/**
+ * Selects a boolean from the form state which is an indicator
+ * whether this form is in a valid and complete state.
+ * @return {bool}
+ */
+export const selectCanSaveForm = ({ form }) => {
+  const areAllValid = Object.values(form).every(({ isValid }) => isValid);
+  const allHaveValues = Object.values(form).every(({ hasValue }) => hasValue);
```
mark-prins: :nit minor naming comment - to be strictly consistent `allAreValid` and `allHaveValues` 

I went down a small rabbit hole of wondering if it is ok to separate validity and 'has a value' - a required field is only valid if it has a value, and optional fields don't need a value, so the `hasValue` isn't required.

src/reducers/FormReducer.js
```diff
@@ -0,0 +1,58 @@
...
+
+      const stateData = state[field];
+
+      const { isRequired, validator } = stateData;
+
+      const newStateData = {
+        ...stateData,
+        value,
+        hasValue: isRequired ? !!value : true,
+        isValid: validator ? validator(value) : true,
```
mark-prins: Re a comment I made elsewhere: I'm wondering what value `hasValue` adds - and whether it would be better to have:
```
      const isValid = (isRequired ? !!value : true) && (validator ? validator(value) : true);

      const newStateData = {
        ...stateData,
        value,
        isValid,
      };
```

joshxg: db050fbe668dd962dca695f1c0b72f12cf35aff0

You got me thinking.. hopefully that's a good thing..

I've just used a validator function for the required fields. You're right, a field being required really is just another variable in the 'validity' of that field, so it does seem right to just use the same 'validator' concept for all of this, leaving the `isRequired` field really just for a UI hint. 

Hopefully this is better?

PR #1731, opened 2019-12-18
#1712 Dashboard button on Menu Page

### LINKED ISSUE #1712: Dashboard button on MenuPage

Need to add a Dashboard button to the `MenuPage`

When enabling the dashboard, the dashboard button should shown on the main page, similar to how dispensary/vaccines is working. 

Modules introduced in this PR: https://github.com/openmsupply/mobile/pull/1488

Will need to finalise a `[store]customData` field which when synced to desktop, will enable the dashboard module and the button.

## Implementation

- Will need to add a `usingDashboardModule` field to `ModulesReducer`
- Add a `Dashboard` button to the menu page when dashboard module is enabled.

N/A

## Additional context

N/A

diegoevangelisti:
Fixes #1712 

## Change summary

- A button named `Dashboard` was added to the modules section.  It turns visible when `usesDashboardModule` is TRUE
- `usingDashboard` was added to `ModulesReducer`

## Testing

A complete testing will be taken place before merging feature/dashboard branch to develop.

As discussed with @joshxg, in the future if many more modules are added and are set as visibible, could ended up in not having enough space unless some UI changes.


src/pages/MenuPage.js
```diff
@@ -235,8 +239,13 @@ const mapDispatchToProps = dispatch => ({
 const mapStateToProps = state => {
   const { modules, user } = state;
   const { currentUser } = user;
-  const { usingDispensary } = modules;
-  return { usingDispensary, usingModules: usingDispensary, isAdmin: currentUser?.isAdmin };
+  const { usingDashboard, usingDispensary } = modules;
```
joshxg: you can destructure `usingModules` here as it is being set in state in the `ModulesReducer` (line 26 in modulesReducer.js), then you don't have to do the check in line 247

In fact, you could just do a `return {...modules, isAdmin: currentUSer?.isAdmin }`  - but I kind of like taking the values out .. it is more explicit and helpful for understanding. What do you think?

diegoevangelisti: Yes, I agree! This is a great way to reduce code and keep it clear at the same time.
I would rather go with the one you mentioned but keeping the destructure for modules in this page. 
Good to know we have the chance to just update `const modules = { usingDashboard, usingDispensary, usingVaccines };` in `ModulesReducer` and we don't need to make modifications here:
```
  const { modules, user } = state;
  const { currentUser } = user;
  return { ...modules, isAdmin: currentUser?.isAdmin };
};
```
Anyway, I agree that it would be good to have ` { usingDashboard, usingDispensary, usingVaccine } = modules` in this page to improve understanding. Otherwise let me know and we took those lines out.
Thanks btw :-)

joshxg: I agree.. explicit is best :)

src/pages/MenuPage.js
```diff
@@ -237,13 +237,14 @@ const mapDispatchToProps = dispatch => ({
 });
 
 const mapStateToProps = state => {
-  const { modules, user } = state;
+  const { modules, usingModules, user } = state;
   const { currentUser } = user;
-  const { usingDashboard, usingDispensary } = modules;
+  const { usingDashboard, usingDispensary, usingVaccines } = modules;
```
joshxg: I don't think `usingModules` is defined here.

The redux shape or the `state` variable would be equal to

```
{
    sync: {...}
    pages: {...}
    user: {..}
    navigation: {...}
    ....
    modules: {
        [whatever is returned from `ModuleReducer`,
         usingDashboard,
         usingDispensary,
         usingVaccines,
         usingModules,
    }
,
}
```

So 

```
const { modules, user } = state;
const { usingDashboard, usingDispensary, usingVaccines, usingModules } = modules;
```

diegoevangelisti: Makes sense, I missed that one before! thanks 😄

PR #1743, opened 2019-12-18
#1740 Add PieChart component 

### LINKED ISSUE #1740: PieChart component

Need a Pie chart component for the mobile dashboard.

A component similar the Pie chart created here: https://github.com/openmsupply/mobile/blob/49d8ab764c85ec5e35699661da92aad8c3e1b03a/src/widgets/ReportChart.js

## Implementation

As above.

N/A

## Additional context

N/A

diegoevangelisti:
Fixes #1740

## Change summary

Add `PieChart` to show the pie chart graphs on the dashboard. 

- It uses React Hooks API.
- It shares a similar skeleton than future components `LineChart` and `BarChart`

## Testing

Many lines of code were taken from this [branch](https://github.com/openmsupply/mobile/blob/49d8ab764c85ec5e35699661da92aad8c3e1b03a/src/widgets/ReportChart.js). Further changes may be taken into consideration.

### DIFF
--- a/src/widgets/PieChart.js
+++ b/src/widgets/PieChart.js
@@ -1,9 +1,10 @@
+/* eslint-disable react/forbid-prop-types */
 /**
  * mSupply Mobile
  * Sustainable Solutions (NZ) Ltd. 2018
  */
 
-import React, { useState, useEffect } from 'react';
+import React, { useState } from 'react';
 import { View, StyleSheet } from 'react-native';
 import PropTypes from 'prop-types';
 import { VictoryLabel, VictoryPie } from 'victory-native';
@@ -18,47 +19,43 @@ import { APP_FONT_FAMILY, GREY } from '../globalStyles';
  *                               options are BarChart, LineChart and PieChart.
  * @prop  {data}         array   An array of {x, y} datapoints to plot.
  * @prop  {width}        number  The width of the parent container.
- * @prop  {height}       number  The height of the parent continer.
+ * @prop  {height}       number  The height of the parent container.
  */
 
-export const ReportChart = ({ report: initialReport }, props) => {
-  const [report, setReport] = useState(initialReport);
-  const [width, setWidth] = useState(null);
-  const [height, setHeight] = useState(null);
-
-  // Same affect as componentDidUpdate
-  useEffect(() => {
-    if (report.id !== props.report.id) setReport(props.report);
-  });
+export const PieChart = ({ title, type, data }) => {
+  const [state, setState] = useState({ height: null, width: null });
+  const { height, width } = state;
 
   // Victory Native sizes are set using absolute values. Parents dimensions are used to
   // calculate relative values for width and height for each chart.
+
   const onLayout = event => {
-    setWidth(event.nativeEvent.layout.width);
-    setHeight(event.nativeEvent.layout.height);
+    const newStateObj = {
+      height: event.nativeEvent.layout.width,
+      width: event.nativeEvent.layout.height,
+    };
+    setState(newStateObj);
+  };
+  const {
+    padVertical,
+    padHorizontal,
+    innerRadius,
+    labelRadius,
+    padAngle,
+    colorScale,
+    style,
+  } = victoryStyles.pieChart;
+
+  const padding = {
+    top: height * padVertical,
+    bottom: height * padVertical,
+    right: width * padHorizontal,
+    left: width * padHorizontal,
   };
 
-  const renderPieChart = () => {
-    const {
-      padVertical,
-      padHorizontal,
-      innerRadius,
-      labelRadius,
-      padAngle,
-      colorScale,
-      style,
-    } = victoryStyles.pieChart;
-
-    const padding = {
-      top: height * padVertical,
-      bottom: height * padVertical,
-      right: width * padHorizontal,
-      left: width * padHorizontal,
-    };
-
-    const widthPadded = width * (1 - padHorizontal);
-
-    return (
+  const widthPadded = state.width * (1 - padHorizontal);
+  return (
+    <View style={localStyles.ChartContainer} onLayout={onLayout}>
       <VictoryPie
         width={width}
         height={height}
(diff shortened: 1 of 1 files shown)


src/widgets/PieChart.js
```diff
@@ -0,0 +1,116 @@
...
+ */
+
+import React, { useState, useEffect } from 'react';
+import { View, StyleSheet } from 'react-native';
+import PropTypes from 'prop-types';
+import { VictoryLabel, VictoryPie } from 'victory-native';
+import { APP_FONT_FAMILY, GREY } from '../globalStyles';
+
+/**
+ * A charting widget for displaying Pie chart report.
```
joshxg: Docstrings imo should be a summary to help other programs HOW to use something, rather than WHAT something is. If I am reading this file, probably I want to use your new `PieChart` component.. then I already really know what this is going to be, I just don't know how to use it yet!

src/widgets/PieChart.js
```diff
@@ -0,0 +1,116 @@
...
+ * @prop  {string}       id      The report ID.
+ * @prop  {string}       title   The title of the report.
+ * @prop  {string}       type    The type of chart to use to display the report,
+ *                               options are BarChart, LineChart and PieChart.
+ * @prop  {data}         array   An array of {x, y} datapoints to plot.
+ * @prop  {width}        number  The width of the parent container.
+ * @prop  {height}       number  The height of the parent continer.
+ */
+
+export const ReportChart = ({ report: initialReport }, props) => {
```
joshxg: this is two parameters, not one. I suggest just destructuring all:

`export const ReportChart = ({ report: initialReport, title, type, data, width, height}) => { .. }`

Then, in the rest of the component you don't need to use `props.title` etc.. just `title`

src/widgets/PieChart.js
```diff
@@ -0,0 +1,116 @@
...
+  const [report, setReport] = useState(initialReport);
+  const [width, setWidth] = useState(null);
+  const [height, setHeight] = useState(null);
+
+  // Same affect as componentDidUpdate
+  useEffect(() => {
+    if (report.id !== props.report.id) setReport(props.report);
+  });
```
joshxg: The previous component was trying to manage being passed different reports and re-rendering. I.e. when you click on the side bar it would render a new chart.

In this case, your `PieChart` is a single component that won't really change, so I don't think this is neccessary at all.

src/widgets/PieChart.js
```diff
@@ -0,0 +1,116 @@
...
+ * @prop  {string}       title   The title of the report.
+ * @prop  {string}       type    The type of chart to use to display the report,
+ *                               options are BarChart, LineChart and PieChart.
+ * @prop  {data}         array   An array of {x, y} datapoints to plot.
+ * @prop  {width}        number  The width of the parent container.
+ * @prop  {height}       number  The height of the parent continer.
+ */
+
+export const ReportChart = ({ report: initialReport }, props) => {
+  const [report, setReport] = useState(initialReport);
```
joshxg: keeping values in state as well as props is generally an anti-pattern.

If you think about it, it doesn't really make sense to have the value passed to your component also held in state.. you're just duplicating the data for no real reason!

src/widgets/PieChart.js
```diff
@@ -0,0 +1,116 @@
...
+  // Same affect as componentDidUpdate
+  useEffect(() => {
+    if (report.id !== props.report.id) setReport(props.report);
+  });
+
+  // Victory Native sizes are set using absolute values. Parents dimensions are used to
+  // calculate relative values for width and height for each chart.
+  const onLayout = event => {
+    setWidth(event.nativeEvent.layout.width);
+    setHeight(event.nativeEvent.layout.height);
```
joshxg: I think keep these in an object `{ height: x, width: y }`

In almost every use case, we will update these at the same time. But, when we call the function that is returned from `useState` (in the case `setWidth` or `setHeight`) this will cause the function to re-render. So, we would be re-render twice!

src/widgets/PieChart.js
joshxg: Also I think you should call this `PieChart`

src/widgets/PieChart.js
```diff
@@ -0,0 +1,116 @@
...
+    switch (report.type) {
+      case 'PieChart':
+        return renderPieChart();
+      default:
+        return null;
+    }
+  };
+  return (
+    <View style={localStyles.ChartContainer} onLayout={onLayout}>
+      {renderVisualisation()}
```
joshxg: You haven't taken out most of this logic which was there when this component was switching between different charts.

There's basically no need for the functions `renderVisualisation` or `renderPieChart` - in fact `renderPieChart` is basically what this component should be.

So I suggest, take the code from `renderPieChart` out of the function, and place it inside the `PieChart` component.. and basically remove the rest! hopefully that makes sense

src/widgets/PieChart.js
diegoevangelisti: Agree, we should be more explicit here. Anyway I was planning to don't touch it now and create an issue to update all docstrings later, given instructions after having a bigger picture of what to write that can be really useful! :-)

src/widgets/PieChart.js
diegoevangelisti: Thanks, I took it out!

src/widgets/PieChart.js
diegoevangelisti: Thanks. In fact, what we are getting is just title, type and data as props, so `export const PieChart = ({ title, type, data }) => { .. }` may be enough. 
Done here 0f3f2a7

src/widgets/PieChart.js
diegoevangelisti: Corrected here 0f3f2a7
Not totally sure about this one:
`const [state, setState] = useState({ height: null, width: null });`

src/widgets/PieChart.js
diegoevangelisti: Done here: 0f3f2a7
Also not 100% sure about this, maybe there's a better way to do it: 
```
const onLayout = event => {
    const newStateObj = {
      height: event.nativeEvent.layout.width,
      width: event.nativeEvent.layout.height,
    };
    setState(newStateObj);
  };
```

src/widgets/PieChart.js
diegoevangelisti: Its name can cause confusions, what do you think? Maybe `heightAndWidth` or `layoutSizes`, etc can represent them better than just `state`?

src/widgets/PieChart.js
diegoevangelisti: I made those changes. Thanks btw ;-)

src/widgets/PieChart.js
diegoevangelisti: 100% agree, better to avoid confusions. Issue added here #1745

src/widgets/PieChart.js
```diff
@@ -0,0 +1,100 @@
...
+ * @prop  {string}       title   The title of the report.
+ * @prop  {string}       type    The type of chart to use to display the report,
+ *                               options are BarChart, LineChart and PieChart.
+ * @prop  {data}         array   An array of {x, y} datapoints to plot.
+ * @prop  {width}        number  The width of the parent container.
+ * @prop  {height}       number  The height of the parent container.
+ */
+
+export const PieChart = ({ title, type, data }) => {
+  const [state, setState] = useState({ height: null, width: null });
```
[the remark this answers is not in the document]
diegoevangelisti: Fits really well! Changed here 92fbcef

diegoevangelisti:
> There is a hook `useLayoutEffect`: https://reactjs.org/docs/hooks-reference.html#uselayouteffect
> 
> And you can also create custom hooks, which means you can share all this logic between different components (as you will do this for every component, instead of rewriting the same code): https://reactjs.org/docs/hooks-custom.html

Wow, that's great. I will give it a try later, sure! Thanks :-)

PR #1746, opened 2019-12-19
#1737 SimpleLabel & Seperator components

### LINKED ISSUE #1737: SimpleLabel component

Various places in the prescription form need a label

A simple display component 

## Implementation

Simple component taking some text and possibly style parameters for size etc.

NA

## Additional context

N/A

joshxg:
Fixes #1737 

## Change summary

- Adds a `SimpleLabel` component:

[This is three `SimpleLabel`'s]
image

- Also trying to sneak in a `Seperator` component 
image

## Testing
N/A

- Trying to break components down to more basic parts
- Will need in the future a better project layout - `widgets` is getting a bit nutty!

### DIFF
--- a/src/widgets/Seperator.js
+++ b/src/widgets/Seperator.js
@@ -1,0 +1,57 @@
+/* eslint-disable react/no-unused-prop-types */
+/**
+ * mSupply Mobile
+ * Sustainable Solutions (NZ) Ltd. 2019
+ */
+
+import React from 'react';
+import { View } from 'react-native';
+import PropTypes from 'prop-types';
+
+import { GREY } from '../globalStyles/index';
+
+/**
+ * Simple Separator display component rendering a line.
+ *
+ * @prop {Number} width            The width of the seperator
+ * @prop {String} length           The length of the seperator as a percentage string.
+ * @prop {Number} marginBottom     Number of pixels for the bottom margin.
+ * @prop {Number} marginTop        Number of pixels for the bottom margin.
+ * @prop {Number} marginHorizontal Number of pixels for the top & bottom margin.
+ */
+export const Seperator = props => {
+  const { mainContainerStyle, borderContainerStyle } = localStyles(props);
+  return (
+    <View style={mainContainerStyle}>
+      <View style={borderContainerStyle} />
+    </View>
+  );
+};
+
+Seperator.defaultProps = {
+  width: 1,
+  length: '100%',
+  marginBottom: 5,
+  marginTop: 0,
+  marginHorizontal: 0,
+};
+
+Seperator.propTypes = {
+  width: PropTypes.number,
+  length: PropTypes.string,
+  marginBottom: PropTypes.number,
+  marginTop: PropTypes.number,
+  marginHorizontal: PropTypes.number,
+};
+
+const localStyles = ({ width, length, marginBottom, marginTop, marginHorizontal }) => ({
+  mainConatinerStyle: {
+    flex: 1,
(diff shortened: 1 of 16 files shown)


src/widgets/Seperator.js
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Simple Separator display component rendering a line.
+ *
+ * @prop {Number} width            The width of the seperator
+ * @prop {String} length           The length of the seperator as a percentage string.
+ * @prop {Number} marginBottom     Number of pixels for the bottom margin.
+ * @prop {Number} marginTop        Number of pixels for the bottom margin.
+ * @prop {Number} marginHorizontal Number of pixels for the top & bottom margin.
+ */
+export const Seperator = props => {
```
Chris-Petty:
```suggestion
export const Separator = props => {
```
Also the file name. Probably worth search/replace in the repo.

src/widgets/Seperator.js
```diff
@@ -0,0 +1,57 @@
...
+Seperator.propTypes = {
+  width: PropTypes.number,
+  length: PropTypes.string,
+  marginBottom: PropTypes.number,
+  marginTop: PropTypes.number,
+  marginHorizontal: PropTypes.number,
+};
+
+const localStyles = ({ width, length, marginBottom, marginTop, marginHorizontal }) => ({
+  mainConatinerStyle: {
```
Chris-Petty:
```suggestion
  mainContainerStyle: {
```

src/widgets/Seperator.js
```diff
@@ -0,0 +1,57 @@
...
+ * @prop {Number} width            The width of the seperator
+ * @prop {String} length           The length of the seperator as a percentage string.
+ * @prop {Number} marginBottom     Number of pixels for the bottom margin.
+ * @prop {Number} marginTop        Number of pixels for the bottom margin.
+ * @prop {Number} marginHorizontal Number of pixels for the top & bottom margin.
+ */
+export const Seperator = props => {
+  const { mainContainerStyle, borderContainerStyle } = localStyles(props);
+  return (
+    <View style={mainContainerStyle}>
```
Chris-Petty: I could be wrong, but you may be able to do this component without a container view? Miss naming the style might support this notion 😄

Chris-Petty: That said, given it's just a border, couldn't you just put a border under the label container?

src/widgets/Seperator.js
joshxg: fec4ca9ff6ddf924a88a96d48137875124ac2dbd - includes style spelling error fix

src/widgets/Seperator.js
joshxg: fec4ca9ff6ddf924a88a96d48137875124ac2dbd - includes style spelling error fix

src/widgets/Seperator.js
joshxg: - Need the container to be able to do non full length like:
image

- In my opinion we really need more smaller components to reuse. I'm quite frequently writing the same things where having generic display components would help a lot. It's incredibly noticeable the benefits when working in a project which has a decent library of simple reusable presentation components v this project where there are minimal!

- For this component specifically, it's seems to be a lot easier to chuck in a little `<Seperator/> when you need a quick grouping line rather than messing around with styles.. padding/margins etc etc and, making it work for all uses of the component

Chris-Petty: I see, generally agree with having a library of reusable small components in the repo, and fair enough

PR #1747, opened 2019-12-19
#1742 Create Line chart component

### LINKED ISSUE #1742: LineChart component

Need a Line chart component for the mobile dashboard.

A component similar the Line chart created here: https://github.com/openmsupply/mobile/blob/49d8ab764c85ec5e35699661da92aad8c3e1b03a/src/widgets/ReportChart.js

## Implementation

As above.

N/A

## Additional context

N/A

diegoevangelisti:
Fixes #1742 

## Change summary

Add `LineChart` component to show line chart graphs on the dashboard. 

- It uses React Hooks API.
- It shares a similar skeleton than `LineChart`, `BarChart` and `ReportTable`

## Testing

### DIFF
--- a/src/widgets/LineChart.js
+++ b/src/widgets/LineChart.js
@@ -77,7 +77,6 @@ const localStyles = StyleSheet.create({
     justifyContent: 'center',
   },
 });
-
 const victoryStyles = {
   axisX: {
     fixLabelOverlap: true,
@@ -88,20 +87,11 @@ const victoryStyles = {
     },
   },
   axisY: {
-    fixLabelOverlap: false,
     style: {
       axis: { stroke: LIGHT_GREY },
       ticks: { stroke: DARK_GREY },
       tickLabels: { fontFamily: APP_FONT_FAMILY, fill: GREY },
     },
-  },
-  barChart: {
-    padTop: 0.1,
-    padBottom: 0.15,
-    padLeft: 0.1,
-    padRight: 0.05,
-    padDomain: 0.05,
-    style: { data: { fill: SUSSOL_ORANGE } },
   },
   lineChart: {
     padTop: 0.1,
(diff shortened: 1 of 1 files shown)


src/widgets/LineChart.js
```diff
@@ -0,0 +1,115 @@
...
+  lineChart: {
+    padTop: 0.1,
+    padBottom: 0.15,
+    padLeft: 0.1,
+    padRight: 0.05,
+    dotSize: 3.5,
+    dotStyle: { data: { fill: SUSSOL_ORANGE } },
+    lineStyle: { data: { stroke: SUSSOL_ORANGE } },
+  },
+};
```
joshxg: not sure if this was the same in the last chart also, but a few of these fields aren't used in this component!

For example, `axisX.fixLabelOverlap` isn't used.. which means there is just a `style` field.. so we have some un-needed nesting imo.

Either we can tidy these up for each file, or create a single file for each chart component to import from. I think just tidying it up in each file is the tidiest?

diegoevangelisti: Yes, good eye btw. I'm checking it right now.
e.g. `fixLabelOverlap: false` is the default value, so no needed and also `pieChart` styles doesn't belong to this component, I removed them as well.
I just checked the recently merged `PieChart` component and it looks good.

By now I guess tidying it up in each file is better and later we can think about moving these styles to added files imo. Can add an issue to do it too!

joshxg: perfect!

diegoevangelisti: Thanks! Done here 0103d70
I will definitely add the issue related with these styles and work on polish them.

diegoevangelisti:
> 🎉
> 
> Nice! Try and keep the eslint-disabling to a minimum - all good for this one in your feature branch, but for example - no-unused vars is a good rule! Whats the point of a variable if it's never used? 🤣

Thanks! Sure... In this one I thought it was good to keep e.g. `title` and `type` explicit in case needed, but it's true, we don't use them (at least by now) 😄

Btw, this is the issue to arrange the discussed styles: #1751

PR #1749, opened 2019-12-19
#1748 misc dispensing components

### LINKED ISSUE #1748: Misc components needed for Prescriptions

In order to do #1738 there are a couple more components required

components:
- A circular icon component because circular icons are nice
- A menu icon added to `icons.js`
- 3 layout components: a stepper + a label, a list of details for an item, a dropdown menu + label + burger menu button

this is to create a row here:
image

which would be broken down like:
image

## Implementation

As above

N/A

## Additional context

N/A

joshxg:
Fixes #1748 

## Change summary

- Adds quite a few components..

`CircleButton`:
image

`StepperRow`:
image

`DetailRow`:
image

`DropdownRow`:
image

None are really 100% right, but in general I think they're minor style adjustments that can be iterated on when pulling together

## Testing

N/A

- Really need better ways to create things .. more smaller components to use I think.. like `<FlexRow flex={2}> ...` which is just a `View` with `flexDirection: "row", flex: flex` .. same with colums etc!

### DIFF
--- a/src/widgets/DetailRow.js
+++ b/src/widgets/DetailRow.js
@@ -22,7 +22,7 @@ export const DetailRow = ({ details }) => {
   const Details = React.useCallback(
     () =>
       details.map(({ label, text }) => (
-        <View>
+        <View key={label}>
           <SimpleLabel label={label} size="small" text={text} />
         </View>
       )),
--- a/src/widgets/DetailRow.js
+++ b/src/widgets/DetailRow.js
@@ -22,7 +22,7 @@ export const DetailRow = ({ details }) => {
   const Details = React.useCallback(
     () =>
       details.map(({ label, text }) => (
-        <View key={label}>
+        <View key={label} style={{ flex: 1 }}>
           <SimpleLabel label={label} size="small" text={text} />
         </View>
       )),
@@ -40,6 +40,8 @@ const localStyles = {
   rowContainer: {
     flexDirection: 'row',
     alignItems: 'center',
+    justifyContent: 'flex-start',
+    flex: 1,
   },
 };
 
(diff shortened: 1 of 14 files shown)


src/widgets/DetailRow.js
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * Each SimpleLabel has a label and some other text.
+ *
+ * @prop {Array} details Array of objects in the form {label, text}
+ */
+export const DetailRow = ({ details }) => {
+  const Details = React.useCallback(
+    () =>
+      details.map(({ label, text }) => (
+        <View>
```
mark-prins: should have a `Key` on the `View`

src/widgets/icons.js
```diff
@@ -86,3 +86,13 @@ export const ChevronRightIcon = () => (
 );
 
 export const FavouriteStarIcon = () => <FAIcon name="star-o" color={SUSSOL_ORANGE} size={20} />;
+export const BurgerMenuIcon = () => <EntypoIcon name="menu" color={SUSSOL_ORANGE} size={30} />;
```
mark-prins: three different icon libraries?! does that mean we have an inconsistent look or something?

src/widgets/DetailRow.js
joshxg: e43d55f892a0181d970a74864b87b03a800c0abf

src/widgets/icons.js
joshxg: I've made this `icons` file to try and get it more consistent.. previously we had 3 different magnifying glass icons! But yes.. and I think it's 6 icon libraries.... :)

Issue #1751, opened 2019-12-19
Tidy up styles in new components in feature/dashboard


diegoevangelisti:
## The suggestion

Work on tidying up styles in added components for the new feature/dashboard. Some similar code can be taken to extra dedicated files and then being imported.
Some issues where can be applicable: #1740, #1741, #1742, #1744 

**Why?**
This will create a more tidy and consistence code. Those mentioned component share some styles and it can be a good chance to organize them properly.

## Describe alternatives you've considered

Leave them like that, but that won't be the best case.

## Additional context

For example:
```
const localStyles = StyleSheet.create({
  ChartContainer: {
    width: '75%',
    minHeight: '100%',
    backgroundColor: 'white',
    alignItems: 'center',
    justifyContent: 'center',
  },
});
```
That code is shared by `PieChart`, `BarChart`, `LineChart`, `ReportTable` component.

joshxg:
Closed by #1820

PR #1756, opened 2019-12-19
#1755 Create ReportCell component

### LINKED ISSUE #1755: Create ReportCell component

Need a `ReportCell` component for the `ReportTable` we are implementing in the mobile dashboard.

A component similar the ReportTable created here: https://github.com/openmsupply/mobile/blob/49d8ab764c85ec5e35699661da92aad8c3e1b03a/src/widgets/ReportChart.js

## Implementation

As above.

N/A

## Additional context

N/A

diegoevangelisti:
Fixes #1755 

## Change summary

Add `ReportCell` component to use together with `ReportRow` and `ReportTable` components for showing Table reports on the feature/dashboard. 

- It uses React Hooks API.
- Needed in conjunction with `ReportRow` and `ReportTable` components.

## Testing

### DIFF
--- a/src/widgets/ReportTable/ReportCell.js
+++ b/src/widgets/ReportTable/ReportCell.js
@@ -1,0 +1,41 @@
+/**
+ * mSupply Mobile
+ * Sustainable Solutions (NZ) Ltd. 2018
+ */
+
+import React from 'react';
+import { View, Text, StyleSheet } from 'react-native';
+import PropTypes from 'prop-types';
+import { APP_FONT_FAMILY, BACKGROUND_COLOR } from '../../globalStyles';
+
+export const ReportCell = ({ even, children }) => {
+  const backgroundColor = even ? { backgroundColor: 'white' } : null;
+  return (
+    <View style={[localStyles.container, backgroundColor]}>
+      <Text style={localStyles.cell}>{children}</Text>
+    </View>
+  );
+};
+
+ReportCell.propTypes = {
+  even: PropTypes.bool,
+  children: PropTypes.string.isRequired,
+};
+
+const localStyles = StyleSheet.create({
+  container: {
+    flex: 1,
+    justifyContent: 'center',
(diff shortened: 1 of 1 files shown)


src/widgets/ReportTable/ReportCell.js
```diff
@@ -0,0 +1,41 @@
...
+ReportCell.propTypes = {
+  even: PropTypes.bool,
+  children: PropTypes.string.isRequired,
+};
+
+const localStyles = StyleSheet.create({
+  container: {
+    flex: 1,
+    justifyContent: 'center',
+    paddingLeft: '2%',
```
joshxg: probably better to not use this and just use flexbox .. but we can organise styles later when the component is all together.. much easier!

src/widgets/ReportTable/ReportCell.js
```diff
@@ -0,0 +1,41 @@
...
+ * mSupply Mobile
+ * Sustainable Solutions (NZ) Ltd. 2018
+ */
+
+import React from 'react';
+import { View, Text, StyleSheet } from 'react-native';
+import PropTypes from 'prop-types';
+import { APP_FONT_FAMILY, BACKGROUND_COLOR } from '../../globalStyles';
+
+export const ReportCell = ({ even, children }) => {
```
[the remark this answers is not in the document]
diegoevangelisti: > It's a little more intuitive imo to have a prop `value` so instead, we just do `<ReportCell value={"Some content"}/>` and in the return we return `{value}` rather than `{children}`

I like that idea you just mentioned! and it looks clearer btw. Agree, better I keep it like this now and move forward, then we can work on this important stuff. :-)

src/widgets/ReportTable/ReportCell.js
diegoevangelisti: Agree, we can work on it here: #1751 
Thanks!

Issue #1744, opened 2019-12-19
ReportTable component


diegoevangelisti:
## Is your feature request related to a problem? Please describe.

Need a ReportTable component for the mobile dashboard.

## Describe the solution you'd like

A component similar the ReportTable created here: https://github.com/openmsupply/mobile/blob/49d8ab764c85ec5e35699661da92aad8c3e1b03a/src/widgets/ReportChart.js

## Implementation

As above.

## Describe alternatives you've considered

N/A

## Additional context

N/A

diegoevangelisti:
Regarding this one, I am going to create `ReportTableCell`, `ReportTableRow` and `ReportTable` components, so I reckon the best way is to add a couple of issues to split those tasks. I am working on it right now.

diegoevangelisti:
@joshxg almost done with this one (I reckon) but maybe you can give me a hand to understand how `const renderItem = ({ item, rowIndex }) => { ... }` is receiving { item, rowIndex } because I can't see the renderItem having those arguments when making the call. Thanks!
Btw this is the function I'm mentioning:
```
export const ReportTable = ({ rows, header }) => {

  const renderItem = ({ item, rowIndex }) => {
    return <ReportRow rowData={item} rowIndex={rowIndex} />;
  };

  const renderHeader = () => {
    return <ReportRow rowData={header} isHeader rowIndex={0} />;
  };

  // TODO: KeyExtractor should be altered to not use the index.
  render() {
    return (
      <View style={localStyles.container}>
        {renderHeader()}
        <FlatList
          data={rows}
          renderItem={renderItem}
          keyExtractor={(_, rowIndex) => `${rowIndex}`}
        />
      </View>
    );
  }
}
```

joshxg:
https://facebook.github.io/react-native/docs/flatlist#renderitem

`FlatList` will call the `renderItem` function you pass as a prop for every element in `data`.

For example:

```

const renderItem = (props) => {
    const {item, index} = props;
    const {id, value} = item;
    return <View><Text>{index}:</Text>{id} -- <Text><Text>Value</Text></View>
}
<FlatList data={[{id: 1, value: "test one"}, {id:2, value: "test two"}]} renderItem={renderItem}/>
```

This will return a list looking like:

```
0: 1 -- test one
1: 2 -- test two
```

-----------

I dunno if thats clear.. I'll give another try:

`FlatList` will call the `renderItem` function passed like: 

```
[...]

PR #1758, opened 2019-12-20
#1744 Create ReportTable component

### LINKED ISSUE #1755: Create ReportCell component

diegoevangelisti:
Fixes #1755 

## Change summary

Add `ReportTable` component to use together with `ReportRow` and `ReportCell` components for visualising Table reports on the feature/dashboard. 

- It uses React Hooks API.
- Needed in conjunction with `ReportRow` and `ReportCell` components.

## Testing

### DIFF
--- a/src/widgets/ReportTable/ReportTable.js
+++ b/src/widgets/ReportTable/ReportTable.js
@@ -1,91 +1,37 @@
 /* eslint-disable react/prop-types */
-/* eslint-disable no-unused-vars */
 /* eslint-disable react/forbid-prop-types */
 /**
  * mSupply Mobile
  * Sustainable Solutions (NZ) Ltd. 2018
  */
 
-import React, { useState } from 'react';
-import { View, StyleSheet } from 'react-native';
+import React from 'react';
+import { StyleSheet, FlatList, View } from 'react-native';
 import PropTypes from 'prop-types';
-import { VictoryChart, VictoryBar, VictoryAxis } from 'victory-native';
-import { APP_FONT_FAMILY, GREY, LIGHT_GREY, DARK_GREY, SUSSOL_ORANGE } from '../../globalStyles';
+import { LIGHT_GREY } from '../../globalStyles';
+import { ReportRow } from './ReportRow';
 
-export const ReportTable = ({ title, type, data }) => {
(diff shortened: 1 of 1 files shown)


src/widgets/ReportTable/ReportTable.js
```diff
@@ -0,0 +1,37 @@
...
+import React from 'react';
+import { StyleSheet, FlatList, View } from 'react-native';
+import PropTypes from 'prop-types';
+import { LIGHT_GREY } from '../../globalStyles';
+import { ReportRow } from './ReportRow';
+
+export const ReportTable = ({ rows, header }) => {
+  const renderItem = ({ item, index }) => <ReportRow rowData={item} rowIndex={index} />;
+
+  const renderHeader = () => <ReportRow rowData={header} isHeader rowIndex={0} />;
```
joshxg: This is fine, but there is also this prop on `FlatList` just for interest!  https://facebook.github.io/react-native/docs/flatlist#listheadercomponent

src/widgets/ReportTable/ReportTable.js
```diff
@@ -0,0 +1,37 @@
+/* eslint-disable react/prop-types */
```
joshxg: this is for the `renderItem` call? Theres some info here about that if you're interested https://github.com/yannickcr/eslint-plugin-react/issues/2325

diegoevangelisti: Yes, I have a problem with that indeed. If I declared them then `es-lint` says I'm not using them 😅

src/widgets/ReportTable/ReportTable.js
diegoevangelisti: I made this change here: 7218d46
Hope it works, otherwise I can come back to previous stage :-) thanks for this!

joshxg: I think that should work!

PR #1760, opened 2019-12-20
#1713 Create ReportChart component

### LINKED ISSUE #1713: ReportChart component

Need a chart component for displaying charts/graphs for the mobile dashboard 

A component similar to https://github.com/openmsupply/mobile/blob/49d8ab764c85ec5e35699661da92aad8c3e1b03a/src/widgets/ReportChart.js

However, I *think* a better implementation would be to separate out the different charts - pie/line/bar into their own components, rather than having a single component.

## Implementation

As above. Whoever the implementer is can decide or we can discuss in the issue whether individual components or a single component is best.

N/A

## Additional context

N/A

diegoevangelisti:
Fixes #1713

## Change summary

This component manages which is the report that is needed to be render (PieChart, BarChart, LineChart, TableReport) according to its `Type`.

## Testing

### DIFF
--- a/src/widgets/ReportChart.js
+++ b/src/widgets/ReportChart.js
@@ -1,0 +1,81 @@
+/* eslint-disable react/forbid-prop-types */
+/**
+ * mSupply Mobile
+ * Sustainable Solutions (NZ) Ltd. 2018
+ */
+
+import React, { useState } from 'react';
+import { View, StyleSheet } from 'react-native';
+import PropTypes from 'prop-types';
+import { PieChart } from './PieChart';
+import { BarChart } from './BarChart';
+import { LineChart } from './LineChart';
+import { ReportTable } from './ReportTable';
+
+/**
+ * A charting widget for displaying reports as bar charts, line charts and pie graphs.
+ *
+ * @prop  {string}       id      The report ID.
+ * @prop  {string}       title   The title of the report.
+ * @prop  {string}       type    The type of chart to use to display the report,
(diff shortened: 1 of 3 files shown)


src/widgets/ReportChart.js
```diff
@@ -0,0 +1,64 @@
...
+
+  // Victory Native sizes are set using absolute values. Parents dimensions are used to
+  // calculate relative values for width and height for each chart.
+  const onLayout = event => {
+    const newDimensionsObj = {
+      height: event.nativeEvent.layout.width,
+      width: event.nativeEvent.layout.height,
+    };
+    setDimensions(newDimensionsObj);
+  };
```
joshxg: I think if we're doing this here we should *only* do it here, not in the children. So, that would mean passing to the charts i.e.

```
const BarChart = ({width, height, report}) => {
 ....
}
```

and using that here :

```
case 'BarChart':
    return <BarChart report={report} width={width} height={height}/>
```

There doesn't seem to be a reason to do it twice

Alternatively, you can just keep it in the chart components, and have them do the check for width/height in all of them, but I think it's nicer to do it the one time in here

diegoevangelisti: I agree! that's what we discussed yesterday afternoon and totally happy with keeping that in the parent and take it out from children components.

diegoevangelisti: I will create a issue to remove width and height calculation on children components as well.

joshxg: ok sure!

PR #1762, opened 2019-12-20
#1761 Create ReportSideBarItem component

### LINKED ISSUE #1761: ReportSideBarItem component to use in ReportSideBar

Need a list of items on the `ReportSideBar` component to can select which report to visualize.

Design a component to be used in conjunction with ReportSidebar as the rendered items.

Ideas taken from here (old branch): https://github.com/openmsupply/mobile/blob/49d8ab764c85ec5e35699661da92aad8c3e1b03a/src/widgets/ReportChart.js

diegoevangelisti:
Fixes #1761 

## Change summary

A new component named `ReportSideBarItem` was created in order to render each report (item) to be selected to visualise.

 - It uses React Hooks API.
- Needed in conjunction with `ReportSideBar`

## Testing

### DIFF
--- a/src/widgets/ReportSideBar/ReportSideBarItem.js
+++ b/src/widgets/ReportSideBar/ReportSideBarItem.js
@@ -1,0 +1,90 @@
+/**
+ * mSupply Mobile
+ * Sustainable Solutions (NZ) Ltd. 2018
+ */
+
+import React from 'react';
+import { View, Text, TouchableOpacity, StyleSheet } from 'react-native';
+import PropTypes from 'prop-types';
+import Icon from 'react-native-vector-icons/FontAwesome';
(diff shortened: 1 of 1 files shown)


diegoevangelisti:
> 🎉 🎉
> 
> logically looks all good! There would be a couple of minor adjustments I would make regarding styles (CapitalCase instead of camelCase).. also I think generally using arrays with styles to combine them is a pretty old way which causes a couple of performance issues, but it's very minor and probably easier to show in person.
> 
> Also a couple of hook optimizations we could use (`useCallback` for the `onPressItem`), but again very minor and easier to do a little later and with in person explanations

Great review as always! Sure, I will be happy to discuss about those changes later 🙂

PR #1763, opened 2019-12-20
#1714 Create ReportSideBar component

diegoevangelisti:
Fixes #1714 

## Change summary

This component manages the list of items available to be render, using its children `ReportSideBarItem` component.

## Testing

Further testing will be taken once the feature/dashboard is intended to be merged into develop.

### Related areas to think about

Codebase taken from past work: [branch](https://github.com/openmsupply/mobile/blob/49d8ab764c85ec5e35699661da92aad8c3e1b03a/src/widgets/ReportChart.js). Further changes may be taken into consideration.

PR #1764, opened 2019-12-20
#1716 Crete DashboardPage component

### LINKED ISSUE #1716: DashboardPage component

Need an actual `DashboardPage` to finish off the feature!

A `DashboardPage` component which brings together other issues!

## Implementation

MobileDashboard-2019-12-05 at 4 44 45 PM

A `DashboardPage` component which brings together the `ReportSidebar`, `ReportChart` and `DashboardReducer` and is navigated to through the `DashboardButton` on the `MenuPage`!

N/A

## Additional context

N/A

diegoevangelisti:
Fixes #1716

## Change summary

A `DashboardPage` component was created in order to manage all children components previously created to render the different kind of charts.

## Testing


src/widgets/ReportSideBar/ReportSideBar.js
```diff
@@ -0,0 +1,75 @@
...
+    margin: 0,
+  },
+});
+
+ReportSideBar.propTypes = {
+  item: PropTypes.objectOf(PropTypes.object).isRequired,
+  data: PropTypes.PropTypes.shape([]).isRequired,
+  onPressItem: PropTypes.func.isRequired,
+  selectedItemIndex: PropTypes.number.isRequired,
+  dimensions: PropTypes.objectOf(PropTypes.object).isRequired,
```
joshxg: These propTypes too!

src/widgets/DashboardPage.js
```diff
@@ -0,0 +1,82 @@
...
+        .map((report, index) => ({
+          id: report.reportID,
+          index,
+          title: report.title,
+          type: report.type,
+          date: report.date,
+          data: JSON.parse(report.data),
+        }))
+    );
+  });
```
joshxg: `useEffect` need a dependencies list to manage when it should be called. At the moment, this will run every time the component renders, and will call `setReports` which will re-render the component.. causing an infinite loop!

needs to be `useEffect(() => { ... }, [])` in order to only be set when the component mounts.

src/widgets/DashboardPage.js
```diff
@@ -0,0 +1,82 @@
...
+        .objects('Report')
+        .snapshot()
+        .map((report, index) => ({
+          id: report.reportID,
+          index,
+          title: report.title,
+          type: report.type,
+          date: report.date,
+          data: JSON.parse(report.data),
+        }))
```
joshxg: I think we don't really need this `map`, doesn't really seem to be do anything!

src/widgets/DashboardPage.js
```diff
@@ -0,0 +1,82 @@
...
+const localStyles = StyleSheet.create({
+  sidebarDimensions: {
+    width: '25%',
+    height: '100%',
+  },
+});
+
+DashboardPage.propTypes = {
+  database: PropTypes.objectOf(PropTypes.object()).isRequired,
+  settings: PropTypes.objectOf(PropTypes.object()).isRequired,
```
joshxg: these `propTypes` are a bit odd to me - I think I missed some that you've added. Should just need `PropTypes.object.isRequired` ?

src/widgets/ReportSideBar/ReportSideBar.js
```diff
@@ -0,0 +1,75 @@
...
+ * Sustainable Solutions (NZ) Ltd. 2018
+ */
+
+import React from 'react';
+import PropTypes from 'prop-types';
+import { View, ListItem, FlatList, Text, StyleSheet } from 'react-native';
+import { APP_FONT_FAMILY, GREY } from '../../globalStyles';
+import { pageInfoStrings } from '../../localization/pageInfoStrings';
+
+export const ReportSideBar = ({ data, onPressItem, selectedItemIndex, dimensions }) => {
```
joshxg: Looks like this was merged into another branch, that's why it's coming up here https://github.com/openmsupply/mobile/pull/1763 if you were unsure!

src/widgets/ReportSideBar/ReportSideBar.js
diegoevangelisti: The thing, both `PropTypes.array.isRequired` and `PropTypes.object.isRequired` are being rejected by es-lint. I found those solutions [here](https://github.com/yannickcr/eslint-plugin-react/issues/2079)
But I can change them and disable es-lint for those lines!

src/widgets/ReportSideBar/ReportSideBar.js
diegoevangelisti: Yes, that happened because I created this branch from the other one and brang those files that then merged on feature/dashboard and never update this one. Thanks for noticing!

src/widgets/DashboardPage.js
diegoevangelisti: Now that you mentioned that, I see it! Done here: c0f78a3

src/widgets/DashboardPage.js
diegoevangelisti: Nice, thanks! I added `useEffect(() => { ... }, [reports])`. That way, I _think_ that `useEffect` will only re-run if `reports` changes. Is that correct?

src/widgets/DashboardPage.js
diegoevangelisti: Same as above 😄. I changed is to `PropTypes.object.isRequired` and disabled es-lint for that here 2fd3321

src/widgets/DashboardPage.js
joshxg: why did you change this to `then`? I don't really understand this? I meant to have this line as just `database.objects("Report")` - what is the `then` doing?

src/widgets/DashboardPage.js
```diff
@@ -0,0 +1,83 @@
...
+        .then((report, index) => ({
+          id: report.reportID,
+          index,
+          title: report.title,
+          type: report.type,
+          date: report.date,
+          data: JSON.parse(report.data),
+        }))
+    );
+  }, [reports]);
```
joshxg: in the `useEffect`, we `setReports`, changing the `reports` variable. Then, we run the `useEffect` when `reports` changes. See a problem with this? :)

joshxg:
Have just added a couple of minor adjustments here to get it runnable and not throwing errors when loading:

image


🎉 🎉

src/pages/DashboardPage.js
```diff
@@ -0,0 +1,59 @@
...
+
+import { ReportSideBar } from '../widgets/ReportSideBar';
+import { ReportChart } from '../widgets/ReportChart';
+
+import { UIDatabase } from '../database/index';
+
+import globalStyles from '../globalStyles';
+
+export const DashboardPage = () => {
+  const [reports] = useState(UIDatabase.objects('Report'));
```
diegoevangelisti: This is a great solution, much easier and understandable (rather than my `UseEffect()` nightmare 😄)

Issue #1782, opened 2019-12-24
Prescriber/Patient columns incorrect


joshxg:
## Is your feature request related to a problem? Please describe.

Currently the patient and prescriber tables in the `DispensaryPage` are potentially not correct.

## Describe the solution you'd like

For them to be correct..

## Implementation

Change the columns to be...

Patient: Code, firstName, LastName, DOB, Dispense
Prescriber: Code, firstName, lastName 

Will also need to change the `Dispense` column icon..

## Describe alternatives you've considered

N/A

## Additional context

N/A

joshxg:
Closed by #1791

PR #1794, opened 2019-12-25
#1792 Edit prescription line quantities

### LINKED ISSUE #1792: Edit quantities of prescription lines

Need to be able to edit quantities of trans lines for a prescription

Hook the callback up for the stepper component

## Implementation

N/A

N/A

## Additional context

N/A

joshxg:
### BRANCHED FROM #1782 

Fixes #1792 

## Change summary
- Adds a prescription action for updating quantities
- Adds pass through of a dispatcher for the added action to the stepper of each row
- Adds some state to the stepper as having this update through the reducer and all the components didn't have a good feel, it was slightly too slow to have a responsive UI for updates every 50ms. Possibly this is why you don't see this kind of feature too often! Anyway, state has been added to update itself, while still passing the value back, just not causing a state update at the parent to propagate down. Causes more confusing code, but worth it for this exception?

editQuantities

## Testing

N/A

N/A

### DIFF
--- a/src/widgets/StepperInput.js
+++ b/src/widgets/StepperInput.js
@@ -15,6 +15,9 @@ import { AddIcon, MinusIcon } from './icons';
  * The buttons can also be held, increase by one every 50ms. The increment amount
  * then grows exponentially every 500ms, reseting to one after releasing.
  *
+ * Component uses state to update itself, and a ref for storing the current value. This
+ * is a bit of a hack, but is much more responsive to the rapid updates of the input.
+ *
  * @prop {String|Number} value          The current numerical value.
  * @prop {Func}          onChangeText   Callback for updating the value.
  * @prop {Number}        lowerLimit     The lower limit of the numerical range.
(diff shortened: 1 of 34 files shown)


src/widgets/StepperInput.js
```diff
@@ -23,26 +26,33 @@ import { AddIcon, MinusIcon } from './icons';
  *
  */
 export const StepperInput = ({ value, onChangeText, lowerLimit, upperLimit, textInputStyle }) => {
+  const [, setCurrentValueState] = React.useState(Number(value));
   const currentValue = React.useRef(Number(value));
   const currentAdjustmentAmount = React.useRef(1);
   const valueAdjustmentInterval = React.useRef();
   const adjustmentIncreaseInterval = React.useRef();
 
+  const parseNumber = number => Math.max(Math.min(number, upperLimit), lowerLimit);
```
mark-prins: very nifty!

src/widgets/StepperInput.js
```diff
@@ -23,26 +26,33 @@ import { AddIcon, MinusIcon } from './icons';
...
   const onUpdate = newValue => {
-    currentValue.current = newValue;
-    onChangeText(newValue);
+    const updateValue = parseNumber(newValue);
+    currentValue.current = updateValue;
+    setCurrentValueState(updateValue);
+    onChangeText(updateValue);
   };
 
   const decrementValue = () => {
```
mark-prins: or.. 

```
const decrementValue = () => onUpdate(currentValue.current - currentAdjustmentAmount.current);

const incrementValue = () => onUpdate(currentValue.current + currentAdjustmentAmount.current);
```

src/widgets/StepperInput.js
```diff
@@ -15,6 +15,9 @@ import { AddIcon, MinusIcon } from './icons';
  * The buttons can also be held, increase by one every 50ms. The increment amount
  * then grows exponentially every 500ms, reseting to one after releasing.
  *
+ * Component uses state to update itself, and a ref for storing the current value. This
+ * is a bit of a hack, but is much more responsive to the rapid updates of the input.
```
mark-prins: maybe just me, but I don't see it as that hacky. For a component like this it is sensible to do the rendering based on local state. And my understanding of the `useRef` is that it is perfect for this sort of thing.

src/widgets/StepperInput.js
joshxg: nice thanks #1821

src/widgets/StepperInput.js
joshxg: I guess more of a creative use of tools? 🤣

Issue #1783, opened 2019-12-24
Patient and prescriber labels in Prescription wizard


joshxg:
## Is your feature request related to a problem? Please describe.

In the Prescription wizard, there should be a patient and prescriber labels. There aren't any!

## Describe the solution you'd like

Labels for:
- Patient, with icons handling touches which open either the patient edit modal or the patient history modal.
- Prescriber: with an icon for opening the prescriber edit modal.

## Implementation

N/A

## Describe alternatives you've considered

N/A

## Additional context

![image](https://user-images.githubusercontent.com/35858975/71390194-38d09400-2664-11ea-9f5e-7c23a1cd15dc.png)

joshxg:
Fixed by #1797

PR #1797, opened 2019-12-27
#1793 Prescriber SearchBar

### LINKED ISSUE #1793: Prescriber step needs a search bar

The first step in a prescription to select a prescriber should have a search bar

A search bar which searches on first or last name of a prescriber

## Implementation

N/A

N/A

## Additional context

N/A

joshxg:
#### BRANCHED FROM #1783 

Fixes #1793 

## Change summary

- Extracts `PrescriberSelect` to it's own file
- Adds `Searchbar ` `searchTerm` logic
- Adds `SearchBar` component
- Adds a 'create prescriber' button

## Testing

N/A

N/A

### DIFF
--- a/src/widgets/FlexView.js
+++ b/src/widgets/FlexView.js
@@ -9,8 +9,8 @@ import { View } from 'react-native';
 import PropTypes from 'prop-types';
 
 /**
- * Simple layout component. For use when needing a simple View with flex
- * direction of row.
+ * Simple layout component. For use when needing a simple View with flex.
+ *
  * @param {React.node} children       Children of this component
  * @param {Number}     flex           The flex amount i.e. 1
  * @param {Object}     style          An additional styles object.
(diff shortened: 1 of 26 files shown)


src/widgets/FlexView.js
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Simple layout component. For use when needing a simple View with flex.
+ *
+ * @param {React.node} children       Children of this component
+ * @param {Number}     flex           The flex amount i.e. 1
+ * @param {Object}     style          An additional styles object.
+ */
+export const FlexView = ({ children, flex, style }) => {
+  const internalStyle = React.useMemo(() => ({ flex, ...style }));
```
mark-prins: I can't think of a case where the style would change dynamically - if it could, then this would be needed?

```suggestion
  const internalStyle = React.useMemo(() => ({ flex, ...style }), [style]);
```

joshxg: oo nice - but actually it's worse and won't do anything.. `useMemo` iterates over the deps list, and checks each elements diff. But since there is no list, there is always a diff. whoops #1821

PR #1806, opened 2019-12-29
#1800 only next with items

### LINKED ISSUE #1800: Disabled the Next button until items have been chosen in a prescription
## Describe the bug

Currently can proceed through the 'checkout' with no items. 

### To reproduce

1. Create a new prescription
2. don't add any items and click next

### Expected behaviour

`NEXT` should be disabled until at least one item has been chosen.

### Proposed Solution

N/A

### Version and device info

Dispensing development

### Additional context

N/A

joshxg:
#### BRANCHED FROM #1800

Fixes #1800 

## Change summary
- Adds a selector which checks if the current prescription has items and quantity applied.
- Adds a `createRecord` parameter to set an initial value of the `TransactionItem`.. this has been set to 1 for a prescription, and also adjusted the lower limit of the stepper.

## Testing

N/A

N/A

### DIFF
--- a/src/widgets/StepperRow.js
+++ b/src/widgets/StepperRow.js
@@ -60,7 +60,7 @@ const localStylez = {
 };
 
 StepperRow.defaultProps = {
-  lowerLimit: 0,
+  lowerLimit: 1,
   upperLimit: 99999,
   labelSize: 'large',
 };
--- a/src/actions/FinaliseActions.js
+++ b/src/actions/FinaliseActions.js
@@ -1,0 +1,14 @@
+/**
+ * mSupply Mobile
+ * Sustainable Solutions (NZ) Ltd. 2019
+ */
+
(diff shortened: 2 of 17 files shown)


src/widgets/StepperRow.js
```diff
@@ -60,7 +60,7 @@ const localStylez = {
 };
```
mark-prins: wait, `localStylez`?? how did that slip past?

joshxg: whoops will get that fixed up 🤣

joshxg:
thoughts were:
- If it's in the 'cart' it would have a quantity.. most/all 'cart' type interfaces I have tried out recently have a default of 1 when you add it.
- Could potentially be customizable for an item..?
- could save a press in a lot of cases?

Issue #1815, opened 2019-12-29
PrescriptionReducer tidy up


joshxg:
## Is your feature request related to a problem? Please describe.

`PrescriptionReducer` was thrown together and is untidy in that the actions, action creators and reducer are in the same file (not bad, just not consistent), as well as inconsistent use of actions

## Describe the solution you'd like

- Tidy up `PrescriptionReducer`
- Investigate connecting other `Prescription` prefixed components to redux to take advantage of this store

## Implementation

N/A

## Describe alternatives you've considered

N/A

## Additional context

N/A

joshxg:
https://github.com/openmsupply/mobile/pull/1796#discussion_r361877438

joshxg:
Closed by #1816

PR #1816, opened 2019-12-29
#1813 Simple directions implementation

### LINKED ISSUE #1813: Add simple directions to a script item

Cannot currently add any directions to items in a script.

Ability to add direections to items in a script!

## Implementation

N/A

N/A

## Additional context

N/A

joshxg:
Fixes #1813 

## Change summary

- Adds a simple directions implementation with a drop down for an items directions, and a textinput to type.

image

- Have created #1815 to tidy up `PrescriptionReducer.js`
- Sneaked in a fix in `incomingSyncUtils` to cast correctly.
- Added localization to issue
- Adds the correct field to display in the summary
- Adds a TextInput rather than a label for the displayed direction. Mainly this is for the simple implementation, as I dont see how this could work well when abbreviations are implemented. I think a modal would be needed which would have an additional drop down to show some directions to add, or something similar.
- The ScrollView wrap is to handle shifting the `TextInput` out of the way/above the keyboard

## Testing

N/A

- Potentially better to just have `PrescriptionX` components all connected to redux. Actually pretty confident it would be! Not a huge shift

### DIFF
--- a/src/widgets/DropdownRow.js
+++ b/src/widgets/DropdownRow.js
@@ -28,6 +28,7 @@ import { SUSSOL_ORANGE, APP_FONT_FAMILY, GREY } from '../globalStyles/index';
  * @prop {String} dropdownTitle     The title of the drop down menu when opened.
  * @prop {Bool}   useSecondaryMenu  Indicator whether the secondary menu should show.
  * @prop {Number} iconSize          The size of the drop down icon.
+ * @prop {String} placeholder       Placeholder string value, when no value has been chosen/entered.
  */
 export const DropdownRow = ({
   currentOptionText,
@@ -37,6 +38,7 @@ export const DropdownRow = ({
   useSecondaryMenu,
   secondaryCallback,
   iconSize,
+  placeholder,
 }) => {
   const DropDownMenuIcon = React.useCallback(() => {
     const iconColor = options.length ? SUSSOL_ORANGE : GREY;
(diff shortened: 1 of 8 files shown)


src/widgets/DropdownRow.js
```diff
@@ -66,7 +63,13 @@ export const DropdownRow = ({
         />
       </View>
       <View style={localStyles.flexNine}>
-        <SimpleLabel size={labelSize} text={currentOptionText} />
+        <TextInput
+          onChangeText={onSelection}
+          value={currentOptionText}
+          underlineColorAndroid={SUSSOL_ORANGE}
+          style={{ flex: 1, fontFamily: APP_FONT_FAMILY }}
+          placeholder="Usage direction"
```
mark-prins: this is very specific within a general component. A temporary thing?
I was expecting props to enable the text input and supply the placeholder

src/widgets/PopoverDropDown.js
```diff
@@ -65,7 +67,7 @@ const DropDown = ({ anchorRef, isVisible, options, onSelection, onClose, title }
       onRequestClose={onClose}
       arrowStyle={arrowStyle}
       backgroundStyle={popoverBackgroundStyle}
-      placement="bottom"
+      placement="auto"
```
mark-prins: perhaps the component should change its name from `PopoverDropDown`?? 😆

src/widgets/DropdownRow.js
joshxg: whoops! thanks : 1ac1107690ce2800fccf792bf8eb9e23906ea09c

src/widgets/PopoverDropDown.js
joshxg: 🤣  hmm.. `PopoverDropAuto` just doesn't seem to have the same ring to it

joshxg:
Also removed some styles 
a74ed3f

Issue #1828, opened 2019-12-31
Add indicator incoming sync


wlthomson:
## Is your feature request related to a problem? Please describe.

Add incoming sync logic for `ProgramIndicator`, `IndicatorAttribute`, `IndicatorValue` records.

## Describe the solution you'd like

See above.

## Implementation

Update `incomingSyncUtils` to handle indicator tables.

## Describe alternatives you've considered

N/A. 

## Additional context

See epic #1685.

wlthomson:
Fixed by #1836.

PR #1837, opened 2020-01-03
#1823 add indicator database objects

### LINKED ISSUE #1823: Add indicators database objects

Add datatypes for 4.08 indicators. 

A direct mapping between desktop and mobile:

- `indicatorValue`
- `indicatorAttribute`
- `programData`

### Indicator

```
ProgramIndicator.schema = {
  name: 'ProgramIndicator',
  primaryKey: 'id',
  properties: {
    id: 'string',
    code: { type: 'string', default: 'placeholderCode' },
    program: 'MasterList',
    isActive: { type: 'bool', default: false }
  }
}
```

### IndicatorAttribute

```
IndicatorAttribute.schema = {
  name: 'IndicatorAttribute',
  primaryKey: 'id',
  properties: {
    id: 'string',
    indicator: 'Indicator',
    description: { type: 'string', default: 'placeholderDescription' },
    code: { type: 'string', default: 'placeholderCode' },
    index: { type: 'int', default: 0 },
    isRequired: { type: 'bool', default: true },
    valueType: { type: 'string', default: 'string' },
    valueDefault: { type: 'string', default: 'string' },
    axis: { type: 'string': default: 'row' },
    is_active: { type: 'bool', default: false }
  }
}
```

### IndicatorValue 

```
IndicatorValue.schema = {
  name: 'IndicatorValue',
  primaryKey: 'id',
  properties: {
    id: 'string',
    facilityId: 'string',
    period: 'Period',
    column: 'IndicatorAttribute'
    row: 'IndicatorAttribute'
    value: type: 'string'
  },
};
```

N/A.

## Additional context

See epic #1685.

## Discussion:

- Initial brain dump of proposed solution. Feedback/thoughts would be great.
- Much of the info being synced is not editable. Is there an elegant way of making realm models/fields immutable or editable only via a setter? Could be good for some models/fields which should be mobile read-only, e.g. `IndicatorValue.facilityId` (always the mobile store), or `IndicatorAttribute` and `programIndicator` records?

wlthomson:
Fixes #1823.

Branched from #1828.

## Change summary

Add indicator tables and incoming sync mappings.

## Testing

- [ ] `ProgramIndicator` records sync to mobile.
- [ ] `IndicatorAttribute` records sync to mobile.
- [ ] `IndicatorValue` records sync to mobile.

Bare bones sync functionality to get the ball rolling, will probably need some fixing up.


src/database/DataTypes/IndicatorAttribute.js
```diff
@@ -0,0 +1,41 @@
...
+export default IndicatorAttribute;
+
+IndicatorAttribute.schema = {
+  name: 'IndicatorAttribute',
+  primaryKey: 'id',
+  properties: {
+    id: 'string',
+    indicator: 'ProgramIndicator',
+    description: { type: 'string', default: 'placeholderDescription' },
+    code: { type: 'string', default: 'placeholderCode' },
```
joshxg: wondering why defaults for these?

src/database/DataTypes/IndicatorValue.js
```diff
@@ -0,0 +1,33 @@
...
+export class IndicatorValue extends Realm.Object {}
+
+export default IndicatorValue;
+
+IndicatorValue.schema = {
+  name: 'IndicatorValue',
+  primaryKey: 'id',
+  properties: {
+    id: 'string',
+    facilityId: 'string',
```
joshxg: this isn't really needed since there will only ever be this stores values in the database. Also introduces terminology (Facility) which isn't used in the mobile code base which I for some reason really don't like 🤣 If we do keep it can we please just call it `storeId`?

src/sync/incomingSyncUtils.js
```diff
@@ -306,9 +306,32 @@ export const createOrUpdateRecord = (database, settings, recordType, record) =>
...
   switch (recordType) {
     case 'IndicatorAttribute': {
+      internalRecord = {
+        id: record.ID,
+        indicator: database.getOrCreate('ProgramIndicator', record.indicator_ID),
+        description: record.description,
+        code: record.code,
+        index: parseNumber(record.index),
+        isRequired: parseBoolean(record.is_required),
+        valueType: record.data_type && record.data_type.value,
```
joshxg: think you can `record?.data_type?.value`

src/sync/incomingSyncUtils.js
```diff
@@ -306,9 +306,32 @@ export const createOrUpdateRecord = (database, settings, recordType, record) =>
...
+      };
+      database.update(recordType, internalRecord);
       break;
     }
     case 'IndicatorValue': {
+      const recordValue = record.value && JSON.parse(record.value);
+      internalRecord = {
+        id: record.ID,
+        facilityId: record.facility_ID,
+        period: database.getOrCreate('Period', record.period_ID),
```
joshxg: I think this will fail if you try to create a `Period` as it needs a `PeriodSchedule` - not sure the best move here.. seems like you need to create some placeholder `Period`, otherwise you won't be able to link up later.. but you can't create the `PeriodSchedule`. Hmm, no suggestion just yet!

src/sync/incomingSyncUtils.js
```diff
@@ -306,9 +306,32 @@ export const createOrUpdateRecord = (database, settings, recordType, record) =>
...
+        valueType: record.data_type && record.data_type.value,
+        valueDefault: record.data_type && record.data_type.default,
+        axis: record.axis,
+        isActive: parseBoolean(record.is_active),
+      };
+      database.update(recordType, internalRecord);
       break;
     }
     case 'IndicatorValue': {
+      const recordValue = record.value && JSON.parse(record.value);
```
joshxg: `value` is required and this has the possibility of assigning a null/undefined. I think `(record.value && JSON.parse(record.value)) || "" ` might be a bit safer

src/sync/incomingSyncUtils.js
```diff
@@ -306,9 +306,32 @@ export const createOrUpdateRecord = (database, settings, recordType, record) =>
...
+      internalRecord = {
+        id: record.ID,
+        facilityId: record.facility_ID,
+        period: database.getOrCreate('Period', record.period_ID),
+        column: database.getOrCreate('IndicatorAttribute', record.column_ID),
+        row: database.getOrCreate('IndicatorAttribute', record.row_ID),
```
joshxg: hm, same as above these `IndicatorAttribute` records require a `ProgramIndicator` which isn't going to be created if they can't be getted, so I believe it'll throw. We need to propogate the creation of the `ProgramIndicator` on top of this. There's no helpers or mechanisms for this at the moment for this

src/database/DataTypes/IndicatorValue.js
wlthomson: I agree, I'm not sure what the reasoning was for facility, but I'd like to change it on desktop, too. I'll change it here for now and piggy back the change to desktop when I next get a chance to update the structure.

src/database/DataTypes/IndicatorAttribute.js
wlthomson: `IndicatorAttribute.description`, `IndicatorAttribute.code` are both user displayed fields on tables, so my thought is that having the field blank would potentially be a bit confusing? See for reference, e.g. `Item.code` and `Item.name`.

src/sync/incomingSyncUtils.js
wlthomson: As discussed, simplest solution is to make `Period.periodSchedule` optional. Will just hotfix for now, but feature branch can be patched with any period logic needed to guard against regressions.

wlthomson: Fixed by 0e7e1e2b.

src/sync/incomingSyncUtils.js
wlthomson: Fixed by d926c4c6.

src/sync/incomingSyncUtils.js
wlthomson: Good catch, e7f9bbccdc2a2ac8fc471a33acaa4f58565c7a34.

src/sync/incomingSyncUtils.js
wlthomson: Nice. f0e6d5088ba704dc6ec77e6c22beded334b6890e.

src/database/DataTypes/IndicatorValue.js
wlthomson: b7b32832.

PR #1839, opened 2020-01-04
#1825 Patient DOB Picker

### LINKED ISSUE #1825: Dispensing Patient D.O.B should use a date picker
## Describe the bug

At the moment, date of birth displayed in the patient modal is a `toString()` of the `dob` in the db. So it displays as "wed 9th .." etc. But, we need for validation a dob in the form `xx/xx/xxxx`. And, imo, a date picker is far better than typing!

### To reproduce

Development bug as described above

### Expected behaviour

N/A

### Proposed Solution

N/A

### Version and device info

N/A

### Additional context

N/A

joshxg:
Fixes #1825 

## Change summary

- Adds a date picker for entering date of births and corrects the validation

## Testing

N/A

N/A

### DIFF
--- a/src/widgets/FormDateInput.js
+++ b/src/widgets/FormDateInput.js
@@ -53,7 +53,7 @@ export const FormDateInput = ({
     const newState = {
       isValid: validity,
       inputValue: newValue,
-      pickerSeedValue: validity ? moment(newValue, 'DD/MM/YY').toDate() : new Date(),
+      pickerSeedValue: validity ? moment(newValue, 'DD/MM/YYYY').toDate() : new Date(),
       datePickerOpen: pickerVisibility,
     };
     setInputState(newState);
@@ -73,7 +73,7 @@ export const FormDateInput = ({
   const onChangeDatez = ({ nativeEvent }) => {
     const { timestamp } = nativeEvent;
     if (!timestamp) return;
-    const newDate = moment(new Date(timestamp)).format('D/M/Y');
+    const newDate = moment(new Date(timestamp)).format('DD/MM/YYYY');
     onUpdate(newDate, true);
   };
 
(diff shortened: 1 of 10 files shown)


src/widgets/FormDateInput.js
```diff
@@ -0,0 +1,152 @@
...
+  // Do not restrict input, but provide feedback to the user.
+  const onChangeTextCallback = React.useCallback(
+    newValue => {
+      const newValueIsValid = onValidate ? onValidate(newValue) : true;
+      onUpdate(newValue, newValueIsValid);
+    },
+    [onValidate]
+  );
+
+  const onChangeDatez = ({ nativeEvent }) => {
```
mark-prins: wait, who wrote this? *Datez*?

src/widgets/FormDateInput.js
```diff
@@ -0,0 +1,152 @@
...
+    const { timestamp } = nativeEvent;
+    if (!timestamp) return;
+    const newDate = moment(new Date(timestamp)).format('DD/MM/YYYY');
+    onUpdate(newDate, true);
+  };
+
+  const openDatePicker = () => setInputState(state => ({ ...state, datePickerOpen: true }));
+
+  return (
+    <FlexRow flex={1}>
```
mark-prins: Is there a reason for `{1}` and not `"1"`? isn't that ever so slightly more expensive to render than a plain string?

joshxg: I'm not sure what you mean - I'm guessing you're thinking `{1}` is a JS object? It's just JSX syntax for the parser. If we want a litteral object we need to `{{}}`.. Unless there's something else? Also `flex` needs a number, not a string

src/widgets/FormDateInput.js
joshxg: I think it was chris !

4d34304

PR #1871, opened 2020-01-08
#1866 Modify setQueryText argument

### LINKED ISSUE #1866: AutoCompleteSelector is not filtering with new SearchBar
## Describe the bug

The AutoCompleteSelector using the new SearchBar it is not working properly while being tested in a Tablet (see details below). 

### To reproduce

- [ ] Download the APK and install it the tablet (or use the AVD)
- [ ] Go to customer invoice
- [ ] Click on "new invoice"
- [ ] Search for customers
- [ ] See error.

### Expected behaviour

The filter should be working properly (e.g. After typing `Baltimor` the rest of the options should hide).

### Proposed Solution

Mentioned by @joshxg 
https://github.com/openmsupply/mobile/blob/1223ef686342e963aadcc51fcb87d8ce3a593c9c/src/widgets/modalChildren/AutocompleteSelector.js#L34 this line is a problem. Should be:

`const inputValue => setQueryText(inputValue)`

### Version and device info

Tablet model Lenovo XB-X304L
Android version 7.1.1
APK: mSupplyMobile-3_2_0-rc1-universal-release.apk

diegoevangelisti:
Fixes #1866 

## Change summary

Now `onChangeText` is directly calling `setQueryText` and setting up the new value of `queryText`. 

## Testing

- [ ] Download the APK and install it the tablet (or use an AVD)
- [ ] Go to customer invoice
- [ ] Click on "new invoice"
- [ ] Search for a customer.
- [ ] See the filter is working ok.

### DIFF
--- a/src/widgets/modalChildren/AutocompleteSelector.js
+++ b/src/widgets/modalChildren/AutocompleteSelector.js
@@ -31,6 +31,8 @@ export const AutocompleteSelector = ({
 }) => {
   const [queryText, setQueryText] = useState('');
 
+  const onChangeText = React.useCallback(inputValue => setQueryText(inputValue), []);
+
   /**
    * Filters a realm results object. Creates two realm results A, B
    * by two query strings. And concat A to B - A.
@@ -102,7 +104,7 @@ export const AutocompleteSelector = ({
         autoCorrect={false}
         autoFocus
         color={WHITE}
-        onChangeText={inputValue => setQueryText(inputValue)}
+        onChangeText={onChangeText}
         placeholder={placeholderText}
         placeholderTextColor="white"
         style={[localStyles.text, localStyles.searchBar]}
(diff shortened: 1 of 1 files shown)


src/widgets/modalChildren/AutocompleteSelector.js
```diff
@@ -104,7 +102,7 @@ export const AutocompleteSelector = ({
         autoCorrect={false}
         autoFocus
         color={WHITE}
-        onChangeText={onChangeText}
+        onChangeText={inputValue => setQueryText(inputValue)}
```
joshxg: I prefer this to be defined as it was before. This is because whenever a re-render occurs, anonymous functions like this are re-created. It's a tiny overhead, but it's good to ensure we are always defining functions explicitly.

So as above,

```
const onChangeText = inputValue => setQueryText(inputValue)
...
onChangeText={onChangeText}
```

To be even better, we can use a react hook, and memoize the function, to cause even less overhead!

```
const onChangeText = React.useCallback(inputValue => setQueryText(inputValue), [])
...
onChangeText={onChangeText}
```

What this does is - React will store an object that will hold this function. Then, whenever re-renders occur, react will just give us back the same function, rather than creating a new one.

diegoevangelisti: Nice suggestion and great to remember the use of memoization. 
Done here 566f930

Issue #1878, opened 2020-01-09
Dashboard fields are incorrect


joshxg:
## Describe the bug

Currently, we have that the "data" object on a `Report` object from mSupply is `data` where in fact it is `json`.

### To reproduce

Development bug

### Expected behaviour

Rename occurences of `"data"` for `Report` to `json`

### Proposed Solution

N/A

### Version and device info

Dashboard development

### Additional context

N/A

diegoevangelisti:
Closed by #1881

PR #1881, opened 2020-01-09
#1880 Get values from data object

### LINKED ISSUE #1880: Reports' data structure is not similar than desktop one
## Describe the bug

Before syncing actual reports, some demo data was created predicting data structures that don't match with the real one expected to be arriving. Different nesting mainly.

### To reproduce

Check the difference between `ReportDemoData.js` and the JSON structure available in desktop reports, which is the following:

- BarChart

```
  {
    name: 'TopFacilities',
    data: [
      {
        label: 'Top Facilities Supplied last 90 days',
        values: [
          {
            x: 'Hufflepuff Health Centre',
            y: 41,
          },
          {
            x: '',
            y: 0,
          },
        ],
      },
    ],
  }
```
- LineChart

```
{
    name: 'ExpiringStock',
    data: [
      {
        label: 'Cumulative',
        values: [
          {
            x: '1 month',
            y: 0,
          },
          {
            x: '2 months',
            y: 6205,
          },
          {
            x: '3 months',
            y: 9730,
          },
          {
            x: '4 months',
            y: 9747.644,
          },
          {
            x: '5 months',
            y: 9747.644,
          },
          {
            x: '6 months',
            y: 9993.794,
          },
          {
            x: '7 months',
            y: 11354.794,
          },
          {
            x: '8 months',
            y: 12752.294,
          },
          {
            x: '9 months',
            y: 12772.604,
          },
          {
            x: '10 months',
            y: 12772.604,
          },
          {
            x: '11 months',
            y: 12772.604,
          },
          {
            x: '12 months',
            y: 61500.188,
          },
        ],
      },
      {
        label: 'Monthly',
        values: [
          {
            x: '1 month',
            y: 0,
          },
          {
            x: '2 months',
            y: 6205,
          },
          {
            x: '3 months',
            y: 3525,
          },
          {
            x: '4 months',
            y: 17.644,
          },
          {
            x: '5 months',
            y: 0,
          },
          {
            x: '6 months',
            y: 246.15,
          },
          {
            x: '7 months',
            y: 1361,
          },
          {
            x: '8 months',
            y: 1397.5,
          },
          {
            x: '9 months',
            y: 20.31,
          },
          {
            x: '10 months',
            y: 0,
          },
          {
            x: '11 months',
            y: 0,
          },
          {
            x: '12 months',
            y: 48727.584,
          },
        ],
      },
    ],
  }
```
- TableReport

```
{
    name: 'TransactionPerStore',
    data: {
      rows: [
        ['General Warehouse', 0, 0],
        ['Hogwarts Dispensary', 0, 0],
        ['Hufflepuff Health Centre', 0, 8],
        ['Gryffindor District Store', 0, 0],
        ['Slytherin Health Centre', 0, 0],
        ['Ravenclaw District Store', 0, 0],
        ['Gryffindor Dispensary', 0, 0],
        ['This Store Is To Add A New One And Long One ', 0, 0],
        ['Testing Programs Store', 0, 0],
      ],
      header: ['Store', 'Week', 'Month'],
    },
}
```
- PieChart

```
{
    name: 'TopFacilities',
    data: [
      {
        label: 'Top Facilities Supplied last 90 days',
        values: [
          {
            x: 'Guido Valadares National Hospital',
            y: 135886,
          },
          {
            x: 'District Services Ermera',
            y: 28444,
          },
        ],
      },
    ],
  }
```

### Expected behaviour

Data structure for all reports should respect the one used in the desktop version.

### Proposed Solution

Update all reports to use the corresponding structure.

diegoevangelisti:
Fixes #1880 and #1878 

## Change summary

According to the data structure obtained from the desktop version, the actual data needed is more nested than we thought. This changes are intending to solve that wrong assumption.
So here we destructure `data` to get the actual values.
Aclaration: No need to do it in `ReportTable`

## Testing

Further testing will be taking place all together for the feature/dashboard branch.

`ReportDemoData` should be updated to can test with the new structure.

### DIFF
--- a/src/sync/incomingSyncUtils.js
+++ b/src/sync/incomingSyncUtils.js
@@ -504,7 +504,7 @@ export const createOrUpdateRecord = (database, settings, recordType, record) => 
       try {
         const parsedData = JSON.parse(data);
         const shouldSetData = checkIsObject(parsedData);
-        internalRecord = { id, title, type, _data: shouldSetData ? data : null };
+        internalRecord = { id, title, type, _data: shouldSetData ? data.data : null };
         database.update(recordType, internalRecord);
       } catch (error) {
         // Throw to parent, for now
--- a/src/sync/incomingSyncUtils.js
+++ b/src/sync/incomingSyncUtils.js
@@ -252,7 +252,7 @@ export const sanityCheckIncomingRecord = (recordType, record) => {
       canBeBlank: ['units', 'comment', 'order_number'],
     },
     Report: {
-      cannotBeBlank: ['ID', 'title', 'type', 'data'],
+      cannotBeBlank: ['ID', 'title', 'type', 'json'],
       canBeBlank: [],
     },
   };
@@ -500,11 +500,11 @@ export const createOrUpdateRecord = (database, settings, recordType, record) => 
       break;
     }
     case 'Report': {
-      const { ID: id, title, type, data } = record;
+      const { ID: id, title, type, json } = record;
       try {
-        const parsedData = JSON.parse(data);
+        const parsedData = JSON.parse(json);
         const shouldSetData = checkIsObject(parsedData);
-        internalRecord = { id, title, type, _data: shouldSetData ? data.data : null };
+        internalRecord = { id, title, type, _data: shouldSetData ? json.data : null };
         database.update(recordType, internalRecord);
       } catch (error) {
         // Throw to parent, for now
--- a/src/sync/incomingSyncUtils.js
+++ b/src/sync/incomingSyncUtils.js
@@ -504,7 +504,12 @@ export const createOrUpdateRecord = (database, settings, recordType, record) => 
       try {
         const parsedData = JSON.parse(json);
         const shouldSetData = checkIsObject(parsedData);
-        internalRecord = { id, title, type, _data: shouldSetData ? json.data : null };
+        internalRecord = {
+          id,
+          title,
+          type,
+          _data: shouldSetData ? JSON.stringify(parsedData.data) : null,
+        };
         database.update(recordType, internalRecord);
       } catch (error) {
         // Throw to parent, for now
--- a/src/widgets/PieChart.js
+++ b/src/widgets/PieChart.js
@@ -11,6 +11,7 @@ import { VictoryLabel, VictoryPie } from 'victory-native';
 import { APP_FONT_FAMILY, GREY } from '../globalStyles';
 
 export const PieChart = ({ width, height, data }) => {
+  const { values } = data;
   const {
     padVertical,
     padHorizontal,
@@ -39,7 +40,7 @@ export const PieChart = ({ width, height, data }) => {
       labelRadius={widthPadded * labelRadius}
       colorScale={colorScale}
       labelComponent={<VictoryLabel style={style} />}
-      data={data}
+      data={values}
     />
   );
 };
--- a/src/widgets/PieChart.js
+++ b/src/widgets/PieChart.js
@@ -10,8 +10,8 @@ import { VictoryLabel, VictoryPie } from 'victory-native';
 
 import { APP_FONT_FAMILY, GREY } from '../globalStyles';
 
-export const PieChart = ({ width, height, json }) => {
-  const { values } = json;
+export const PieChart = ({ width, height, data }) => {
+  const { values } = data;
   const {
     padVertical,
     padHorizontal,
@@ -48,7 +48,7 @@ export const PieChart = ({ width, height, json }) => {
 PieChart.propTypes = {
   width: PropTypes.number.isRequired,
   height: PropTypes.number.isRequired,
-  json: PropTypes.array.isRequired,
+  data: PropTypes.array.isRequired,
 };
 
 const victoryStyles = {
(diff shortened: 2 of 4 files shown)


src/sync/incomingSyncUtils.js
```diff
@@ -504,7 +504,7 @@ export const createOrUpdateRecord = (database, settings, recordType, record) =>
       try {
         const parsedData = JSON.parse(data);
         const shouldSetData = checkIsObject(parsedData);
-        internalRecord = { id, title, type, _data: shouldSetData ? data : null };
+        internalRecord = { id, title, type, _data: shouldSetData ? data.data : null };
```
joshxg: Lets just do this one here too if you don't mind: https://github.com/openmsupply/mobile/issues/1878

If you see here: https://github.com/sussol/msupply/pull/4985 - actually it is not called `data`! it is `json` .. so above we need to change `data` to `json` and also it should be `json.data`.

also here: https://github.com/openmsupply/mobile/blob/8490cbec79a165ea9e0d362c69b0b3a8ca448d67/src/sync/incomingSyncUtils.js#L255

diegoevangelisti: Great that you noticed that! 
Sure, I'' work on that now. I didn't see #1878 because it wasn't in the Project Dashboard (Found you're a human been after all haha)

joshxg: 🤖

andreievg: > Found you're a human been after all haha
no way 🤯

diegoevangelisti: Done here 6ac1fc6

diegoevangelisti: @joshxg 
I guess this one `_data` should be `_json` as well, don't you think? I missed that

src/sync/incomingSyncUtils.js
```diff
@@ -500,11 +500,11 @@ export const createOrUpdateRecord = (database, settings, recordType, record) =>
...
     }
     case 'Report': {
-      const { ID: id, title, type, data } = record;
+      const { ID: id, title, type, json } = record;
       try {
-        const parsedData = JSON.parse(data);
+        const parsedData = JSON.parse(json);
         const shouldSetData = checkIsObject(parsedData);
-        internalRecord = { id, title, type, _data: shouldSetData ? data : null };
+        internalRecord = { id, title, type, _data: shouldSetData ? json.data : null };
```
joshxg: sorry my mistake - `json` will actually be a string here. So, we will need to parse the string - for example `parsedData.data`. BUT we also need to make sure we are setting a STRING as `_data`, NOT an object!

src/sync/incomingSyncUtils.js
joshxg: I personally hate the name `json` for a field 🤣  So I think `data` is fine. What do you think?

src/sync/incomingSyncUtils.js
diegoevangelisti: Ok, so apart from checking `parsedData.data` we can add something like `checkIsString` to check`json` as well, and if it's all good we assigned that to `_data` (or `_json`)
Am I right?

joshxg: Hmm it does get a bit confusing. Consider this simplified case, where I think `testData` is a simple version of the `json` field and `testOne` is `json.data`:

```
const testData = { testOne: [{ testTwo: 2 }] };
const testDataAsString = JSON.stringify(testData);
const parsedTestData = JSON.parse(testDataAsString);
console.log(typeof parsedTestData.testOne);
```

then what gets logged?

src/sync/incomingSyncUtils.js
diegoevangelisti: Yeah haha let's keep that `_data`, at least as much as we can

src/sync/incomingSyncUtils.js
diegoevangelisti: haha I guess we will see the original `testData.testOne`, but with a couple of shakings, right?

diegoevangelisti: Great, thanks! I reckon it's solved here: 01ad0a9

src/widgets/PieChart.js
```diff
@@ -10,7 +10,8 @@ import { VictoryLabel, VictoryPie } from 'victory-native';
 
 import { APP_FONT_FAMILY, GREY } from '../globalStyles';
 
-export const PieChart = ({ width, height, data }) => {
+export const PieChart = ({ width, height, json }) => {
```
joshxg: hmm.. you seem to have renamed a bunch of these props to `json` (I'm not sure why), but you are still passing as `data` here: https://github.com/openmsupply/mobile/blob/f8b25f317a6b1d6e3a000c9d52f8e715b01ea817/src/widgets/ReportChart.js#L31-L38

I would just revert back to `data`?

diegoevangelisti: My mistake! Yes, I guess that's the best option, since we decided to keep it like `data`, otherwise I need to rename to `json` all the rest but not the best option.
Thanks btw😅

diegoevangelisti: Hopefully this is well done here bf1b7bd 😄

PR #1905, opened 2020-01-12
#1899 Change the way values are being obtained from data

### LINKED ISSUE #1899: Change values to values[0] in the data prop
## Describe the bug

Data coming from the `json` object includes values, which are not being properly obtained the way it is setted right now. 

### Expected behaviour

Charts should be rendered properly (it is not working right now).

### Proposed Solution

Change data props of `PieChart`, `BarChart`, `LineChart` components from `data = { values }` to `data = { values[0] }`

### Additional context

No need to modify `TableReport`

diegoevangelisti:
Fixes #1899 

## Change summary

Based on the data structure received from dashboard reports, the correct way to read what to graph should be accessing to `values`, which can be one or more array with `x` and `y` values.

## Testing

Further testing will be taking place when the dashboard feature is ready.

### DIFF
--- a/src/widgets/BarChart.js
+++ b/src/widgets/BarChart.js
@@ -12,7 +12,6 @@ import { VictoryChart, VictoryBar, VictoryAxis } from 'victory-native';
 import { APP_FONT_FAMILY, GREY, LIGHT_GREY, DARK_GREY, SUSSOL_ORANGE } from '../globalStyles';
 
 export const BarChart = ({ data, width, height }) => {
-  const { values } = data;
   const renderYAxis = () => <VictoryAxis dependentAxis style={victoryStyles.axisY} />;
   const renderXAxis = () => {
     const tickTruncate = label => (label.length > 11 ? `${label.slice(0, 11)}...` : label);
@@ -33,9 +32,11 @@ export const BarChart = ({ data, width, height }) => {
 
   return (
     <VictoryChart width={width} height={height} padding={padding} domainPadding={domainPadding}>
-      <VictoryBar style={style} data={values[0]} />
-      {renderXAxis()}
+      {data.map(({ values }) => (
+        <VictoryBar style={style} data={values} />
+      ))}
       {renderYAxis()}
+      {renderXAxis()}
     </VictoryChart>
   );
 };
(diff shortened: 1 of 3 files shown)


src/widgets/BarChart.js
```diff
@@ -33,7 +33,7 @@ export const BarChart = ({ data, width, height }) => {
 
   return (
     <VictoryChart width={width} height={height} padding={padding} domainPadding={domainPadding}>
-      <VictoryBar style={style} data={values} />
+      <VictoryBar style={style} data={values[0]} />
```
wlthomson: This will throw an error if `values.length == 0` 😢 . A safer alternative is to declare a variable by deconstructing:

```
const [value = null] = values;

return (
  ...
     <VictoryBar style={style} data={value} />
  ...
)
```

Using this syntax, `value` will default to `null` if `values.length == 0`.

diegoevangelisti: Thanks Will! I was making more changes related with this PR that can be found here:
f5067e2
dba8a2c 
e46724a

I found that data structure can include more than one array of values (mainly on line charts), so I added a map over data to render as many i.e. lines as necessary in the same graph.

src/widgets/LineChart.js
```diff
@@ -36,10 +35,14 @@ export const LineChart = ({ width, height, data }) => {
   };
   return (
     <VictoryChart width={width} height={height} padding={padding}>
-      <VictoryScatter size={dotSize} style={dotStyle} data={values} />
-      <VictoryLine style={lineStyle} data={values} />
-      {renderXAxis()}
+      {data.map(({ values }) => (
+        <VictoryScatter size={dotSize} style={dotStyle} data={values} />
+      ))}
```
joshxg: You can use hooks here to memoize:

`const scatterLine = React.useMemo(() => data.map...., [data])`

then below.. `{scatterLine}`

Although not 100% needed, good to know

diegoevangelisti:
> Looks good. I wonder if you tested with an empty array? (For example, if you just use `{[]}`, which could happen if `data` is empty .. which it shouldn't be, but could be!)
> 
> nice work! 🎉

Thanks both @joshxg and @wlthomson for noticing the empty data issue. I guess I will need to display a message to show `Empty Report` or something similar. I will create an issue for that so I can move on with labels now

src/widgets/LineChart.js
diegoevangelisti: Nice! I'll give it a try later, thanks for sharing this

Issue #1903, opened 2020-01-12
Update indicators to use column codes


wlthomson:
## Is your feature request related to a problem? Please describe.

mSupply desktop indicators functionality has been updated to correctly set `[indicator_attribute]` column codes (see https://github.com/sussol/msupply/pull/5080, https://github.com/sussol/msupply/pull/5107).

## Describe the solution you'd like

Update `getIndicatorTableData` helpers to use column codes.

## Implementation

See above.

## Describe alternatives you've considered

N/A.

## Additional context

N/A.

wlthomson:
Closed #1911.

PR #1914, opened 2020-01-13
Add indicator attributes to outgoing sync

### LINKED ISSUE #1913: Add indicators outgoing sync

Indicators values can be edited on mobile. Values need to be synced back to desktop.

Add outgoing sync logic.

## Implementation

Add `IndicatorValue` to `recoreTypesSynced` and `generateSyncData.

N/A.

## Additional context

N/A.

wlthomson:
Fixes #1913. Branches off #1903.

## Change summary

Adds `IndicatorValue` to outgoing sync.

## Testing

Easiest way to test that a value is synced to desktop is to place a breakpoint into desktop mSupply incoming sync logic. If on a build, record browser can be used.

- [ ] Updated indicator values are synced to desktop. 

N/A.

### DIFF
--- a/src/sync/outgoingSyncUtils.js
+++ b/src/sync/outgoingSyncUtils.js
@@ -73,6 +73,16 @@ const safeGet = (record, path) => {
  */
 const generateSyncData = (settings, recordType, record) => {
   switch (recordType) {
+    case 'IndicatorValue': {
+      return {
+        ID: record.id,
+        facility_ID: settings.get(THIS_STORE_ID),
+        period_ID: record.period.id,
+        column_ID: record.column.id,
+        row_ID: record.row.id,
+        value: record.value,
+      };
+    }
     case 'ItemBatch': {
       return {
         ID: record.id,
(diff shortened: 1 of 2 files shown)


src/sync/outgoingSyncUtils.js
```diff
@@ -73,6 +73,16 @@ const safeGet = (record, path) => {
  */
 const generateSyncData = (settings, recordType, record) => {
   switch (recordType) {
+    case 'IndicatorValue': {
+      return {
+        ID: record.id,
+        facility_ID: settings.get(THIS_STORE_ID),
+        period_ID: record.period.id,
+        column_ID: record.column.id,
+        row_ID: record.row.id,
```
joshxg: is it too OTT to do `record?.period.id` etc?

wlthomson: I think we had this convo before (with positions reversed), with the question being whether it is good practice to null-check a non-optional realm field. I don't feel particularly strongly either way:

Pros:
- More "defensive" code.

Cons:
= Implication that `null` is an acceptable assignment, which in this case isn't true.

So, IMO, it's probably better to get an error in this situation rather than silently "fail" and potentially mess up server data?

PR #1920, opened 2020-01-14
#1898 Add labels to charts

### LINKED ISSUE #1898: Add Labels to Charts
## Describe the bug

The `json` data arriving from the desktop version includes label information from all charts, so that can be added to explain in a better way the illustrated data.

### To reproduce

See that all rendered charts are not showing any labels.

### Expected behaviour

Show labels on charts 😃

diegoevangelisti:
Fixes #1898

## Change summary

Graphs were needed to have its corresponding labels showing related data, specially for `LineChart` with more than one line per graph.
`label` information is part of the data received from desktop. For`LineChart` graphs with more than one line, the intention is to explicitly show what information each line represents.

Examples:

Screen Shot 2020-01-14 at 2 11 40 PM
Screen Shot 2020-01-14 at 2 11 54 PM
Screen Shot 2020-01-14 at 2 12 05 PM
Screen Shot 2020-01-14 at 2 12 17 PM

## Testing

Further testing will take place when the dashboard feature is ready.

### DIFF
--- a/src/widgets/BarChart.js
+++ b/src/widgets/BarChart.js
@@ -7,7 +7,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-import { VictoryChart, VictoryBar, VictoryAxis } from 'victory-native';
+import { VictoryChart, VictoryBar, VictoryAxis, VictoryLabel } from 'victory-native';
 
 import { APP_FONT_FAMILY, GREY, LIGHT_GREY, DARK_GREY, SUSSOL_ORANGE } from '../globalStyles';
 
@@ -33,7 +33,12 @@ export const BarChart = ({ data, width, height }) => {
   return (
     <VictoryChart width={width} height={height} padding={padding} domainPadding={domainPadding}>
       {data.map(({ values }) => (
-        <VictoryBar style={style} data={values} />
+        <VictoryBar
+          style={style}
+          data={values}
+          labels={({ datum }) => datum.y}
+          labelComponent={<VictoryLabel style={victoryStyles.labelStyle} />}
+        />
       ))}
       {renderYAxis()}
       {renderXAxis()}
@@ -66,4 +71,5 @@ const victoryStyles = {
     padDomain: 0.05,
     style: { data: { fill: SUSSOL_ORANGE } },
   },
+  labelStyle: { fontFamily: APP_FONT_FAMILY, fill: GREY },
 };
--- a/src/widgets/LineChart.js
+++ b/src/widgets/LineChart.js
@@ -6,7 +6,13 @@
 
 import React from 'react';
 import PropTypes from 'prop-types';
-import { VictoryAxis, VictoryChart, VictoryLine, VictoryScatter } from 'victory-native';
+import {
+  VictoryAxis,
+  VictoryChart,
+  VictoryLabel,
+  VictoryLine,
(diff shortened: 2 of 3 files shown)


diegoevangelisti:
> This looks great. I think it would be good to change the colours of the labels to match the colour of the line it is for. Specifically for the line chart, but could do for the pie also. Would be good to put into this PR as it seems to me to be quite a small change? If we say a limit of 5 lines, then we can just use 5 different colours and use a lookup table
> 
> ```
> const COLOURS = {
> 0: "#FFFFF"
> 1: "#CDCDCD"
> ...
> }
> ```
> 
> and use the index when mapping
> 
> ```
> data.map(({values}, index) => <VictoryLabel style={{...victoryStyles.labelStyle, color: COLOURS[index] .... } })
> ```
> 
> What do you think?

Sounds great, thanks! If fact, I was thinking about how to implement those different colors (as you said, specially for LineChart). I'll work on this now, try to make it for line and pie as well.

diegoevangelisti:
Part done here: 2da6921
I'm struggling with changing label colors for PieChart based on the fact that `index` remains stable, doesn't change as in LineChart.
I was trying to take `colorScale` to label styles, even defining a new array of colours for `colorScale` (instead of using `colorScale: warm`) and trying to use those but I couldn't find the way to related them.
Any ideas?

joshxg:
@diegoevangelisti hmm not really 🤣 

Do you mind chucking up some screen shots?

diegoevangelisti:
haha Sure!
That's the code for VictoryPie component, with its `colorScale` defined as `warm`(can be defined as an array of colours like `const colorScale = [ SUSSOL_ORANGE, GREY, ... ]`. And since mapping `data` returns just one array of `values`, if I include `index`, it's going to be just 0, so even if I declare `const COLOURS = { 0: ..., 1: ... }`, it'll take just the first color.
That's the complete situation

```
return data.map(({ values }) => (
    <VictoryPie
      width={width}
      height={height}
      padding={padding}
      padAngle={padAngle}
      innerRadius={widthPadded * innerRadius}
      labelRadius={widthPadded * labelRadius}
      colorScale={colorScale}
      labels={({ datum }) => `${datum.x} : ${datum.y}`}
      labelComponent={<VictoryLabel style={style} />}
      data={values}
    />
  ));
```
```
const victoryStyles = {
  pieChart: {
    padVertical: 0.15,
    padHorizontal: 0.15,
    innerRadius: 0.2,
    labelRadius: 0.375,
    padAngle: 2.5,
    colorScale: 'warm',
    style: { fontFamily: APP_FONT_FAMILY, fill: GREY },
  },
};
```

joshxg:
@diegoevangelisti 

Screen shots of the graphs line etc with the new colours 🤣 

Regarding the pie. That all makes pretty good sense, really. In general I think the labels are pretty sufficiently separated so it should be ok?

diegoevangelisti:
> @diegoevangelisti
> 
> Screen shots of the graphs line etc with the new colours 🤣
> 
> Regarding the pie. That all makes pretty good sense, really. In general I think the labels are pretty sufficiently separated so it should be ok?

Sweet! I added a graph that includes two lines for the same LineChart with different colours on top of the page (after editing the comment, so maybe so missed that one). Is that what you mean?😁

PR #1942, opened 2020-01-16
#1781 replace confirm modal with new confirm modal

### LINKED ISSUE #1781: NewConfirmModal to replace ConfirmModal
## The suggestion

Currently there are two components `NewConfirmModal` and `ConfirmModal` each doing the same thing. `ConfirmModal` should be finally deleted and replaced completely with `NewConfirmModal`

**Why?**
Having both is not good!

N/A

## Additional context

`ConfirmModal` is used in `DemoUserModal` and `finaliseModal`

Rebecca367:
Fixes #1781 

## Change summary

replace confirmModal with NewConfirmModal in related files and delete confirmModal.js accordingly.

## Testing

- [ ] log-out and see if app still running properly

If there are files that haven't been replaced with newConfirmModal.


joshxg:
@Rebecca367 

Does this look right?

image


4.5k added lines is a lot!

confusedmanheadsctratch


I think you need to change the base to develop

Rebecca367:
No, it's not right. obsolutely!

PR #1964, opened 2020-01-19
#1963 add cash register module

### LINKED ISSUE #1963: Add cash register module

Add cash register to modules.

See above.

## Implementation

- Add `"cash_register"` to `navStrings`.
- Add `usingCashRegister` to `MenuPage` props.
- Add `usingCashRegister` to `ModulesReducer`.
- Add `CASH_REGISTER_MODULE` to `settings`.

N/A.

## Additional context

See epic #1917.

wlthomson:
Fixes #1963.

## Change summary

Add cash register module (no navigation).

## Setup

- Add `"usesCashRegisterModule"` to store custom fields.

- [ ] `"Cash Register"` module appears in menu page.

N/A.


wlthomson:
> Sorry - we've got a conflict. I've screwed you. I had a PR from the other day which added the `usingPayments` to the `ModuleReducer`.
> 
> Without my PR I'd still suggest that we use `usingPayments` rather than just `cashRegister` as I think it is the larger, more encompassing super set of the functionality?

Sounds good. 

Only thought is whether keeping the modules separate may actually be a good idea? As far as I can see the biggest negative is two custom fields at config time, and I **think** there may be utility in separating out the cash handling from dispensary payments. On desktop I think there are different permissions for some of these use cases, so I'm thinking there could theoretically be situations where you may want to give a store functionality for receiving payments in the dispensing window but not for cash handling via the register?

joshxg:
Yup buy desktop also has a whole bunch of other functionality - for example patient receipts. Where as we aren't implemented cash receipts, and have just the functionality of being able to do things in the cash register and invoices etc.

I think we've both got decent arguments for future uses - so I think the question becomes what is the easiest to change - going from two to one, or one to two? I *think* going from two to one is harder?

PR #1972, opened 2020-01-21
#1953 add indicators search bar

### LINKED ISSUE #1953: Search bar doesn't work as expected for Indicators
## Describe the bug

When filtering an item using a program requisition with Indicators, the search bar is not filtering properly. See error here, when I am typing a random and long string, the filter is not being applied.

Screen Shot 2020-01-17 at 10 40 36 AM

### To reproduce

1. Go to 'Suppliers Requisitions'
2. Click on 'New Requisition'
3. Create a Program order
4. Select Indicators and check the table is visible.
4. Try to filter using the Search bar
5. See error.

### Expected behaviour

I reckon the placeholder should be `Search by: Description` (or `Search by: Code` ?) and then after typing something the filter should work and show proper options.

Same it's happening for option HIV or REGIMEN.

### Proposed Solution

Check that the search it is not trying to filter by ITEM (based on the actual placeholder) instead of Description or Code.

### Version and device info

- App version: All
- Tablet model: All
- OS version: All

wlthomson:
Fixes #1953.

## Change summary

This PR completes `3.2.0` release functionality 😄 🍾 . 

A few notes on improvements from the past few PRs:

- Functionality:
  - customer requisitions show indicator values.
  - searchbar works for indicators.
- Performance:
  - toggling and editing a lot faster.
- UX:
  - datatable stays in the same place when items/indicators toggled.
  - searchbar works for indicators.
- Code:
  - refactored a few badly designed methods.
  - more use of realm queries.
  - more use of selectors.
  - indicator helper methods cleaned up, code is also a lot tidier and easier to follow.

Did some testing and everything seems to be working pretty well. Due to a few refactors and replacing JS array methods with realm queries, the performance is a lot better now (although only tested on a small data file).

## Testing

### Supplier requisitions

- [ ] Supplier requisition indicators search bar filters indicator rows by description.
- [ ] Supplier requisition indicators search bar filters indicator rows by code. 
- [ ] Supplier requisition item search bar filters as previously.
- [ ] Supplier requisition search bar is reset when page is toggled.

### Customer requisitions

- [ ] Customer requisition indicators search bar filters indicator rows by description.
- [ ] Customer requisition indicators search bar filters indicator rows by code. 
- [ ] Customer requisition item search bar filters as previously.
- [ ] Customer requisition search bar is reset when page is toggled.

N/A.


src/pages/dataTableUtilities/reducer/tableReducers.js
```diff
@@ -144,9 +159,14 @@ export const toggleShowFinalised = state => {
...
 
-export const showIndicators = state => ({ ...state, showIndicators: true });
-
-export const hideIndicators = state => ({ ...state, showIndicators: false });
+export const toggleIndicators = state => {
+  const { showIndicators } = state;
+  return {
+    ...state,
+    showIndicators: !showIndicators,
+    searchTerm: '',
```
joshxg: cant test this at the moment could be wrong:

If you: 1) filter with the search bar while toggled on items 2) toggle to indicators 3) toggle back to items - are all the items still filtered, with nothing in the search bar?

Similar with sorting, but that might not matter to much.

wlthomson: Tested and works. I believe changing the `searchTerm` prop programmatically triggers the `SearchBar` `onChangeText` callback, so the behaviour you're describing doesn't occur.

PR #1981, opened 2020-01-22
#1906 enforce indicator types

### LINKED ISSUE #1906: Enforce indicator types

Indicator value updates are currently untyped.

Associated `[indicator_attribute]value_type` should be used for enforcing type constraints.

## Implementation

Update `editIndicatorValue` cell action.

N/A.

## Additional context

N/A.

wlthomson:
Fixes #1906.

## Change summary

Adds type enforcement to indicator data entry.

## Testing

- [ ] HIV input cells can only be set to positive integer values.
- [ ] HIV input cell values are persisted after reopening requisition.
- [ ] Requisition value input cells can only be set to the row data type.
- [ ] Requisition comment input cells can be any string.

Currently, indicators only supports string and numerical (interpreted as positive integer) data. Would be good in future to add support for negative integers, real numbers, booleans etc.

### DIFF
--- a/src/sync/incomingSyncUtils.js
+++ b/src/sync/incomingSyncUtils.js
@@ -132,7 +132,6 @@ export const sanityCheckIncomingRecord = (recordType, record) => {
   const requiredFields = {
     IndicatorAttribute: {
       canBeBlank: [
-        'code',
         'description',
         'index',
         'is_required',
@@ -141,7 +140,7 @@ export const sanityCheckIncomingRecord = (recordType, record) => {
         'axis',
         'is_active',
       ],
-      cannotBeBlank: ['indicator_ID'],
+      cannotBeBlank: ['code', 'indicator_ID'],
     },
     IndicatorValue: {
       cannotBeBlank: ['facility_ID', 'period_ID', 'column_ID', 'row_ID'],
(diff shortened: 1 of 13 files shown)


src/sync/incomingSyncUtils.js
```diff
@@ -337,7 +336,7 @@ export const createOrUpdateRecord = (database, settings, recordType, record) =>
         period: database.getOrCreate('Period', record.period_ID),
         column: indicatorColumn,
         row: indicatorRow,
-        value: record.value ?? '',
+        _value: record.value ?? '',
```
joshxg: should this default to `"0"` for numeric types

wlthomson: Currently all initialisation for `IndicatorValue` records is handled on creation, where the initial value can inferred from the column, row types). When the value is being synced in, there's no guarantee the `indicatorColumn` or `indicatorRow` has been synced yet (although likely it has), but also if that record has a `null` or `undefined` value,  it means somethings gone wrong big time (as it should never be able to get to that state).

wlthomson:
> Request changes to check if this be into a patch? And therefore master?

Good point. What do you think about making a patch release branch off master?

joshxg:
sure sounds good

PR #1989, opened 2020-01-23
#1957 Localized Currency

### LINKED ISSUE #1957: Localized currency

When displaying currency, it should be localized

I think a singleton wrapper around `currency` object, which somehow is set with the correct localization options

## Implementation

N/A

N/A

## Additional context

N/A

joshxg:
Fixes #1957 

## Change summary

- Adds a wrapper around the `currency` object to be able to use and set some configs passed on the current language

## Testing

N/A

N/A

### DIFF
--- a/src/localization/currency.js
+++ b/src/localization/currency.js
@@ -1,0 +1,34 @@
+/**
+ * mSupply Mobile
+ * Sustainable Solutions (NZ) Ltd. 2020
+ */
+
+import currency from 'currency.js';
+import { LANGUAGE_CODES } from '.';
+
+const CURRENCY_CONFIGS = {
+  DEFAULT: {
+    decimal: '.',
+    separator: ',',
+    pattern: '!#',
+    symbol: '$',
+    formatWithSymbol: true,
+  },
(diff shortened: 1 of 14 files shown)


src/localization/currency.js
```diff
@@ -0,0 +1,34 @@
...
+
+const CURRENCY_CONFIGS = {
+  DEFAULT: {
+    decimal: '.',
+    separator: ',',
+    pattern: '!#',
+    symbol: '$',
+    formatWithSymbol: true,
+  },
+  [LANGUAGE_CODES.FRENCH]: {
```
wlthomson: Nice! 

Definitely think its a good idea to localise decimals etc., only slight reservation is from a UI point of view re. the currency symbol and whether it might suggest to the user that we handle multiple currencies (as far as I'm aware we don't currently store transacts with any currency related metadata and don't handle conversions etc.).

joshxg: hmm I guess that's a good point - I wonder if it's better to just not use any symbol... hmm..

PR #1998, opened 2020-01-23
#1982 Validate dashboard's reports

### LINKED ISSUE #1982: Errors can occur while plotting a dashboard report with a wrong data structure
## Describe the bug

As mentioned in the docs related with `dashboard_report`, https://docs.msupply.org.nz/playground:2020-01_sworup_dashboard, If no `reportType` parameter is provided, table type is set by default.

Since the `json` structure can be one for pie/line/bar and a different one for table (main difference is the header), in case of assigning i.e. a table type (as default) to a bar chart, can occur in errors while trying to plot a graph in the new dashboard feature. It will result in trying to represent a table but received a bar chart `json` structure instead.

## To Reproduce

1. Create a dashboard report without a `reportType`
2. Make sure the json structure is the correct one for a Bar chart.
3. Run the report as usual
4. Sync in mobile
5. Go to `DashboardPage`
6. Try to open the mentioned report
7. See error

### Expected behaviour

The plotted report should have a proper json structure.

### Proposed Solution

Check that the `reportType` and the json structure matches. It can be done verifying if there's a header in the json (means the real type is table) and that the `reportType` is table. If not, show an error message instead of trying to plot.

### Version and device info

- App version: All
- Tablet model: All
- OS version: All

diegoevangelisti:
Fixes #1982

## Change summary

Because `dashboard_store_reports` coming from desktop can have a different data structure or report type as the expected, a report validation was added. Basically here we use the report type to contrast the data structure with a known validated json schema (using an external dependency named [ajv](https://www.npmjs.com/package/ajv)).

## Setup

Instructions to generate reports:

1. Make Dashboard feature visible in main page:

- [ ]  In Desktop go to: **Special -> Stores -> Select the current store -> Custom Fields** and add the following: 
Screen Shot 2020-01-25 at 11 44 09 AM

- [ ] Sync
- [ ] Check that the `Dashboard` button in visible in mobile and can navigate to its corresponding page.

2. Add reports

- [ ] In Desktop go to: **Admin -> Dashboard**
- [ ] Select one of the reports and click on `Duplicate`
- [ ] Open the Duplicated report and add the following properties:
Screen Shot 2020-01-25 at 11 54 42 AM

- [ ] Go to `json` tab and click on `Run Report`
- [ ] Now duplicate this created report (make 3 copies) and modify its reportType property to `BarChart`, `PieChart` and `Table`
- [ ] Run all reports (click on `json` tab -> `Run Report` in each of them)

3. Edit json structures of each generated report
- [ ] Download all json structures [here](https://drive.google.com/drive/folders/11YO89TChF6rTW36emyQGpnHCdkZ8-P5R?usp=sharing)
- [ ] In Desktop go to Design Mode, execute the method `record_browser` (see image below):
Screen Shot 2020-01-25 at 12 05 58 PM

- [ ] Look for `dashboard_store_report` and click on `All records`
- [ ] Click on Next to navigate through all previously created records one by one
- [ ] For each report, edit the field: `json` with its corresponding structure (previously download)
- [ ] After editing, click on `Update` and then `Save` (repeat the process for each report).
Screen Shot 2020-01-25 at 12 11 36 PM

- [ ] In mobile, Sync again and go to `DashboardPage`
- [ ] See the reports

## Testing

Many more test cases can be added in order to test this properly, but anyway I _think_ the following are the main ones:

Test 1 - Edit json structure to make the invalid

- [ ] Edit the json structure of all reports (check previous steps above).
Example: remove "label" and its value of TableReport, delete one "x" or "y" property and its values of a PieChart, delete "name" and its value of a BarChart, etc.
- [ ] Sync in mobile
- [ ] Check that the modified report/s is/are plot correct- means modifications were rejected
- [ ] Make as many changes an possible in the json structure of each reporta and repeat the process. 

Test 1 - Edit data in reports without changing the structure

- [ ] Edit some data of all reports (check previous steps above).
Example: change value of a "x" or "y" property in PieChart/BarChart/LineChart, modify a row value or header value in TableReport.
- [ ] Sync in mobile
- [ ] Check that the modified report/s is/are correct - means modifications were accepted

I _believe_ all expected errors were taking into account while defining which are the required parameters, but maybe I'm missing some 😅

### DIFF
--- a/src/sync/incomingSyncUtils.js
+++ b/src/sync/incomingSyncUtils.js
@@ -556,14 +556,13 @@ export const createOrUpdateRecord = (database, settings, recordType, record) => 
       const { ID: id, title, type, json } = record;
       try {
         const parsedData = JSON.parse(json);
-        const shouldSetData = checkIsObject(parsedData);
-        const validatedReport = validateReport(parsedData, type);
-        if (validatedReport) {
+        const shouldSetData = checkIsObject(parsedData) ? validateReport(parsedData, type) : false;
+        if (shouldSetData) {
           internalRecord = {
             id,
             title,
             type,
-            _data: shouldSetData ? JSON.stringify(parsedData.data) : null,
+            _data: JSON.stringify(parsedData.data),
           };
           database.update(recordType, internalRecord);
         }
--- a/src/sync/incomingSyncUtils.js
+++ b/src/sync/incomingSyncUtils.js
@@ -14,7 +14,7 @@ import {
 import { CHANGE_TYPES, generateUUID } from '../database';
 import { deleteRecord } from '../database/utilities';
 import { SETTINGS_KEYS } from '../settings';
-import { checkIsObject, validateReport } from '../utilities';
+import { validateReport } from '../utilities';
 
 const { THIS_STORE_ID, THIS_STORE_TAGS, THIS_STORE_CUSTOM_DATA } = SETTINGS_KEYS;
 
@@ -556,8 +556,7 @@ export const createOrUpdateRecord = (database, settings, recordType, record) => 
       const { ID: id, title, type, json } = record;
       try {
         const parsedData = JSON.parse(json);
-        const shouldSetData = checkIsObject(parsedData) ? validateReport(parsedData, type) : false;
-        if (shouldSetData) {
+        if (validateReport(parsedData, type)) {
           internalRecord = {
             id,
             title,
--- a/src/sync/incomingSyncUtils.js
+++ b/src/sync/incomingSyncUtils.js
@@ -14,7 +14,7 @@ import {
 import { CHANGE_TYPES, generateUUID } from '../database';
 import { deleteRecord } from '../database/utilities';
 import { SETTINGS_KEYS } from '../settings';
-import { checkIsObject } from '../utilities';
+import { checkIsObject, validateReport } from '../utilities';
 
 const { THIS_STORE_ID, THIS_STORE_TAGS, THIS_STORE_CUSTOM_DATA } = SETTINGS_KEYS;
 
@@ -557,13 +557,16 @@ export const createOrUpdateRecord = (database, settings, recordType, record) => 
       try {
         const parsedData = JSON.parse(json);
         const shouldSetData = checkIsObject(parsedData);
-        internalRecord = {
-          id,
-          title,
-          type,
-          _data: shouldSetData ? JSON.stringify(parsedData.data) : null,
-        };
-        database.update(recordType, internalRecord);
+        const validatedReport = validateReport(parsedData, type);
+        if (validatedReport) {
+          internalRecord = {
+            id,
+            title,
+            type,
+            _data: shouldSetData ? JSON.stringify(parsedData.data) : null,
+          };
+          database.update(recordType, internalRecord);
+        }
       } catch (error) {
         // Throw to parent, for now
         throw error;
--- a/src/utilities/validateReport.js
+++ b/src/utilities/validateReport.js
@@ -24,6 +24,7 @@ const tableReportSchema = {
         },
         formatters: {
           type: 'array',
+          items: { type: 'string' },
         },
       },
       required: ['header', 'rows', 'formatters'],
@@ -40,20 +41,26 @@ const otherReportSchema = {
     data: {
       type: 'array',
       items: {
-        label: { type: 'string' },
-        values: {
-          type: 'array',
-          items: {
-            type: 'object',
-            properties: {
-              x: { type: ['string', 'number'] },
-              y: { type: ['string', 'number'] },
-            },
-            required: ['x', 'y'],
+        type: 'object',
+        properties: {
+          label: { type: 'string' },
+          values: {
+            type: 'array',
+            items: { $ref: '#/definitions/plotValues' },
           },
         },
         required: ['label', 'values'],
       },
+    },
+  },
+  definitions: {
+    plotValues: {
+      type: 'object',
+      properties: {
+        x: { type: ['string', 'number'] },
+        y: { type: ['string', 'number'] },
+      },
+      required: ['x', 'y'],
     },
   },
   required: ['name', 'data'],
--- a/src/utilities/validateReport.js
+++ b/src/utilities/validateReport.js
@@ -9,6 +9,15 @@
 import Ajv from 'ajv';
 
 // Expected json structure for Table reports
+/**
+ *   {
+ *      data: {
+ *        header: [...],
+ *        rows: [...],
+ *        formatters: [...],
+ *      },
+ *   }
+ */
 const tableReportSchema = {
   type: 'object',
   properties: {
@@ -17,6 +26,7 @@ const tableReportSchema = {
       properties: {
         header: {
           type: 'array',
+          items: { type: 'string' },
         },
         rows: {
           type: 'array',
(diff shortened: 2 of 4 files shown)


src/utilities/validateReport.js
```diff
@@ -0,0 +1,87 @@
...
+        x: { type: ['string', 'number'] },
+        y: { type: ['string', 'number'] },
+      },
+      required: ['x', 'y'],
+    },
+  },
+  required: ['name', 'data'],
+};
+
+export const validateReport = (parsedData, type) => {
```
joshxg: I'd be tempted to use a structure a little more flexible to changes in the future:

```
const TABLE_SCHEMAS = {
    Table: TableReportSchema,
    Pie: DefaultReportSchema,
    Line: DefaultReportSchema,
    Bar: DefaultReportSchema,
}
```

Then your validate report would be more 

`export const validateReport = (data, type) => return ajv.validate(TABLE_SCHEMAS[type] || {}, data)`

No default case here, so not sure if `ajv.validate({}, data)` would always return false (for when the type does not exist)

Seems a little nicer to me, but it's functionally the same!

src/utilities/validateReport.js
```diff
@@ -0,0 +1,87 @@
...
+
+// Expected json structure for Table reports
+const tableReportSchema = {
+  type: 'object',
+  properties: {
+    data: {
+      type: 'object',
+      properties: {
+        header: {
+          type: 'array',
```
joshxg: does this get an `items`?

src/utilities/validateReport.js
```diff
@@ -0,0 +1,87 @@
...
+        rows: {
+          type: 'array',
+          items: { type: 'array' },
+        },
+        formatters: {
+          type: 'array',
+          items: { type: 'string' },
+        },
+      },
+      required: ['header', 'rows', 'formatters'],
```
joshxg: formatters is not required

src/utilities/validateReport.js
```diff
@@ -0,0 +1,87 @@
...
+ * on its data structure and type and return a boolean.
+ *
+ * @param   {object}  parsedData  The data to check.
+ * @param   {string}  type        Type of Report to check.
+ */
+
+import Ajv from 'ajv';
+
+// Expected json structure for Table reports
+const tableReportSchema = {
```
joshxg: I'd potentially be tempted to put a docstring on each schema which had a quick basic example of the schema - it's quite hard to read (and can't be helped) but for scanning.. i.e.:

```
{
    data: { 
        header: [...],
        rows: [...],
        formatters: [...],
    },
}
```

for quickly scanning, it's nice to have a little reference for the shape for any large object

src/sync/incomingSyncUtils.js
```diff
@@ -557,13 +557,16 @@ export const createOrUpdateRecord = (database, settings, recordType, record) =>
...
         const parsedData = JSON.parse(json);
         const shouldSetData = checkIsObject(parsedData);
-        internalRecord = {
-          id,
-          title,
-          type,
-          _data: shouldSetData ? JSON.stringify(parsedData.data) : null,
-        };
-        database.update(recordType, internalRecord);
+        const validatedReport = validateReport(parsedData, type);
```
joshxg: does the `validatedReport` now remove the need for `shouldSetData`?

Is it also not even needed now to parse the data? Maybe we could just do that in the `validateReport` helper?

diegoevangelisti:
> 🎉
> 
> Nice!
> 
> I really like this a lot!! A few suggestions - mostly opinions so do make sure you consider them carefully :)
> 
> I have not tested this yet and I'd like to - but will probably do it tomorrow hehe. Got any tips for testing? I have forgotten already how to set it all up and if you can tell me the dashboard methods to generate each type of report?

Sure, I'll update this adding all the steps to add different reports and with its corresponding data structure (so you can have fun tomorrow 😂) Good you noticed that

src/utilities/validateReport.js
diegoevangelisti: This is great, I was thinking about something to group all different reports and avoid adding as many cases as reports we have (not a problem with four, but a good practice for longer lists).

src/utilities/validateReport.js
diegoevangelisti: I didn't know that haha nice! I'll remove the required

src/utilities/validateReport.js
diegoevangelisti: Yeah, I missed that. Sure, this line should be added: `items: { type: 'string' }`
Thanks!

src/sync/incomingSyncUtils.js
diegoevangelisti: I was thinking about something like this:

```
case 'Report': {
      const { ID: id, title, type, json } = record;
      try {
        const parsedData = JSON.parse(json);
        const shouldSetData = checkIsObject(parsedData) ? validateReport(parsedData, type) : false;
        if (shouldSetData) {
          internalRecord = {
            id,
            title,
            type,
            _data: JSON.stringify(parsedData.data)
          };
          database.update(recordType, internalRecord);
        }
      } catch (error) {
        // Throw to parent, for now
        throw error;
      }
      break;
    }
```

src/utilities/validateReport.js
diegoevangelisti: Done here 
e7779b1

src/utilities/validateReport.js
diegoevangelisti: Done here e7779b1

src/utilities/validateReport.js
diegoevangelisti: Done here e7779b1

src/utilities/validateReport.js
diegoevangelisti: Done here e7779b1

src/sync/incomingSyncUtils.js
diegoevangelisti: Done here df1e5b9
Open to suggestions anyway 😁

src/sync/incomingSyncUtils.js
```diff
@@ -556,14 +556,16 @@ export const createOrUpdateRecord = (database, settings, recordType, record) =>
...
         const parsedData = JSON.parse(json);
-        const shouldSetData = checkIsObject(parsedData);
-        internalRecord = {
-          id,
-          title,
-          type,
-          _data: shouldSetData ? JSON.stringify(parsedData.data) : null,
-        };
-        database.update(recordType, internalRecord);
+        const shouldSetData = checkIsObject(parsedData) ? validateReport(parsedData, type) : false;
```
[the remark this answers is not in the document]
diegoevangelisti: You're right, `validateReport` is doing all the job here😄

diegoevangelisti:
> Just a change with the condition I believe to simplify it.
> 
> Also - a suggestion for the catch clause, and possibly the validation
> 
> This function here:
> https://github.com/openmsupply/mobile/blob/744668de82c1cbdf20cf712ccf96459e41bdb939/src/sync/Synchroniser.js#L94-L100
> 
> This reports to bugsnag, any errors that occur during sync. So anywhere where anyone is using mSupply mobile, if there is a sync problem, we log it to bugsnag with the body and headers from the server.
> 
> Would be good to combine everything once you understand that method https://docs.bugsnag.com/platforms/javascript/ feel free to make a new issue for this if you'd rather.
> 
> Some pseudo:
> 
> ```
> try {
>   *parse json*
>    if (validateRecord)
>         *integrate record into mobile*
>    else
>         throw error("couldnt validate - maybe validator method can tell us why")
> }
> catch(error){
>      *log to bugsnag*
> }
> ```
> 
> Then we can know whenever we are getting sent incorrect records - and we can remote into the server that is sending the wrong data and fix it!

Thanks for this, great idea. I'd rather open a new issue regarding this so it can be integrated properly after the validation changes are approved!

src/sync/incomingSyncUtils.js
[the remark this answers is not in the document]
diegoevangelisti: Done here: 76cefa3

PR #2043, opened 2020-01-28
#2037 fix cash register amount input

### LINKED ISSUE #2037: Fix cash register transaction amount input

The cash register modal transaction amount bottom modal input is not suitable for entering numeric data.

- Update keyboard to numeric.
- Validate input as currency.

## Implementation

See above.

N/A.

## Additional context

See epic #1917.

wlthomson:
Fixes #2037.

## Change summary

Updates cash register input to use `currency` module. Still a pretty basic input, but looking better.

## Testing

- [ ] Can only set numerical cash in/out quantities.
- [ ] Transaction totals display correctly in cash transaction page.

Still a lot of room for improvement here, ideally expanding `BottomCurrencyModal` to use a similar component to `CurrencyRowInput`.

### DIFF
--- a/src/widgets/bottomModals/BottomCurrencyEditor.js
+++ b/src/widgets/bottomModals/BottomCurrencyEditor.js
@@ -1,0 +1,63 @@
+/* eslint-disable import/prefer-default-export */
+import React from 'react';
+import { PropTypes } from 'prop-types';
+import { TextInput, StyleSheet } from 'react-native';
+
+import { BottomModalContainer } from './BottomModalContainer';
+import { OnePressButton } from '../OnePressButton';
+import globalStyles from '../../globalStyles';
+
+const {
+  modalTextInput,
+  modalText,
+  modalButtonText,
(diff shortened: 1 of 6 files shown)


src/widgets/bottomModals/BottomCurrencyEditor.js
```diff
@@ -0,0 +1,63 @@
...
+const {
+  modalTextInput,
+  modalText,
+  modalButtonText,
+  modalOrangeButton,
+  button,
+  buttonText: buttonTextStyle,
+} = globalStyles;
+
+export const BottomCurrencyEditor = React.memo(
```
joshxg: I think this would be better overall as `CurrencyEditor` - just like `TextEditor` where you can wrap it in either the `BottomModalContainer` or the `ModalContainer`. All it would take would be to remove the `BottomModalContainer` and wrap this component in it. Better to do that even if it doesn't really work out of the box imo

PR #2049, opened 2020-01-29
#2046 localize dates

### LINKED ISSUE #2046: Localize dates
## Describe the bug

Dates aren't localized

### To reproduce

See a date when using french language

### Expected behaviour

Dates should be localized

### Proposed Solution

N/A

### Version and device info

N/A

### Additional context

N/A

joshxg:
Fixes #2046 

## Change summary

- Adds a function to set the date locale of `moment` when changing the language of the app.

## Testing

N/A

N/A

### DIFF
--- a/src/settings/MobileAppSettings.js
+++ b/src/settings/MobileAppSettings.js
@@ -2,7 +2,7 @@ import { Settings } from 'react-native-database';
 
 import { SETTINGS_KEYS } from './index';
 import { DEFAULT_LANGUAGE } from '../localization/index';
-import { setCurrentLanguage } from '../localization/utilities';
+import { setCurrentLanguage, setDateLocale } from '../localization/utilities';
 import { UIDatabase } from '../database';
 import { setCurrencyLocalisation } from '../localization/currency';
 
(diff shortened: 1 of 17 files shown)


src/settings/MobileAppSettings.js
```diff
@@ -30,6 +30,7 @@ class MobileAppSettings extends Settings {
     const currentLanguage = this.get(SETTINGS_KEYS.CURRENT_LANGUAGE);
     setCurrentLanguage(currentLanguage);
```
wlthomson: Might be missing some info, but I can't think of a situation where you would want to set these to different languages. If that's the case, I think could make this a bit tidier by calling them all in `setCurrentLanguage`?

src/settings/MobileAppSettings.js
```diff
@@ -30,6 +30,7 @@ class MobileAppSettings extends Settings {
...
     setCurrencyLocalisation(currentLanguage);
+    setDateLocale(currentLanguage);
```
wlthomson: I think this is more setting "format" rather than locale. I think this has a bit of potential for confusion, e.g. between a currency format and an actual unit of currency, or a date format and a timezone. Maybe something like `setCurrencyLanguage`, `setDateLanguage`, or `setCurrencyFormat(currentLanguage)`, `setDateFormat(currentLanguage)`?

joshxg: yeah you're right - will fix soon

PR #2064, opened 2020-01-30
Fix outgoing sync for payment and receipt transactions

### LINKED ISSUE #2063: Payment and receipt transaction totals do not sync

See #2062.

Update `outgoingSyncUtils` with cases for `Payment` and `Receipt` transaction types.

## Implementation

See above.

An alternative is to generate placeholder `TransactionItem` and `TransactionBatch` records associated with the payment record, but this is not done in desktop, so consistency seems like the best way to go here.

## Additional context

N/A.

wlthomson:
Fixes #2063.

## Change summary

Updated outgoing sync to correctly handle totals for  `payment` and `receipt` transactions (i.e. cash transactions with no associated `TransactionItem` or `TransactionBatch` records).

## Testing

- [ ] Create a cash in transaction for a non-zero amount via the mobile cash register. Sync and open the transaction in the desktop cash register. The amount should be equal to that entered previously.
- [ ] Create a cash out transaction for a non-zero amount via the mobile cash register. Sync and open the transaction in the desktop cash register. The amount should be equal to that entered previously.

N/A.

### DIFF
--- a/src/sync/outgoingSyncUtils.js
+++ b/src/sync/outgoingSyncUtils.js
@@ -173,6 +173,8 @@ const generateSyncData = (settings, recordType, record) => {
       };
     }
     case 'Transaction': {
+      const isCashReconciliation = record.type === 'payment' || record.type === 'receipt';
+      const total = String(isCashReconciliation ? record.total : record.totalPrice);
       return {
         ID: record.id,
         name_ID: record.otherParty && record.otherParty.id,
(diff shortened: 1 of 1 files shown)


src/sync/outgoingSyncUtils.js
```diff
@@ -173,6 +173,8 @@ const generateSyncData = (settings, recordType, record) => {
       };
     }
     case 'Transaction': {
+      const isCashReconciliation = record.type === 'payment' || record.type === 'receipt';
```
joshxg: could just do `record.totalPrice || record.total`

wlthomson:
> If this works, sweet - but I think better solution is to set `total` and basically remove `totalPrice`. Mainly for historic. Would just need to change `finalise` method to basically use `totalPrice` to set `total`. Using `totalPrice` will mean finalised `Transaction` totals will change

I agree. That would also remove the need to generate the `TransactionItem` to link the customer invoice/supplier credit record to the `TransactionBatch`.

src/sync/outgoingSyncUtils.js
wlthomson: I went with explicit mainly for readability but also just because I think it's a bit safer in terms of potential regressions or weird edge cases. I agree with above re. removing `totalPrice`, so my inclination is to keep it this way and consider it as just a short term workaround for UAT.

PR #2072, opened 2020-01-30
#2045 add cash transaction name searchbar

### LINKED ISSUE #2045: Add cash transaction name searchbar

Cash transaction names bottom modal requires a searchbar. 

Add a searchbar to the bottom modal.

## Implementation

Wrap `MultiSelectList` in a bottom modal container, e.g. `BottomMultiSelectList`.

N/A.

## Additional context

Epic #1917.

wlthomson:
Fixes #2045.

## Change summary

- Replace name bottom modal with full screen modal.
- Add searchbar to name modal.

## Testing

- [ ] Name selection modal opens on press.
- [ ] Name selection modal correctly updates name.
- [ ] Name selection modal searchbar correctly filters names.

N/A.


src/widgets/modals/CashTransactionModal.js
```diff
@@ -160,80 +161,79 @@ export const CashTransactionModal = ({ onConfirm }) => {
...
-          data={reasons}
-          keyToDisplay={CASH_TRANSACTION_KEYS.REASON}
-          onPress={onSubmitReason}
-          highlightValue={reason?.title}
+      </View>
+      <ModalContainer title={placeholderTextName} isVisible={isNameModalOpen}>
+        <AutocompleteSelector
+          options={names}
+          queryString="name CONTAINS[c] $0"
+          sortKeyString="name"
```
joshxg: Would be good to use the `renderLeftText` and `renderRightText` I think. For a patient, using `${lastName}, ${firstName}` and for anyone else just `${name}`. Alternatively you could make the `Name.toString()` do the work for that.

A `renderRightText` for showing the `code` and possibly any other details - `D.O.B` maybe?

src/widgets/modals/CashTransactionModal.js
```diff
@@ -160,80 +161,79 @@ export const CashTransactionModal = ({ onConfirm }) => {
...
-        isOpen={isReasonModalOpen}
-        modalStyle={localStyles.bottomModalContainerStyle}
-      >
-        <GenericChoiceList
-          data={reasons}
-          keyToDisplay={CASH_TRANSACTION_KEYS.REASON}
-          onPress={onSubmitReason}
-          highlightValue={reason?.title}
+      </View>
+      <ModalContainer title={placeholderTextName} isVisible={isNameModalOpen}>
```
joshxg: need to provide an `onClose` prop. I think `fullScreen` would be a bit better too

src/widgets/modals/CashTransactionModal.js
wlthomson: Nice. 58567d32597e762d194d28447967defa54f4fe78, 87f4d1141f5eddf2d1625aa849a35fdeeaf68f80.

src/widgets/modals/CashTransactionModal.js
wlthomson: acaf42cab8a8c69fe48c990e935dbcb76a3f8529. Have omitted `fullScreen` for now, just due to the fact that the parent modal isn't fullscreen, which makes it a bit jarring. Open to changing this though.

src/database/utilities/constants.js
```diff
@@ -21,3 +21,12 @@ export const NUMBER_SEQUENCE_KEYS = {
...
   SUPPLIER_INVOICE_NUMBER: 'supplier_invoice_serial_number',
 };
+
+export const NAME_TYPE_KEYS = {
+  INVENTORY_ADJUSTMENT: 'inventory_adjustment',
+  FACILITY: 'facility',
+  PATIENT: 'patient',
+  BUILD: 'build',
+  STORE: 'store',
+  REPACK: 'repack',
```
joshxg: awesome!

src/widgets/modals/CashTransactionModal.js
```diff
@@ -38,7 +43,7 @@ export const CashTransactionModal = ({ onConfirm }) => {
   const [isReasonModalOpen, setIsReasonModalOpen] = useState(false);
   const [isDescriptionModalOpen, setIsDescriptionModalOpen] = useState(false);
 
-  const names = useMemo(() => UIDatabase.objects('Name'));
+  const names = useMemo(() => UIDatabase.objects('Name').filtered('isVisible == true'));
```
joshxg: sorry missed this before - Should probably remove the mobile stores name, and also `InventoryAdjustment`, `build` and `repack` names. Could potentially use all `isCustomer`, `isSupplier` and `isPatient` names, I'm pretty sure. Possibly worth a `UIDatabase` case.

src/widgets/modals/CashTransactionModal.js
```diff
@@ -60,6 +65,25 @@ export const CashTransactionModal = ({ onConfirm }) => {
   const onChangeText = useMemo(() => text => setDescriptionBuffer(text));
   const onChangeAmount = useMemo(() => value => setAmountBuffer(value));
 
+  const renderNameLeftText = useCallback(({ firstName, lastName, name: fullName, isPatient }) =>
+    isPatient ? `${lastName}, ${firstName}` : `${fullName}`
+  );
+
+  const renderNameRightText = useCallback(({ type: nameType, dateOfBirth }) => {
+    switch (nameType) {
+      case NAME_TYPE_KEYS.FACILITY:
```
joshxg: You can also have `type === 'store'` (internal suppliers) - facility also is really terminology which isn't used very often? I think using just the names `code` would be all good, and what you have for a patient if a patient?

src/widgets/modals/CashTransactionModal.js
```diff
@@ -154,80 +180,85 @@ export const CashTransactionModal = ({ onConfirm }) => {
...
+        onClose={onCloseName}
       >
-        <GenericChoiceList
-          data={reasons}
-          keyToDisplay={CASH_TRANSACTION_KEYS.REASON}
-          onPress={onSubmitReason}
-          highlightValue={reason?.title}
+        <AutocompleteSelector
+          options={names}
+          queryString="name CONTAINS[c] $0"
```
joshxg: would possibly want to make it `firstName contains[c]... OR name... OR lastName....` -- alternatively we could make the `name` for a patient equal to `lastName, firstName` rather than `""` as it is now

src/widgets/modals/CashTransactionModal.js
wlthomson: 5498cbe109c2a4a5b0d05cc5392abb00a9828b2f.

src/widgets/modals/CashTransactionModal.js
wlthomson: f15e29f43a35879ef971b538970dcb166f17c1e5.

I think facility is generally used as a generic term for an internal or external supplier, so trying to keep consistent with that. Not sure what you mean re. patient (there's a key case for that)?

src/widgets/modals/CashTransactionModal.js
wlthomson: e7996746d24ab5f37b3199defc234850c436e4cf.

src/widgets/modals/CashTransactionModal.js
joshxg: `.filtered("isSupplier == $0 || isCustomer == $0 || isPatient == $0",true)` is that nicer? I'm not sure but it certainly isn't worth even the time it took me to type it!

src/widgets/modals/CashTransactionModal.js
joshxg: That's then a contradiction - if a facility is an external store, then you've just added both potentially internal and external suppliers (the terms used in mobile - i.e. in `UIDatabase`) to have "facility" next to it. 

I think using `isSupplier` or `isPatient` fields would be more suitable - really, we just need to check if `isPatient` and render the extra details.

src/widgets/modals/CashTransactionModal.js
joshxg: I forgot about this - in desktop `[name]name` is already set to this for patients. The `createRecord` case has also now been updated https://github.com/openmsupply/mobile/blob/5c86d528ff49ef4cef189881277181ff0fa29236/src/database/utilities/createRecord.js#L123-L136

so this isn't needed

wlthomson: Can we get these changes into `#feature-dispensary` (alternatively can we get this approved and resolve at merge)?

joshxg: the link above is from `#feature-dispensary` ? confused sorry

src/widgets/modals/CashTransactionModal.js
wlthomson: Typo above: "external" = "internal or external" store.

Cash register needs to handle customers as well as suppliers. So I think using the `isX` is good, but will need to include `isCustomer`.

src/widgets/modals/CashTransactionModal.js
wlthomson: It's been merged now 👍 .

src/widgets/modals/CashTransactionModal.js
wlthomson: 45e11b63aaa999da1a4bb3f42c982bef8cf574c4.

src/widgets/modals/CashTransactionModal.js
wlthomson: A bit nicer, but equally arguable that should only parametrise queries when parameter is variable (constant here). Will leave for now but don't feel strongly either way.

wlthomson:
@joshxg Rerequesting review (can't do it via UI).

joshxg:
@wlthomson approved but there is still the code that isn't needed in incoming sync utils

PR #2079, opened 2020-01-31
#2069 LogOut timer

### LINKED ISSUE #2069: Auto-log out after some time frame

From @craigdrown 

IC Requirement: Automatically log a user out after a time frame of X for security reasons. Data should not be lost which was being worked on.

Log the user out after X time

## Implementation

Could use store custom data `logoutTimeoutPeriodInSecondsOrMinutes` to set the period..

- Use [AppState](https://facebook.github.io/react-native/docs/appstate#addeventlistener) to hook into the app going into the background - log the user out if so. Android would go to sleep after X amount of time, putting the app into the background (I THINK, UNTESTED)
- Use a package like https://www.npmjs.com/package/redux-idle-monitor to track activity, have a schedule that is reset each time an event happens.
- Use the above package but combine it with the current user authentication scheduler (tries to re-auth against the server every X minutes, or the local DB if no connection available) - I think not good

Not having this in the dispensary/IC branch 🤣 

## Additional context

@wlthomson @mark-prins @Chris-Petty thoughts?

joshxg:
Fixes #2069 

## Change summary

- Adds a hook into the app state:
    - When put into the background, store the time in state
    - When coming back into the forground, if the current time is greater than the time held+3minutes - log the user out.
- Had to use `ModalContainer` for `LoginModal` because the state of the modal kept playing up and not opening correctly.

## Testing

- [ ] When the app is in the background for >3 minutes, the user is logged out
- [ ] When logging back in after being logged out, the app state is the sme as when you left it
- [ ] When the app is in the background for < 3 minutes, the user is not logged out

Decided to just go with this quite basic implementation cause it was the quickest to implement


src/actions/UserActions.js
```diff
@@ -3,12 +3,33 @@
...
+
+const login = user => ({ type: USER_ACTION_TYPES.LOG_IN, payload: { user } });
+const logout = () => ({ type: USER_ACTION_TYPES.LOG_OUT });
+const setTime = () => ({ type: USER_ACTION_TYPES.SET_TIME });
+
+const active = () => (dispatch, getState) => {
+  const { user } = getState();
+  const { time } = user;
+
+  const tenMinutesLater = moment(time).add(3, 'minutes');
```
wlthomson:
```suggestion
  const tenMinutesLater = moment(time).add(10, 'minutes');
```

wlthomson: Could probably make this a more generic identifier, e.g. `activityTimeout`?

joshxg:
Renamed and used `threeMinutesLater` - I think if we want this to be more generic, it'll be a pref coming from desktop

Issue #2104, opened 2020-02-02
Refactor payment code-path


joshxg:
## Is your feature request related to a problem? Please describe.

The payment code-path is a little convoluted and not very extensible

## Describe the solution you'd like

Not sure!

## Implementation

Not sure!

## Describe alternatives you've considered

Not sure!

## Additional context

Not sure!

PR #2105, opened 2020-02-02
#2099 Store insurance details

### LINKED ISSUE #2099: Store insurance discount amount and rate on a Transaction
## Describe the bug

Currently deriving a `Transaction` insurance discount amount and rate through the relation. Need to store the values for history

### To reproduce

Dispensing Development bug

### Expected behaviour

Dispensing Development bug

### Proposed Solution

Dispensing Development bug

### Version and device info

Dispensing Development bug

### Additional context

Dispensing Development bug

joshxg:
Fixes #2099 

## Change summary

- Correctly assigns insurance details to a finalise prescription. I've made #2104 to refactor this code to be less weird/horrible as the amount of parameters is very msupplyesque but it works so going with it for now
- Added insurance stuff to outgoing sync including the historic insurance details
- Fixed a couple of prop type and key warnings

Basically, this makes syncing prescriptions with insurance correct

I.e. I can now get this in desktop after syncing a prescription with a payment of $10, with 25% insurance 🎉 

image

## Testing

N/A

N/A

### DIFF
--- a/src/utilities/modules/dispensary/pay.js
+++ b/src/utilities/modules/dispensary/pay.js
@@ -36,7 +36,16 @@ import { UIDatabase } from '../../../database';
  * @param {Number} scriptTotal The total amount of the script.
  */
 
-export const pay = (currentUser, patient, script, cashAmount, scriptTotal) => {
+export const pay = (
+  currentUser,
+  patient,
+  script,
+  cashAmount,
+  scriptTotal,
+  subtotal,
+  discountAmount,
+  discountRate
+) => {
   if (!patient.isPatient) throw new Error('Patient is not a patient');
   if (!script.isPrescription) throw new Error('Script is not a script');
   if (!(script.items.length > 0)) throw new Error('Script is empty');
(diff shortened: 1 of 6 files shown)


src/utilities/modules/dispensary/pay.js
```diff
@@ -80,5 +89,9 @@ export const pay = (currentUser, patient, script, cashAmount, scriptTotal) => {
   if (usingCredit && creditToAllocate > 0) throw new Error('Credit not fully allocated');
 
   // Ensure the script is finalised once paid.
+  script.subtotal = subtotal;
+  script.total = subtotal;
```
mark-prins: just a question: should the total be the subtotal - discount amount?

Kind of related, there's now this: https://github.com/sussol/msupply/issues/5085

joshxg: Yeah - at this point I have no idea what implementation to do. Same with https://github.com/sussol/msupply/issues/5086 :(

Issue #2134, opened 2020-02-04
Align total + subtotal Transaction fields with desktop


joshxg:
## Is your feature request related to a problem? Please describe.

`[transact]total` and `[transact]subtotal` fields have been slightly changed in desktop:

related PR: https://github.com/sussol/msupply/pull/5366

## Describe the solution you'd like

Align fields with the spec:
https://wiki.sussol.net/doku.php/msupply:specifications:customers:receipts_payments

## Implementation

N/A

## Describe alternatives you've considered

N/A

## Additional context

N/A

joshxg:
Closed by #2139

PR #2136, opened 2020-02-05
#2062 cash in cash out totals incorrect

### LINKED ISSUE #2062: Custoemr Credit made from a cash in is incorrect
## Describe the bug

Custoemr Credit made from a cash in is incorrect

### To reproduce

Dispensary development bug

### Expected behaviour

- Should have correct total, subtotal - the negative amount of the total of the RC

### Proposed Solution

N/A

### Version and device info

N/A

joshxg:
Fixes #2062 

## Change summary

- Save the transaction total amount on a Transaction rather than trying to use the derived value for anything other than a `ps` or `rc`, 'sc', 'cc' when finalising. Avoiding regressions by keeping the same behaviour for the existing types. The only problem is for the `ci` created for a cash out which would technically get through and have it's `total` reset, but the finalise method isn't used for that type. I don't think that's a good enough reason, so I think this still needs some thinking - it is working for now though.. :(
- Can then use `total` field in `outgoingSync` for all cases.
- Using `total` field in all cases fixes this for the `cc` and `ci`

## Testing

N/A

- There's probably so more nicer cleanup to do here. I think using a different method to determine the total for cash transactions would be nice? I.e. if no `Item`, try see if there are any `batch` or something like that.

### DIFF
--- a/src/database/DataTypes/Transaction.js
+++ b/src/database/DataTypes/Transaction.js
@@ -450,6 +450,7 @@ export class Transaction extends Realm.Object {
     if (!this.isConfirmed) this.confirm(database);
 
     this.status = 'finalised';
+    this.total = this.totalPrice;
     database.save('Transaction', this);
   }
 }
--- a/src/database/DataTypes/Transaction.js
+++ b/src/database/DataTypes/Transaction.js
@@ -116,6 +116,10 @@ export class Transaction extends Realm.Object {
     return this.mode === 'dispensary' && this.type === 'customer_invoice';
   }
 
+  get isCashTransaction() {
+    return this.type === 'payment' || this.type === 'receipt';
+  }
+
   /**
    * Get if transaction is an external supplier invoice.
    *
@@ -449,8 +453,14 @@ export class Transaction extends Realm.Object {
     }
     if (!this.isConfirmed) this.confirm(database);
 
+    // Finding the totalPrice propogates through `TransactionItem` records down to the batch
+    // level, deriving the full cost of the Transaction. Cash transactions do not use the Item
+    // level, so the total can't be derived the same.
+    if (!this.isCashTransaction) {
+      this.total = this.totalPrice;
+    }
+
     this.status = 'finalised';
-    this.total = this.totalPrice;
     database.save('Transaction', this);
   }
 }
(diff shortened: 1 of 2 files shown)


src/database/DataTypes/Transaction.js
```diff
@@ -449,6 +457,13 @@ export class Transaction extends Realm.Object {
     }
     if (!this.isConfirmed) this.confirm(database);
 
+    // Finding the totalPrice propogates through `TransactionItem` records down to the batch
+    // level, deriving the full cost of the Transaction. Cash transactions and credits are
+    // created finalised, having their total already set.
+    if (!this.isCashTransaction || !this.isCredit) {
```
wlthomson: Should this be `!(this.isCashTransaction || this.isCredit)`

src/sync/outgoingSyncUtils.js
```diff
@@ -188,21 +185,18 @@ const generateSyncData = (settings, recordType, record) => {
...
         mode: record.mode,
         prescriber_ID: record.prescriber && record.prescriber.id,
-        total,
-        foreign_currency_total: total,
+        total: String(record.total),
+        foreign_currency_total: String(record.total),
         their_ref: record.theirRef,
         confirm_date: getDateString(record.confirmDate),
-        subtotal: total,
+        subtotal: String(record.total),
```
wlthomson: Nite: might be nicer to do something like:

```
const total = String(record.total);
return {
  ...
  total,
  foreign_currency_total: total,
  ...
  subtotal: total.
  ...
};
```

Or, something I just found out works in JS:


```
const foreign_currency_total = subtotal = total = String(record.total);
return {
  ...
  total,
  foreign_currency_total,
  ...
  subtotal,
  ...
};
```

😆 .

src/database/DataTypes/Transaction.js
joshxg: embares

https://github.com/openmsupply/mobile/pull/2136/commits/e7870f102c8abf3ecfd77e57f4a961c1c1f633b4

src/sync/outgoingSyncUtils.js
joshxg: about to do a new PR for #2134 that changes more things regarding total, so will change this up there

PR #2147, opened 2020-02-06
#2142 supplier column for stocktakes

### LINKED ISSUE #2142: Supplier column on StocktakeBatchModal

When creating a new stocktake batch, you cannot assign a supplier. 

A column to select a supplier

## Implementation

@craigdrown @mark-prins (sorry, thoughts?)

- Not sure if this should be under any sort of customisation. I am thinking it shouldn't
- Maybe default to the mobiles  stores supplying store?
- Should no supplier be allowed? Doesn't seem to make sense
- Should possible suppliers be only the stores supplying store + external suppliers?

N/A

## Additional context

N/A

joshxg:
Fixes #2142 

## Change summary

- Not 100% sure if this is all correct 🤣 
- Supplier invoices created on mobile weren't adding their supplier to the batch name - have adjusted that in the `createRecord` case for an `ItemBatch` by passing the supplier as an extra parameter. Not sure if this was intentional or a mistake.
- Used the extra parameter to pass the `inventory_adjustment` name when creating a stocktake batch, which will set that as the default (same as desktop behaviour) - not sure why it wasn't doing this originally.
- Added the `supplier` field to `StocktakeBatch` to set it in the underlying `ItemBatch` after the `Stocktake` has been finalised.
- Allows choosing a supplier through the regular supplier select modal by clicking on the supplier cell - selection is any supplier or the inventory adjustment name

Pic of ICs `StocktakeBatch` edit now .. 🤔 

image

## Testing
N/A

N/A


src/pages/dataTableUtilities/getPageInitialiser.js
```diff
@@ -211,19 +212,28 @@ const stocktakesInitialiser = () => {
...
-  isAscending: true,
-  modalKey: '',
-  modalValue: null,
-  columns: getColumns(ROUTES.CUSTOMER_INVOICE),
-  getPageInfoColumns: getPageInfoColumns(ROUTES.CUSTOMER_INVOICE),
-});
+const stocktakeBatchInitialiser = stocktakeItem => {
+  const thisStoreNameId = UIDatabase.getSetting(SETTINGS_KEYS.THIS_STORE_NAME_ID);
+  const suppliers = UIDatabase.objects('Name').filtered(
+    '(isVisible == true AND isSupplier == true) AND id != $0 OR (type == "inventory_adjustment")',
```
mark-prins: I'm not sure about the intention of that first set of brackets:

```suggestion
    '(isVisible == true AND isSupplier == true AND id != $0) OR (type == "inventory_adjustment")',
```

PR #2158, opened 2020-02-06
#2152 New insurances not showing

### LINKED ISSUE #2152: Adding insurance doesn't update the drop down
## Describe the bug

After adding an insurance policy, the policy is not added to the drop down

### To reproduce

As described in https://github.com/openmsupply/mobile/issues/2108#issuecomment-582219634

### Expected behaviour

N/A 

### Proposed Solution

N/A 

### Version and device info

N/A 

### Additional context

N/A

joshxg:
Fixes #2152 

## Change summary

- The bug fix was the fix in the `DropDrown` memoization dependency list
- Went off road and used `headerValue` prop in each of the drop downs which look liiiike:

image

image

image

## Testing

N/A

N/A


joshxg:
I agree - have had a little play - have to use android native styles and I've never really used em - can't get a bit of left padding so im giving up 🤣 

image

PR #2159, opened 2020-02-06
#2096 add cash register search bar

### LINKED ISSUE #2096: Add cash register search bar

Add search bar to cash register.

Add a search bar component for filtering transactions by name.

## Implementation

Nothing tricky, just a case of making a few tweaks to an existing page search bar component.

N/A.

## Additional context

See epic #1917.

wlthomson:
Fixes #2096.

## Change summary

Adds search bar component to cash register.

Filtering is persisted for payments and receipts (i.e. not reset on toggling).

Pros:
  - Guessing that a common use case would be to view payments and receipts for a particular supplier/customer. The search bar resetting could make this frustrating,

Cons:
  - Search bar filtering is reset on toggle on other pages, not following this pattern is bad UX (rule of consistency).

@joshxg @mark-prins @Chris-Petty @diegoevangelisti Thoughts?

## Testing

- [ ] Can filter search bar transactions by name.
- [ ] Filtering is persisted when toggling payments/receipts.

N/A.


src/pages/dataTableUtilities/reducer/tableReducers.js
```diff
@@ -39,15 +40,37 @@ export const filterData = (state, action) => {
...
   const searchTerm = payload.searchTerm?.trim();
 
-  // Indicator filtering is performed on component re-render.
-  if (usingIndicators && showIndicators) {
-    return {
-      ...state,
-      searchTerm,
-    };
+  // Data filtering may be deferred to state-to-prop selectors.
+  switch (route) {
```
joshxg: wouldn't creating a new action and reducer case be a bit cleaner here?

src/widgets/SimpleLabel.js
```diff
@@ -80,6 +86,8 @@ SimpleLabel.defaultProps = {
   label: '',
   labelAlign: 'left',
   textAlign: 'left',
+  labelBackground: 'transparent',
+  textBackground: 'transparent',
```
joshxg: I think we should just use `PageInfo` rather than simple label here, just like other pages?

image


Also since we may need to show multiple balances i.e. for each payment types, it'll be nice using `PageInfo` I think

PR #2172, opened 2020-02-07
#2162 fix cash register not filtering reasons

### LINKED ISSUE #2162: Cash transactions not using the specific transaction cash reasons
## Describe the bug

There is a specific type of reason which should be used for cash in reasons

image

### To reproduce

Dispensary development

### Expected behaviour

As above

### Proposed Solution

Create a `UIDatabase` case for this specific type and use that rather than `Options`

### Version and device info

Dispensary development

### Additional context

Dispensary development

wlthomson:
Fixes #2162.

## Change summary

Fixes cash register reason filtering.

## Testing

- [ ] Cash out transaction reasons only display `Options` records with type `"newCashOutTransaction"`.

N/A.

### DIFF
--- a/src/widgets/modals/CashTransactionModal.js
+++ b/src/widgets/modals/CashTransactionModal.js
@@ -42,7 +42,7 @@ export const CashTransactionModal = ({ onConfirm }) => {
   const [isDescriptionModalOpen, setIsDescriptionModalOpen] = useState(false);
 
   const names = useMemo(() => UIDatabase.objects('CashTransactionName'));
-  const reasons = useMemo(() => UIDatabase.objects('Options'));
+  const reasons = useMemo(() => UIDatabase.objects('CashTransactionReason'));
   const type = useMemo(
     () => (isCashIn ? CASH_TRANSACTION_TYPES.CASH_IN : CASH_TRANSACTION_TYPES.CASH_OUT),
     [isCashIn]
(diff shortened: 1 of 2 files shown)


src/widgets/modals/CashTransactionModal.js
```diff
@@ -42,7 +42,7 @@ export const CashTransactionModal = ({ onConfirm }) => {
   const [isDescriptionModalOpen, setIsDescriptionModalOpen] = useState(false);
 
   const names = useMemo(() => UIDatabase.objects('CashTransactionName'));
-  const reasons = useMemo(() => UIDatabase.objects('Options'));
+  const reasons = useMemo(() => UIDatabase.objects('CashTransactionReason'));
```
joshxg: this needs a dependency list, and the one above - `[]`

wlthomson: Good catch, de740a0e079c65cdf001008e3d98583422ff2a7e.

joshxg: https://github.com/openmsupply/mobile/blob/2ca29f7680485d3922950388f263c7b7b78c0477/src/widgets/modals/CashTransactionModal.js#L63-L64 

https://github.com/openmsupply/mobile/blob/2ca29f7680485d3922950388f263c7b7b78c0477/src/widgets/modals/CashTransactionModal.js#L66-L83

Maybe more that I didn't see !

PR #2197, opened 2020-02-10
Fix zero amount cash transactions

### LINKED ISSUE #2193: Do not allow cash transactions for zero amount

Cash transactions can currently be made for an amount of `0.00`.

Do not allow a cash in/cash out to be confirmed for `0.00`.

## Implementation

Update `isValidTransaction` conditional to require amounts larger than zero.

N/A.

## Additional context

N/A.

wlthomson:
Fixes #2193.

## Change summary

Disable `Confirm` button for transactions with zero amounts.

## Testing

- [ ] Enter an amount of `0.00` for a new cash transaction. The confirm button should be greyed out and unclickable.

N/A.


src/widgets/modals/CashTransactionModal.js
```diff
@@ -48,12 +48,10 @@ export const CashTransactionModal = ({ onConfirm }) => {
...
   );
 
-  const isValidTransaction = useMemo(() => !!name && !!amount && (isCashIn || !!reason), [
-    name,
-    amount,
-    reason,
-    isCashIn,
-  ]);
+  const isValidTransaction = useMemo(
+    () => !!name && !!amount && amount > 0 && (isCashIn || !!reason),
```
joshxg: if amount was `0` then `!!amount == false` and this expression is the same as before, no?

wlthomson: It explicitly coerces the `amount` to an `int`. Not a nice solution, but under time constraints does the job.

wlthomson: It explicitly coerces the `amount` to an `int`. Not a nice solution, but under time constraints does the job.

wlthomson: 24ccda1 (temporary fix, needs a refactor asap).

PR #2207, opened 2020-02-11
#2203 cant select item directions

### LINKED ISSUE #2203: Can't select an ItemDirection
## Describe the bug

`ItemDirection`s in the drop down aren't selectable 

### To reproduce

Dispensing development bug

### Expected behaviour

Dispensing development bug

### Proposed Solution

Dispensing development bug

### Version and device info

Dispensing development bug

### Additional context

Dispensing development bug

joshxg:
Fixes #2203 

## Change summary

- Couldnt set item directions when quantity of an item was 0 - as it had no batches, and the note was being set on the batches!
- Only do abbreviation expansion after a space
- Small UI update

## Testing

N/A

N/A


src/widgets/PrescriptionCartRow.js
```diff
@@ -116,7 +116,7 @@ PrescriptionCartRowComponent.propTypes = {
   itemName: PropTypes.string.isRequired,
   totalQuantity: PropTypes.number.isRequired,
   itemCode: PropTypes.string.isRequired,
-  totalPrice: PropTypes.string.isRequired,
+  price: PropTypes.string.isRequired,
```
wlthomson: A bit confused by this, is this name change related to the main PR?

PR #2238, opened 2020-02-12
#2232 cash register balance filter payment types

### LINKED ISSUE #2232: Cash register balance should only include cash transaction totals

The cash register balance currently does not filter transactions by payment type. This differs from desktop behaviour which only sums transactions with cash payment types.

Filter transactions by payment type before summing totals.

## Implementation

See above.

N/A.

## Additional context

N/A.

wlthomson:
Fixes #2232.

## Change summary

Makes mobile cash register behaviour consistent with desktop.

## Testing

- [ ] Cash register balance is only effected by payments and receipts of `paymentType.code=='cash'`.

N/A.

### DIFF
--- a/dev_scripts/upload.sh
+++ b/dev_scripts/upload.sh
@@ -7,7 +7,7 @@
 
 BUGSNAG_URL=https://upload.bugsnag.com/react-native-source-map
 BUGSNAG_KEY=16a680e189b1e5f03f28665870f1401f
-APP_VERSION=4.0.0-rc3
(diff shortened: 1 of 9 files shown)


wlthomson:
> I believe the idea on desktop was to have a drop down with a list of payment types to filter by, and the balance updates for each payment type. This'll do for now?

Yeah I believe that's correct. The thinking here was that the cash register is for "cash", so by default you'll probably want to see the cash balance. Next step is to add the dropdown to bring it into full compliance with desktop 👍 .

PR #2245, opened 2020-02-12
Add cash register page info columns

### LINKED ISSUE #2244: Update cash register to use page info component

See above.

Display cash register balance with `PageInfo` component .

## Implementation

Add a `selectCashRegisterInfoColumns` selector, pass columns directly as props to `CashRegisterPage`.

N/A.

## Additional context

N/A.

wlthomson:
Fixes #2244.

## Change summary

Tidies up the cash register UI and makes consistent with other pages.

## Testing

No functional changes.

N/A.

### DIFF
--- a/src/pages/CustomerRequisitionPage.js
+++ b/src/pages/CustomerRequisitionPage.js
@@ -94,6 +94,8 @@ export const CustomerRequisition = ({
     isFinalised,
   ]);
 
+  console.log(pageInfoColumns);
+
   const getCallback = useCallback(colKey => {
     switch (colKey) {
       case 'suppliedQuantity':
--- a/src/selectors/cashRegister.js
+++ b/src/selectors/cashRegister.js
@@ -4,6 +4,7 @@
  */
 
 import { createSelector } from 'reselect';
+import { pageInfoStrings } from '../localization';
 import currency from '../localization/currency';
 import { CASH_TRANSACTION } from '../utilities/modules/dispensary/constants';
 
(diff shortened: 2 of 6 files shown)


src/pages/CustomerRequisitionPage.js
```diff
@@ -94,6 +94,8 @@ export const CustomerRequisition = ({
     isFinalised,
   ]);
 
+  console.log(pageInfoColumns);
```
joshxg: !

src/selectors/cashRegister.js
```diff
@@ -74,3 +75,12 @@ export const selectBalance = createSelector(
...
 );
+
+export const selectCashRegisterInfoColumns = createSelector([selectBalance], currentBalance => [
+  [
+    {
+      title: `${pageInfoStrings.current_balance}:`,
+      info: currentBalance,
+    },
+  ],
+]);
```
joshxg: you reckon put all the page info sets in selectors? 

The current way makes a lot of the page info columns reuseable

wlthomson: Just for a quick fix, this ended up being a bit quicker. I think they can be mixed though (still use the reusability we have, but derive the info columns from state using selectors). Not super worried, but its nice to shift as much to selectors as possible and the current way relies on storing a function in state, which I think we are trying to avoid?

joshxg: yeah you're right that isn't great !

PR #2279, opened 2020-02-12
#2218 SupplierCredit crash

### LINKED ISSUE #2218: App crash - return to supplier
## Describe the bug

On confirmation of a supplier return the app crashes

### To reproduce

Raised during UAT - the tablet demoing is syncing to demoivory.msupply.org using the zambakro store.

1. Go to a finalised supplier invoice
2. Return
3. Enter quantities and click **Confirm**
4. App closes

### Version and device info

- App version: 4.0.0-RC4
- Tablet model: Lenovo Tab 4 10
- OS version: Android

joshxg:
Fixes #2218 

## Change summary

- Attempt #2!
- Pushing returns the count.. not the array.. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/push :) heheehe.....

## Testing

- [ ] Creating a `SupplierCredit` where there are more than one batch being returned to the same supplied does not crash the app
- [ ] Creating a `SupplierCredit` from a `SupplierInvoice`, where all batches are having all amounts returned does not crash the app

Where to start...........

### DIFF
--- a/src/actions/SupplierCreditActions.js
+++ b/src/actions/SupplierCreditActions.js
@@ -41,7 +41,7 @@ const create = () => (dispatch, getState) => {
     const { id: supplierId } = itemBatch.itemBatch?.supplier || {};
     const suppliersGroup = groupings[supplierId];
 
-    if (suppliersGroup) return { ...groupings, [supplierId]: suppliersGroup.push(itemBatch) };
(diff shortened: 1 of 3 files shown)


joshxg:
@mark-prins 

It should be good for any situation now!

PR #2321, opened 2020-02-15
#2273 Missing TransactionBatch when syncing

### LINKED ISSUE #2273: Sync: While syncing from tablet got error
## Describe the bug
As per client email in below link, while syncing the stock take got an error.
https://support.msupply.org.nz/cerb6/profiles/ticket/MB-6183-WP/conversation
### To reproduce
For now we are unable to reproduce if we get any clue then we will update.

### Expected behaviour

Must work properly

### Proposed Solution

### Version and device info

- App version:3.2.1
- Tablet model:
- OS version:

joshxg:
Fixes #2273 

## Change summary

- Steps to reproduce the error:

1. Create a Stocktake with > 1 line.
2. Go to the batch editing of a line, and add a batch.
3. Edit the quantity of the batch.
4. Leave the Stocktake and Sync.
5. Go back to the Stocktake, and go to the batch you edited previously
6. ONLY edit the expiry date [NOTE: This cell saves on LOSING FOCUS]
7. Leave the Stocktake and go sync again.

Why:

- Used the wrong action for updating the `ExpiryDate`. All this did was make the `database.save()` method, use `database.save("TransactionBatch", ...)` rather than `database.save("StocktakeBatch", ...)`. 
- When `.save()` is called, we trigger the `SyncQueue` event handler for the pass parameters (In this instance, `recordType = "TransactionBatch`, and the actual record is the `StocktakeBatch`. 
- The `SyncQueue` first looks for (any) record in the `SyncOut` queue which has the `ID` of the object we are trying to save [Note: Doesn't look for TYPE + ID, JUST ID)
- If it finds it, it updates it - otherwise, it creates a `SyncOut`.
- If there is already a `StocktakeBatch` record in the `SyncOut` for the `StocktakeBatch` we edited the date of, then it would just update it (i.e. after creation, after editing quantity).
- Otherwise, when we revisit after syncing, there is no `SyncOut`, so we create a new one where the `RecordType` is `TransactionBatch`
- Then, when we sync, that is where we check both ID AND TYPE - and we can't find the `TransactionBatch` with the ID, as it is really a `StocktakeBatch`

## Testing

Using the reproduction steps above - make sure you can recreate the problem on the current MASTER. Then, try the steps again on this branch and we can ensure it does not happen again.

Left the error throwing as even though this may really hurt some data in the future by completely stopping sync, I think this is a good case of error throwing. If we did not have it we wouldn't have found this for a long time, potentially which could have potentially caused some more unintended consequences!

The bugsnag adding is ugly, yeah

### DIFF
--- a/src/sync/outgoingSyncUtils.js
+++ b/src/sync/outgoingSyncUtils.js
@@ -17,7 +17,7 @@ import {
   TRANSACTION_BATCH_TYPES,
   TRANSACTION_TYPES,
 } from './syncTranslators';
-import { CHANGE_TYPES } from '../database';
+import { UIDatabase, CHANGE_TYPES } from '../database';
 import { SETTINGS_KEYS } from '../settings';
 
 const { THIS_STORE_ID, SYNC_URL, SYNC_SITE_NAME } = SETTINGS_KEYS;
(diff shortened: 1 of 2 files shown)


wlthomson:
### Testing 

- [x] Error is not thrown on step 7.

PR #2327, opened 2020-02-15
#2252 Maximum date to date picker

### LINKED ISSUE #2252: New Patient/Prescriber form -> DOB field doesn't alert when adding a future date
## Describe the bug

There is a warning message "Must be a date before today, in the format DD/MM/YYY" when incorrectly entering a DOB.  However, it only shows up if entering a date via text input -- if you select a future date from the Date Picker, no warning is given.

### To reproduce

1. Go to Dispensary
2. Select "New patient"
3. Select the Date Picker in the Date of Birth field.
4. Select a future date.
5. No Invalid message appears below the date entry.
6. Focus on the date.
7. Re-type the same date using the keyboard.
8. Now the warning appears.

### Expected behaviour

Future dates chosen from the Date Picker should give the same Invalid message as when typed in from the keyboard.

### Version and device info

- App version: 4.0.0 RC5
- Tablet model: Lenovo TB-X304L
- OS version: Android 7.1.1

### Additional context

Picked up during testing [here](https://github.com/openmsupply/mobile/issues/2106#issuecomment-585014049).

joshxg:
Fixes #2252 

## Change summary

- Adds use of the `maximumDate` to RNDatePicker

## Testing

- [ ] cannot set a future date for a patients DOB when creating or editing using the date picker

N/A

### DIFF
--- a/src/widgets/FormDateInput.js
+++ b/src/widgets/FormDateInput.js
@@ -111,6 +111,7 @@ export const FormDateInput = React.forwardRef(
               mode="date"
               display="spinner"
               value={pickerSeedValue}
+              maximumDate={new Date()}
             />
           )}
           <FormInvalidMessage isValid={isValid} message={invalidMessage} />


src/widgets/FormDateInput.js
```diff
@@ -111,6 +111,7 @@ export const FormDateInput = React.forwardRef(
               mode="date"
               display="spinner"
               value={pickerSeedValue}
+              maximumDate={new Date()}
```
wlthomson: NIT: this is going to run the constructor on every render of the `DateTimePicker`. Unlikely that the day will tick over while the user is editing, so think a good compromise is to recalculate the max date when the date picker is closed/opened.

 `const maximumDate = useMemo(() => new Date(), [datePickerOpen]);`

joshxg: On top of that would cause a re-render on `DateTimePicker` as it will be a new reference (I'm assuming `DateTimePicker` is memoized) - but technically none of this occurs if it's not open - and if it's open then it won't be re-rendering from `FormDateInput`! 

So the memoization would do nothing (it will run the constructor and render once per update of `datePickerOpen`)

PR #2328, opened 2020-02-16
#2237 Number of Prescribers as a parameter to selectors

### LINKED ISSUE #2237: When you add your first prescriber, it does not appear in the list
## Describe the bug

When you add your first prescriber, it does not appear in the list immediately. I had to search for it (it didn't show in the search) and then deleted my search term and it appeared. 

I have no screenshots because I can't delete prescribers and reproduce, sorry! 

It seems to work fine after the first one is added, ie. when adding the 2nd prescriber. 

### To reproduce

1. On a datafile with no prescribers, select a patient and click 'Dispense'
2. On the window with an empty prescriber listbox, click 'New Prescriber'
3. Add a prescriber and return to the listbox window - it will still be empty

### Expected behaviour

I think it should appear in the listbox when added 

### Proposed Solution

### Version and device info

- App version:4.0.0 RC4
- Tablet model:
- OS version:

joshxg:
Fixes #2237 

## Change summary

- Make the number of prescribers a parameter to the selector so that it recalculates if it changes

## Testing

- [ ] Adding a prescriber in the `PrescriberSelect` component refreshes the list of prescribers

N/A

### DIFF
--- a/src/selectors/prescriber.js
+++ b/src/selectors/prescriber.js
@@ -21,14 +21,22 @@ export const selectSearchTerm = ({ prescriber }) => {
   return searchTerm;
 };
 
-export const selectFilteredPrescribers = createSelector([selectSearchTerm], searchTerm => {
-  const [lastName, firstName] = searchTerm.split(',').map(name => name.trim());
+export const selectNumberOfPrescribers = () => {
+  const numberOfPrescribers = UIDatabase.objects('Prescriber').length;
+  return numberOfPrescribers;
+};
 
-  const queryString = 'lastName BEGINSWITH[c] $0 AND firstName BEGINSWITH[c] $1';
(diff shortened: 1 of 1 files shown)


src/selectors/prescriber.js
```diff
@@ -21,14 +21,22 @@ export const selectSearchTerm = ({ prescriber }) => {
   return searchTerm;
 };
 
-export const selectFilteredPrescribers = createSelector([selectSearchTerm], searchTerm => {
-  const [lastName, firstName] = searchTerm.split(',').map(name => name.trim());
+export const selectNumberOfPrescribers = () => {
+  const numberOfPrescribers = UIDatabase.objects('Prescriber').length;
+  return numberOfPrescribers;
```
wlthomson: Can make this a one-liner?

joshxg: 89fp90

PR #2350, opened 2020-02-18
#2335 Filtering by having a supplier

### LINKED ISSUE #2335: Only show batches with a supplier for supplier credits

For `SupplierCredit`s batches belonging to `invad`, `build` etc show

Only show batches from `internal` suppliers (?)

## Implementation

N/A

N/A

## Additional context

N/A

joshxg:
Fixes #2335 

## Change summary

- Filters batches to ensure they have a supplier. Query is probably a bit much but I'm not 100% on the rules of the 'misc names' yet - will find out!

## Testing

N/A

N/A

### DIFF
--- a/src/reducers/SupplierCreditReducer.js
+++ b/src/reducers/SupplierCreditReducer.js
@@ -68,11 +68,11 @@ export const SupplierCreditReducer = (state = initialState(), action) => {
 
       const item = UIDatabase.get('Item', itemId);
 
-      const query = 'numberOfPacks > 0 && supplier != null && supplier.isVisible == true';
+      const batchesWithASupplierAndStock = 'numberOfPacks > 0 && supplier.isVisible == true';
 
       return {
         ...state,
-        batches: mapBatchToObject(item?.batches?.filtered(query)),
+        batches: mapBatchToObject(item?.batches?.filtered(batchesWithASupplierAndStock)),
         open: true,
         item,
         type: 'supplierCreditFromItem',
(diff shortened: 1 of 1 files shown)


src/reducers/SupplierCreditReducer.js
```diff
@@ -68,9 +68,11 @@ export const SupplierCreditReducer = (state = initialState(), action) => {
 
       const item = UIDatabase.get('Item', itemId);
 
+      const query = 'numberOfPacks > 0 && supplier != null && supplier.isVisible == true';
```
wlthomson: Is the `supplier != null` needed here (`supplier.isVisible == true` only if `supplier != null`)?

src/reducers/SupplierCreditReducer.js
```diff
@@ -68,9 +68,11 @@ export const SupplierCreditReducer = (state = initialState(), action) => {
...
+
       return {
         ...state,
-        batches: mapBatchToObject(item?.batches?.filtered('numberOfPacks > 0') ?? []),
+        batches: mapBatchToObject(item?.batches?.filtered(query)),
```
wlthomson: Think where possible it would be good to use more "self-documenting" identifiers for query string variables? E,g, `queryNonEmptyVisibleBatches`?

src/reducers/SupplierCreditReducer.js
joshxg: wowow 
108b0a6 
embares

src/reducers/SupplierCreditReducer.js
joshxg: went with this I don't like either 🤣 
108b0a6

PR #2356, opened 2020-02-19
#2355 disable cash register if no payment types defined

### LINKED ISSUE #2355: Cash register throws error if no payment types defined
## Describe the bug

If no payment types are defined, the cash register throws an error. Arguably is a config issue, but could theoretically occur due to an incomplete sync etc. In any case, best to play it safe and add some defensive code. 

### To reproduce

1. Setup a datafile with no payment types defined.
2. Enable mobile cash register (`usesCashRegisterModule`).
3. Open cash register on mobile.

### Expected behaviour

My current thinking is that the cash register is pretty tightly bound to payment types, so probably shouldn't appear as a module unless there is at least one payment type defined?

### Proposed Solution

Small change to `modulesReducer`:

```
const usingPaymentTypes = UIDatabase.objects('PaymentType').length > 0;

const usingCashRegister = checkModule(SETTINGS_KEYS.CASH_REGISTER_MODULE) && usingPaymentTypes;
```

### Version and device info

- App version: v4.0.0
- Tablet model: Emulator
- OS version: API 29.

### Additional context

N/A.

wlthomson:
Fixes #2355.

## Change summary

Disables cash register if no payment types defined.

## Testing

- [ ] If `usesCashRegisterModule` is `true` but no payment types defined, cash register button is not displayed.
- [ ] If `usesCashRegisterModule` is `true` and at least one payment types defined, cash register button is displayed.

Possibly needed for prescription functionality also?

### DIFF
--- a/src/reducers/ModulesReducer.js
+++ b/src/reducers/ModulesReducer.js
@@ -26,20 +26,20 @@ import { SYNC_TRANSACTION_COMPLETE } from '../sync/constants';
 const checkModule = key => UIDatabase.getSetting(key).toLowerCase() === 'true';
 
 const initialState = () => {
+  const usingInsurance = UIDatabase.objects('InsuranceProvider').length > 0;
+  const usingPaymentTypes = UIDatabase.objects('PaymentType').length > 0;
+  const usingPrescriptionCategories = UIDatabase.objects('PrescriptionCategory').length > 0;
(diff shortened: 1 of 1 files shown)


wlthomson:
> I personally don't think payment types should be IC specific, personally - but since they seem to be..
> 
> tada

Hmm, I think the idea is to limit the changes just to IC as they are the target for the next release (also at the moment there isn't a UI to edit them). Either way though, no payment types = no cash register :disappointed: .

PR #2361, opened 2020-02-20
#2340 prescribers editable only from this store

### LINKED ISSUE #2340: Only allow editing of prescribers from the mobile store

Related: #2339 

Potentially allowing prescribers from other stores into the database. Need to limit editing to only those which are from the mobile store.

- Disable the editing form for a prescriber not from the mobile store.
- Still allow opening the edit modal, to be able to see all details
- In outgoing sync, if the prescriber is not from the store, don't sync it out.

## Implementation

As above

N/A

## Additional context

N/A

joshxg:
Fixes #2340 

## Change summary

- Adds `fromThisStore` to `Prescriber` following the same pattern as `NameStoreJoin`
- Disables editing of a prescriber if the prescriber is not `fromThisStore`
- In the unlikely event of trying to sync a prescriber out: don't!

## Testing

- [ ] Cannot edit a prescriber that hasn't been created on the mobile store

N/A

### DIFF
--- a/src/selectors/prescriber.js
+++ b/src/selectors/prescriber.js
@@ -45,5 +45,13 @@ export const selectSortedPrescribers = createSelector(
 
 export const selectPrescriberModalOpen = ({ prescriber }) => {
   const { isCreatingPrescriber, isEditingPrescriber } = prescriber;
+
   return isCreatingPrescriber || isEditingPrescriber;
+};
+
+export const selectCanEditPrescriber = ({ prescriber }) => {
+  const { currentPrescriber } = prescriber;
+  const { fromThisStore } = currentPrescriber ?? {};
+
+  return fromThisStore ?? true;
 };
--- a/src/widgets/FormInputs/FormDateInput.js
+++ b/src/widgets/FormInputs/FormDateInput.js
@@ -19,6 +19,23 @@ import { CircleButton } from '../CircleButton';
 import { FormLabel } from './FormLabel';
 import { FormInvalidMessage } from './FormInvalidMessage';
 
+/**
+ * Form input for entering dates. Renders a TextInput as well as a button, opening a
+ * native Date picker.
+ *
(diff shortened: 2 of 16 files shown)


src/selectors/prescriber.js
```diff
@@ -45,5 +45,13 @@ export const selectSortedPrescribers = createSelector(
...
   const { isCreatingPrescriber, isEditingPrescriber } = prescriber;
+
   return isCreatingPrescriber || isEditingPrescriber;
 };
+
+export const selectCanEditPrescriber = ({ prescriber }) => {
+  const { currentPrescriber } = prescriber;
+  const { fromThisStore } = currentPrescriber ?? {};
+
+  return fromThisStore ?? true;
```
wlthomson: Shouldn't this default to `false`?

src/widgets/FormInputs/FormDateInput.js
```diff
@@ -10,15 +10,32 @@ import moment from 'moment';
...
+ * Form input for entering dates. Renders a TextInput as well as a button, opening a
+ * native Date picker.
+ *
+ * @prop {Date}   value                    A date or moment - The initial date.
+ * @prop {Bool}   isRequired               Indicator whether this date is required.
+ * @prop {Bool}   isDisabled               Indicator whether this input is disabled.
+ * @prop {Func}   onValidate               Callback to validate the input.
+ * @prop {Func}   onChangeDate             Callback for date changes[called only with a valid date].
+ * @prop {Func}   onSubmit                 Callback after submitting a date from text.
+ * @prop {String} label                    Label for the input
```
wlthomson: Nit: `.` :laughing:

src/selectors/prescriber.js
joshxg: default to true for when the modal is opened and there isn't actually a prescriber (creation)

PR #2370, opened 2020-02-21
#2341 only allow editing of patients from this store

### LINKED ISSUE #2341: Only allow editing of patients from the mobile store

Allowing patients from other stores to be synced to mobile. Disallow editing of patient records from other stores.

- Disable the editing form for a patient not from the mobile store.
- Still allow opening the edit modal, to be able to see all details
- In outgoing sync, if the patient is not from the store, don't sync it out.
- Potentially add a `fromThisStore` field to `name`

## Implementation

## Additional context

joshxg:
Fixes #2341 

## Change summary

- Makes patients editable only if theyre from the mobile store.
- Rejects any patients not from the mobile store in incoming sync.

## Testing

- [ ] Patients are not editable if they're not created in the mobile store
- [ ] Patients are editable which have been created in this mobile store

N/A

### DIFF
--- a/src/selectors/patient.js
+++ b/src/selectors/patient.js
@@ -45,3 +45,10 @@ export const selectPatientModalOpen = ({ patient }) => {
   const { viewingHistory } = patient;
   return [isCreating || isEditing, viewingHistory];
 };
+
+export const selectCanEditPatient = ({ patient }) => {
+  const { currentPatient } = patient;
+  const { thisStoresPatient } = currentPatient ?? {};
+
+  return thisStoresPatient ?? true;
+};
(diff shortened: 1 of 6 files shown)


src/selectors/patient.js
```diff
@@ -45,3 +45,10 @@ export const selectPatientModalOpen = ({ patient }) => {
   const { viewingHistory } = patient;
   return [isCreating || isEditing, viewingHistory];
 };
+
+export const selectCanEditPatient = ({ patient }) => {
+  const { currentPatient } = patient;
+  const { thisStoresPatient } = currentPatient ?? {};
```
wlthomson: `const { thisStoresPatient = false } = currentPatient ?? {};`

src/selectors/patient.js
```diff
@@ -45,3 +45,10 @@ export const selectPatientModalOpen = ({ patient }) => {
...
+
+  return thisStoresPatient ?? true;
```
wlthomson: Do we want to default to `true` here? Shouldn't we be defensive and default to `false`?

PR #2373, opened 2020-02-21
#2249 Shortcutting when reusing a reused number

### LINKED ISSUE #2249: Crashing when trying to reuse an invoice number
## Error in mSupply Mobile

**Error** in **MainActivity**
Sequence customer_invoice_serial_number already reusing 567

[View on Bugsnag](https://app.bugsnag.com/sustainable-solutions-nz-ltd/msupply-mobile/errors/5e436f6e0dc0300019ee9a94?event_id=5e436f6e0055feefa1d50000&i=gh&m=ci)

## Stacktrace

    src/database/DataTypes/NumberSequence.js:67 - value
    src/database/utilities/createRecord.js:436 - createNumberToReuse
    src/database/utilities/createRecord.js:806 - createRecord
    src/database/DataTypes/NumberSequence.js:52 - value
    src/database/utilities/createRecord.js:26 - number
    src/database/DataTypes/Transaction.js:52 - value
    src/database/UIDatabase.js:231 - value
    src/navigation/actions.js:86 - 
    src/database/UIDatabase.js:247 - value
    src/navigation/actions.js:84 - 
    src/pages/MenuPage.js:246 - toDispensary

[View full stacktrace](https://app.bugsnag.com/sustainable-solutions-nz-ltd/msupply-mobile/errors/5e436f6e0dc0300019ee9a94?event_id=5e436f6e0055feefa1d50000&i=gh&m=ci)

joshxg:
Fixes #2249 

## Change summary

- Edge case when switching an existing desktop site to a mobile site 

## Testing

N/A

N/A

### DIFF
--- a/src/database/utilities/createRecord.js
+++ b/src/database/utilities/createRecord.js
@@ -426,6 +426,13 @@ const createNumberSequence = (database, sequenceKey) =>
  * @param  {number}          number
  */
 const createNumberToReuse = (database, numberSequence, number) => {
+  const alreadyReusing = database
+    .objects('NumberToReuse')
+    .query('numberSequence == $0 && number == $1', numberSequence, number);
+
+  // If we are already reusing this number - shortcut return
+  if (alreadyReusing.length) return;
+
   const numberToReuse = database.create('NumberToReuse', {
     id: generateUUID(),
     numberSequence,


src/database/utilities/createRecord.js
```diff
@@ -426,6 +426,13 @@ const createNumberSequence = (database, sequenceKey) =>
  * @param  {number}          number
  */
 const createNumberToReuse = (database, numberSequence, number) => {
+  const alreadyReusing = database
```
wlthomson: Hmm, might be clearer to:

const `isNumberReused = database.objects(...).query(...).length > 0;`

andreievg:
@joshxg, what's the use case for the error in the issue.
Looks like it happens when transaction is deleted, and we are trying to add number to be reused (that is already in the reused table). But how come we have transaction number in the reused numbers already ? Is it sync issue ?  (in that case maybe we should be fixing it in sync)

joshxg:
@andreievg 

Has been coming up a bit - i wrote a big thing in a reply but looks like I didn't comment and now i've lost it :( sorry!

Basically: If you transfer a desktop site to be a mobile site, there are transactions which have been created, but there are no number records because they are locally on desktop sites. So deleting transactions before you start using numbers..

I don't think this is a very good solution, but apparently having a few mistakes in invoice numebrs etc isn't too big of a deal?

andreievg:
@joshxg 
> Basically: If you transfer a desktop site to be a mobile site, there are transactions which have been created, but there are no number records because they are locally on desktop sites. So deleting transactions before you start using numbers..

I am a little bit confused (although i trust that the code would resolve this issue), just want to make sure there is no side effects. 

https://github.com/openmsupply/mobile/blob/cb7efd9fc28ca92fe57fb184399d0a4a0b05accf/src/database/DataTypes/NumberSequence.js#L65-L70

The error is thrown when reuse number already exists, right ? And you've mentioned there are no number records, thus the confusion.

joshxg:
@andreievg 

I believe it's from moving a store to and from being a desktop/mobile store [Has come up now from people testing on data files and making desktop sites mobile sites etc] and the difference between desktop syncing numbers and mobile not - i.e. mobile store has the reused number, syncs it, transactions made on desktop when you switch the site - they don't sync - create on mobile etc.. I suppose you could say a data file issue, but at the same time you should, as a user be able to switch between a site being mobile and desktop? I would be happy to leave this out as I think throwing errors is good. 

However it does make the app unusable and doesn't really let the user recover from it if it did occur.. hence the 'fix' - Justified to myself that since desktop doesn't sync numbers etc then if we 'reinit' desktop sites then the numbers get a bit messy. Let me know and I'm happy to close this and investigate further to find proper reproduction steps a bit later!

andreievg:
Ahh thanks for explanation. @joshxg 

This fix makes sense now.

I think our sync for desktop and mobile should align though, if there are reproduction steps, it would be great to make mSupply issue (no rush btw)

joshxg:
sweet - the main issue is that mobile has `[number]` and `[number_to_reuse]` in both incoming AND outgoing sync. Where as desktop satellites do NOT sync either of these - they are just local and NEVER sync to the server - I think possibly because of reinitialization and maybe mobile does that more??

PR #2407, opened 2020-02-27
#2380 stop syncing with incompatible servers

### LINKED ISSUE #2380: Limitations when interacting with an incompatible server

Some versions of mobile are incompatible with older versions.

1. *Incompatible schemas*: For example if there exists a table `[X]` on desktop V5.0 - and in a new version of desktop V5.1 introduces field `[X]fieldn` and a mobile site tries to send this record with the `[X]fieldn` (undefined, or defined) - the record will be rejected. Since push occurs first - sync then stops.
2. *Differences in values / general backwards incompatibilities*: For example - if `[transact]total` starts to get calculated differently in a new V5.1 than V5.0 - then the versions are not compatible.

note: As we use semantic versioning, this means a major version pump

Ideally: Shouldn't even be able to install the new mobile version.
Practically: Stop sync from working if the server is not compatible. Bug snag then error.

## Implementation

1. Send a field in responses from the server for example [here](https://github.com/sussol/msupply/blob/18548c9f92a6ae7d859cc1aa142de60fc8c0018f/vc_methods/REST_get_records.txt#L2235-L2236) which gives the servers version. The mobile version then sets a flag - allowing/disallowing sync to occur.
    - Pros:
        - Simple to implement
    - Cons:
        - Not a very nice solution!
        - Adds overhead to all(?) fetches [for example: if you only do it on login - you might not be connected to the internet when logging in]

2. Create an endpoint(s) `/version` which could tell you the servers version, incompatibilities, compatible versions etc (I dunno)
    - Pros:
        - Very extensible
        - Could use to query all servers on a server so we can get an up-to-date picture etc
    - Cons:
        - Not too quick
        - Possibly functionality that nicole's hub would be doing?

3. Bump the rest api end point
    - Pros:
        - RESTful! - semantically correct
        - Don't need to send an auth request on every request
    - Cons:
        - Time [Will this actually take extra time?]

4. Mobile sends it's version - server validates
    - Pros:
        - Fast to implement
        - Reduce the number of queries
    - Cons:
        - Desktop tracks what versions of other applications it is compatible with

## Alternatives

- Make the sync API (either mobile or desktop) more backwards compatible. It might be easier for desktop to just integrate a record which has 'extra' fields. However this still doesn't solve the case of 'general' backwards incompatibilities.

## Additional context

Specifically, this is for mobile version `4.0.0` being incompatible with `<4.09` on desktop. So, if any desktop work needs to be done, it would need to be done very quickly.

joshxg:
Fixes #2380 

## Change summary

- Stops syncing with servers which are not compatible!

## Testing

- [ ] Cannot sync version `4.0.0` mobile with a server < `4.09` [current develop at the time of this pr]

N/A

### DIFF
--- a/src/authentication/SyncAuthenticator.js
+++ b/src/authentication/SyncAuthenticator.js
@@ -62,7 +62,7 @@ export class SyncAuthenticator {
     const [majorAppVersion] = appVersion.split('.');
     const { ServerVersion: serverVersion } = responseJson;
 
-    if (serverVersion >= SERVER_COMPATIBILITIES[majorAppVersion]) {
+    if (serverVersion < SERVER_COMPATIBILITIES[majorAppVersion]) {
       throw new Error(authStrings.incompatible_server);
     }
 
--- a/src/authentication/SyncAuthenticator.js
+++ b/src/authentication/SyncAuthenticator.js
@@ -44,13 +44,13 @@ export class SyncAuthenticator {
    *                                       any error message as a second parameter
    * @return {none}
    */
-  async authenticate(serverURL, username, password) {
+  async authenticate(serverURL, username, password, hashedPassword) {
     if (serverURL.length === 0) throw new Error('Enter a server URL');
     if (username.length === 0) throw new Error('Enter the sync site username');
-    if (password.length === 0) throw new Error('Enter the sync site password');
+    if (!password && !hashedPassword) throw new Error('Enter the sync site password');
 
     // Hash the password.
-    const passwordHash = hashPassword(password);
+    const passwordHash = hashedPassword ?? hashPassword(password);
 
     const authURL = `${serverURL}${AUTH_ENDPOINT}`;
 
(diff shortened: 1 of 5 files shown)


joshxg:
Have made https://github.com/openmsupply/mobile/issues/2408  to make the checking better. I.e. to not have to add a major with a lower bound for every new major release, even if it is not different to the last one.. although that means you shouldn't really be releasing a major sooo.....

joshxg:
image

joshxg:
@alainsussol 

If possible: 

- `"The mSupply Server you're trying to sync with is not compatible with your current version. Please contact your administrator."`
- `"Le serveur mSupply avec lequel vous essayez de vous synchroniser n'est pas compatible avec votre version actuelle. Veuillez contacter votre administrateur."`


This will show up when trying to sync with an old mSupply server. That french makes sense in my area, what about yours?

alainsussol:
@joshxg 

The french translation is good if that was the question :)

PR #2455, opened 2020-03-03
#2433 fix insurance policy number editing

### LINKED ISSUE #2433: Insurance fields are incorrect
## Describe the bug

Insurance policies aren't quite right, after reading the wiki: https://sussol.net/wiki/doku.php/msupply:specifications:customers:insurance

- Can edit policy numbers after having been saved
- Policy numbers need to be unique as a full number per provider. Not just a unique personal number, which validation checks for currently

### To reproduce

1. Go to a prescription
2. Try to make an insurance policy

### Expected behaviour

As above

### Proposed Solution

N/A

### Version and device info

- App version: v4.0.0RC15
- Tablet model: N/A
- OS version: N/A

### Additional context

N/A

wlthomson:
Fixes #2433.

## Change summary

This prevents policy numbers from being edited after saved, which is a constraint currently enforced on desktop.

## Testing

- [ ] Create a new insurance policy and save. The policy number is editable.
- [ ] Reopen the previously saved insurance policy. The policy number is not editable.

Still to do:

- Lock down provider and policy type. 
- Prevent duplicate policy numbers.

### DIFF
--- a/src/utilities/formInputConfigs.js
+++ b/src/utilities/formInputConfigs.js
@@ -55,6 +55,7 @@ const FORM_INPUT_CONFIGS = seedObject => ({
     isRequired: true,
     label: formInputStrings.first_name,
     invalidMessage: formInputStrings.must_not_be_empty,
+    isEditable: true,
   },
   [FORM_INPUT_KEYS.LAST_NAME]: {
     type: 'text',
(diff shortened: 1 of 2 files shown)


src/utilities/formInputConfigs.js
```diff
@@ -55,6 +55,7 @@ const FORM_INPUT_CONFIGS = seedObject => ({
     isRequired: true,
     label: formInputStrings.first_name,
     invalidMessage: formInputStrings.must_not_be_empty,
```
mark-prins: I can't help thinking that `isReadOnly` would have been much easier!
Would default to `false` and so only two lines need to be added

wlthomson: Hmm, fair point. This is a bit of a quick fix and likely to be something revisited later anyway, so nothing finalised yet!

PR #2475, opened 2020-03-05
#2474 react navigation dynamic config

### LINKED ISSUE #2474: react-navigation update: Dynamic configuration
## The suggestion

After upgrading to the v5 `react-navigation` api the `StackNavigator` needs to be declared dynamically rather than staticly

**Why?**
It's far better!

N/A

## Additional context

For example: https://reactnavigation.org/docs/hello-react-navigation/

We need to declare the stack navigation like:

```
<NavigationContainer>
    <Stack.Navigator>
      <Stack.Screen name="MenuPage" component={HomeScreen} />
      <Stack.Screen name="SupplierRequisitionPage" component={HomeScreen} />
      <Stack.Screen name="SupplierRequisitionsPage" component={HomeScreen} />
      ....
    </Stack.Navigator>
</NavigationContainer>
```

We previously used something like [this](https://github.com/openmsupply/mobile/blob/f33d9d04b4680f3b1a89c5d1c56c3595892c49bd/src/mSupplyMobileApp.js#L311-L322):

```
<ReduxNavigator
    state={navigationState}
    dispatch={dispatch}
    screenProps={{
        database: UIDatabase,
        settings: Settings,
        currentUser,
        routeName: navigationState.routes[navigationState.index].routeName,
        runWithLoadingIndicator: this.runWithLoadingIndicator,
        isInAdminMode,
    }}
/>
```

where the state was controlled by redux and the props passed in `screenProps` are not really used anymore (might be somewhere, will have to double check them all).

However we also had mixed in the component `PageContainer` - which hooked into redux and passed extra props [here](https://github.com/openmsupply/mobile/blob/f33d9d04b4680f3b1a89c5d1c56c3595892c49bd/src/pages/PageContainer.js#L15-L27) and each page was actually wrapped in a `PageContainer` like [here](https://github.com/openmsupply/mobile/blob/f33d9d04b4680f3b1a89c5d1c56c3595892c49bd/src/pages/index.js#L42)

The `PageContainer` component tries to pass a bunch of props - I don't even know what most of them are - and they certainly aren't really used too much and can fetched from elsewhere, I believe.

So a recap: 
- We use a `StackNavigator` which is wrapped in a component which is hooked into redux which has the config (possible screens that can be navigated to) defined staticly. 
- Each screen is wrapped in a `PageContainer` component which passes various props to the 'actual'  screen/page component. 
- The goal of this issue is to define the config for `StackNavigator` dynamically - removing the need for the redux wrapper (it doesn't work anymore) and the `PageContainer` wrapper

joshxg:
#### BRANCHED FROM : #2469 

Fixes #2474 

## Change summary

- Gets navigating up and running again!
- Decided to do a middleware on the redux store- previously, the actions dispatched to the store would cause the navigator to navigate accordingly. This doesn't work anymore. For the smallest amount of changes, the middleware will trigger the correct navigation methods when the old actions are dispatched - essentially replicating the same behaviour
- Since this navigating is actually after the dispatching of the redux action which sets up the state I've added some defensive code to pages so that they are fine with essentially having undefined state (A lot less changes than I anticipated!)

## Testing

N/A

- This breaks a few things: Previously navigating would cause `mSupplyMobileApp` to re-render, which essentially re-renders the whole app. This made a lot of things 'work correctly'
- Namely the navigation bar: This will no longer re-render when navigating as it is currently at the same level as the navigator. This will be the next thing to fix
- Finalising is also broken, as the logic for that is extremely convoluted and depends on the current route etc etc. Will be fixed after the navigation bar.
- I don't like listing the 20 pages in a row like this in the render method, so keen to just have a mapping of an object of `routeName: component` to components


src/mSupplyMobileApp.js
```diff
@@ -308,8 +321,66 @@ class MSupplyMobileAppContainer extends React.Component {
...
+              name={ROUTES.STOCKTAKE_EDITOR}
+              component={StocktakeEditPage}
+            />
+
+            <MainStackNavigator.Screen name={ROUTES.DISPENSARY} component={DispensingPage} />
+            <MainStackNavigator.Screen name={ROUTES.PRESCRIPTION} component={PrescriptionPage} />
+
+            <MainStackNavigator.Screen name={ROUTES.REALM_EXPLORER} component={RealmExplorer} />
+            <MainStackNavigator.Screen name={ROUTES.SETTINGS} component={SettingsPage} />
+            <MainStackNavigator.Screen name={ROUTES.DASHBOARD} component={DashboardPage} />
```
wlthomson: Just due to the number of components here, might be nice to programmatically generate these, e.g.:

```
ROUTES.map(route => <MainStackNavigator.Screen name={route} component={COMPONENTS[route]} />
```

src/navigation/RootNavigator.js
```diff
@@ -0,0 +1,35 @@
...
+
+const navigate = (screenName, screenProps) =>
+  rootNavigatorRef.current?.navigate(screenName, screenProps);
+
+const replace = (screenName, screenProps) =>
+  rootNavigatorRef.current?.replace(screenName, screenProps);
+
+const goBack = () => rootNavigatorRef.current?.goBack();
+
+const canGoBack = () => !!rootNavigatorRef.current?.canGoBack();
```
wlthomson: Is the `!!` needed here?

src/pages/CustomerInvoicePage.js
```diff
@@ -200,9 +200,9 @@ export const CustomerInvoice = ({
 };
 
 const mapDispatchToProps = (dispatch, ownProps) => {
-  const { transaction } = ownProps || {};
-  const { otherParty } = transaction;
-  const hasMasterLists = otherParty.masterLists.length > 0;
+  const { transaction } = ownProps?.navigation?.route?.param || {};
+  const { otherParty } = transaction || {};
```
wlthomson: `??`.

lol

src/pages/CustomerRequisitionPage.js
```diff
@@ -307,9 +307,9 @@ const mapDispatchToProps = (dispatch, ownProps) =>
   getPageDispatchers(dispatch, ownProps, 'Requisition', ROUTES.CUSTOMER_REQUISITION);
 
 const mapStateToProps = state => {
-  const { pages } = state;
-  const { customerRequisition } = pages;
-  const { usingIndicators, showIndicators } = customerRequisition;
+  const { pages = {} } = state;
```
wlthomson: `state ?? {}`?

src/pages/SupplierRequisitionPage.js
```diff
@@ -387,9 +387,9 @@ const mapDispatchToProps = (dispatch, ownProps) => {
 };
 
 const mapStateToProps = state => {
-  const { pages } = state;
-  const supplierRequisition = pages[ROUTES.SUPPLIER_REQUISITION];
-  const { usingIndicators, showIndicators } = supplierRequisition;
+  const { pages = {} } = state;
```
wlthomson: See above.

src/mSupplyMobileApp.js
joshxg: See PR comment regarding this ! waiting a bit to see if there is something nicer that can be done, potentially

src/navigation/RootNavigator.js
joshxg: if it's undefined then you cant go back so a bool is nice!

src/pages/CustomerInvoicePage.js
joshxg: heh, everytime! will fix them up in the next PR

PR #2459, opened 2020-03-03
#2458 fix insurance policy type editing

wlthomson:
Fixes #2458.

## Change summary

Fixes #2458.

Branched off #2457.

## Testing

Steps to reproduce or otherwise test the changes of this PR:

- [ ] Create a new insurance policy. The policy type can be toggled.
- [ ] Reopen the saved insurance policy. The policy type cannot be toggled.

### Related areas to think about

Still to do:

- Prevent duplicate policy numbers.

Issue #2458, opened 2020-03-03
Disable editing policy type for saved policies


wlthomson:
## Is your feature request related to a problem? Please describe.

Related to #2433. Insurance policy type should not be editable after a policy is saved.

## Describe the solution you'd like

See above.

## Implementation

Add to changes in #2455.

## Describe alternatives you've considered

N/A.

## Additional context

See #2433.

joshxg:
Closed by #2459

nishaSussol:
Tested in mobile version 4.0.1

## Tests
- [x] Create a new insurance policy. The policy type can be toggled. **-PASS-**
- [x] Reopen the saved insurance policy. The policy type cannot be toggled. **-PASS-**

PR #2481, opened 2020-03-05
#2460 add duplicate policy number validation

### LINKED ISSUE #2460: Prevent duplicate policy numbers

All policies for a given provider should have distinct policy numbers. This is currently enforced on desktop but not on mobile.

When entering a new policy, the policy can only be saved if the policy number does not already exist for that provider.

## Implementation

Tidiest way of doing this at the moment is to share state between inputs by passing a `form` prop to `Form` components (e.g. `FormTextInput`). This can then be passed as an optional parameter to the validation function.

N/A.

## Additional context

N/A.

wlthomson:
Fixes #2460.

Branches off #2459.

## Change summary

Took a bit of time to get there, but a minor refactor with a bit of a workaround to make policy number validation consistent with desktop behaviour.

## Testing

- [ ] If a duplicate policy number is entered, both the personal and family form inputs provide a warning.
- [ ] If either personal or family form inputs are updated such that the policy number is no longer equal to an existing policy number, both personal and family form inputs warnings are no longer displayed.
- [ ] When testing the previous two tests, the personal AND family warnings are synchronised, i.e. either both display a warning, or neither.

Not the tidiest fix, I think may need a bit of refactoring when more time is available.


mark-prins:
Hm. I don't think that it fixes #2469 more likely #2458

wlthomson:
Update: had to do some rebasing and cherry-picking to rewrite the history to be based on `master` not `develop`. No code changes.

wlthomson:
@mark-prins Have addressed both your usage things. Will wait for you to give it another look before merging.

I agree re. the message, but if we want to patch we will have to also get it translated, so maybe best for a separate issue?

Key thing is that without doing some refactoring, there isn't a way to specify more than one error message per input, so I think the strategy might be to get this in so we are enforcing the constraints we need to for the patch, and getting the rest of it sorted out in the next release?

Rebecca367:
Nicely done!

## Testing

- [x] If a duplicate policy number is entered, both the personal and family form inputs provide a warning.
- [x] If either personal or family form  puts are updated such that the policy number is no longer equal to an existing policy number, both personal and family form inputs warnings are no longer displayed.
- [x] When testing the previous two tests, the personal AND family warnings are synchronised, i.e. either both display a warning, or neither.
Screen Shot 2020-03-24 at 3 34 22 PM

PR #2491, opened 2020-03-06
#2477 navigation bar not re rendering

### LINKED ISSUE #2477: react-navigation update: NavigationBar not re-rendering
## The suggestion

After upgrading to v5, the `NavigationBar` component doesn't re-render when navigating

**Why?**

Because navigating doesn't cause `mSupplyMobileApp` to re-render

N/A

## Additional context

- I think the solution here is to put the `NavigationBar` into the `Navigator`, so it is a child, rather than having it at the same level. But not 100% what to do !

joshxg:
Fixes #2477 

## Change summary

- Back navigation working now
- Still hooked into the store through middleware, to limit code changes (We hook into the back event to do a few things - for example, refresh the customer invoiceS page after navigating back from the customer invoice page - so that the status can correctly show finalised if you finalised in the customer invoice page, because the finalisation logic does not really follow the data flow).

- Attempt 1: was to just use a custom header in the root navigators props, you can specify a component for the header. However the component didn't re-render when swapping pages. So titles wouldn't update when switching between pages correctly.
- Attempt 2: Do the same as above, but connect the NavigationBar to redux, so that when navigating, the navigate action would cause the state to update, triggering a re-render on the navigation bar and subsequently the title to refresh. This worked *too* well, and the title was updating before actually fully navigating/transitioning. Also, by not dispatching a back event (just using `RootNavigator.goBack()`, there was no updating going on at all when navigating back.
- Attempt 3: dispatch a `goBack` action to make sure updates were done when navigating back. Also at this point the whole navigation bar was a waste of time, so just removed it!
- Attempt 4: Same as 3, except also doing the hardware back button. Extracted so could use the same logic for both.

## Testing

N/A

- Finalising is still broken


src/Store.js
```diff
@@ -8,8 +8,8 @@ import { persistStore, persistReducer } from 'redux-persist';
 import { applyMiddleware, createStore } from 'redux';
 import thunk from 'redux-thunk';
 
+import { navigationMiddleware } from './navigation';
```
joshxg: just put this in navigation folder - thought it was better but now not so sure

src/mSupplyMobileApp.js
```diff
@@ -68,43 +60,12 @@ import { PrescriptionPage } from './pages/PrescriptionPage';
 import { SettingsPage } from './pages/SettingsPage';
 import { DashboardPage } from './pages/DashboardPage';
 import { SupplierRequisitionPage } from './pages/SupplierRequisitionPage';
+import { navigationStyles } from './globalStyles/navigationStyles';
 
 const SYNC_INTERVAL = 10 * 60 * 1000; // 10 minutes in milliseconds.
 const AUTHENTICATION_INTERVAL = 10 * 60 * 1000; // 10 minutes in milliseconds.
 
 class MSupplyMobileAppContainer extends React.Component {
-  handleBackEvent = debounce(
```
joshxg: this is moved to be in the `navigation/actions` file, `goBack` action. Handles whether or not to go back or close the app

src/mSupplyMobileApp.js
```diff
@@ -137,16 +98,12 @@ class MSupplyMobileAppContainer extends React.Component {
   }
 
   componentDidMount = () => {
-    BackHandler.addEventListener('hardwareBackPress', this.handleBackEvent);
-
```
joshxg: moved to `main.js`, the real root

src/mSupplyMobileApp.js
```diff
@@ -266,11 +211,6 @@ class MSupplyMobileAppContainer extends React.Component {
     );
   };
 
-  renderPageTitle = () => {
```
joshxg: This is in `BackButton.js` now

src/mSupplyMobileApp.js
```diff
@@ -283,10 +223,19 @@ class MSupplyMobileAppContainer extends React.Component {
     );
   };
 
+  getOptions = () => ({
```
joshxg: these options are 'defaults' for screens within the navigator. Only called once. Looking for a better place to put it.

src/main.js
```diff
@@ -4,18 +4,23 @@
...
+import { backHandler, rootNavigatorRef } from './navigation';
 
 import MSupplyMobileApp from './mSupplyMobileApp';
 
-function App() {
+const App = () => {
+  React.useEffect(() => {
+    BackHandler.addEventListener('hardwareBackPress', backHandler(store));
+    return () => BackHandler.removeEventListener('hardwareBackPress', backHandler(store));
+  }, []);
```
joshxg: This only occurs when this component first mounts and then the return function is run when it's unmounted. Injecting the store into the back handler function to avoid a cycle, mainly

src/navigation/RootNavigator.js
```diff
@@ -14,22 +14,41 @@
...
+import { StackActions } from '@react-navigation/native';
 
 export const rootNavigatorRef = React.createRef();
 
 const navigate = (screenName, screenProps) =>
   rootNavigatorRef.current?.navigate(screenName, screenProps);
 
 const replace = (screenName, screenProps) =>
-  rootNavigatorRef.current?.replace(screenName, screenProps);
+  rootNavigatorRef.current?.dispatch(StackActions.replace(screenName, screenProps));
```
joshxg: No `navigation.replace` method .. but can dispatch the action.. go figure

src/navigation/RootNavigator.js
```diff
@@ -14,22 +14,41 @@
...
+  return routes[routes.length - 1]?.name ?? '';
+};
+
+const prevRouteName = () => {
+  if (!rootNavigatorRef.current?.getRootState()) return '';
+
+  const routes = rootNavigatorRef.current?.getRootState()?.routes ?? [];
+
+  return routes[routes.length - 2]?.name ?? '';
+};
```
joshxg: simple current/prev route names - prevroute name is used in the `Back` action to set up the next current route name.. hoping to get rid of this in the next PR

src/navigation/actions.js
```diff
@@ -30,6 +34,26 @@ import { ROUTES } from './constants';
...
+        dispatch({
+          ...NavigationActions.back(),
+          payload: {
+            prevRouteName: RootNavigator.prevRouteName(),
+          },
+        });
+        dispatch(PrescriptionActions.deletePrescription());
+      });
+    });
+  }
```
joshxg: With rendering logic dependent on navigation, I think it's a lot simpler to have this in redux - it's a few more changes to make a specific action for all the current reducers which are checking for navigating back.

src/navigation/utilities.js
```diff
@@ -0,0 +1,73 @@
...
+ * Sustainable Solutions (NZ) Ltd. 2020
+ */
+
+import { RootNavigator } from './RootNavigator';
+import { goBack } from './actions';
+import { ROUTES } from './constants';
+
+import { navStrings } from '../localization';
+
+export const SCREEN_TITLES = {
```
joshxg: I don't like this for getting a screens title - open to suggestions!

src/navigation/actions.js
```diff
@@ -30,6 +34,26 @@ import { ROUTES } from './constants';
  *
  */
 
+export const goBack = () => dispatch => {
+  if (!RootNavigator.canGoBack()) BackHandler.exitApp();
+  else {
+    UIDatabase.write(() => {
+      const prescriptions = UIDatabase.objects('Prescription').filtered('status != "finalised"');
+      UIDatabase.delete('Transaction', prescriptions);
```
wlthomson: But confused by this, is this the only side-effect performed on the `goBack` action?

src/navigation/utilities.js
wlthomson: Hmm, I don't think this is too bad? Can't really think of another way of doing it that's any tidier tbh.

src/widgets/BackButton.js
```diff
@@ -0,0 +1,54 @@
...
+
+import { textStyles } from '../globalStyles';
+
+const mapStateToProps = state => {
+  const { pages = {} } = state;
+  const { currentRoute = '' } = pages;
+  const { [currentRoute]: currentPageState = {} } = pages;
+  const { pageObject } = currentPageState;
+
+  return { title: getRouteTitle(pageObject, currentRoute) };
```
wlthomson: Hmm, I think this is elegant and concise, but I feel like "error handling" by deferring to passing `null` to `getRouteTitle` is a bit confusing. You might think OTT, so feel free to disagree, but something like this has more explicit branches which would make debugging/testing easier:

```suggestion
  const defaultProps = { title: '' };
  const { pages = {} } = state;
  const { currentRoute } = pages;
  if (!currentRoute) return defaultProps;
  const { [currentRoute]: currentPageState } = pages;
  if (!currentPageState) return defaultProps;
  const { pageObject } = currentPageState;
  if (!pageObject) return defaultProps;
  return { ...defaultProps, title:  getRouteTitle(pageObject, currentRoute) };
```

joshxg: this definitely seems odd and not to your regular "one path no early returns" arguments ! hehhee

joshxg: Your problem is with calling `getRouteTitle` with null so wouldn't something like:

`return {title: currentRoute ? getRouteTitle(.. ) : ""` address it?

src/navigation/actions.js
joshxg: there are other pure effects which other reducers are acting on - main one is the `PagesReducer`

src/Store.js
mark-prins: I prefer it in `./navigation`

src/widgets/BackButton.js
wlthomson: As above, just my personal option but:

My thought was more that unless it's a pattern throughout the codebase, it's probably more consistent to prevent passing `null` parameters in the caller, rather than assuming the function will handle `null` parameters and return a sensible value. If I didn't know the contract that `getRouteTitle` enforces, I have to double check that `getRouteTitle` behaves sensibly when it is called with basically no parameters. IMO this is the sort of thing that can cause bugs and something to be careful of when function params aren't typed.

PR #2503, opened 2020-03-08
#2501 add is_edited field to outgoing stocktake batch sync

### LINKED ISSUE #2501: Enter qty of stocktake is not shown in the mSupply desktop app
## Describe the bug
Creating stocktake in the mobile with quantity different than its snapshot quantity just shows blank in the desktop app.

### To reproduce

1. Create a new stocktake in the mobile.
2. Put a quantity different than its snapshot quantity. Then force sync to the mSupply desktop app.
3. Open a stocktake in the desktop app.
4. See the enter quantity shows now blank.

### Expected behaviour
Quantity should show up if counted no. of packs are different than its snapshot quantity.

### Proposed Solution
Had a quick check in the desktop app and found that `stockt_ake_line.is_edited` field didn't populated while assigning quantity in the mobile. The field should get True when actual quantity is different than its snapshot quantity in the mobile.

### Version and device info

- App version: 4.0.0

wlthomson:
Fixes #2501.

## Change summary

Adds `is_edited` to `stocktakeBatch` outgoing sync.

### Setup

- Using mobile, create a new stocktake.
 - Enter a quantity different than its snapshot quantity. 
- Sync to desktop.
- Open the stocktake in desktop.

### Test cases 

- [ ] The entered quantity is shown correctly (i.e. the same as entered on mobile).

N/A.

### DIFF
--- a/src/sync/outgoingSyncUtils.js
+++ b/src/sync/outgoingSyncUtils.js
@@ -184,6 +184,7 @@ const generateSyncData = (settings, recordType, record) => {
         Batch: record.batch,
         item_ID: record.itemId,
         optionID: record.option && record.option.id,
+        is_edited: record.hasBeenCounted,
       };
     }
     case 'Transaction': {


wlthomson:
> works - but as discussed if IC cloud server is on 4.07+ (whatever patch this `is_edited` is in), then we might have to repush all stocktake batches of finalised stocktakes.. which is not ideal :( Lets hope they're not!

IC cloud server is `v4.06.05`, so think we're safe here!

PR #2531, opened 2020-03-11
#2523 prevent applying inactive policies

### LINKED ISSUE #2523: Insurance policy display even after making it inactive
## Describe the bug
After an insurance policy has been made inactive, it still shows up in policies list in dispensary

### To reproduce
1. First, make any of the existing insurance policy inactive in the desktop, and Sync.
2. Now, in mobile, Go to Dispensary.
3. Select a patient. Choose a prescriber and select any items from the list and click next.
4. Select the insurance policy tab.
5. See error : the inactive insurance policy still shows up in the list.

### Expected behaviour
The inactive insurance policy should not show up in the list.

### Version and device info
- Tablet model: 4.0.1

### Additional context
 found while testing #2481

wlthomson:
Fixes #2523.

## Change summary

- Update insurance policy selectors to display if policies active or not. 
- Do not apply discount rate for inactive policies.

## Testing

- [ ] When viewing an inactive policy in the dropdown, `"(inactive)"` is appended to the policy name.
- [ ] If an inactive policy is selected, the insurance discount rate is `"0%"`.
- [ ] If an inactive policy is selected, the insurance discount amount is `"0.00"`.
- [ ] If an inactive policy is selected, the insurance discount is not applied to the subtotal or total.

N/A.

### DIFF
--- a/src/widgets/PaymentSummary.js
+++ b/src/widgets/PaymentSummary.js
@@ -1,7 +1,7 @@
 /* eslint-disable react/forbid-prop-types */
 /**
  * mSupply Mobile
- * Sustainable Solutions (NZ) Ltd. 2019
+ * Sustainable Solutions (NZ) Ltd. 2020
  */
 
 import React from 'react';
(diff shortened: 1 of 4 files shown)


src/widgets/PaymentSummary.js
```diff
@@ -109,9 +109,14 @@ const PaymentSummaryComponent = ({
   usingInsurance,
   usingPaymentTypes,
 }) => {
-  const policyNumbers = React.useMemo(() => insurancePolicies.map(policy => policy.policyNumber), [
-    insurancePolicies,
-  ]);
+  const policyNumbers = React.useMemo(
```
mark-prins: is there a reason for the change?  I like the terse version, so am just wondering if I'm missing something

wlthomson: No reason. Was initially going to apply the policy number mapping here, but then moved it to the selector. 94c61daa9a484f4e517714b525636827e4c11b08.

PR #2548, opened 2020-03-16
#2545 map route params to props

### LINKED ISSUE #2545: Map route.params to props

React Navigation is passing params by `route.params` - there are instances where components ( such as `PrescriptionPage` ) is expecting params to come in via props.

Map the parameters passed via `route.params` to the `props` parameter in the component

mark-prins:
Fixes #2545 

## Change summary

In the end probably not worth a separate PR. I see that this is well trodden ground already.
Just added parsing of `route.params` to props for the PrescriptionPage - preserves the prop specification and type checking while passing in the params from the route.

## Testing

- [ ] Create a prescription. This was failing after selection of the prescriber.


mark-prins:
> hm dunno whats gone on with the diff here? But what I think you changed looks good!

I went and branched off develop to begin with - and that has picked up all of the merges. Rather than untangle that, I'll just update the original branch. sigh.

PR #2564, opened 2020-03-23
#2516 refactor prefs

### LINKED ISSUE #2516: Use Prefs instead of custom data for preferences

Desktop is changing here: https://github.com/sussol/msupply/issues/5762

- Deprecating custom data usage for preferences
- using the pref table to store preferences for mobile stores as well as desktop stores

- Currently after sync, there is code in `PostSyncProcessor` which evaluates the mobile stores `custom_data` stringified object.
- This stores each key in `Settings`. For example: `USES_VACCINE_MODULE: true` is placed in settings after having received `usesVaccineModule: true` in the store custom data.
- The solution required is to determine preferences from a `pref` record, rather than custom data. We can then either continue to store the key:value pairs in `Settings` or store them in a new `Preference` table.

## Implementation

Rather than using `PostSyncProcessor`, a potential solution would be to just read the `Pref` records as they are synced, and store them in the database. Not sure if we want to be selective on the prefs we store? i.e. only store the ones we have decided to accept or not

N/A

## Additional context

So my questions on the implementation:

- Use `Settings` or `Preference` table? `Settings` is easier but I feel like it's piggy backing on something that it wasn't really designed for. I think there are differences between Sett
- Just save all `pref` records or be selective?

wlthomson:
Fixes #2516.

## Change summary

Removes custom data logic and adds logic for `[pref]item="store_preferences"`.

The following `custom_data` keys have been moved to prefs:

- `usesVaccineModule`
- `usesDispensaryModule`
- `usesDashboardModule`
- `usesCashRegisterModule`
- `usesPaymentModule`
- `usesPatientTypes`
- `usesHideSnapshotColumn`
- `monthlyConsumptionEnforceLookBackPeriod`
- `monthlyConsumptionLookBackPeriod`
- `monthsLeadTime`

### Setup

- Ensure you are using mSupply `>= v4.10-RC9`.
- The above preferences should be visible in the store preferences tab.
- Note that in order to sync store preferences, you must change at least one of the preferences AFTER initial sync (see https://github.com/sussol/msupply/issues/5862).
- Ensure the `View/edit preferences` checkbox is enabled for the user/store (required to view synced `pref` data).
- Ensure at least one payment type is defined (see `paymentType` table).
- To examine a preference, go to `Settings -> Realm Explorer` and type `"Pref"` in the search bar. 

### Test cases 

- [ ] Enable preference `Uses Vaccine Module` and sync with mobile. 
  - [ ] The `Pref` `usesVaccineModule` is set to `true`.
- [ ] Disable preference `Uses Vaccine Module` and sync with mobile.
   - [ ] The `Pref` `usesVaccineModule` is set to `false`.
- [ ] Enable preference `Uses Dispensary Module` and sync with mobile. 
  - [ ] The `Pref` `usesDispensaryModule` is set to `true`.
  - [ ] The `"Dispensary"` tab is visible in the bottom-right of the menu page.
- [ ] Disable preference `Uses Dispensary Module` and sync with mobile. 
  - [ ] The `Pref` `usesDispensaryModule` is set to `false`.
  - [ ] The `"Dispensary"` tab is not visible in the bottom-right of the menu page.
- [ ] Enable preference `Uses Dashboard Module` and sync with mobile. 
  - [ ] The `Pref` `usesDashboardModule` is set to `true`.
  - [ ] The "`Dashboard`" tab is visible in the bottom-right of the menu page.
- [ ] Disable preference `Uses Dashboard Module`and sync with mobile. 
  - [ ] The `Pref` `usesDashboardModule` is set to `false`.
  - [ ] The "`Dashboard`" tab is not visible in the bottom-right of the menu page.
- [ ] Enable preference `Uses Cash Register Module` and sync with mobile. 
  - [ ] The `Pref`  `usesCashRegisterModule` is set to `true`.
  - [ ] The `"Cash Register"` tab is visible in the bottom-right of the menu oage.
- [ ] Disable preference `Uses Cash Register Module` and sync with mobile. 
  - [ ] The `Pref` `usesCashRegisterModule` is set to `false`.
  - [ ] The `"Cash Register"` tab is not visible in the bottom-right of the menu page.
- [ ] Enable preference `Uses Payment Module` and sync with mobile.
  - [ ] The `Pref`  `usesPaymentModule` is set to `true`.
  - [ ] Payments/prices are enabled for supplier invoices, stocktakes and prescriptions.
- [ ] Disable preference `Uses Payment Module` and sync with mobile. 
  - [ ] The `Pref` `usesPaymentModule` is set to `false`.
  - [ ] Payments/prices are disabled for supplier invoices, stocktakes and prescriptions.
- [ ] Enable preference `Uses Patient Types` and sync with mobile.
  - [ ] The `Pref`  `usesPatientTypes` is set to `true`.
  - [ ] The patient type dropdown is displayed in the prescription wizard.
- [ ] Disable preference `Uses Patient Types` and sync with mobile.
  - [ ] The `Pref` `usesPatientTypes` is set to `false`.
  - [ ] The patient type dropdown is not displayed in the prescription wizard.
- [ ] Enable preference `Uses Hide Snapshot Column` and sync with mobile. 
  - [ ] The `Pref`  `usesHideSnapshotColumn` is set to `true`.
  - [ ] The snapshot quantity column is NOT displayed for stocktakes.
- [ ] Disable preference `Uses Hide Snapshot Column` and sync with mobile.
  - [ ] The `Pref` `usesHideSnapshotColumn` is set to `false`.
  - [ ] The snapshot quantity column is displayed for stocktakes.
- [ ] Set preference `"Monthy consumption lookback period"` to a non-negative numerical value.
  - [ ] The `Pref` `monthlyConsumptionLookBackPeriod` is set to the given value.
  - [ ] The lookback period is calculated as expected for requisitions (@andreievg, @joshxg, @mark-prins).
- [ ] Enable preference `Enforce look back period` and sync with mobile.
  - [ ] The `Pref` `monthlyConsumptionEnforceLookBackPeriod` is set to `true`.
  - [ ] The lookback period is calculated as expected for requisitions (@andreievg, @joshxg, @mark-prins).
- [ ] Disable preference `Enforce look back period` and sync with mobile.
  - [ ] The `Pref` `monthlyConsumptionEnforceLookBackPeriod` is set to `false`.
  - [ ] The lookback period is calculated as expected for requisitions (@andreievg, @joshxg, @mark-prins).
- [ ] Set preference `"Months lead time"` to a positive numerical value.
  - [ ] The `Pref` `monthsLeadTime` is set to the given value.
  - [ ] The lead time is calculated as expected for supplier requisitions (@andreievg, @joshxg, @mark-prins).

Looks like `monthlyConsumptionEnforceLookBackPeriod` was not added as part of the recent pref changes on desktop. @andreievg, @joshxg, @mark-prins Worth making an issue for this?


src/sync/incomingSyncUtils.js
```diff
@@ -193,6 +193,14 @@ export const sanityCheckIncomingRecord = (recordType, record) => {
       cannotBeBlank: ['name', 'number_to_use'],
       canBeBlank: [],
     },
+    Pref: {
+      cannotBeBlank: ['ID', 'store_ID', 'item', 'data'],
+      canBeBlank: ['user_ID', 'network_ID'],
+    },
+    PrefBlob: {
+      cannotBeBlank: ['ID', 'store_id', 'item', 'type', 'blob'],
+      canBeBlank: ['user_ID', 'network_id'],
```
andreievg: we don't integrated pref blob records, was wondering why we would need to do sanity check on them

src/sync/incomingSyncUtils.js
```diff
@@ -193,6 +193,14 @@ export const sanityCheckIncomingRecord = (recordType, record) => {
       cannotBeBlank: ['name', 'number_to_use'],
       canBeBlank: [],
     },
+    Pref: {
+      cannotBeBlank: ['ID', 'store_ID', 'item', 'data'],
```
andreievg: was wondering, is there a chance that some global preference would also sync to mobile and cause sync to halt because it doesn't have store_ID ? Maybe it's more stable to check store_ID in `case 'Pref': {`

src/sync/incomingSyncUtils.js
```diff
@@ -632,6 +640,27 @@ export const createOrUpdateRecord = (database, settings, recordType, record) =>
       numberSequence.addNumberToReuse(numberToReuse);
       break;
     }
+    case 'Pref': {
+      const { item } = record;
+      if (item === 'store_preferences') {
+        try {
+          const recordData = JSON.parse(record?.data);
```
andreievg: no need for optional chaining here

src/sync/incomingSyncUtils.js
```diff
@@ -632,6 +640,27 @@ export const createOrUpdateRecord = (database, settings, recordType, record) =>
...
+          const recordKeys = Object.keys(recordData);
+          recordKeys.forEach(key => {
```
andreievg: Not required change (stylistic preference), but whenever I need to iterate through key values:
Object.entries(recordDate).forEach(([key, value]) => {
}

src/database/DataTypes/Pref.js
```diff
@@ -0,0 +1,72 @@
...
+export class Pref extends Realm.Object {
+  get type() {
+    return PREF_TYPES[this.id];
+  }
+
+  get value() {
+    switch (this.type) {
+      case TYPES.BOOL:
+        try {
+          return !!JSON.parse(this.data);
```
andreievg: was wondering what is the use case of needing `!!` here ?

src/database/UIDatabase.js
```diff
@@ -262,9 +262,14 @@ class UIDatabase {
...
 
+  getPreference(key) {
+    const preference = this.database.get('Pref', key);
+    return preference?.value;
+  }
+
   getSetting(key) {
     const setting = this.database.get('Setting', key, 'key');
-    return (setting && setting.value) || '';
+    return setting?.value ?? '';
```
andreievg: not part of the issue, please revert

src/database/UIDatabase.js
```diff
@@ -262,9 +262,14 @@ class UIDatabase {
     return this.database.write(...args);
   }
 
+  getPreference(key) {
+    const preference = this.database.get('Pref', key);
+    return preference?.value;
```
andreievg: Hmm, what happens to all the callers of getPreference if preference is not defined ?

src/sync/incomingSyncUtils.js
```diff
@@ -632,6 +640,27 @@ export const createOrUpdateRecord = (database, settings, recordType, record) =>
...
+        } catch (error) {
+          // Silently ignore malformed prefs.
+        }
+      }
+      break;
+    }
+    case 'PrefBlob': {
+      // Ignore pref_blob records.
+      break;
+    }
```
joshxg: Is this for being explicit that we ignore pref blobs? I'm thinking it might be a little nicer to just leave it out

src/sync/PostSyncProcessor.js
```diff
@@ -86,39 +83,6 @@ export class PostSyncProcessor {
     this.settings.set(LAST_POST_PROCESSING_FAILED, 'false');
   };
 
-  /**
```
joshxg: niiice

src/database/UIDatabase.js
joshxg: gonna say the opposite - nice!

src/navigation/actions.js
```diff
@@ -402,17 +401,8 @@ export const createSupplierRequisition = ({
...
-  const customData = Settings.get(SETTINGS_KEYS.THIS_STORE_CUSTOM_DATA);
-
-  // CustomData is a stringified JSON object.
-  const parsedCustomData = customData ? JSON.parse(customData) : '';
-
   // Months lead time has an effect on daysToSupply for a requisition.
-  const monthsLeadTime = parsedCustomData.monthsLeadTime
-    ? Number(parsedCustomData.monthsLeadTime.data)
-    : 0;
+  const monthsLeadTime = UIDatabase.getPreference(PREF_KEYS.MONTHS_LEAD_TIME);
```
joshxg: Exact same behaviour but to be more explicit about the value passed to `createRecord`, could `monthsLeadTime = ...getPref... ?? 0`

src/reducers/ModulesReducer.js
```diff
@@ -24,21 +24,20 @@ import { SYNC_TRANSACTION_COMPLETE } from '../sync/constants';
...
   const usingSupplierCreditCategories = UIDatabase.objects('SupplierCreditCategory').length > 0;
 
-  const usingDashboard = checkModule(SETTINGS_KEYS.DASHBOARD_MODULE);
-  const usingDispensary = checkModule(SETTINGS_KEYS.DISPENSARY_MODULE);
-  const usingVaccines = checkModule(SETTINGS_KEYS.VACCINE_MODULE);
-  const usingCashRegister = checkModule(SETTINGS_KEYS.CASH_REGISTER_MODULE) && usingPaymentTypes;
-  const usingPayments = checkModule(SETTINGS_KEYS.PAYMENT_MODULE);
-  const usingPatientTypes = checkModule(SETTINGS_KEYS.PATIENT_TYPES);
-  const usingHideSnapshotColumn = checkModule(SETTINGS_KEYS.HIDE_SNAPSHOT_COLUMN);
+  const usingDashboard = UIDatabase.getPreference(PREF_KEYS.DASHBOARD_MODULE);
```
joshxg: could be pedantic and explicitly type each call for the `using..` variables as you expect them to be bools rather than null (super pedantic)

I couldnt find a gif about being pedantic so you get this

chef

src/database/DataTypes/Pref.js
```diff
@@ -0,0 +1,72 @@
...
+  [PREF_KEYS.VACCINE_MODULE]: TYPES.BOOL,
+  [PREF_KEYS.DISPENSARY_MODULE]: TYPES.BOOL,
+  [PREF_KEYS.DASHBOARD_MODULE]: TYPES.BOOL,
+  [PREF_KEYS.CASH_REGISTER_MODULE]: TYPES.BOOL,
+  [PREF_KEYS.PAYMENT_MODULE]: TYPES.BOOL,
+  [PREF_KEYS.PATIENT_TYPES]: TYPES.BOOL,
+  [PREF_KEYS.HIDE_SNAPSHOT_COLUMN]: TYPES.BOOL,
+};
+
+export class Pref extends Realm.Object {
```
joshxg: this my most annoying one: I wish this was `Preference` 🤣

src/database/DataTypes/Pref.js
```diff
@@ -0,0 +1,72 @@
...
+  PATIENT_TYPES: 'usesPatientTypes',
+  HIDE_SNAPSHOT_COLUMN: 'usesHideSnapshotColumn',
+};
+
+const TYPES = {
+  NUMERIC: 'numeric',
+  BOOL: 'bool',
+};
+
+export const PREF_TYPES = {
```
joshxg: posssssibly `database/utilities/constants`?

src/sync/syncTranslators.js
```diff
@@ -68,6 +68,8 @@ export const RECORD_TYPES = new SyncTranslator({
   TransactionBatch: 'trans_line',
   User: 'user',
   Unit: 'unit',
+  Pref: 'pref',
+  PrefBlob: 'pref_blob',
```
joshxg: as above I think `PrefBlob` is for explicitness? but I think it's confusing me more then helping but could be just me!

src/sync/incomingSyncUtils.js
```diff
@@ -632,6 +640,27 @@ export const createOrUpdateRecord = (database, settings, recordType, record) =>
...
+        try {
+          const recordData = JSON.parse(record?.data);
+          const recordKeys = Object.keys(recordData);
+          recordKeys.forEach(key => {
+            const prefData = JSON.stringify(recordData[key] ?? {});
+            internalRecord = { id: key, data: prefData };
+            database.update(recordType, internalRecord);
+          });
+        } catch (error) {
+          // Silently ignore malformed prefs.
```
joshxg: good place for some snaggin' in a new issue I reckon?

andreievg:
> re optional chaining/null coals: Have been using them for quite awhile now!

Ohh indeed, @wlthomson oops, ignore that optional chaining question

src/database/UIDatabase.js
andreievg: Food for though, I think it would be better to cast preference type somewhere here rather then in 'Pref' data type.

mark-prins:
> Looks like `monthlyConsumptionEnforceLookBackPeriod` was not added as part of the recent pref changes on desktop

Screen Shot 2020-03-24 at 4 09 37 PM

I don't know about that preference - but the above would suggest that it doesn't need to be migrated?

src/sync/incomingSyncUtils.js
wlthomson: Yeah explicit for now but also just a placeholder for later. Did the work and didn't think it was worth removing. Currently waiting on Adrian for info on any pref blobs we might need to sync. Seems like you and @andreievg on the same page on this one so will go ahead and remove!

wlthomson:
> > Looks like `monthlyConsumptionEnforceLookBackPeriod` was not added as part of the recent pref changes on desktop
> 
> Screen Shot 2020-03-24 at 4 09 37 PM
> 
> I don't know about that preference - but the above would suggest that it doesn't need to be migrated?

Good spot. Not sure whether it should be added to desktop or removed from here? Thoughts?

mark-prins:
> Good spot. Not sure whether it should be added to desktop or removed from here? Thoughts?

given that I don't know what it's for - nor why it is required.. my vote is to remove.

andreievg:
@wlthomson @mark-prins 

From memory, it was requested by IC. Before this mobile would use either lookBackPeriod or first transaction (whichever one is closer to current date), so there was mass confusion why lookBackPeriod wasn't enforced as denominator for new mobile sites. I suggest we keep it and add it to desktop changes

src/database/UIDatabase.js
mark-prins: I don't see the problem - it's a much neater implementation

src/database/schema.js
```diff
@@ -188,7 +190,7 @@ export const schema = {
     User,
     Currency,
   ],
-  schemaVersion: 15,
+  schemaVersion: 17,
```
mark-prins: what happened to poor old 16?

mark-prins:
> From memory, it was requested by IC. Before this mobile would use either lookBackPeriod or first transaction (whichever one is closer to current date), so there was mass confusion why lookBackPeriod wasn't enforced as denominator for new mobile sites. I suggest we keep it and add it to desktop changes

Thanks @andreievg - the mists are clearing, and I vaguely remember now. IIRC the request from Côte d'Ivoire was to *always* use the lookback period. We implemented the other because it is our default and it makes logical sense. So really, this should just be tied to the IC customisation. which we don't do in mobile. So yeah, sorry. leave it in. 🤷‍♂ 

While I'm looking at it.. the naming of this doesn't make sense to me:

`addedRecently = itemAddedDate.isBefore(lookbackDate)`

If something is before the lookbackDate then is is earlier than it and therefore older? Not 'added recently'?? ( https://momentjs.com/docs/#/query/is-before/ )

wlthomson:
@mark-prins @andreievg Thanks for clarification. See issue https://github.com/sussol/msupply/issues/5868, will do that ASAP and add to tests.

wlthomson:
> While I'm looking at it.. the naming of this doesn't make sense to me:

addedRecently = itemAddedDate.isBefore(lookbackDate)

If something is before the lookbackDate then is is earlier than it and therefore older? Not 'added recently'?? ( https://momentjs.com/docs/#/query/is-before/ )

Could possibly be wrong as not an expert on this functionality, but possibly not just a identifier name but also logic might be wrong here? I think the intended logic is:

```
IF enforcing lookback THEN use lookback date
ELSE
  IF item added after lookback date, use item added date
  ELSE use lookback date
```

So I think maybe it should be something like:

```
const itemNewerThanLookback = itemAddedDate.isAfter(lookbackDate)
const startDate = amcEnforceLookback ||  !itemNewerThanLookback ? lookbackDate : itemAddedDate;
```

Does that seem right?

src/database/UIDatabase.js
andreievg: So does that mean, anytime we are working on an issue and editing a file that can be nice and neater with use of optional chaining and null coalescing operator, we should put an effort into finding and converting to that syntax ?

I really want to link some PRs here (desktop sync) that all four of us have reviewed where all four of us concluded that 'surgical' and 'targeted' approach to code change should have been exercised (it's been a soft PR 'guideline' in desktop for awhile).

(not saying I don't like the syntax btw, even if technically `??` is not the same as `||`)

joshxg:
@wlthomson yeah looks like a logical error for a new issue - should be `isAfter` (and maybe a variable name change if it helps). then `addedRecently` is true if it was added before the look back date. which means that if you're enforcing the lookback, or if the item was first added before the lookback, then use the lookback date. Otherwise, use the date the item was added

src/database/UIDatabase.js
joshxg: > So does that mean, anytime we are working on an issue and editing a file that can be nice and neater with use of optional chaining and null coalescing operator, we should put an effort into finding and converting to that syntax ?

ya!

> I really want to link some PRs here (desktop sync) that all four of us have reviewed where all four of us concluded that 'surgical' and 'targeted' approach to code change should have been exercised (it's been a soft PR 'guideline' in desktop for awhile).

I agree with it on desktop - should touch as little as possible haha that's the wild west. But doing refactors is how you avoid that situation and since we don't really do scheduled refactors, I think as you go is a great solution. I think even bigger refactors in a PR are all good if ya mention it in the PR comment!

wlthomson:
@joshxg Nice, thanks for clarification :+1:.

src/database/UIDatabase.js
```diff
@@ -50,6 +52,34 @@ const translateToCoreDatabaseType = type => {
   }
 };
 
+const isBooleanData = data => {
```
mark-prins: nice! ( I thought when first reviewing that it would just work anyway, but all the same I really like being explicit and using typing )

src/database/UIDatabase.js
mark-prins: yep. agree that in desktop we should be super careful and not go wild with refactoring. We're planning to deprecate that codebase and there isn't a test framework to give us any safety when refactoring.

Also agree that in small and specific cases that it is actually good to do in an active codebase. Keeps it clean. Yes, the danger is that you introduce bugs and those sort of bugs are actually the hardest to find later because they don't directly relate to the issue.

In this case though- the change is simple, functionally equivalent and in an area that will be directly tested by the testing for this issue.

wlthomson:
Updated tests to include `monthlyConsumptionEnforceLookBackPeriod`. See https://github.com/sussol/msupply/pull/5875.

src/sync/incomingSyncUtils.js
wlthomson: Nice! #2673.

src/database/UIDatabase.js
wlthomson: thankyou

src/sync/incomingSyncUtils.js
wlthomson: 615d1b2a42f7545707699d21d693733763b1c32e.

src/database/UIDatabase.js
wlthomson: Yeah, I agree, see 48c5aaab4a36171a98dc48edf55a688804713fab.

src/database/DataTypes/Pref.js
wlthomson: Just to force a cast to `bool`, but more explicit now!

src/sync/incomingSyncUtils.js
wlthomson: Yeah, much nicer. Can't think why I went with this option! 27d82b07c910015038d36b761ee6136d127fc75c.

src/database/UIDatabase.js
andreievg: I am glad we having this conversation actually, because I was and still am of a different opinion.  

@joshxg 
> I agree with it on desktop - should touch as little as possible haha that's the wild west. But doing refactors is how you avoid that situation

I haven't seen any evidence that refactoring of code outside of issues scope is how we avoid that situation. I have seen cases where refactoring of code outside of issue scope cased regression

>  I think even bigger refactors in a PR are all good if ya mention it in the PR comment!

Why would it be all good if it's not solving the issue it addresses ?

@mark-prins 
> We're planning to deprecate that codebase and there isn't a test framework to give us any safety when refactoring.

I don't believe we have one in mobile too, and are starting to get quite a few permutation of preferences / settings, which is already making testing harder.

> Also agree that in small and specific cases that it is actually good to do in an active codebase. Keeps it clean.

If there is code that is obviously needs a clean up, we can make an issue to clean it up. I don't actually agree that existing code was not clean (yes it neater and nicer with the change though, but that's not saying it wasn't clean, or not safe or less readable before)

> In this case though- the change is simple, functionally equivalent and in an area that will be directly tested by the testing for this issue.

The change itself is simple yes, and it's functionally equivalent  (even though `??'` is not quite the same as `||`) testing wise however, it's not part of preferences testing and this issue testing. 

Btw, I am pretty sure optional chaining was made for deeper then one nested object property check, and there was no need to use null coalescent here, '||' would look just as neat and further preserves functionality.

My conclusion is still
* Don't touch things that are not solving the issue, the focus should be on solving the issue, not making code base neater and nicer.
* If you notice something 'wrong' with existing code, make another issue for it
* In existing code where things are obviously not broken (and look perfectly fine), It should be up to stakeholders to decide if it's worth refactoring (issue)

Sounds like this conversation is getting too big for this issue, we should archive it somewhere (like wiki) once we reached a consensus.

By the way, this PR has refactoring (great refactoring) which is within the scope of the issue, which are completely fine.

And lastly, I do realise I am spending quite a bit of time commenting on one line of code, but it's beyond that actually, it's one area of PR 'guidelines' that needs to be agree on and clarified (by us 4)

src/sync/incomingSyncUtils.js
wlthomson: Yeah, suppose not as the `catch` would catch the `TypeError`.  b9eb872f1876e6978720fb0b4b5956fc3c5b1ff9.

src/sync/incomingSyncUtils.js
wlthomson: AFAIK, it the sanity check rules are broken the record is just silently ignored (no halting of syncing). Correct me if wrong?

src/sync/incomingSyncUtils.js
wlthomson: Fair point. 615d1b2a42f7545707699d21d693733763b1c32e.

src/navigation/actions.js
wlthomson: 9082642ad73a41a14c55a0096c15c0c53573560e. Also added explicit default assignment behaviour with pref type refactor (see `parseNumericData` etc).

src/reducers/ModulesReducer.js
wlthomson: 9eb4b118cbfa521678f4c0b701a38c63a07a6a7a :+1: .

src/database/DataTypes/Pref.js
wlthomson: Yeah, I think better. Not sure why I went with `Pref`, maybe subconcious corrupted by desktop :laughing: . 4f954526932b44a2a4660a24a9e86d2c8ae1bd9f.

src/database/DataTypes/Pref.js
wlthomson: 39cae3c95162058314f1126de2441865b1bbff39.

src/sync/syncTranslators.js
wlthomson: 615d1b2a42f7545707699d21d693733763b1c32e.

src/database/schema.js
wlthomson: Oops. Fixed (somewhere in the other commits!).

src/reducers/ModulesReducer.js
```diff
@@ -24,21 +24,23 @@ import { SYNC_TRANSACTION_COMPLETE } from '../sync/constants';
...
   const usingSupplierCreditCategories = UIDatabase.objects('SupplierCreditCategory').length > 0;
 
-  const usingDashboard = checkModule(SETTINGS_KEYS.DASHBOARD_MODULE);
-  const usingDispensary = checkModule(SETTINGS_KEYS.DISPENSARY_MODULE);
-  const usingVaccines = checkModule(SETTINGS_KEYS.VACCINE_MODULE);
-  const usingCashRegister = checkModule(SETTINGS_KEYS.CASH_REGISTER_MODULE) && usingPaymentTypes;
-  const usingPayments = checkModule(SETTINGS_KEYS.PAYMENT_MODULE);
-  const usingPatientTypes = checkModule(SETTINGS_KEYS.PATIENT_TYPES);
-  const usingHideSnapshotColumn = checkModule(SETTINGS_KEYS.HIDE_SNAPSHOT_COLUMN);
+  const usingDashboard = Boolean(UIDatabase.getPreference(PREFERENCE_KEYS.DASHBOARD_MODULE));
```
joshxg: `!!`? :trollface:

src/sync/incomingSyncUtils.js
andreievg: I meant record?.data would check that record is undefined right? (at this stage record has to be defined)

src/sync/incomingSyncUtils.js
andreievg: From memory it was halting sync, I did a quick check, didn't see any non store specific prefs syncing

src/database/UIDatabase.js
andreievg: I'll move this to another issue, resolving conversation and 2vs1 so will approve this change

src/reducers/ModulesReducer.js
wlthomson: myboy

src/database/UIDatabase.js
```diff
@@ -50,6 +52,34 @@ const translateToCoreDatabaseType = type => {
...
+
+const isNumericData = data => !isNaN(data);
+
+const isNonNegativeNumericData = data => isNumericData(data) && Number(data) >= 0;
+
+const parseBooleanData = data => (isBooleanData(data) ? JSON.parse(data) : false);
+
+const parseNumericData = data => (isNumericData(data) ? JSON.parse(data) : 0);
+
+const parseNonNegativeNumericData = data => (isNonNegativeNumericData(data) ? JSON.parse(data) : 0);
```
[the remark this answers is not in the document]
wlthomson: giphy

src/database/UIDatabase.js
```diff
@@ -50,6 +52,34 @@ const translateToCoreDatabaseType = type => {
...
+  switch (data.toLowerCase()) {
+    case 'true':
+    case 'false':
+      return true;
+    default:
+      return false;
+  }
+};
+
+const isNumericData = data => !isNaN(data);
```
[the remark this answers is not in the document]
wlthomson: Yeah, I think so, I think maybe something for a tidy up issue as I believe we have a few of these sort of `parseX` methods now?

wlthomson:
@andreievg You able to re-review? Would be good to get this into `develop` :+1:.

wlthomson:
> Looks good @wlthomson, thanks.
> 
> btw, the reason I suggested that preferences value typing should move to UIDatabase.js, was because you can then enforce relative default type return when calling getPreference (so that all future users of getPreference can expect correct valueType, even if preference doesn't exist)

Yeah, I understood and agree on the most part. Currently, still always returns `null` if preference or `data` does not exist. I agree that having a default would be good in some instances, but I'm a bit hesitant to do it for all instances unless there's a convention that ensures that e.g. all numerical preferences default is `0`, all booleans is `false` etc.

Related thing I had in mind was that by returning `null`, the user can differentiate non-existent preferences from preferences with the "default" value. E.g. if I have an optional boolean preference: if user set to `true` display `X`, if user set to `false` display `Y`, if user has not set do not display anything. In this situation, the getter always returning a default would prevent this. Not suggesting this is a good design, just saying the convention might need to be documented (maybe it is and I'm unaware of it!) if we want to go forward that way.

Does that all make sense?

andreievg:
> Yeah, I understood and agree on the most part. Currently, still always returns null if preference or data does not exist. I agree that having a default would be good in some instances, but I'm a bit hesitant to do it for all instances unless there's a convention that ensures that e.g. all numerical preferences default is 0, all booleans is false etc.

Sure. In desktop we no have default values for each of the prefs individually, which I think is the way to go

> Related thing I had in mind was that by returning null, the user can differentiate non-existent preferences from preferences with the "default" value. E.g. if I have an optional boolean preference: if user set to true display X, if user set to false display Y, if user has not set do not display anything. In this situation, the getter always returning a default would prevent this. Not suggesting this is a good design, just saying the convention might need to be documented (maybe it is and I'm unaware of it!) if we want to go forward that way.

Mostly agree, apart from boolean having three states, if more then two states it would become a numeric or text with some values being 'not-defined'.

I guess I don't quite see what the need was for generic way of validating and typing prefs if they still can be null (and require further validation and typing when the generic way is used).

wlthomson:
> Sure. In desktop we now have default values for each of the prefs individually, which I think is the way to go

Ah OK, so default by pref rather than by type? I think that would solve my concerns.

> I guess I don't quite see what the need was for generic way of validating and typing prefs if they still can be null (and require further validation and typing when the generic way is used).

I think we're basically on the same page here, I think it was more just caution on my part not to assume too much re. desired behaviour. Will make a new issue.

andreievg:
Great thanks @wlthomson, can continue discussion there. Would be great to have a clear picture of generic way for prefs in mobile

PR #2566, opened 2020-03-23
#2540 add patient api ui

wlthomson:
Fixes #2540.

Branches off #2567. 

## Change summary

Just a rough start to get things going. UI looks like this:

#2540-add-patient-api-ui

## Testing

Steps to reproduce or otherwise test the changes of this PR:

- [ ] When toggling patient/prescriber datasets, the lookup button updates correctly (`"Lookup patient/Lookup prescriber"`).
- [ ] When the `"Lookup patient"` button is pressed, an empty modal is opened with the same title.
- [ ] When the `"Lookup prescriber"` button is pressed, an empty modal is opened with the same title.
- [ ] When the modal is closed, the user is returned to the dispensary page without any change to the previous state.

### Related areas to think about

Still to do:

- Add modal: #2541.
- Add lookup logic: #2542.

PR #2567, opened 2020-03-23
#2565 add text prop to page button memo

### LINKED ISSUE #2565: PageButtonComponent memoisation does not evaluate text prop
## Describe the bug

`PageButtonComponent` is exported as a memoised component, but does not currently evaluate the `text` prop. The effect of this is that if a buttons text is updated but the button callback is not, the button will not rerender.

### To reproduce

The bug cannot be reproduced currently, getting ahead of #2540!

### Expected behaviour

`propsAreEqual` should evaluate the `text` prop.

### Proposed Solution

Add the `text` prop to the  `propsAreEqual` evaluation.

### Version and device info

- App version: v4.0.2.
- Tablet model: Android emulator.
- OS version: API 29.

### Additional context

No existing bugs related to this AFAIK (found while working on #2540), so marking as `5.0.0` to easily branch off for #2540.

wlthomson:
Fixes #2565

## Change summary

Just a quick fix to the page button memoisation logic.

## Testing

Tests included in #2566.

Done as part of #2566, but felt like something that should be a separate PR!

### DIFF
--- a/src/widgets/PageButton.js
+++ b/src/widgets/PageButton.js
@@ -39,12 +39,14 @@ export function PageButtonComponent(props) {
 }
 
 const propsAreEqual = (
-  { isDisabled: prevIsDisabled, onPress: prevOnPress },
-  { isDisabled: nextIsDisabled, onPress: nextOnPress }
+  { isDisabled: prevIsDisabled, onPress: prevOnPress, text: prevText },
+  { isDisabled: nextIsDisabled, onPress: nextOnPress, text: nextText }
 ) => {
   const isDisabledAreEqual = prevIsDisabled === nextIsDisabled;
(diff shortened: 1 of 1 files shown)


diegoevangelisti:
## Testing

- [X] When toggling patient/prescriber datasets, the lookup button updates correctly (`"Lookup patient/Lookup prescriber"`). **-PASS-**
- [X] When the `"Lookup patient"` button is pressed, an empty modal is opened with the same title. **-PASS-**
- [X] When the `"Lookup prescriber"` button is pressed, an empty modal is opened with the same title. **-PASS-**
- [X] When the modal is closed, the user is returned to the dispensary page without any change to the previous state. **-PASS-**

Issue #2561, opened 2020-03-19
confirm button cut-off in the middle when it's on description mode while creating new transaction 


Rebecca367:
## Describe the bug

Confirm button partially cut-off when it's on description mode while creating a new transaction 
<img width="813" alt="Screen Shot 2020-03-19 at 10 38 37 PM" src="https://user-images.githubusercontent.com/57238075/77053008-89bd1200-6a32-11ea-9585-3046d0b76a6e.png">

### To reproduce

1. Go to 'cash register'
2. Click on 'new transaction'
3. Scroll down to type in 'description'

### Proposed Solution
N/A

### Version and device info

- App version: v4.0.2
- Tablet model: emulator
- OS version: macOS

### Additional context
Need more testing on this version due to some new behaviours appears.

joshxg:
The confirm button shouldn't be able to be clicked when the description bottom modal is up, as far as I know - maybe @wlthomson can confirm.

So I think the solution here is possible to either bring the button higher or increase the width of the bottom modal?

Rebecca367:
@joshxg @wlthomson 
I think 1.5 confirm button is confusing, so maybe better not show the middle confirm button?
The confirm button at right hand side is responsive to confirm description and it's working. 
It all depends on everybody's priority, I guess.

wlthomson:
Yeah both solutions work, but the easiest fix is to just hide the main confirm button when any modal is open (gets rid of the possibility of both being displayed at once). 

Agree with @joshxg this isn't super high priority, but def suitable for the next patch release.

wlthomson:
Fixed by #2561.

nishaSussol:
## TESTING
Tested in mobile version 4.0.3

### Tests
- [x] When making a new cash transaction, the "Confirm" button is displayed at the bottom of the modal. **-PASS-**
- [x] When making a new cash transaction, the "Confirm" button is not displayed at the bottom of the modal if another submodal is open (e.g. click "Enter a description"). **-PASS-**

PR #2535, opened 2020-03-11
#2476 modal titles not centred

mark-prins:
Fixes #2476 

## Change summary

Centred. Took the issue on so that I could fix a French translation - we had the `Supplier Requisitions` button showing `Invoices requisitions`

## Testing

Steps to reproduce or otherwise test the changes of this PR:

- [ ] Open a modal and see if centred
- [ ] Check the french translation of the supplier requisitions button

Issue #2532, opened 2020-03-11
Headline of "Patient Detail" when create a new patient need to move to center 


Rebecca367:
## Describe the bug

Headline of "Patient Detail" when create a new patient is not centered, so it needs to be moved to center. 

### To reproduce

Steps to reproduce the behaviour:

1. Go to ''Dispensary"

2. Click on 'New Patient'

### Expected behaviour

Headline should be placed in the center

### Proposed Solution
 Change spacing in modale to center

### Version and device info

- App version: v4.0.0rc
- Tablet model: emulator
- OS version: mac

### Additional context

<img width="835" alt="Screen Shot 2020-03-11 at 10 00 36 PM" src="https://user-images.githubusercontent.com/57238075/76456585-13ba1900-643c-11ea-8910-3ebeda070510.png">

joshxg:
appears to be a problem with all modal titles - so we can just use a single issue: https://github.com/openmsupply/mobile/issues/2476 

And I will close this one

Issue #2476, opened 2020-03-05
Modal titles are not centered


Rebecca367:
## Describe the bug
current select master list is not in the middle of row.
It's better to be in the middle of row.

![Screen Shot 2020-03-05 at 2 46 43 PM](https://user-images.githubusercontent.com/57238075/75939866-5b96e880-5ef0-11ea-85cb-fd5428e4514b.png)

A clear and concise description of what the bug is. Use screenshots!

### To reproduce

1. Go to 'customer invoice'
2. Click on 'new invoice'
3. Click on add master list items

### Expected behaviour
Header should be in the middle of row and supposedly with bigger font

### Proposed Solution
Get into Modal String to change font and distance from left and right?!

### Version and device info

- App version: 4.0.0 rc10
- Tablet model: emulator
- OS version:

joshxg:
Closed by #2535

PR #2570, opened 2020-03-24
#2554 fix modal strings not centred

### LINKED ISSUE #2554: Modal string is not centred( create new patient details and create a new prescriber)
## Describe the bug

Modal string is not centred when you create a new patient and asking for patient details and the same as to create a new prescriber.
Some related issues been closed but issues still remain.
https://github.com/openmsupply/mobile/issues/2532#issuecomment-597830571

Screen Shot 2020-03-17 at 1 50 45 PM
Screen Shot 2020-03-17 at 1 51 18 PM
Screen Shot 2020-03-19 at 10 14 34 AM

### To reproduce

1. Go to 'dispensary'
2. Click on 'new patient' and "new prescriber"
3. See error

### Expected behaviour

modal string to be centered

### Proposed Solution

Make textAlign to be centered.

### Version and device info

- App version: v4.0.1
- Tablet model: emulator
- OS version: 

### Additional context
similar to https://github.com/openmsupply/mobile/issues/2476

wlthomson:
Fixes #2554.

Branched from #2561.

## Change summary

Just a small style change to centre modal titles.

## Testing

- [ ] Modal titles are centred (e.g. create new patient, create new prescriber).

N/A.

### DIFF
--- a/src/widgets/modals/ModalContainer.js
+++ b/src/widgets/modals/ModalContainer.js
@@ -165,6 +165,7 @@ const localStyles = StyleSheet.create({
     fontFamily: APP_FONT_FAMILY,
     color: 'white',
     fontSize: 20,
+    textAlign: 'center',
   },
   closeButtonContainer: {
     flex: 1,


mark-prins:
I'm confused? I've done this already: https://github.com/openmsupply/mobile/pull/2535

wlthomson:
@mark-prins Yeah, the issue came up in testing on `master`. I saw your fix and just copied it here so we can close out the issue. Probably could have just left it for develop, but it was a bit distracting working with it on `master`!

I think good to keep bug fixes to patches and leave `develop` only for new functionality (fixes will get merged back into `develop` anyway). Since testing is on `master`, if a bug is fixed on `develop`, the tester won't see it and will probably raise it (I think what happened here?) Either way, this is targeted as a patch fix, which I plan to release today, so still worth getting in IMO.

wlthomson:
> in that case.. great work! looks good to me ;)

:laughing: . 

Was thinking maybe the above is a bit too strict. It's probably a losing game trying to change processes and enforce all non-functionality changes as patches, so rephrasing as "keep bug fixes to patches **where possible**" :smiley:.

mark-prins:
> > in that case.. great work! looks good to me ;)
> 
> 😆 .
> 
> Was thinking maybe the above is a bit too strict. It's probably a losing game trying to change processes and enforce all non-functionality changes as patches, so rephrasing as "keep bug fixes to patches **where possible**" 😃.

I agree with your `master` / `develop` and `feature` / `patch` distinctions. 
And this PR is a good reminder, thanks.

Rebecca367:
@wlthomson @mark-prins 
I think that I have issued one issue similar to these as well.
https://github.com/openmsupply/mobile/issues/2476

It's pretty hard to keep track of feature, develop and master branch and when to merge what to what...at least it's confusing to newbie, especially we don't know each step of work each other  is doing?! So one issue been solved but after a while the same issue occur again. It causes frustration over time.

wlthomson:
> @wlthomson @mark-prins
> I think that I have issued one issue similar to these as well.
> #2476
> #2532
> 
> It's pretty hard to keep track of feature, develop and master branch and when to merge what to what...at least it's confusing to newbie, especially we don't know each step of work each other is doing...maybe can find a solution for people's life easier?

I can understand for people new to git its a bit tricky, but our system is quite simple (simpler than most open-source software projects I've contributed to!). The processes are used because they make life easier, not harder! Removing our branching structure would make it much harder to work as a team on separate pieces of work (especially now that we are remote!) and also harder for our support team to document things and ensure customers are running compatible versions of desktop and mobile.

We're using "semantic versioning", which means we have three "types" of release: major, minor, and patch (e.g. `4.0.2` has a major version of `4`, a minor of `0`, and a patch of `2`).

The `master` branch is the current release. We target all releases at `master`. 

As we want to reduce the amount of merging into the main branches (`develop` and `master`), we prepare a release by merging related work into a `release` branch. In the case of a patch, that branch consists of a number of PRs. For minor/major, we usually just branch directly off `develop` (which will contain recently merged `features` etc., see below). 

When we are ready to release, we merge the release into `master` and `master` is merged back into `develop` (`develop` should always have everything `master` has in it).

That's the main workflow. Since you're not 100% comfortable with it, I would suggest not merging anything unless you're sure! When doing testing, if you are testing on `master` you are essentially testing a version that we feel can be released to customers (and may be already being used). If you are testing on `develop`, you are testing our "internal" version which has not yet been released.

Feature branches are not "core" to the workflow, they are just a tidy way of reducing the amount of "traffic" going into `develop`. Instead of constantly merging into `develop`, we make a `feature` branch and merge all related work into that. When the feature is ready, we merge the `feature` into `develop`. 

Hope that helps!

Rebecca367:
@wlthomson 
Thanks for explaining all these to me. It definitely helps. Would you recommend testers to test on master branch to get the most accurate results that users might be get after using mobile?

wlthomson:
@Rebecca367 It depends on the situation. Most of the time, you'll want to be testing on `develop`. Once something is in `master`, it's ready for customers and should already be tested. So everything in `master` should already be tested. But because bugs slip through, we will sometimes need patches (as explained above). So the only time you'll be testing on `master` is if you're testing a patch (remember that patches don't go to `develop` first, they go straight to `master`). 

Makes sense?

 The main thing is to make sure you wait until the work has been merged before you test it. If it's a PR for a patch, you'll want to check it's made its way to `master` before testing it.

Rebecca367:
@wlthomson 
cool, thank you for the explanation again. It's clear enough for me now.

PR #2581, opened 2020-03-26
Highlight/bold a row in Customer Invoice

### LINKED ISSUE #1657: Bold/ Highlight the current row
**Describe the bug**
Currently when editing a cell in a row, there is not enough visual feedback of what row the the user has focused.

**To reproduce**

Go to a unfinalised customer invoice
Add a few items if there aren't any
Focus textinput of quantity field of an item
WANT: the row that focused cell is in to be highlighted.

**Expected behaviour**
The row currently with focus in it should be highlighted in some manner. Either highlight the row by changing the background colour. BONUS: Get fancy and highlight the row and bold the text in the cell with focus if you like.

**Solution**
This will require some in-depth knowledge of the page and data table state management.

1. Add action constant "focusCell"
Create an action creator function in cellActions.js. Action payload should have rowKey and columnKey
2. Add a reducer function in cellReducers.js that immuntably sets: dataState.row[action.rowKey].focusedColumn = action.columnKey (likely catch, if the focus has moved from one row to another, need to clear the focusedColumn from the previous rowState object)
3. Add a function onFocusCell to getPageDispatchers that dispatches the above action.
TextInputCell.js needs prop for onFocus that passes into the TextInput props. The component should make a method that calls it with rowKey and columnKey.
4. DataTableRow.js needs to check rowState to see if it's "focusedColumn" has a value and highlight the row if so.
Version and device info
App version: all
Tablet model: all
OS version: all
Additional context
Original description:

When editing a stocktake in particular, but also requisitions etc., it can be easy for a user to enter the quantity on the row above or below the one they mean to. The contrasting lines help with that, but apparently not enough. Would be great to highlight or bold the currently edited item to make it clearer!

Rebecca367:
Fixes #1657 

## Change summary

Highlight a row on customer invoice when user focus on rows in customer invoice

## Testing

- [ ] Click on customer invoice
- [ ] Create a new invoice
- [ ] Click on a row 

Is it also a good feature apply on Supplier Invoice and Current Stock and Stocktake?


Rebecca367:
This feature is not right yet, issue this PR is for people to collaborate and for myself to access to those commits I have committed before. Welcome to make suggestions everyone.

mark-prins:
Hi @Rebecca367 - there isn't an issue created, so I'm not sure what the intention is or how this change came about. perhaps just create an issue to make things clearer?

mark-prins:
https://github.com/openmsupply/mobile/issues/1657#issuecomment-604728202

src/pages/dataTableUtilities/actions/cellActions.js
```diff
@@ -26,6 +26,11 @@ import { selectPageState } from '../../../selectors/pages';
  *
  * @param {String} rowKey Key of the row to edit.
  */
+export const focusCell = (rowKey, columnKey) => ({
+  type: ACTIONS.FOCUS_CELL,
+  payload: { rowKey, columnKey },
```
joshxg: I think you should only need the row key here as we would only need to know which row is being focused to highlight it. We also need the route for these actions i.e. like `refreshRow`

```
export const focusCell = (rowKey, route) => ({
  type: ACTIONS.FOCUS_CELL,
  payload: { rowKey, route },
```

Also the constant isn't defined

src/pages/dataTableUtilities/reducer/cellReducers.js
```diff
@@ -26,6 +26,11 @@ export const refreshIndicatorRow = state => {
   return { ...state, indicatorRows: currentIndicator.rows };
 };
 
+export const focusCell = (state, action) => {
+  const { dataState } = state;
+  dataState.row[action.rowKey].focusedColumn = action.columnKey;
+};
```
joshxg: you would need to make `dataState` have a new reference by making it a new set, otherwise the equality check for rendering will return equal for the table. Then the `rowKey` for the specific row in the set will need a new reference, otherwise the row will not re-render

i.e.

```
const newDataState = new Map(dataState);
const rowState = newDataState.get(rowKey);
newDataState.set(rowKey, { ...rowState, isFocused: true });
```

I think this might be better in the `rowReducers.js` file as it's more for a row rather than a cell, what do you think?

src/widgets/DataTable/DataTableRow.js
```diff
@@ -120,6 +120,7 @@ const DataTableRow = React.memo(
                   cellTextStyle={editableCellText}
                   rowIndex={rowIndex}
                   placeholder={placeholder}
+                  focusedColumn={SUSSOL_ORANGE}
```
joshxg: hmm the `TextInputCell` doesn't have a prop `focusedColumn` - I think this should be `onFocus` prop which you can pass a function which dispatches the `focusCell` action, is that what you're goin for?

src/widgets/DataTable/TextInputCell.js
```diff
@@ -55,7 +56,7 @@ const TextInputCell = React.memo(
 
     const { focusNextCell, getRefIndex, getCellRef } = React.useContext(RefContext);
     const refIndex = getRefIndex(rowIndex, columnKey);
-
+    const onFocus = () => onFocus(rowKey, columnKey);
```
joshxg: hmm I think you need to get the `onFocus` function from a prop and then you can create the function here

src/widgets/InfoBadge.js
```diff
@@ -4,6 +4,7 @@ import PropTypes from 'prop-types';
 
 import { View, Text } from 'react-native';
 import Popover from 'react-native-popover-view';
+// eslint-disable-next-line import/no-unresolved
```
joshxg: whats this for? 😱

PR #2582, opened 2020-03-26
#1364 Bluetooth module

### LINKED ISSUE #1364: Bluetooth low energy module

Need to add the BLE module created for vaccine module

Add:

https://github.com/openmsupply/mobile/tree/vaccine-development/android/app/src/main/java/com/msupplymobile/Ble

N/A

## Additional context

N/A

joshxg:
Fixes #1364 

## Change summary

- Just a port of the blue tooth module from the previous develop branch.
- I can see a few places where I'd like to improve a few things, but going to leave it for now - it's working and sending packets. I
- I think improvements can be done when we make a npm package?

## Testing

N/A [Will do tests when the feature branch is complete]

Little pic maybe helpful, maybe not. As above there are improvements needed/could do but the general idea of having specific 'device' concrete classes which relate to a specific physical sensor which is composed of a parser for a specific physical sensor implementation

https://drive.google.com/file/d/1nhtpZNPP1AWvaGzPhOVHqGETEOvAO0Mp/view?usp=sharing


android/app/src/main/java/com/msupplymobile/Ble/BleDevice.java
```diff
@@ -0,0 +1,77 @@
...
+    
+    protected BleDeviceListener bleDeviceListener;
+    protected ScanResult scanResult;
+
+    public BleDevice(ScanResult scanResult){
+        this.scanResult = scanResult;
+    }
+
+    /**  Registers a listener. Currently supports just one listener */
+    public void registerListener(BleDeviceListener bleLDeviceistener){
```
andreievg: a bit pedantic with spelling

mark-prins: wow, good spotting!

android/app/src/main/java/com/msupplymobile/Ble/BlueMaestroDevice.java
```diff
@@ -0,0 +1,200 @@
...
+     */
+    @Override
+    public WritableMap toObject(){
+        WritableMap asObject = this.parseAdvertisement();
+        asObject.putString("name", getName());
+        asObject.putString("macAddress", getAddress());
+        asObject.putInt("rssi", getRssi());
+        if (command.equals("*logall")){
+            asObject.putArray("logs", this.parseLogs());
+        }
```
andreievg, line 85: i think if command != "" we should add "commandString" or something analog

android/app/src/main/java/com/msupplymobile/Ble/BleManager.java
```diff
@@ -0,0 +1,202 @@
...
+        return "BleManager";
+    }
+
+    /** 
+     * Callback for handling commands sent to devices,
+     * whether it was succcesful, or not.
+     * Overrides BleDeviceListener
+     */
+    @Override
+    public void BleCallback(boolean result){
```
andreievg, line 52: little bit iffy about this one, the only 'better' way i can think of here is to just return the device and have it as a single object in 'data'. So different return to JS if we are scanning for devices ({devices: [devices]}), or getting result of command for one device ({device: {single device info}} <- actually i think that would make more sense. what cha reckon ?

Issue #1359, opened 2019-10-15
Add Location, ItemBatchLocationMovement data types


joshxg:
## Is your feature request related to a problem? Please describe.

Need to add `Location` and `ItemBatchLocationMovement` as well as `batch` changes for location tracking

## Describe the solution you'd like

Add:

![image](https://user-images.githubusercontent.com/35858975/77504098-46e0bb80-6ec4-11ea-875b-aced46a2f144.png)

## Describe alternatives you've considered

N/A

## Additional context

N/A

joshxg:
Closed by #2584

PR #2585, opened 2020-03-26
#1360 Items and doses

### LINKED ISSUE #1360: Add doses to `Item` and `..Batches`

Update `Item`, `ItemBatch`, `TransactionBatch` and `StocktakeBatch` data types 

Add:

image

N/A

## Additional context

N/A

joshxg:
### BRANCHED FROM: #1359 

Fixes #1360 

## Change summary

- Adds `TransactionBatch.doses`
- Adds `StocktakeBatch.doses`
- Adds `ItemBatch.doses`
- Adds `Item.dosesPerVial`
- Adds `Item.doses`

## Testing

N/A

N/A


src/database/DataTypes/Item.js
```diff
@@ -248,6 +248,8 @@ Item.schema = {
     unit: { type: 'Unit', optional: true },
     defaultRestrictedLocationType: { type: 'LocationType', optional: true },
     directions: { type: 'linkingObjects', objectType: 'ItemDirection', property: 'item' },
+    dosesPerVial: { type: 'double', default: 0 },
```
joshxg: do we like `dosesPerVial`?

src/database/DataTypes/TransactionBatch.js
```diff
@@ -185,6 +185,7 @@ TransactionBatch.schema = {
     type: { type: 'string', optional: true },
     linkedTransaction: { type: 'Transaction', optional: true },
     location: { type: 'Location', optional: true },
+    doses: 'double?',
```
joshxg: Was just copying the `numberOfPacks` definition - hoping there is either a reason and just following that

src/database/DataTypes/Item.js
mark-prins: `dosesPerVial` for me

mark-prins: no, wait. the `ItemBatch` has `doses` and I think the two should be consistent. I was choosing `perVial` because the specificity is helpful - but I'm now reconsidering and think that the general `doses` could be more useful

joshxg:
@mark-prins 

`Item.dosesPerVial` would be how many doses per 1 quantity of an item in a pack/batch. `ItemBatch.doses` is how many doses we have left.

Example: `ItemA` has `dosesPerVial = 2`. So, an `ItemBatch` with `quantity = 1` has `doses = quantity * `item.dosesPerVial`

Does that make anything clearer and would you still think `doses`?

mark-prins:
> Does that make anything clearer and would you still think `doses`?

It's just that `dosesPerVial` is specific to a unit type - it's implying that the medicine we are referring to is in a vial.  Which is true for things that come in doses, because that's vaccines which are liquid, AFAIK. For more info we'd have to ask a pharmacist 😃 as a general rule though - we have an item, which has a unit that tells us if the thing is a tablet, capsule, vial etc and then we have a property which says `dosesPerVial`. 

It's a minor point - but it means you can have a unit type of tablet and then specify the 'dosesPerVial' for that. `doses` works for me, but you could also go with `dosesPerUnit`

joshxg:
Thanks! Went with doses!

PR #2586, opened 2020-03-26
#1365 VVMStatus tracking types

### LINKED ISSUE #1365: Add `VaccineVialMonitorStatus` & `VaccineVialMonitorStatusLog` data types

Add `VaccineVialMonitorStatus` & `VaccineVialMonitorStatusLog` data types

Add:

image

N/A

## Additional context

N/A

joshxg:
### NOTE: Branched from #1360 

Fixes #1365 

## Change summary

- Adds `VaccineVialMonitorStatus`
- Adds `VaccineVialMonitorStatusLog`
- Adds `..Batch` field updates

## Testing

N/A

N/A


src/database/DataTypes/VaccineVialMonitorStatus.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * Sustainable Solutions (NZ) Ltd. 2020
+ * mSupply Mobile
+ */
+
+import Realm from 'realm';
+
+export class VaccineVialMonitorStatus extends Realm.Object {}
```
joshxg: I mean it's a long name.. but.. I dunno..!

src/database/DataTypes/VaccineVialMonitorStatus.js
```diff
@@ -0,0 +1,23 @@
...
+  name: 'VaccineVialMonitorStatus',
+  primaryKey: 'id',
+  properties: {
+    id: 'string',
+    description: { type: 'string', optional: true },
+    code: { type: 'string', optional: true },
+    level: { type: 'double', default: 0 },
+    shouldDispense: { type: 'bool', default: false },
+    isActive: { type: 'string', default: false },
+  },
```
joshxg: so: 
- `level` is to be able to order the status' when displaying. But also I am realizing we also want to assosciate an icon with it.. so.. ideas?
-`shouldDispense` is I think not needed because we're not auto dispensing?
- `isActive` is just for your standard soft deletion mechanism for system data.. seems better to have it mandatory for any system data rather than coming back in 1-2 years to do it..
- `description`/`code`.. self explanatory.. but .. needed?

Are these going to be user defined or system defined ? We could allow for a choice maybe, "use 4 level system" or something if we ever need ?

src/database/DataTypes/VaccineVialMonitorStatusLog.js
```diff
@@ -0,0 +1,21 @@
...
+export class VaccineVialMonitorStatusLog extends Realm.Object {}
+
+VaccineVialMonitorStatusLog.schema = {
+  name: 'VaccineVialMonitorStatusLog',
+  primaryKey: 'id',
+  properties: {
+    id: 'string',
+    status: 'VaccineVialMonitorStatus',
+    date: { type: 'date', default: new Date() },
+    itemBatch: 'ItemBatch',
```
joshxg: should this have a `Option`?

src/database/DataTypes/VaccineVialMonitorStatus.js
mark-prins: ...but not as long as `VaccineVialMonitorStatusLog`!

src/database/DataTypes/VaccineVialMonitorStatus.js
mark-prins: probs not for me to comment on.. but fwiw:
- code isn't very helpful, as level is more of an identifier for the status
- some sort of mapping for icons? have an icon with a property of `level` to match them up? or an `icon` prop
- can't see why `shouldDispense` is needed
- system defined makes sense to me 
- a way of translating the descriptions would be helpful

😃

src/database/DataTypes/VaccineVialMonitorStatusLog.js
mark-prins: uh, pardon my ignorance. why? I'm missing something..

joshxg: So that we can see the full 'ledger' or historic vvm status' and when they changed of an item batch, an item batch can have many 'logs', where the most recent one is the current status. So you can see when a vaccine has transitioned from level 1 -> 2 -> 3 -> 4 and when. Make sense? Thoughts?

src/database/DataTypes/VaccineVialMonitorStatus.js
joshxg: > a way of translating the descriptions would be helpful

Hmm.. maybe using `code`?

mark-prins: pfft..! you and your crazy ideas 🙄

src/database/DataTypes/VaccineVialMonitorStatusLog.js
mark-prins: I'm still confused, but I won't comment on a merged PR because that could be confusing in itself.

I get the many 'logs' attached to an item batch and the use of a ledger display. I don't see why `Option` is mentioned or what that should do. but don't worry, I don't think I need to know.

joshxg: ohhh - sorry. And also sorry for replying to you telling me not to reply. I meant should the 'log' have a 'reason' - i.e. make a new category of 'reasons' for vvm status changes so a user could select why the vvm status changed.. i.e. 'fridge exploded', 'left fridge door open oops!' etc..

sorrrrrry 🤣

mark-prins: oh, yeah, right. that makes sense - thanks! 
and now I've replied too.. oops! In that case, yes. `Option` would be good

PR #2612, opened 2020-03-28
#2611 add patient api form changes

### LINKED ISSUE #2611: Update FormControl logic to handle search forms

As described in #2609, the patient API search UI uses the existing `FormControl` component for entering search parameters. This requires some additions to the existing logic.

Update `FormControl` logic to handle requirements for searching for a record (as well as creating/editing a record).

## Implementation

First round will probably be a bit cludgey, but idea is to see what works and then potentially refactor towards the end if needed:

- Add `SEARCH_X` cases to `formInputConfigs` (same as existing cases, but different validation logic).
- Update `FormControl` component to render different buttons when used for searching (i.e. save button is search button, no cancel button). 

N/A.

## Additional context

See epic #2446.

wlthomson:
Fixes #2611.

Branched off #2566.

## Change summary

Not the nicest solution, a bit of a cludge at this stage! 

Adds specific form configs for search use-cases (see self-review for more info).

## Testing

N/A.

N/A.


src/utilities/formInputConfigs.js
```diff
@@ -57,6 +62,16 @@ const FORM_INPUT_CONFIGS = seedObject => ({
     invalidMessage: formInputStrings.must_not_be_empty,
     isEditable: true,
   },
+  [FORM_INPUT_KEYS.SEARCH_FIRST_NAME]: {
+    type: 'text',
+    initialValue: '',
+    key: 'firstName',
+    validator: () => true,
+    isRequired: false,
```
wlthomson: As searching, there is not requirement for a `firstName` to be specified. At a later stage may be worth adding some validation logic, but good chance of a refactor later down the line, so I think best to leave it simple for now!

src/utilities/formInputConfigs.js
```diff
@@ -57,6 +62,16 @@ const FORM_INPUT_CONFIGS = seedObject => ({
...
+    label: formInputStrings.first_name,
+    invalidMessage: '',
```
wlthomson: See above, no `invalidMessage` specified as no way for the input to be invalid!

src/utilities/formInputConfigs.js
```diff
@@ -67,6 +82,16 @@ const FORM_INPUT_CONFIGS = seedObject => ({
     invalidMessage: formInputStrings.must_not_be_empty,
     isEditable: true,
   },
+  [FORM_INPUT_KEYS.SEARCH_LAST_NAME]: {
```
wlthomson: See above, same use case as `firstName`.

src/utilities/formInputConfigs.js
```diff
@@ -92,6 +117,22 @@ const FORM_INPUT_CONFIGS = seedObject => ({
     label: formInputStrings.date_of_birth,
     isEditable: true,
   },
+  [FORM_INPUT_KEYS.SEARCH_DATE_OF_BIRTH]: {
+    type: 'date',
+    initialValue: '',
+    key: 'dateOfBirth',
+    invalidMessage: formInputStrings.must_be_a_date,
+    isRequired: false,
+    validator: input => {
```
wlthomson: Similar to existing date logic, but `validator` allows a blank input to enable the use case where a `dateOfBirth` is not specified.

src/utilities/formInputConfigs.js
```diff
@@ -157,6 +198,16 @@ const FORM_INPUT_CONFIGS = seedObject => ({
     invalidMessage: formInputStrings.must_be_between_0_and_50,
     isEditable: true,
   },
+  [FORM_INPUT_KEYS.SEARCH_REGISTRATION_CODE]: {
```
wlthomson: See above, same logic as described for `firstName`.

src/utilities/formInputConfigs.js
```diff
@@ -167,6 +218,16 @@ const FORM_INPUT_CONFIGS = seedObject => ({
     label: formInputStrings.personal_policy_number,
     isEditable: !seedObject,
   },
+  [FORM_INPUT_KEYS.SEARCH_POLICY_NUMBER_PERSON]: {
```
wlthomson: See above, same logic as described for `firstName`.

src/utilities/formInputConfigs.js
```diff
@@ -177,6 +238,16 @@ const FORM_INPUT_CONFIGS = seedObject => ({
     label: formInputStrings.family_policy_number,
     isEditable: !seedObject,
   },
+  [FORM_INPUT_KEYS.SEARCH_POLICY_NUMBER_FAMILY]: {
```
wlthomson: See above, same logic as described for `firstName`.

src/widgets/FormControl.js
```diff
@@ -54,10 +55,7 @@ const FormControlComponent = ({
     setRefs({ length: inputConfig.length });
   }, []);
 
-  const onSaveCompletedForm = React.useCallback(() => onSave(completedForm), [
-    onSave,
-    completedForm,
-  ]);
+  const onSaveCompletedForm = useCallback(() => onSave(completedForm), [onSave, completedForm]);
```
wlthomson: Oops, no actual logic changes here (just changed from `React.useCallback`).

src/widgets/FormControl.js
```diff
@@ -46,6 +46,7 @@ const FormControlComponent = ({
   onSave,
   inputConfig,
   isDisabled,
+  isSearchForm,
```
wlthomson: A bit of a cludge, but the simplest solution I could think of for now!

src/widgets/FormControl.js
```diff
@@ -169,6 +167,36 @@ const FormControlComponent = ({
       }
     );
 
+  const Buttons = useCallback(
```
wlthomson: Again, a bit of a cludge, but gets the job done for now!

src/widgets/FormInputs/FormDateInput.js
```diff
@@ -60,8 +60,8 @@ export const FormDateInput = React.forwardRef(
 
     const [inputState, setInputState] = React.useState({
       isValid: true,
-      inputValue: moment(initialValue).format('DD/MM/YYYY'),
-      pickerSeedValue: initialValue.toDate(),
+      inputValue: initialValue.isValid() ? moment(initialValue).format('DD/MM/YYYY') : '',
+      pickerSeedValue: initialValue.isValid() ? initialValue.toDate() : new Date(),
```
wlthomson: Another change which is a bit cludgey. See above re. date logic, had to make this change here to allow a blank date input.

joshxg: `initialValue?.isValid()` prob a good idea

src/widgets/FormControl.js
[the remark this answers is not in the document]
wlthomson: Yeah, as commented, this "solution" is more of a temporary cludge to get things working to make sure the whole thing is feasible. For that reason I think best to leave it for now (if OK with you). 

I'll probably need to think of a better way of doing it. I agree with the props can be more granular, but my thinking was that it might be a bit of a smell that we have to specify props to a component to tell it what children to render (I think also, even with flex, there are often niggly issues with margins etc. which will prob require passing style props). 

I was thinking perhaps of separating out the buttons from component altogether? In any case, I think maybe something to return to in a new issue once the rest of the UI/functionality is in decent shape?

src/utilities/formInputConfigs.js
```diff
@@ -29,17 +29,22 @@ import { formInputStrings } from '../localization';
 
 const FORM_INPUT_KEYS = {
   FIRST_NAME: 'firstName',
+  SEARCH_FIRST_NAME: 'firstName',
```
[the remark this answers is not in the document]
wlthomson: Good catch! Thought I'd put an obvious error in there for you to find to keep you on your toes! It was COMPLETELY ON PURPOSE.

ohno

e064a475d248ad91d9301696ce2842d9bb322968

Issue #2446, opened 2020-03-02
Patient lookup API 


joshxg:
## Description

Epic for managing mobile work for patient lookup API. 

Overview:

- Add an API to desktop which allows mobile to query for patients (e.g. by name, DOB, insurance number). 
- Update the mobile dispensing page to allow prescribers to lookup patients who are not "local" to their store/site.
- Add logic to update the visibility for patient/prescriber which are added to the current store. 
- Update the dispensing wizard to enforce certain constraints when prescribing to these patients, e.g. cannot edit patient details, cannot edit patient insurance etc.

Constraints:

- The prescriber/patient home store should not be changed. 
- The prescriber/patient should be "read-only" on the satellite site:
  - New prescriptions/policies can be created.
  - Existing records cannot be edited (including linked records, e.g. insurance policies).
- Record visibility should be updated on the server on incoming sync (i.e. mobile outgoing sync).

## Issues

- [x] Add patient lookup UI to dispensing page. #2540.
- [x] Add patient lookup modal to dispensing page #2541.
- [x] Add patient lookup logic to dispensing page. #2542.
- [x] Update prescription tabs to enforce read-only constraints for non-local patients. #2543.
- [x] Add patient API network helpers. #2616.
- [x] Add bugsnag to catch runtime API errors. #2632.
- [x] Add onPress logic for selecting records from API search. #2633.
- [x] Update search form components. #2645.
- [x] Add patient API network authentication logic. #2651.
- [x] Move network helper calls to thunks. #2654.
- [x] Add logic for creating insurance policies. #2657.
- [x] Extract patient API record creation helpers. #2664.
- [x] Add insurance policy creation logic thunks. #2678.
- [x] Add response code handling to API network helpers. #2681
- [x] Lookup modal may cause unmounted state changes. #2687.
- [x] Read-only constraints not being consistently applied. #2689.

## Specification docs

- Desktop epic: https://github.com/sussol/msupply/issues/5821.
- Internal wiki (in progress): https://sussol.net/wiki/doku.php/msupply:specifications:customers:lookup_api.
- Discussion issues:  #2544.

wlthomson:
I think I'm missing a bit of background info on this, so if any of this is wrong, let me know.

AFAICS, this functionality can be broken down into 3 components:

1. desktop side API.

- design the API specs (e.g. endpoint, request/response format). 
- add logic for looking up a specified patient (or patients?), and return that patient(s) details in a given format.

2. mobile side UI, logic.

[...]

PR #2615, opened 2020-03-28
#2609 add patient api search form

### LINKED ISSUE #2609: Add SearchForm component

As part of the patient API, a search form is required to enter search parameters and view server records.

- A form for entering patient/prescriber details, similar to existing new patient/new prescriber form.
- A list for displaying results, similar to existing data table or modal lists.

## Implementation

- Add new `SearchForm` component.
- Use existing `FormControl` component for form.
- Use RN `FlatList` component for displaying results.

N/A.

## Additional context

See epic #2446.

wlthomson:
Fixes #2609.

Branched off #2612.

## Change summary

Adds basic search form for retrieving and displaying patient/prescriber records. 

Uses a combo of `FormControl` and `FlatList` to provide a search interface and list for rendering results. Still just minimal functionality and UI, but gives a good idea of the basic design.

## Testing

N/A.

N/A.

### DIFF
--- a/src/widgets/modals/SearchForm.js
+++ b/src/widgets/modals/SearchForm.js
@@ -17,143 +17,145 @@ import { TouchableOpacity } from 'react-native-gesture-handler';
 import { queryPatientApi, queryPrescriberApi } from '../../utilities/network/lookupApi';
 
 const RECORD_TYPES = {
-    PATIENT: 'patient',
-    PRESCRIBER: 'prescriber',
-}
+  PATIENT: 'patient',
+  PRESCRIBER: 'prescriber',
+};
 
 const FORM_CONFIGS = {
-    [RECORD_TYPES.PATIENT]: 'searchPatient',
-    [RECORD_TYPES.PRESCRIBER]: 'searchPrescriber',
-}
+  [RECORD_TYPES.PATIENT]: 'searchPatient',
+  [RECORD_TYPES.PRESCRIBER]: 'searchPrescriber',
+};
 
 const LIST_CONFIGS = {
-    [RECORD_TYPES.PATIENT]: [ 'firstName', 'lastName', 'dateOfBirth' ],
-    [RECORD_TYPES.PRESCRIBER]: [ 'firstName', 'lastName', 'registrationCode' ],
-}
+  [RECORD_TYPES.PATIENT]: ['firstName', 'lastName', 'dateOfBirth'],
+  [RECORD_TYPES.PRESCRIBER]: ['firstName', 'lastName', 'registrationCode'],
+};
 
 const SearchListItemColumn = ({ value, type }) => (
-    <View style={localStyles.columnContainer}>
-        <Text style={localStyles.text}>
-            {type === 'date' ? value.toDateString() : value}
-        </Text>
-    </View>
+  <View style={localStyles.columnContainer}>
+    <Text style={localStyles.text}>{type === 'date' ? value.toDateString() : value}</Text>
+  </View>
 );
 
 const SearchListItem = ({ listItem, listConfig }) => {
-    const onPress = item => console.log(item);
-    const columns = listConfig.map(({ key, type}) => {
-        const value = listItem[key];
-        return <SearchListItemColumn key={key} value={value} type={type}/>;
-    });
-    return (
-        <TouchableOpacity onPress={onPress} style={localStyles.rowContainer}>
-            {columns}
-        </TouchableOpacity>
+  const onPress = item => console.log(item);
+  const columns = listConfig.map(({ key, type }) => {
+    const value = listItem[key];
+    return <SearchListItemColumn key={key} value={value} type={type} />;
+  });
+  return (
+    <TouchableOpacity onPress={onPress} style={localStyles.rowContainer}>
+      {columns}
+    </TouchableOpacity>
+  );
+};
+
+export const SearchForm = ({ dataSource }) => {
+  const [data, setData] = useState([]);
+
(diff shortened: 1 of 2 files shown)


src/widgets/modals/SearchForm.js
```diff
@@ -0,0 +1,159 @@
...
+    listConfig: PropTypes.array.isRequired,
+}
+
+SearchListItemColumn.propTypes = {
+    value: PropTypes.oneOfType([PropTypes.string, PropTypes.instanceOf(Date)]).isRequired,
+    type: PropTypes.string.isRequired,
+}
+
+  
+const localStyles = StyleSheet.create({
```
wlthomson: Don't like the reliance on absolute values for the styles here, but gets the job done for now. Planning to address some of these in a dedicated UI tidy up issue once the main functionality is implemented.

src/widgets/modals/SearchForm.js
```diff
@@ -0,0 +1,159 @@
...
+
+const SearchListItemColumn = ({ value, type }) => (
+    <View style={localStyles.columnContainer}>
+        <Text style={localStyles.text}>
+            {type === 'date' ? value.toDateString() : value}
+        </Text>
+    </View>
+);
+
+const SearchListItem = ({ listItem, listConfig }) => {
```
joshxg: I think it might've been a better idea to adapt the current `ResultRow` component which is used in the `AutoCompleteSelector` (even though that would need a refactor and would be more work)

src/widgets/modals/SearchForm.js
```diff
@@ -0,0 +1,159 @@
...
+import { queryPatientApi, queryPrescriberApi } from '../../utilities/network/lookupApi';
+
+const RECORD_TYPES = {
+    PATIENT: 'patient',
+    PRESCRIBER: 'prescriber',
+}
+
+const FORM_CONFIGS = {
+    [RECORD_TYPES.PATIENT]: 'searchPatient',
+    [RECORD_TYPES.PRESCRIBER]: 'searchPrescriber',
```
joshxg: I think putting this `FORM_CONFIGS` in with the `formInputConfigs`, then if the `dataSource` is just one of these constants I think it would be a lot simpler.

src/widgets/modals/SearchForm.js
```diff
@@ -0,0 +1,159 @@
...
+    if (!RECORD_TYPES.includes(dataSource)) return null;
+
+    const configKey = useMemo(() => FORM_CONFIGS[dataSource], [dataSource]);
+    const formConfig = useMemo(() => getFormInputConfig(configKey), [configKey]);
+
+    const listConfig = useMemo(() => {
+        const keys = LIST_CONFIGS[dataSource];
+        const keyTypes = keys.reduce((acc, key) => ({ ...acc, [key]: formConfig.find(config => config.key === key)?.type }), {});
+        return keys.map(key => ({ key, type: keyTypes[key] }));
+    }, [formConfig]);
```
joshxg: this is all needed because of the date? I think it might be nicer to have the response have the details formatted prior to getting to this component, if that makes sense? There is already complexity I imagine where the request is being managed, so I think putting it there?

src/widgets/modals/SearchForm.js
wlthomson: I did consider that, but came to the conclusion that you said re. it being a bit more work at this stage. Kind of still trying to do the minimum changes to existing code just to get a demo working. Once everything works, will definitely revisit suggestions for refactors like this!

src/widgets/modals/SearchForm.js
wlthomson: Good point, I do think that would be a nicer way to do it, but probably best for a separate issue as it will touch other components (I'm thinking one issue to put all the constants into `formInputConfigs` and update all components with references to e.g. `'patient'` to use them).

src/widgets/modals/SearchForm.js
wlthomson: At the moment just the date, yeah, but would also be needed for e.g. numbers, etc. My thinking was that its not so much the formatting of the response, it's more how to display it in a list. As more info will be being returned than is displayed, it seemed a good idea to separate the view from the data, and leave how to display the data to the component. Also a bit cleaner as if the row is selected, the data doesn't have to be converted back again to enter the record into the database.

Not sure if that's a good argument or not :man_shrugging:. I'm cool with making the change if you still think its better!

src/widgets/modals/SearchForm.js
```diff
@@ -0,0 +1,159 @@
...
+        switch (dataSource) {
+            case RECORD_TYPES.PATIENT: {
+                queryPatientApi(params).then(patientData => setData(patientData));
+                break;
+            }
+            case RECORD_TYPES.PRESCRIBER: {
+                queryPrescriberApi(params).then(prescriberData => setData(prescriberData));
+                break;
+            }
+            default: queryPatientApi(params);
```
mark-prins: I just don't see the need? I know the line will never be called.. but if it was, then it would query the API and do nothing with the response

src/widgets/modals/SearchForm.js
wlthomson: Updated epic #2446 with related issue.

src/widgets/modals/SearchForm.js
wlthomson: I agree, I think this was probably just me not wanting to leave the default blank, but I think it's probably a candidate for a bugsnag call!

Removed for now. 689d6fe9db6844b44c8823ab21c7398be62cb2ce.

Issue #2580, opened 2020-03-25
Sensor class methods


joshxg:
## Is your feature request related to a problem? Please describe.

Add the `Sensor` data type methods

## Describe the solution you'd like

Add the following methods:

![image](https://user-images.githubusercontent.com/35858975/77505417-f3706c80-6ec7-11ea-9251-bff492ece8b0.png)

## Implementation

N/A

## Describe alternatives you've considered

N/A

## Additional context

Previous `Sensor` data type: https://github.com/openmsupply/mobile/blob/vaccine-development/src/database/DataTypes/Sensor.js

joshxg:
Closed by #2622

PR #2623, opened 2020-03-29
#2579 Item methods [D4]

### LINKED ISSUE #2579: Additional Item methods

Need a few `Item` data type methods

Add the following methods:

image

## Implementation

N/A

N/A

## Additional context

Dependent on `ItemBatch`

joshxg:
#### NOTE: Branched from #2580 

Fixes #2579 

## Change summary

- Adds `Item` methods!

## Testing

N/A

- Don't like the use of reduce but since this should always be a pretty small amount of transactions (since the most recent requisition, so either monthly - or if not ordering monthly, should probably have a small amount of transactions), it should be ok. Would be possible to just use `UIDatabase` but I think leaving it like this and seeing if there is a performance problem later is the better decision
- I think it might be better to use only finalised requisitions however the last implementation didn't, so have left it

### DIFF
--- a/src/database/DataTypes/Item.js
+++ b/src/database/DataTypes/Item.js
@@ -282,14 +282,14 @@ export class Item extends Realm.Object {
   closedVialWastage(fromDate) {
     if (!this.isVaccine || !fromDate) return 0;
 
-    const customerCreditTransactionItems = this.transactionItems.filtered(
+    const supplierCreditTransactionItems = this.transactionItems.filtered(
       'transaction.confirmDate > $0 && transaction.type == $1 && transaction.otherParty.name == $2',
       fromDate,
       'supplier_credit',
       'inventory_adjustment'
     );
 
-    const totalDosesPossible = customerCreditTransactionItems.reduce(
+    const totalDosesPossible = supplierCreditTransactionItems.reduce(
       (dosesPossible, { batches }) => dosesPossible + batches.sum('numberOfPacks') * this.doses,
       0
     );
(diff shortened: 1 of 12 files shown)


src/database/DataTypes/Item.js
```diff
@@ -168,6 +168,18 @@ export class Item extends Realm.Object {
     return this.unit?.units ?? generalStrings.not_available;
   }
 
+  /**
+   * @return {Date} The date of the most recent confirmed requisition for this item.
+   */
+  get lastRequisitionDate() {
+    const mostRecentRequisitionItem = this.requisitionItems.sorted(
```
mark-prins: should this perhaps be filtering for confirmed reqs? The comment indicates that we should expect confirmed only.
should it sort by `confirmDate` rather than `entryDate`?

src/database/DataTypes/Item.js
```diff
@@ -228,6 +240,63 @@ export class Item extends Realm.Object {
...
+  /**
+   * Closed vial wastage differs from open vial wastage in that closed vial wastage
+   * is 'standard wastage'.
+   * @param {Date} fromDate
+   * @param {Number} the number of DOSES wasted.
+   */
+  closedVialWastage(fromDate) {
+    if (!this.isVaccine || !fromDate) return 0;
+
+    const customerCreditTransactionItems = this.transactionItems.filtered(
```
mark-prins: Should this only be negative inventory adjustments?
And is the assumption that any invad is a 'wastage'?

PR #2642, opened 2020-03-31
#2638 additional breach methods

### LINKED ISSUE #2638: Add TemperatureBreach and TemperatureBreachConfiguration additional methods

Would be good to have a couple more methods for temperature breaches

Methods:

- `TemperatureBreach.willContinueBreach(TemperatureLog)` : takes a temperature breach and determines if it would carry on the breach
- `TemperatureBreachConfiguration.createBreach(database, location, temperatureLog)` : Takes a location and temperature log and if the temperature log should create a breach, creates one

## Implementation

N/A

N/A

## Additional context

N/A

joshxg:
Fixes #2638 

## Change summary

- Adds `TemperatureBreach.willContinueBreach` method : This will be used when creating a `TemperatureLog` - if the `Location` is currently in a breach, can simply call this with the created `TemperatureLog` to check if the breach should continue.
- Adds `TemperatureBreachConfiguration.createBreach` method : currently this will be used on a `TemperatureLog` where the current location is not in breach. Couple q's on opinions below!

## Testing
N/A

The config `createBreach` method could be broken up into three methods:

```
`temperatureWithinBreachRange = (temperatureLog) => *checks if temperature within breach temperature range*`

`temperatureOfProximalLogsWithinBreachRange = (temperatureLog) => *checks if proximal logs duration and are all in the breach temperature range*

`createBreach = (temperatureLog) => *creates breach, sets all proximal logs*

```

However you would when creating a breach need to fetch the proximal logs twice, rather than once. Not sure which way is better. I have a feeling this second way is actually better but decided to PR and get any opinions and get it off my mind for a few hours so that I can come back and hate it. Oh well, it'll be basically the same logic/code !

### DIFF
--- a/src/database/DataTypes/TemperatureBreachConfiguration.js
+++ b/src/database/DataTypes/TemperatureBreachConfiguration.js
@@ -4,8 +4,69 @@
  */
 
 import Realm from 'realm';
+import moment from 'moment';
+import { createRecord } from '../utilities';
 
-export class TemperatureBreachConfiguration extends Realm.Object {}
+export class TemperatureBreachConfiguration extends Realm.Object {
+  /**
+   * With the provided temperatureLog, determines if the passed location
+   * was in a breach at the time of the temperature log and if so, will
+   * create a temperature breach record. Also adds to each of the temperature
+   * logs which are currently part of that breach, the related breach
+   * record.
+   *
+   * A breach has a duration and minimum and maximum temperatures defined by this
+   * configuration - for example above 8 degrees for 2 hours. To determine if a
+   * location has currently been breached, look back from the temperature logs
+   * timestamp the duration of this configuration and determine if every temperature
+   * log since that time has been outside of the threshold.
+   *
+   * @param {Realm}          database App-wide database interface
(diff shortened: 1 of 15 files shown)


src/database/DataTypes/TemperatureBreachConfiguration.js
```diff
@@ -4,8 +4,69 @@
...
+   * configuration - for example above 8 degrees for 2 hours. To determine if a
+   * location has currently been breached, look back from the temperature logs
+   * timestamp the duration of this configuration and determine if every temperature
+   * log since that time has been outside of the threshold.
+   *
+   * @param {Realm}          database App-wide database interface
+   * @param {Location}       location A Location record to create a breach for.
+   * @param {TemperatureLog} temperatureLog A potential log which may create a breach.
+   */
+  createBreach(database, location, temperatureLog) {
```
mark-prins, line 28: I like the suggestion of splitting up this method. Like you say, it is actually fine like it is - I think it would be slightly neater if split into the three methods.

I'd filter the logs in `createBreach` and pass the `proximalLogs` down rather than fetching twice

Issue #2563, opened 2020-03-19
App throws error when click on "dispense" in dispensary module


Rebecca367:
## Describe the bug

App throws error messages when you click on "dispense" for a patient in dispensary. 

<img width="803" alt="Screen Shot 2020-03-19 at 11 31 03 PM" src="https://user-images.githubusercontent.com/57238075/77058660-a0676700-6a3a-11ea-82a0-89f05cccebf1.png">
<img width="838" alt="Screen Shot 2020-03-19 at 11 29 54 PM" src="https://user-images.githubusercontent.com/57238075/77058454-57afae00-6a3a-11ea-89ff-c1493cfb43b2.png">

 
### To reproduce

1. Go to 'dispensary'
2. Create a patient then Click on 'dispense'
3. See error

### Expected behaviour

go to dispense mode

### Proposed Solution

N/A

### Version and device info

- App version: v4.0.2
- Tablet model: emulator
- OS version: macOS

joshxg:
Duplicate of #2537

PR #2652, opened 2020-04-02
#2633 add lookup api on press logic

### LINKED ISSUE #2633: Add on press logic for selecting records from API search

The basic API workflow requires users to select a patient/prescriber from the list of records which match the given search parameters. 

1. User searches for patient/prescriber.
2. User gets a list of matches.
3. User selects the row in the list which matches the patient/prescriber.
4. The user receives feedback that the patient/prescriber has been added. 

## Implementation

- Add an `onPress` handler to handle user selection.
- Add a popup confirmation on selection.

N/A.

## Additional context

N/A.

wlthomson:
Fixes #2633.

## Change summary

Adds logic for handling patient/prescriber selection.

## Testing

- [ ] Selecting a patient record closes the lookup modal and adds that patient to the dispensary patient list.
- [ ] Selecting a prescriber record closes the lookup modal and adds that patient to the dispensary patient list.

Completes the basic logic for looking up and adding a patient/prescriber. Next steps are to tidy up components and migrate logic to redux.

### DIFF
--- a/src/utilities/network/lookupApi.js
+++ b/src/utilities/network/lookupApi.js
@@ -6,7 +6,10 @@
 import { SETTINGS_KEYS } from '../../settings';
 import { UIDatabase } from '../../database';
 
-const { SYNC_URL } = SETTINGS_KEYS;
+import { NAME_TYPES } from '../../sync/syncTranslators';
+import { getOrCreateAddress, parseBoolean, parseDate } from '../../sync/incomingSyncUtils';
+
+const { THIS_STORE_ID, SYNC_URL } = SETTINGS_KEYS;
 
 const RESOURCES = {
   PATIENT: '/api/v4/patient',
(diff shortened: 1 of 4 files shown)


src/widgets/modals/SearchForm.js
```diff
@@ -37,8 +41,7 @@ const SearchListItemColumn = ({ value, type }) => (
   </View>
 );
 
-const SearchListItem = ({ listItem, listConfig }) => {
-  const onPress = item => console.log(item);
+const SearchListItem = ({ listItem, listConfig, onPress }) => {
```
joshxg, line 44: should probs memo this with `React.memo` and the column

src/widgets/modals/SearchForm.js
```diff
@@ -68,8 +71,24 @@ export const SearchForm = ({ dataSource }) => {
   }, [formConfig]);
 
   const renderItem = useMemo(
-    () => ({ item }) => <SearchListItem key={item.id} listItem={item} listConfig={listConfig} />,
-    [listConfig]
+    () => ({ item }) => {
+      const { id: key } = item;
+      const onPress = () => {
```
joshxg, line 76: Should separate this out from within this function so there isn't a new reference each time - so then the `SearchListItem` component can properly be memoized. With this layout there's no caching at all

src/utilities/network/lookupApi.js
```diff
@@ -6,7 +6,10 @@
 import { SETTINGS_KEYS } from '../../settings';
 import { UIDatabase } from '../../database';
 
-const { SYNC_URL } = SETTINGS_KEYS;
+import { NAME_TYPES } from '../../sync/syncTranslators';
+import { getOrCreateAddress, parseBoolean, parseDate } from '../../sync/incomingSyncUtils';
```
joshxg, line 10: should definitely extract these out!

src/utilities/network/lookupApi.js
```diff
@@ -6,7 +6,10 @@
 import { SETTINGS_KEYS } from '../../settings';
 import { UIDatabase } from '../../database';
 
-const { SYNC_URL } = SETTINGS_KEYS;
+import { NAME_TYPES } from '../../sync/syncTranslators';
```
joshxg, line 9: an issue for a `sync/constants` file what you reckon

src/widgets/modals/SearchForm.js
wlthomson: Agree 100%, this is a really bad cludge! Should be fixed in #2563.

src/utilities/network/lookupApi.js
wlthomson: Agreed! Add sync constants is on the patient API epic issues list (no issue created yet, but will do once I work my way to it!).

src/utilities/network/lookupApi.js
wlthomson: Agreed! I think there is a case for extracting out `sync` helpers as well as `type` helpers (we have a lot of run-time type enforcement stuff throughout the app). It's on the list (see below).

src/widgets/modals/SearchForm.js
wlthomson: Maybe incorrect, don't pure components only re-render on updated props (so equivalent to `memo` with all props as deps)?

src/utilities/network/lookupApi.js
```diff
@@ -19,68 +22,219 @@ const SEPARATORS = {
   POLICY_NUMBER: '-',
 };
 
-const getQueryString = ({
+export const createPatientRecord = response => {
```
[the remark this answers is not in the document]
wlthomson: Yeah, agree. The main issue is that the sync creators expect the incoming data to be in "external" format, so I think the tidiest way of doing it might be to make a separate creator helper which both use (and maybe the incoming sync logic maps the exterrnal fields to internal before calling it?).

src/widgets/modals/SearchForm.js
joshxg: pure components yeah - this is functional component (pure component is a class inheriting from `React.PureComponent` - functional components just re-render whenever called. `React.memo(functionalComponent)` makes the functional component equivalent to a `PureComponent`

src/widgets/modals/SearchForm.js
```diff
@@ -68,8 +71,24 @@ export const SearchForm = ({ dataSource }) => {
...
+            break;
+          case RECORD_TYPES.PRESCRIBER:
+            createPrescriberRecord(item);
+            break;
+          default:
+            break;
+        }
+        onClose();
+      };
+      return <SearchListItem key={key} listItem={item} listConfig={listConfig} onPress={onPress} />;
```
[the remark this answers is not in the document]
wlthomson: Good point. e0462aaa.

src/utilities/network/lookupApi.js
```diff
@@ -19,68 +22,219 @@ const SEPARATORS = {
...
+    isPatient,
+    firstName,
+    lastName,
+    dateOfBirth,
+    isVisible: true,
+  };
+  UIDatabase.write(() => UIDatabase.update('Name', patient));
+};
+
+export const createPrescriberRecord = response => {
```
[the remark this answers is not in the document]
wlthomson: Agree. Same as above.

src/widgets/modals/SearchForm.js
wlthomson: cookie

Do you know any reasoning behind the design decision to not make functional components pure by default?

joshxg: nah - at a guess probably just that they were originally for 'smaller' things as classes were the default, so if they now make every functional component pure that would be a pretty massive backwards-incompatible change!

src/utilities/network/lookupApi.js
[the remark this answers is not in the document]
joshxg: hmm I'm a bit slow - I think you're saying `createRecord` expects "external" format (snake_case) - I could be not understanding you though. It's actually just for general use rather than for sync - https://github.com/openmsupply/mobile/blob/d00f1274640307c64fadd8a18df877ba20375954/src/database/utilities/createRecord.js#L113-L142 or for a prescriber: https://github.com/openmsupply/mobile/blob/d00f1274640307c64fadd8a18df877ba20375954/src/database/utilities/createRecord.js#L89-L101

It's super close to being able to just pass your mapped object and removing your methods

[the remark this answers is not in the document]
joshxg: I do agree with a common method for sync/not sync though.. some sort of `deserialize`/`serialize` methods maybe - would be nice for each data type... possibly a `create` static method in each class wou;d be a cleaner approach also.. would at least save us from the soon to be 4d `createRecord` 3000 line file..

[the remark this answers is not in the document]
wlthomson: Yeah, I think we're on the same page here (I wasn't distinguishing properly between `createRecord` and the creation logic in `incomingSyncUtils`).

PR #2653, opened 2020-04-02
#2645 update search form components

### LINKED ISSUE #2645: Update search form components

Placeholder issue for fixing up `SearchForm` components.

- Minor fix ups to `SearchForm` components.
- Update component props, callbacks.
- Hook `SearchForm` into redux store.

## Implementation

See above.

N/A.

## Additional context

See epic #2446.

wlthomson:
Fixes #2645.

Branches off #2633.

## Change summary

Sorry, got a bit carried away with this one!

Still redux thunks and bit of UI stuff to tidy up, but goes a little way to getting things in a bit better shape. Ignore the `TODO`s, they will be removed in the bugsnag PR!

## Testing

No new functionality, probably more efficient to defer tests to the feature PR?

Next steps are to migrate logic to redux and add bugsnag, which should complete the lookup modal and logic. After that, need to add logic for handling data integrity, e.g. dispensing to non-local patients/prescribers, messaging logic for updating visibility, etc.

### DIFF
--- a/src/actions/DispensaryActions.js
+++ b/src/actions/DispensaryActions.js
@@ -8,8 +8,8 @@ export const DISPENSARY_ACTIONS = {
   SORT: 'Dispensary/sort',
   SWITCH: 'Dispensary/switch',
   REFRESH: 'Dispensary/refresh',
-  LOOKUP_RECORD: 'Dispensary/lookupRecord',
-  CANCEL_LOOKUP_RECORD: 'Dispensary/cancelLookupRecord',
+  OPEN_LOOKUP_MODAL: 'Dispensary/openLookupModal',
+  CLOSE_LOOKUP_MODAL: 'Dispensary/closeLookupModal',
 };
 
 const filter = searchTerm => ({ type: DISPENSARY_ACTIONS.FILTER, payload: { searchTerm } });
@@ -20,15 +20,15 @@ const switchDataSet = () => ({ type: DISPENSARY_ACTIONS.SWITCH });
 
 const refresh = () => ({ type: DISPENSARY_ACTIONS.REFRESH });
 
-const lookupRecord = () => ({ type: DISPENSARY_ACTIONS.LOOKUP_RECORD });
+const openLookupModal = () => ({ type: DISPENSARY_ACTIONS.OPEN_LOOKUP_MODAL });
 
-const cancelLookupRecord = () => ({ type: DISPENSARY_ACTIONS.CANCEL_LOOKUP_RECORD });
+const closeLookupModal = () => ({ type: DISPENSARY_ACTIONS.CLOSE_LOOKUP_MODAL });
 
 export const DispensaryActions = {
   filter,
   sort,
   switchDataSet,
   refresh,
-  lookupRecord,
-  cancelLookupRecord,
+  openLookupModal,
+  closeLookupModal,
 };
--- a/src/actions/DispensaryActions.js
+++ b/src/actions/DispensaryActions.js
@@ -3,6 +3,10 @@
  * Sustainable Solutions (NZ) Ltd. 2020
  */
 
+import { batch } from 'react-redux';
+
+import { FORM_ACTIONS } from './FormActions';
+
 export const DISPENSARY_ACTIONS = {
   FILTER: 'Dispensary/filter',
   SORT: 'Dispensary/sort',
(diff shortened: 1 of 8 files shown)


src/selectors/dispensary.js
```diff
@@ -67,3 +85,19 @@ export const selectSortedData = createSelector(
...
+  [selectDataSet, selectLookupFormConfig],
+  (dataSource, formConfig) => {
+    const listKeys = LOOKUP_LIST_CONFIGS[dataSource];
+    const listTypes = listKeys.reduce(
+      (acc, key) => ({ ...acc, [key]: formConfig.find(config => config.key === key)?.type }),
+      {}
+    );
+    return listKeys.map(key => ({ key, type: listTypes[key] }));
+  }
+);
```
joshxg, line 103: just because I often spend a lot of time with these decisions of DRY code: I'm not sure I like the coupling of these configs. I think you could be creating new column definitions and using `getColumns` or am I a bit off ?

src/utilities/formInputConfigs.js
```diff
@@ -27,6 +27,14 @@ import { formInputStrings } from '../localization';
  * }
  */
 
+export const FORM_INPUT_TYPES = {
+  TEXT: 'text',
+  DATE: 'date',
+  DROPDOWN: 'dropdown',
+  TOGGLE: 'toggle',
+  SLIDER: 'slider',
+};
```
joshxg, line 36: nice, been meaning to do this. Thanks!

src/widgets/FormControl.js
```diff
@@ -167,34 +175,40 @@ const FormControlComponent = ({
...
+      <PageButton
+        onPress={onSaveCompletedForm}
+        style={localStyles.saveButton}
+        isDisabled={!canSaveForm || isDisabled}
+        textStyle={localStyles.saveButtonTextStyle}
+        text={saveButtonText}
+      />
+    ),
+    [isDisabled, showCancelButton, canSaveForm, saveButtonText, onSaveCompletedForm]
+  );
```
joshxg, line 190: done now so more of a comment for the future and to see what you reckon: I think it's ok to maybe not memo buttons as default given `PageButton` is memoized on its own, so we're possibly doubling up on all of the prop comparisons

src/actions/DispensaryActions.js
```diff
@@ -20,15 +20,15 @@ const switchDataSet = () => ({ type: DISPENSARY_ACTIONS.SWITCH });
 
 const refresh = () => ({ type: DISPENSARY_ACTIONS.REFRESH });
 
-const lookupRecord = () => ({ type: DISPENSARY_ACTIONS.LOOKUP_RECORD });
+const openLookupModal = () => ({ type: DISPENSARY_ACTIONS.OPEN_LOOKUP_MODAL });
 
-const cancelLookupRecord = () => ({ type: DISPENSARY_ACTIONS.CANCEL_LOOKUP_RECORD });
+const closeLookupModal = () => ({ type: DISPENSARY_ACTIONS.CLOSE_LOOKUP_MODAL });
```
joshxg, line 25: Haven't tested but possibly need other cleanup here such as resetting the form state?

src/selectors/dispensary.js
wlthomson: Definitely could, tbh the whole thing is probably a bit of a smell in general as it's a bit tricky to find the line between how much to use of the `formInputConfig` stuff and how much to redefine. I think ideally it should all be a bit more decoupled, but because of the use of the `FormControl` component and the way it couples with the input configs, I can't see a way of doing it that feels satisfying! 

Separately defining the columns would work though, but as the format of the data you're getting is still determined by the `formInputConfig`, so you're still kind of coupled with it. But I think you're right that it's worth doing. Maybe something to be included in the issue for updating constants?

But

src/widgets/FormControl.js
wlthomson: Hm, yeah that's a good point. I guess the only time you'd want to do it is for parameters passed as expressions (to avoid recalculating them on each render).

src/actions/DispensaryActions.js
wlthomson: Yeah, currently it's always being called as:

```
batch(() => {
  dispatch(FormActions.resetForm());
  dispatch(DispensaryActions.closeLookupModal());
});
```

Good point, should just be included in the action.

wlthomson: 197b1cd1.

PR #2656, opened 2020-04-02
#2598 TemperatureSync modal component

### LINKED ISSUE #2598: TemperatureSyncModal component

Create a `TemperatureSyncModal` component

Two options:
1. Make `SyncModal` more generic to allow for displaying other syncing things
2. Make a new `TemperatureSyncModal`.

Option 1 is a lot nicer, I think, but a decent amount more work. I think option one might be the way to go

## Implementation

Create a `TemperatureSyncModal` component which has a button to force a manual sync of sensor logs and also displays the current progress.

N/A

## Additional context

N/A

joshxg:
Fixes #2598 

## Change summary

- Adds `TemperatureSync` component which will be wrapped in a `ModalContainer` and used as the temperature syncing modal.
- It is more or less exactly the same as `SyncModal` - using the same styles and sub-components. Just slightly different because of showing the progress of individual sensors, rather than the process as a whole

## Testing

N/A

N/A

### DIFF
--- a/src/actions/TemperatureSyncActions.js
+++ b/src/actions/TemperatureSyncActions.js
@@ -13,6 +13,8 @@ import { MILLISECONDS } from '../utilities';
 import { chunk } from '../utilities/chunk';
 
 export const TEMPERATURE_SYNC_ACTIONS = {
+  OPEN_MODAL: 'TemperatureSync/openModal',
+  CLOSE_MODAL: 'TemperatureSync/closeModal',
   ERROR_NO_SENSORS: 'TemperatureSync/errorNoSensors',
 
   START_SYNC: 'TemperatureSync/startSync',
@@ -24,7 +26,7 @@ export const TEMPERATURE_SYNC_ACTIONS = {
 
   DOWNLOAD_LOGS_START: 'TemperatureSync/downloadLogsStart',
   DOWNLOAD_LOGS_COMPLETE: 'TemperatureSync/downloadLogsComplete',
-  DOWNLOAD_LOGS_ERROR: 'TemperatureSync/downloadLogsErorr',
+  DOWNLOAD_LOGS_ERROR: 'TemperatureSync/downloadLogsError',
 
   START_RESETTING_ADVERTISEMENT_FREQUENCY: 'TemperatureSync/startResettingAdvertisementFrequency',
   ERROR_RESETTING_ADVERTISEMENT_FREQUENCY: 'TemperatureSync/errorResettingAdvertisementFrequency',
@@ -36,6 +38,8 @@ export const TEMPERATURE_SYNC_ACTIONS = {
 
   START_SAVING_TEMPERATURE_LOGS: 'TemperatureSync/startSavingTemperatureLogs',
   COMPLETE_SAVING_TEMPERATURE_LOGS: 'TemperatureSync/completeSavingTemperatureLogs',
+
+  UPDATE_SENSOR_PROGRESS: 'TemperatureSync/updateSensorProgress',
 };
 
 const scanStart = () => ({ type: TEMPERATURE_SYNC_ACTIONS.SCAN_START });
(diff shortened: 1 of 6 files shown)


src/actions/TemperatureSyncActions.js
```diff
@@ -197,9 +201,9 @@ const createTemperatureLogs = sensor => dispatch => {
         const breachConfigs = UIDatabase.objects('TemperatureBreachConfiguration');
         breachConfigs.some(breachConfig => breachConfig.createBreach(UIDatabase, location, newLog));
       }
-
-      UIDatabase.delete('SensorLog', groupedSensorLogs);
     });
+
+    UIDatabase.delete('SensorLog', sensorLogsToGroup);
```
joshxg, line 206: this was broken - oops!

src/actions/TemperatureSyncActions.js
```diff
@@ -226,6 +234,8 @@ const syncTemperatures = () => async (dispatch, getState) => {
   for (let i = 0; i < numberOfSensors; i++) {
     const sensor = sensors[i];
 
+    dispatch(updateSensorProgress(sensor));
```
wlthomson, line 237: Out of curiousity, why not `forEach`?

```
const sensors = UIDatabase.objects('Sensor').filtered('location != null').slice();
...
sensors.forEach(sensor => {
  dispatch(updateSensorProgress(sensor));
  ...
}
```

src/reducers/TemperatureSyncReducer.js
```diff
@@ -24,36 +25,70 @@ const initialState = () => ({
...
+
+      return { ...state, currentSensorName };
+    }
+
+    case TEMPERATURE_SYNC_ACTIONS.OPEN_MODAL: {
+      return { ...state, modalIsOpen: true };
+    }
+    case TEMPERATURE_SYNC_ACTIONS.CLOSE_MODAL: {
+      return { ...state, modalIsOpen: false };
+    }
```
wlthomson, line 59: Just a general thing, we've got lots of modals which we display/hide using a generic key like `modalIsOpen`. Lots of them in separate bits of the redux store, but still might be good to make the keys a bit more explicit, e.g. `temperatureSyncModalIsOpen` etc.

src/actions/TemperatureSyncActions.js
joshxg: because of `await` the function would need to be `async` - considered adding an `asyncForEach` to the array prototype, thought that'd be fun but decided against :(

src/reducers/TemperatureSyncReducer.js
joshxg: feel like `temperatureSync.temperatureSyncModalIsOpen` is just quite redundant, can always rename in parts of the app where it's needed? `const { modalIsOpen: .. } = ..` 

Also since we're here. I was wondering what your thoughts are on the number of cases in the reducer. Do you think it would be nicer to have a `updateState` action, where the new state is passed in the payload, rather than unique action cases which define what the state will become?

src/actions/TemperatureSyncActions.js
wlthomson: Ah yeah, I've come up against that too. I believe can fix that by doing something like this (but maybe not too nice):

```
sensors.forEach(async sensor => ....)
```

src/reducers/TemperatureSyncReducer.js
wlthomson: > feel like temperatureSync.temperatureSyncModalIsOpen is just quite redundant, can always rename in parts of the app where it's needed? const { modalIsOpen: .. } = ..

Yeah, I'm thinking more for the cases where one page etc. has multiple modals. Agree there's a trade-off with explicitness and redundancy here, though.

> Do you think it would be nicer to have a updateState action, where the new state is passed in the payload, rather than unique action cases which define what the state will become?

I don't feel strongly either way. My initial thought would be that there's a good chance both would be similar as you'll probably need lots of smaller functions to handle the cases specified in the payload (or a big do-it-all method). I think you'd possibly end up with similar functions but calling them as helpers rather than explicit actions?

PR #2659, opened 2020-04-02
#2658 add lookup list date null check logic

### LINKED ISSUE #2658: Fix patient API throwing error on default date

An error is thrown if the results of a patient query includes a patient with a default `date_of_birth`  ISO date of `"0000-00-00T00:00:00"`,

Add logic for handling default `date_of_birth` fields.

## Implementation

Add `null` check to `SearchForm` date rendering logic.

N/A.

## Additional context

See epic #2446.

wlthomson:
Fixes #2658.

## Change summary

Add logic to handle invalid/default dates.

## Testing

- [ ] Patient search form list renders default date of birth (`00/00/000`) as `N/A`.

N/A.

### DIFF
--- a/src/widgets/modals/SearchForm.js
+++ b/src/widgets/modals/SearchForm.js
@@ -31,7 +31,7 @@ import {
 } from '../../selectors/dispensary';
 
 const SearchListItemColumnComponent = ({ value, type }) => {
-  const valueText = type === 'date' ? value?.toDateString() ?? '' : value;
+  const valueText = type === 'date' ? value?.toDateString() ?? generalStrings.not_available : value;
   return (
     <View style={localStyles.columnContainer}>
       <Text style={localStyles.text}>{valueText}</Text>


src/widgets/modals/SearchForm.js
```diff
@@ -31,7 +31,7 @@ import {
 } from '../../selectors/dispensary';
 
 const SearchListItemColumnComponent = ({ value, type }) => {
-  const valueText = type === 'date' ? value?.toDateString() ?? '' : value;
+  const valueText = type === 'date' ? value?.toDateString() ?? generalStrings.not_available : value;
```
mark-prins, line 34: this would make a good candidate for a helper utility method

wlthomson: Yeah, I'm thinking in the near future it would be good to have all "parsers" in one utility scope. AFAICS, the main use cases are converting from database format and interface format (and also from incoming sync to database format), so something like:

```
const date = new Date();
// Interface wants to display date
const dateString = parsers.dateToString(date);
....
// Database wants to save date
const dateObject = parsers.stringToDate(dateString)
```

Ideal would be to combine all the existing logic for handling int/floats/double conversions, object/JSON conversions etc. into one `parser` namespace (maybe even a static class).

Issue #2544, opened 2020-03-15
Update outgoing sync to prevent syncing out read-only data


wlthomson:
## Is your feature request related to a problem? Please describe.

Adding patient API functionality will involve storing non-local patient and insurance records locally. Although the UI will enforce that these records are read-only, catching these records at outgoing sync only requires a few lines of code and provides an extra layer of security and reduces unnecessary syncing.

## Describe the solution you'd like

Add checks to outgoing sync to prevent syncing read-only records.

## Implementation

- Update `Name` sync out logic to not sync out patients where `isLocal` is `false`.
- Update `InsurancePolicy` sync out logic to not sync out policies associated with non-local patients.

## Describe alternatives you've considered

N/A.

## Additional context

See epic #2446.

wlthomson:
Update: 

- Desktop logic exists to update visibility for incoming `[patient]` records, meaning these records must be synced back to desktop in order for `[nameStoreJoin]` records to be created. 
- Similar logic is planned for `[clinician]` records, meaning these records also need to be synced back to desktop.
- Policies can be created on satellites for non-local patients, but only policies created by the current user can be edited. Solution is to only sync out policies where `enteredBy` is the current user?

joshxg:
> Desktop logic exists to update visibility for incoming [patient] records, meaning these records must be synced back to desktop in order for [nameStoreJoin] records to be created.

Are these the patient records created on mobile? The server creates a name store join record for new `name` records, but also the server needs the `name` record? I think I don't understand lol

> Similar logic is planned for [clinician] records, meaning these records also need to be synced back to desktop.

Again I think I don't understand 🤣 

> Policies can be created on satellites for non-local patients, but only policies created by the current user can be edited. Solution is to only sync out policies where enteredBy is the current user?

I don't agree or think using the user table for anything is a good idea. I think adding `storeID` to insurance name join and making it a hybrid patient/store data table (where policies follow the patient around, but are only editable on the `storeID` store)

wlthomson:
> Are these the patient records created on mobile? The server creates a name store join record for new name records, but also the server needs the name record? I think I don't understand lol

Yeah, so the patient records need to be synced back to desktop so that desktop can see that the patient has been used on mobile so it can update the `nameStoreJoins` (as I understand it, the user might add a patient and not actually dispense, so we can't just rely on transactions).

[...]

Issue #2543, opened 2020-03-15
Update prescription tabs to enforce read-only constraints for non-local patients


wlthomson:
## Is your feature request related to a problem? Please describe.

Patient lookup functionality will enable users to prescribe to non-local patients. To ensure data integrity, users will not be able to edit these patients or associated data (i.e. insurance policies). 

## Describe the solution you'd like

Update the prescription wizard to enforce the following constraints on non-local patients:

- Patient details are read-only.
- Patient insurance policies are read-only.

## Implementation

- Add a getter to `Name`, e.g. `isLocal` which returns if `supplyingStoreId` is current store.
- If prescribing for a patient where `isLocal` is `false`, disable patient, insurance editing.

## Describe alternatives you've considered

N/A. 

## Additional context

See epic #2446.

joshxg:
`Name` has a field `thisStoresPatient` which can be used for this which is set in `incomingSyncUtils`

wlthomson:
@joshxg Ah nice, I wasn't aware of that. Was going to make a getter `isLocal` which just does a match on the current store and the patient's supplying store. Can go with the existing field if already in the schema.

wlthomson:
Fixed by #2660.

DhanyaHerath:
### Testing 

- [x] Patients which are added using the API cannot be edited locally.
- [x] Insurance polices which are added using the API cannot be edited locally.
- [x] Prescribers which are added using the API cannot be edited locally.


Feels very odd to create something (a policy) that immediately becomes uneditable, but it makes sense why :)

wlthomson:
> Feels very odd to create something (a policy) that immediately becomes uneditable, but it makes sense why :)

Yeah, was a bit of a compromise solution to prevent messy edge cases which could cause data integrity problems!

PR #2660, opened 2020-04-02
#2657 add insurance policy lookup record creation logic

### LINKED ISSUE #2657: Add patient API insurance policy creation logic

When a patient is added to the local database using the lookup API, related insurance policy records also need to be created.

Add lookup API logic for creating insurance policy records.

## Implementation

Add insurance policy record creation logic to existing `lookupApi` helpers.

N/A.

## Additional context

See epic #2446.

wlthomson:
Fixes #2657.

## Change summary

Adds insurance policies for patient records added via lookup API.

## Testing

- [ ] When a prescription is made for a patient retrieved using the API, all associated insurance policies appear in the dropdown list.

Should finish up the creation logic, but a lot of duplication occurring. Next steps are to extract out some shared logic from `createRecord` and `incomingSyncUtils`.

### DIFF
--- a/src/utilities/network/lookupApi.js
+++ b/src/utilities/network/lookupApi.js
@@ -7,7 +7,7 @@ import { SETTINGS_KEYS } from '../../settings';
 import { UIDatabase } from '../../database';
 
 import { NAME_TYPES } from '../../sync/syncTranslators';
-import { getOrCreateAddress, parseBoolean, parseDate } from '../../sync/incomingSyncUtils';
+import { getOrCreateAddress, parseBoolean, parseDate, parseNumber } from '../../sync/incomingSyncUtils';
 
 const { THIS_STORE_ID, SYNC_URL } = SETTINGS_KEYS;
 
@@ -20,6 +20,34 @@ const SEPARATORS = {
   QUERY_STRING: '?',
   QUERY_PARAMETERS: '&',
   POLICY_NUMBER: '-',
+};
+
+export const createInsurancePolicyRecord = response => {
+  const {
+    id,
+    policyNumberFamily,
+    policyNumberPerson,
+    type,
+    discountRate,
+    expiryDate,
+    enteredById,
+    nameId,
+    insuranceProviderId,
+    isActive,
+  } = response;
+  const policy = { 
+    id,
+    policyNumberFamily,
+    policyNumberPerson,
+    type,
+    discountRate: parseNumber(discountRate),
+    expiryDate: parseDate(expiryDate),
+    enteredBy: UIDatabase.getOrCreate('User', enteredById),
+    patient: UIDatabase.getOrCreate('Name', nameId),
+    insuranceProvider: UIDatabase.getOrCreate('InsuranceProvider', insuranceProviderId),
+    isActive: parseBoolean(isActive)
+  };
+  UIDatabase.write(() => UIDatabase.update('InsurancePolicy', policy));
 };
 
 export const createPatientRecord = response => {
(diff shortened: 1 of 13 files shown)


src/utilities/network/lookupApi.js
```diff
@@ -73,6 +102,7 @@ export const createPatientRecord = response => {
     isVisible: true,
   };
   UIDatabase.write(() => UIDatabase.update('Name', patient));
+  policies.forEach(policy => createInsurancePolicyRecord(policy));
```
mark-prins, line 105: Just for fun and to remove a tiny bit of code:

```suggestion
  policies.forEach(createInsurancePolicyRecord);
```

src/utilities/network/lookupApi.js
```diff
@@ -156,11 +186,32 @@ const getPrescriberRequestUrl = params => {
...
-  const patientData = responseJson.map(
+const processInsuranceResponse = response => response.map(({
+    ID: id,
+    insuranceProviderID: insuranceProviderId,
+    nameID: nameId,
+    policyNumberFamily,
+    policyNumberPerson,
+    discountRate,
+    expiryDate,
+    isActive: isActive,
```
mark-prins, line 197: unnecessary rename; should just be `isActive`

src/utilities/network/lookupApi.js
```diff
@@ -22,6 +22,34 @@ const SEPARATORS = {
   POLICY_NUMBER: '-',
 };
 
+export const createInsurancePolicyRecord = response => {
```
joshxg, line 25: same comment as before - adding a case to `createRecord` should be what we're doing here!

src/utilities/network/lookupApi.js
```diff
@@ -22,6 +22,34 @@ const SEPARATORS = {
...
+    type,
+    discountRate: parseNumber(discountRate),
+    expiryDate: parseDate(expiryDate),
+    enteredBy: UIDatabase.getOrCreate('User', enteredById),
+    patient: UIDatabase.getOrCreate('Name', nameId),
+    insuranceProvider: UIDatabase.getOrCreate('InsuranceProvider', insuranceProviderId),
+    isActive: parseBoolean(isActive)
+  };
+  UIDatabase.write(() => UIDatabase.update('InsurancePolicy', policy));
+};
```
joshxg, line 51: also if we're integrating insurance policies from other stores we need a mechanism to ensure they're read only. An easy solution would be to add a getter to `InsurancePolicy`?

```
get canEdit(){
    return this.patient.thisStoresPatient
}
```

and then obviously the code on the policy edit to ensure it's read only?

Is there any other patient data being integrated into mobile?

joshxg: and possibly to be extra safe adding a check in outgoing sync to ensure policies which aren't from this stores patients are never synced out

wlthomson: > also if we're integrating insurance policies from other stores we need a mechanism to ensure they're read only

#2543 :).
 
> and possibly to be extra safe adding a check in outgoing sync to ensure policies which aren't from this stores patients are never synced out

#2544 :).

src/utilities/network/lookupApi.js
wlthomson: #2664 :).

src/utilities/network/lookupApi.js
wlthomson: Fixed in #2671 (fcdd2832).

src/utilities/network/lookupApi.js
wlthomson: Oops, fixed in #2671.

Issue #2669, opened 2020-04-04
Create Temperature class for localized temperature formatting


joshxg:
## Is your feature request related to a problem? Please describe.

To easily display temperatures as F or C or whatever for a particular language that's been selected, it would be easier to do this similar to how `moment` or `currency` are used. Except for temperatures, there's not really a widely used package

## Describe the solution you'd like

Create a singleton `temperature` which when passed a number returns and object which represents a temperature.

Has methods `format = (scale = "celsius") => `${this.temp}${degree symbol}${celsisus ? C : F}` kind of thing.

## Implementation

I think a global property which sets temperatures to be formatted to celsius or farenheit 

## Describe alternatives you've considered

N/A

## Additional context

N/A

joshxg:
closed by #3002

Issue #2599, opened 2020-03-27
Add TemperatureSync to NavigationBar


joshxg:
## Is your feature request related to a problem? Please describe.

Need a way to open the temperature sync modal and to show the current progress.

## Describe the solution you'd like

Add the temperature sync status just like regular sync

## Implementation

N/A

## Describe alternatives you've considered

N/A

## Additional context

Not sure what the icon should be.......

![image](https://user-images.githubusercontent.com/35858975/77712219-13bf3900-7038-11ea-9835-7a56eee50127.png)

??

joshxg:
Closed by #2667

PR #2670, opened 2020-04-04
#2595 fridge component

### LINKED ISSUE #2595: Create FridgeDisplay component

Create a component which is an item in a list within the `VaccineModulePage`.

A `FridgeDisplay` component, which renders the current temperature, the number of items in the fridge and the vaccine chart with the temperatures for the particular fridge. This would also potentially have actions for date ranges for temperatures.

## Implementation

N/A

N/A

## Additional context

N/A

joshxg:
### BRANCHED FROM: #2599 

Fixes #2595 

## Change summary

- Adds a couple of simple components to display fridges

fridge2
(the labels on the chart are within the box and been fixed but forgot to take a screen shot. And temperature of the fridge is formatted also)

fridge3

## Testing

N/A

These aren't exactly like the original:

image

Not because I think it looks better but because it's using existing components and it's easier..! If it's worse, then happy to change it to be like the old way but again was just easier 🤣 

Also have made this issue to deal with localized temperatures https://github.com/openmsupply/mobile/issues/2669

The actual chart not being 100% is a seperate issue 🤣


src/widgets/HazardPoint.js
```diff
@@ -0,0 +1,53 @@
+/* eslint-disable react/forbid-prop-types */
```
joshxg, line 1: not sure whats up with the diff - looks like I messed up the order of things. This has already had a PR and been merged :(

src/widgets/FridgeDisplay.js
```diff
@@ -0,0 +1,70 @@
...
+
+      {isActive ? (
+        <VaccineChart
+          minLine={minLine}
+          maxDomain={maxDomain}
+          minDomain={minDomain}
+          maxLine={maxLine}
+          breaches={breaches}
+        />
+      ) : null}
```
wlthomson, line 40: Nit: could maybe wrap this in a `useCallback`?

src/widgets/FridgeDisplayInfo.js
```diff
@@ -0,0 +1,70 @@
...
+  const { description, currentTemperature, temperatureExposure, numberOfBreaches } = fridge;
+  return (
+    <View style={localStyles.container}>
+      <FlexRow flex={1} justifyContent="space-between" alignItems="center">
+        <FlexRow flex={1} justifyContent="space-evenly">
+          <ChevronDownIcon color={isActive ? WHITE : SUSSOL_ORANGE} style={localStyles.icon} />
+          <SimpleLabel
+            label={description}
+            size="large"
+            text={formatTemperature(currentTemperature)}
```
wlthomson, line 31: Nit: could separate this out and wrap in `useMemo` with `currentTemperature` as a dep?

src/widgets/FridgeDisplayInfo.js
```diff
@@ -0,0 +1,70 @@
...
+            <SimpleLabel
+              label={vaccineStrings.breaches}
+              text={numberOfBreaches}
+              textAlign="right"
+              labelAlign="right"
+            />
+          ) : null}
+          <SimpleLabel
+            label={vaccineStrings.temperature_exposure}
+            text={formatTemperatureExposure(temperatureExposure)}
```
wlthomson, line 46: Nit: same as above?

src/widgets/FridgeDisplayInfo.js
```diff
@@ -0,0 +1,70 @@
...
+            textAlign="right"
+            labelAlign="right"
+          />
+          <SimpleLabel
+            label={vaccineStrings.total_stock}
+            text={fridge.totalStock(UIDatabase)}
```
wlthomson, line 52: Nit: same as above 😆 .

src/widgets/FridgeDisplayInfo.js
joshxg: thanks for these - will do a round of performance tweaking if needed. For now I think I'll just hold off. I think it's obvious you disagree but on this one in particular my thoughts are that the difference between an equality check and just calling this method is so minimal it's not worth it? 🤷‍♂ mainly because it's a primitive !

PR #2671, opened 2020-04-04
#2664 extract record creation logic

### LINKED ISSUE #2664: Extract lookup API and sync shared record creation logic

There is quite a bit of duplication of record creation logic in the current lookup API implementation. Idea is to extract out some common logic into helpers.

Add record creation helpers and related data mappings.

## Implementation

See above.

N/A.

## Additional context

See epic #2446.

wlthomson:
Fixes #2664

## Change summary

Updated `createRecord` cases and reduced network API creators to being wrappers over these cases. Tricky to find an optional solution that handles all use cases, but think this is a pretty decent compromise which will look OK once methods moved into thunks rather than utility methods.

Initial thinking was to consolidate all creation logic (including incoming sync), but after spending a bit of time on this, realised its potentially a bit of a bigger job than thought and has quite a lot of surface area for regressions. Thinking is its probably going to be best to do this in parts rather than trying to refactor everything in one go. 

## Testing

N/A (tests in feature PR).

N/A.

### DIFF
--- a/src/database/utilities/createRecord.js
+++ b/src/database/utilities/createRecord.js
@@ -12,6 +12,40 @@ import { NUMBER_SEQUENCE_KEYS } from './constants';
 import { generalStrings } from '../../localization';
 import { SETTINGS_KEYS } from '../../settings';
 
+/**
+ * Return a database Address object with the given address details (reuse if one
+ * already exists).
+ *
+ * @param   {Realm}         database  The local database.
+ * @param   {string}        line1     Line 1 of the address (can be undefined).
+ * @param   {string}        line2     Line 2 of the address (can be undefined).
+ * @param   {string}        line3     Line 3 of the address (can be undefined).
+ * @param   {string}        line4     Line 4 of the address (can be undefined).
+ * @param   {string}        zipCode   Zip code of the address (can be undefined).
+ * @return  {Realm.object}            The Address object described by the params.
+ */
+export const getOrCreateAddress = (database, line1, line2, line3, line4, zipCode) => {
+  let results = database.objects('Address');
+  if (typeof line1 === 'string') {
+    results = results.filtered('line1 == $0', line1);
+  }
+  if (typeof line2 === 'string') {
+    results = results.filtered('line2 == $0', line2);
+  }
+  if (typeof line3 === 'string') {
+    results = results.filtered('line3 == $0', line3);
+  }
+  if (typeof line4 === 'string') {
+    results = results.filtered('line4 == $0', line4);
+  }
+  if (typeof zipCode === 'string') {
+    results = results.filtered('zipCode == $0', zipCode);
+  }
+  if (results.length > 0) return results[0];
+
+  return createRecord(database, 'Address', { line1, line2, line3, line4, zipCode });
+};
+
 // Get the next highest number in an existing number sequence.
 export const getNextNumber = (database, sequenceKey) => {
   const numberSequence = getNumberSequence(database, sequenceKey);
--- a/src/database/utilities/createRecord.js
+++ b/src/database/utilities/createRecord.js
@@ -74,28 +74,42 @@ export const getNumberSequence = (database, sequenceKey) => {
   return sequenceResults[0];
 };
 
-const createInsurancePolicy = (database, policyValues) => {
+/**
+ * Creates a insurance policy record. policyDetails can have the shape:
+ * {
+ *     id, firstName, lastName, registrationCode, address1, address2, isVisible,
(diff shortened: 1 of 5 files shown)


src/database/utilities/createRecord.js
```diff
@@ -82,63 +130,134 @@ const createAddress = (database, { line1, line2, line3, line4, zipCode } = {}) =
...
+
+  const id = prescriberId ?? generateUUID();
+  const thisStoreId = database.getSetting(SETTINGS_KEYS.THIS_STORE_ID)
+  const supplyingStoreId = prescriberStoreId ?? thisStoreId;
+  const fromThisStore = supplyingStoreId === thisStoreId;
+
+  const address = getOrCreateAddress(database, address1, address2);
+
+  const isVisible = true;
+  const isActive = true;
```
joshxg, line 159: visibility is a given but is the `isActive` 100%? Possible to get an `active = false`?

src/utilities/network/lookupApi.js
```diff
@@ -22,88 +23,22 @@ const SEPARATORS = {
...
-    isActive: true,
-    phoneNumber,
-    mobileNumber,
-    emailAddress,
-    fromThisStore,
-  };
-  UIDatabase.write(() => UIDatabase.update('Prescriber', prescriber));
-}
+export const createPrescriberRecord = prescriber =>
+  UIDatabase.write(() => createRecord(UIDatabase, 'Prescriber', prescriber));
```
joshxg, line 32: this and the patient one seem a bit over the top

src/utilities/network/lookupApi.js
```diff
@@ -22,88 +23,22 @@ const SEPARATORS = {
...
+
+export const createPolicyRecord = policy => {
+  const { nameId, enteredById, insuranceProviderId } = policy;
+  const enteredBy = UIDatabase.getOrCreate('User', enteredById);
+  const patient = UIDatabase.getOrCreate('Name', nameId);
+  const insuranceProvider = UIDatabase.getOrCreate('InsuranceProvider', insuranceProviderId);
+  const policyRecord = { ...policy, enteredBy, patient, insuranceProvider };
+  UIDatabase.write(() => createRecord(UIDatabase, 'InsurancePolicy', policyRecord));
```
joshxg, line 40: I dont know exactly where this is being called from, but I generally try to keep writes to the most top of a call stack as I can as you can't start a new write from within one, so it can avoid making changes in the future and subtle/annoying bugs.  Although not too important, creating a new transaction is quite a high proportion of the time spent doing a write, best to try and do as many actions within the same transaction as pos imo

src/database/utilities/createRecord.js
```diff
@@ -82,63 +130,134 @@ const createAddress = (database, { line1, line2, line3, line4, zipCode } = {}) =
...
+    emailAddress,
+    billingAddress,
+    country,
+    supplyingStoreId,
+    thisStoresPatient,
+    isVisible,
   });
 
-  database.save('Patient', patient);
+  database.save('Name', patient);
```
joshxg, line 259: don't need to `save` if you use update

src/database/utilities/createRecord.js
wlthomson: Hmm, yeah, wasn't sure about this one! I think maybe set to given value or `true` if `undefined`?

src/utilities/network/lookupApi.js
wlthomson: Yeah, as above, just waiting to move to thunks!

src/utilities/network/lookupApi.js
wlthomson: Good idea, about to do the thunk work so thinking best solution might be to wrap the policy logic in the same write transaction as the patient.

src/database/utilities/createRecord.js
wlthomson: Nice, out of curiousity, what is the need for `create`? After referencing with `react-native-database`, `update` seems to cover all `create` use-cases (seems to be equivalent of an `upsert`, so creates if primary key non-existent but updates if the record already exists).

src/database/utilities/createRecord.js
wlthomson: 80002c59502022f4737374158f1f38db42bd7b15.

src/database/utilities/createRecord.js
wlthomson: 3f1ef95d0f48b6cdf328930a001b6f50b187ecb1.

PR #2672, opened 2020-04-04
#1384 Vaccine module page

### LINKED ISSUE #1384: VaccineModulePage for vaccine module

Need to transfer the `VaccineModulePage`

Code here: https://github.com/openmsupply/mobile/blob/vaccine-development/src/pages/VaccineModulePage.js

N/A

## Additional context

N/A

joshxg:
### BRANCHED FROM #2595 

Fixes #1384 

## Change summary

- Adds `VaccinePage` component. Just rendering in a flat list each fridge component.

hello1

## Testing

N/A

- The `locationType.description == 'fridge'` for a fridge is not 100% but a 'for now' type situation. I think better to make a location type in desktop startup code with a description or code something like `@@sussol/fridge` or something..?
- The activity indicator display is a bit of a hack but not too bad I think! Displays an activity indicator for a second before rendering a fridge. It's not the exact right timing but I think it's kinda nice to show that rather than etither our normal spinner popping up when navigating, or for navigating to be a bit janky/slow or for there just to show nothing
- Added dates in the reducer n stuff as I want to add some date pickers to choose a range of dates to show temperatures for 
- few style changes that could be made: the headers for breaches, temperature exposure and total stock not lining up etc is annoying. Waiting to see if I should change these styles to the old ones. Also PR getting a bit big so want to PR now and make targetted PRs for the small adjustments

### DIFF
--- a/src/database/DataTypes/Location.js
+++ b/src/database/DataTypes/Location.js
@@ -39,10 +39,10 @@ export class Location extends Realm.Object {
     return { minimumTemperature, maximumTemperature };
   }
 
-  batchesAtTime(database, timestamp = new Date()) {
+  batchesAtTime(database, timestamp) {
     const locationMovements = this.locationMovements.filtered(
       'enterTimestamp <= $0 && (exitTimestamp > $0 || exitTimestamp == null)',
-      timestamp
+      timestamp ?? new Date()
     );
     const queryString = locationMovements.map(({ id }) => `id == "${id}"`).join(' OR ');
     const itemBatches = database.objects('ItemBatch');
--- a/src/database/DataTypes/Location.js
+++ b/src/database/DataTypes/Location.js
@@ -24,7 +24,7 @@ export class Location extends Realm.Object {
   }
 
   get currentTemperature() {
-    const { temperature } = this.mostRecentTemperatureLog;
+    const { temperature } = this.mostRecentTemperatureLog ?? {};
     return temperature;
   }
 
--- a/src/pages/VaccinePage.js
+++ b/src/pages/VaccinePage.js
@@ -40,12 +40,7 @@ export const VaccinePageComponent = ({
         isActive={item.id === selectedFridge.id}
       />
     ),
-    minLine,
-    maxLine,
-    breaches,
-    maxDomain,
-    minDomain,
-    onSelectFridge
+    [minLine, maxLine, breaches, maxDomain, minDomain, onSelectFridge, selectedFridge]
   );
 
   const BlankComponent = React.useCallback(
--- a/src/pages/VaccinePage.js
+++ b/src/pages/VaccinePage.js
@@ -1,16 +1,103 @@
+/* eslint-disable react/forbid-prop-types */
 /**
  * mSupply Mobile
  * Sustainable Solutions (NZ) Ltd. 2016
  */
 
 import React from 'react';
-import { View } from 'react-native';
 import { connect } from 'react-redux';
+import { StyleSheet, Text, FlatList } from 'react-native';
+import PropTypes from 'prop-types';
 
-export const VaccinePageComponent = () => <View />;
+import { FridgeDisplay, FlexView } from '../widgets';
 
-const mapStateToProps = state => state;
+import { selectMinAndMaxLogs, selectMinAndMaxDomains } from '../selectors/fridge';
+import { FridgeActions } from '../actions/FridgeActions';
 
-const mapDispatchToProps = dispatch => ({ dispatch });
+import { APP_FONT_FAMILY } from '../globalStyles';
+import { vaccineStrings } from '../localization';
+
+export const VaccinePageComponent = ({
+  selectedFridge,
+  minLine,
+  maxLine,
+  breaches,
+  maxDomain,
+  minDomain,
+  fridges,
+  onSelectFridge,
+}) => {
+  const Fridge = React.useCallback(
+    ({ item }) => (
+      <FridgeDisplay
+        fridge={item}
+        minLine={minLine}
+        maxLine={maxLine}
+        breaches={breaches}
+        maxDomain={maxDomain}
+        minDomain={minDomain}
+        onSelectFridge={onSelectFridge}
+        isActive={item.id === selectedFridge.id}
+      />
+    ),
+    minLine,
+    maxLine,
+    breaches,
+    maxDomain,
+    minDomain,
+    onSelectFridge
+  );
+
+  const BlankComponent = React.useCallback(
+    () => (
+      <FlexView flex={1} justifyContent="center" alignItems="center">
+        <Text style={localStyles.blankText}>{vaccineStrings.no_fridges}</Text>
+      </FlexView>
+    ),
+    []
+  );
+
+  return fridges.length ? <FlatList data={fridges} renderItem={Fridge} /> : <BlankComponent />;
+};
+
+const mapStateToProps = state => {
+  const { fridge } = state;
+  const { fridges, selectedFridge, breaches } = fridge;
+
+  const { minLine, maxLine } = selectMinAndMaxLogs(state);
+  const { minDomain, maxDomain } = selectMinAndMaxDomains(state);
+  return {
+    minLine,
+    maxLine,
+    minDomain,
+    maxDomain,
+    fridges,
+    selectedFridge,
+    breaches,
+  };
+};
+
+const mapDispatchToProps = dispatch => ({
+  onSelectFridge: fridge => dispatch(FridgeActions.select(fridge)),
+});
+
+VaccinePageComponent.defaultProps = {
+  breaches: [],
+};
+
+VaccinePageComponent.propTypes = {
+  selectedFridge: PropTypes.object.isRequired,
+  minLine: PropTypes.array.isRequired,
+  maxLine: PropTypes.array.isRequired,
+  breaches: PropTypes.array,
+  maxDomain: PropTypes.number.isRequired,
+  minDomain: PropTypes.number.isRequired,
+  fridges: PropTypes.oneOfType([PropTypes.array, PropTypes.object]).isRequired,
+  onSelectFridge: PropTypes.func.isRequired,
+};
+
+const localStyles = StyleSheet.create({
+  blankText: { textAlign: 'center', fontFamily: APP_FONT_FAMILY, fontSize: 20 },
+});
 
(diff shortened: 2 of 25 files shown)


src/database/DataTypes/Location.js
```diff
@@ -24,7 +24,7 @@ export class Location extends Realm.Object {
   }
 
   get currentTemperature() {
-    const { temperature } = this.mostRecentTemperatureLog;
+    const { temperature } = this.mostRecentTemperatureLog ?? {};
```
wlthomson, line 27: Just checking intended behaviour is to return `null` instead of a default?

src/pages/VaccinePage.js
```diff
@@ -1,16 +1,104 @@
...
+  selectedFridge,
+  minLine,
+  maxLine,
+  breaches,
+  maxDomain,
+  minDomain,
+  fridges,
+  onSelectFridge,
+}) => {
+  const Fridge = React.useCallback(
```
wlthomson, line 30: Nitty nit: just for consistency, I've noticed you maybe prefer scoped `React` methods rather than destructuring them on import? I don't have an opinion either way, but I usually use `useCallback` over `React.useCallback`, so if there's a reason or just personal preference for using the `React` namespace I'll start doing it that way too!

src/pages/VaccinePage.js
```diff
@@ -1,16 +1,104 @@
...
+        onSelectFridge={onSelectFridge}
+        isActive={item.id === selectedFridge.id}
+      />
+    ),
+    minLine,
+    maxLine,
+    breaches,
+    maxDomain,
+    minDomain,
+    onSelectFridge
```
wlthomson, line 48: Do we need to add `selectedFridge` to deps here?

src/pages/VaccinePage.js
```diff
@@ -1,16 +1,104 @@
...
+  const BlankComponent = React.useCallback(
+    () => (
+      <FlexView flex={1} justifyContent="center" alignItems="center">
+        <Text style={localStyles.blankText}>{vaccineStrings.no_fridges}</Text>
+      </FlexView>
+    ),
+    []
+  );
+
+  return fridges.length ? <FlatList data={fridges} renderItem={Fridge} /> : <BlankComponent />;
```
wlthomson, line 60: Not sure how es-lint feels, but good to be explicit? 

```suggestion
  return !!fridges.length ? <FlatList data={fridges} renderItem={Fridge} /> : <BlankComponent />;
```

wlthomson, line 60: Not sure of naming of below suggested component, but I think avoids calculating `fridges.length` on every render here?

 ```suggestion
const FridgesList = useCallback(() => !!fridges.length ? <FlatList data={fridges} renderItem={Fridge} /> : <BlankComponent />, [fridges]);
return <FridgesList />;
```

File: src/pages/VaccinePage.js [hunk 5]
```diff
@@ -1,16 +1,104 @@
...
+  return {
+    minLine,
+    maxLine,
+    minDomain,
+    maxDomain,
+    fridges,
+    selectedFridge,
+    breaches,
+  };
+};
```
wlthomson, line 79: Nice!

src/pages/VaccinePage.js
```diff
@@ -1,16 +1,104 @@
...
+  return fridges.length ? <FlatList data={fridges} renderItem={Fridge} /> : <BlankComponent />;
+};
+
+const mapStateToProps = state => {
+  const { fridge } = state;
+  const { fridges, selectedFridge } = fridge;
+  const { breaches } = selectedFridge;
+
+  const { minLine, maxLine } = selectMinAndMaxLogs(state);
+  const { minDomain, maxDomain } = selectMinAndMaxDomains(state);
```
wlthomson, line 69: Nit: guessing behind the scenes there are selectors for `selectMinLogs`, select`selectMaxLogs` etc. (if not, ignore). Maybe wrappers are a bit overkill?

src/selectors/fridge.js
```diff
@@ -0,0 +1,89 @@
...
+ */
+
+import { createSelector } from 'reselect';
+import { UIDatabase } from '../database/index';
+import { chunk } from '../utilities/chunk';
+
+export const selectSelectedFridgeID = ({ fridge }) => {
+  const { selectedFridge = {} } = fridge;
+  const { id } = selectedFridge;
+
```
wlthomson, line 13: Nit: role reversal 😆 ! Whitespace seems a bit redundant here?

src/selectors/fridge.js
```diff
@@ -0,0 +1,89 @@
+/**
+ * mSupply Mobile
+ * Sustainable Solutions (NZ) Ltd. 2020
+ */
+
+import { createSelector } from 'reselect';
+import { UIDatabase } from '../database/index';
+import { chunk } from '../utilities/chunk';
+
```
wlthomson, line 9: Maybe overkill, but could add a selector just for `selectFridge`, and then the other selectors can all `createSelector` with that as a dependency. 

1. Probably not worth it just for this, but would make some of the other selectors tidy one-liners, e.g.:

```
export const selectFridge = ({ fridge = {} }) => fridge;
export const selectFridgeID = createSelector([selectFridge], ({ id }) => id);
```

2. I think would also be an optimisation (other selectors will be memoised and only recalculated if fridge changes)?

src/selectors/fridge.js
```diff
@@ -0,0 +1,89 @@
...
+  return fromDate;
+};
+
+export const selectTemperatureLogsToDate = ({ fridge }) => {
+  const { toDate } = fridge;
+
+  return toDate;
+};
+
+export const selectSelectedFridge = createSelector([selectSelectedFridgeID], selectedFridgeID => {
```
wlthomson, line 29: Hmm, probably should have read down before making above comment!

src/selectors/fridge.js
```diff
@@ -0,0 +1,89 @@
...
+    return temperatureLogs;
+  }
+);
+
+export const selectChunkedTemperatureLogs = createSelector(
+  [selectFridgeTemperatureLogsFromDate],
+  logs => {
+    const { length: numberOfLogs } = logs;
+
+    return numberOfLogs < 30 ? logs : chunk(logs, Math.ceil(numberOfLogs / 30));
```
wlthomson, line 56: Nit: could probably extract this calculation into a utility somewhere (at some point).

src/selectors/fridge.js
```diff
@@ -0,0 +1,89 @@
...
+      const timestamp = Math.min(...timestamps);
+
+      const { minLine, maxLine } = acc;
+
+      return {
+        minLine: [...minLine, { temperature: minTemperature, timestamp }],
+        maxLine: [...maxLine, { temperature: maxTemperature, timestamp }],
+      };
+    },
+    { minLine: [], maxLine: [] }
```
wlthomson, line 78: Nice! This is code art!

src/widgets/FridgeDisplay.js
```diff
@@ -19,25 +21,41 @@ export const FridgeDisplay = ({
...
+        </FlexView>
+      ),
+    [render]
+  );
+
+  return (
+    <View style={containerStyle}>
+      <FridgeDisplayInfo onPress={onSelectFridge} fridge={fridge} isActive={isActive} />
+
+      {isActive ? <Chart /> : null}
```
wlthomson, line 58: Nit: could maybe memoise this? Something like (but with a better name):

```
const FridgeChart = useCallback(() => isActive ? <Chart /> : null, [isActive]);
return return (
    <View style={containerStyle}>
      <FridgeDisplayInfo onPress={onSelectFridge} fridge={fridge} isActive={isActive} />
      <FridgeChart />
    </View>
);
```

src/widgets/FridgeDisplayInfo.js
```diff
@@ -18,10 +18,15 @@ import { SimpleLabel } from './SimpleLabel';
 import { WHITE, SUSSOL_ORANGE } from '../globalStyles';
 import { vaccineStrings } from '../localization';
 
-export const FridgeDisplayInfo = ({ fridge, isActive }) => {
+export const FridgeDisplayInfo = ({ fridge, isActive, onPress }) => {
   const { description, currentTemperature, temperatureExposure, numberOfBreaches } = fridge;
+
+  const Container = isActive ? View : TouchableOpacity;
```
wlthomson, line 24: Nit: probably overkill, but could memoise this 😄 .

src/database/DataTypes/Location.js
joshxg: yeah - default I think is a bit nasty - can't be `0`, `-Infinity` or `Infinity` are just more annoying to check, especially now with `??`

src/pages/VaccinePage.js
joshxg: just easier.. less code to write 🤣 Very explicit and good for grokking since there are other `useX..` custom hooks n stuff .. personal perference no real reason. I'm just lazy

src/pages/VaccinePage.js
joshxg: ya 2f5a123d7ee9f1f61f331d4497e08fc65737b5cd

(look at the commit, theres another funny bug there actually 🤣 )

src/pages/VaccinePage.js
joshxg: eslint doesn't like it actually 🤣

joshxg: I think there's quite a few comments re caching which we might disagree on. Mainly because it isn't really going to offer any performance benefit and imo actually makes it practically harder to read in my opinion.. gotta look in two places+ to see exactly whats being returned. 

I think "memoize all the things" is a bit of a trap and I'm trying to do it sparingly and with purpose when I do as it can just lead to so many small, subtle and annoying bugs! There doesn't seem to be toooo much benefit in this case

src/pages/VaccinePage.js
joshxg: I just did min and max in the one to reduce iterations

src/selectors/fridge.js
joshxg: I need to revisit these selectors - but one main point of using an `id` selector is to be a 'base' selector for a bunch of other selectors to depend on, rather than a reference to an object which is less reliable, especially as it's a realm object!

src/selectors/fridge.js
joshxg: 🤣 

trying to get a balance - I like a space before return although it's starting to get excessive

PR #2674, opened 2020-04-05
#2654 add lookup record creation thunks

### LINKED ISSUE #2654: Move patient API network logic to thunks

Lookup requests are currently called directly from the `SearchForm` component. To be consistent with the mobile architecture, these should be encapsulated in thunks.
 
Add redux logic for patient API network requests.

## Implementation

Most consistent solution is to keep the lookup modal state in the store, and do the heavy lifting in the reducer. At the moment, this might be a bit more work that it's worth. Suggested solution is to keep the parameters in the component state, and inject them into the reducer via the action payload.

N/A.

## Additional context

See epic #2446.

wlthomson:
Fixes #2654.

## Change summary

Update existing patient and prescriber actions and migrate patient API.

## Testing

- [ ] New patients are created successfully.
- [ ] Existing patients are updated successfully.
- [ ] Patients added via the API are created successfully.
- [ ] New prescribers are created successfully.
- [ ] Existing prescribers are updated successfully.
- [ ] Prescribers added via the API are created successfully.

N/A.

### DIFF
--- a/src/actions/PatientActions.js
+++ b/src/actions/PatientActions.js
@@ -27,27 +27,94 @@ const editPatient = patient => ({
   },
 });
 
-const patientUpdate = completedForm => (dispatch, getState) => {
+const patientUpdate = patientDetails => (dispatch, getState) => {
   const { patient } = getState();
   const { currentPatient } = patient;
 
-  if (currentPatient) {
-    const { addressOne: line1, addressTwo: line2 } = completedForm;
-    const { firstName = '', lastName = '', billingAddress = {} } = currentPatient;
-    const patientName = `${lastName}, ${firstName}`;
+  const {
+    id: currentPatientId,
+    code: currentCode,
+    name: currentName,
+    firstName: currentFirstName,
+    lastName: currentLastName,
+    dateOfBirth: currentDateOfBirth,
+    emailAddress: currentEmailAddress,
+    phoneNumber: currentPhoneNumber,
+    billingAddress: currentBillingAddress,
+    country: currentCountry,
+    supplyingStoreId: currentSupplyingStoreId,
+    isActive: currentIsActive,
+    female: currentFemale,
+  } = currentPatient ?? {};
 
-    UIDatabase.write(() => {
-      UIDatabase.update('Address', { ...billingAddress, line1, line2 });
-      UIDatabase.update('Name', { ...currentPatient, ...completedForm, name: patientName });
-    });
-  } else {
-    UIDatabase.write(() => {
-      createRecord(UIDatabase, 'Patient', completedForm);
-    });
-  }
+  const {
+    id: currentBillAddressId,
+    line1: currentLine1,
+    line2: currentLine2,
+    line3: currentLine3,
+    line4: currentLine4,
+    zipCode: currentZipCode,
+  } = currentBillingAddress ?? {};
+
+  const {
+    id: patientId,
+    code: patientCode,
+    firstName: patientFirstName,
+    lastName: patientLastName,
+    dateOfBirth: patientDateOfBirth,
+    emailAddress: patientEmailAddress,
+    phoneNumber: patientPhoneNumber,
+    addressOne: patientLine1,
+    addressTwo: patientLine2,
+    country: patientCountry,
+    female: patientFemale,
+  } = patientDetails ?? {};
+
+  const id = patientId ?? currentPatientId;
+  const code = patientCode ?? currentCode;
+  const firstName = patientFirstName ?? currentFirstName;
+  const lastName = patientLastName ?? currentLastName;
+  const name = `${lastName}, ${firstName}` || currentName;
+  const dateOfBirth = patientDateOfBirth ?? currentDateOfBirth;
+  const emailAddress = patientEmailAddress ?? currentEmailAddress;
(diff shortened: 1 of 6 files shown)


src/actions/PatientActions.js
```diff
@@ -27,27 +27,94 @@ const editPatient = patient => ({
...
+    billAddress3,
+    billAddress4,
+    billPostalZipCode,
+    country,
+    female,
+    supplyingStoreId,
+    isActive,
+  };
+
+  UIDatabase.write(() => createRecord(UIDatabase, 'Patient', patientRecord));
```
joshxg, line 113: I find code like this very interesting - you want to be explicit, but that's so the code is clear and easy to read, but then if it's too long and explicit, it becomes harder to read!

I haven't checked every field but I think something a little more JS could make it a little easier on the eyes

```
const record = { ...currentPatient, ...{... patientDetails, suppliyingStore: UID..(thisstore), name: ... },  
```

src/actions/PrescriberActions.js
```diff
@@ -32,24 +32,70 @@ const editPrescriber = prescriber => ({
...
+    addressTwo,
+    phoneNumber,
+    mobileNumber,
+    emailAddress,
+    female,
+    storeId,
+    isActive,
+  };
+
+  UIDatabase.write(() => createRecord(UIDatabase, 'Prescriber', prescriberRecord));
```
joshxg, line 94: same as above - quite easy to merge objects with JS, so I think taking advantage of that while using comments to describe the objects shape makes for the cleanest code here!

src/database/utilities/createRecord.js
```diff
@@ -17,15 +17,20 @@ import { SETTINGS_KEYS } from '../../settings';
...
+ * @param   {string}        id        Id of the address (can be undefined).
  * @param   {string}        line1     Line 1 of the address (can be undefined).
  * @param   {string}        line2     Line 2 of the address (can be undefined).
  * @param   {string}        line3     Line 3 of the address (can be undefined).
  * @param   {string}        line4     Line 4 of the address (can be undefined).
  * @param   {string}        zipCode   Zip code of the address (can be undefined).
  * @return  {Realm.object}            The Address object described by the params.
  */
-export const getOrCreateAddress = (database, line1, line2, line3, line4, zipCode) => {
+export const getOrCreateAddress = (database, id, line1, line2, line3, line4, zipCode) => {
```
joshxg, line 28: I believe this is going to break the calls in `incomingSyncUtils.js`? Might want to make it the last param? possibly having an object param, but quite a few changes over just making it the last param

src/database/utilities/createRecord.js
```diff
@@ -132,36 +137,44 @@ const createAddress = (database, { line1, line2, line3, line4, zipCode } = {}) =
...
 
   const id = prescriberId ?? generateUUID();
-  const thisStoreId = database.getSetting(SETTINGS_KEYS.THIS_STORE_ID);
-  const supplyingStoreId = prescriberStoreId ?? thisStoreId;
-  const fromThisStore = supplyingStoreId === thisStoreId;
+  const firstName = prescriberFirstName ?? '';
+  const lastName = prescriberLastName ?? '';
+  const registrationCode = prescriberRegistrationCode ?? '';
+
+  const address = getOrCreateAddress(database, addressOne, addressTwo);
```
joshxg, line 164: wrong params?

src/database/utilities/createRecord.js
wlthomson: Ah, forgot to make the changes to the other callers, plan was to just add a `null` param, but that's pretty ugly. Tbh, I think longer term this method can be made a bit nicer, but I think for now making an object param is nicer, it's not called in that many places!

src/actions/PatientActions.js
wlthomson: Yeah, should have been more descriptive in the PR. For now I'm keeping it explicit to make it clear exactly what params are coming from where and going where. The probably with the nice spread syntax is that you can't easily see what parameters are actually being used (I had to cross-reference other files and it was easier in the end just to log everything).  

To answer the question re. explicitness being harder to read, I think key point is that explicitness isn't to make the code nicer or easier to read (it's definitely harder to read and uglier!), the point is to show the data being passed around. IMO, in this case using spread syntax makes it nicer to look at, but doesn't make it easier to read, as the information is all hidden, so it has to be found by reading other methods and logging etc. 

Saying all that, the plan is still to make it nice and tidy once all the reviewing and stuff is done (i.e. at the point where functionality is working well and there's less need for that data to be explicit).

src/actions/PrescriberActions.js
wlthomson: As above, using spread syntax with comments is in the pipeline for the tidyup stage, but I think a bit premature at this point.

src/database/utilities/createRecord.js
wlthomson: cfafe1f96cbfb8f91bc811b521b2b9c81542954b.

src/database/utilities/createRecord.js
wlthomson: cfafe1f96cbfb8f91bc811b521b2b9c81542954b.

PR #2675, opened 2020-04-06
#2543 add isEditable helpers

### LINKED ISSUE #2543: Update prescription tabs to enforce read-only constraints for non-local patients

wlthomson:
Fixes #2543.

Branches off #2674.

## Change summary

Looks like most of the functionality needed is already implemented! Just added a few helpers to make things a bit more explicit.

## Testing

- [ ] Patients which are added using the API cannot be edited locally.
- [ ] Insurance polices which are added using the API cannot be edited locally.
- [ ] Prescribers which are added using the API cannot be edited locally.

N/A.

### DIFF
--- a/src/selectors/insurance.js
+++ b/src/selectors/insurance.js
@@ -17,6 +17,19 @@ export const selectInsurancePolicyDiscountRate = ({ insurancePolicy }) => {
 export const selectInsuranceDiscountRate = ({ insurance }) =>
   selectInsurancePolicyDiscountRate(insurance);
 
+export const selectIsSelectedPolicyEditable = ({ user, insurance }) => {
+  const { selectedInsurancePolicy } = insurance ?? {};
+  const { patient: policyPatient, enteredBy: policyUser } = selectedInsurancePolicy ?? {};
+
+  const { isEditable: isPatientEditable } = policyPatient ?? {};
+
+  const { currentUser } = user ?? {};
+  const { id: currentUserId } = currentUser ?? {};
+  const { id: policyUserId } = policyUser ?? {};
+
+  return isPatientEditable || (currentUserId && policyUserId && currentUserId === policyUserId);
+};
+
 export const selectInsuranceModalOpen = ({ insurance }) => {
   const { isCreatingInsurancePolicy, isEditingInsurancePolicy } = insurance;
   return isCreatingInsurancePolicy || isEditingInsurancePolicy;
--- a/src/selectors/patient.js
+++ b/src/selectors/patient.js
@@ -50,7 +50,6 @@ export const selectPatientModalOpen = ({ patient }) => {
 
 export const selectCanEditPatient = ({ patient }) => {
   const { currentPatient } = patient;
-  const { thisStoresPatient } = currentPatient ?? {};
-
-  return thisStoresPatient ?? true;
+  const { isEditable = true } = currentPatient ?? {};
+  return isEditable;
 };
--- a/src/selectors/patient.js
+++ b/src/selectors/patient.js
@@ -50,6 +50,6 @@ export const selectPatientModalOpen = ({ patient }) => {
 
 export const selectCanEditPatient = ({ patient }) => {
   const { currentPatient } = patient;
-  const { isEditable = true } = currentPatient ?? {};
+  const { isEditable = false } = currentPatient ?? {};
   return isEditable;
 };
--- a/src/database/DataTypes/Name.js
+++ b/src/database/DataTypes/Name.js
@@ -96,6 +96,15 @@ export class Name extends Realm.Object {
   }
 
   /**
+   * Get if this name record can be edited.
+   *
+   * @return  {boolean}
+   */
+  get isEditable() {
+    return this.thisStoresPatient;
+  }
+
+  /**
    * Add master list to name, if it has not already been added.
    *
    * @param  {MasterList}  masterList
--- a/src/database/DataTypes/Prescriber.js
+++ b/src/database/DataTypes/Prescriber.js
@@ -1,6 +1,10 @@
 import Realm from 'realm';
 
-export class Prescriber extends Realm.Object {}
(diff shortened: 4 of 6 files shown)


src/selectors/patient.js
```diff
@@ -50,7 +50,6 @@ export const selectPatientModalOpen = ({ patient }) => {
 
 export const selectCanEditPatient = ({ patient }) => {
   const { currentPatient } = patient;
-  const { thisStoresPatient } = currentPatient ?? {};
-
-  return thisStoresPatient ?? true;
+  const { isEditable = true } = currentPatient ?? {};
```
joshxg, line 53: possibly better to have this as false as the default

wlthomson: Good point. 6d37591.

wlthomson:
@joshxg Have added insurance logic if you want to do a quick rereview.

src/selectors/insurance.js
```diff
@@ -17,6 +17,19 @@ export const selectInsurancePolicyDiscountRate = ({ insurancePolicy }) => {
...
+  const { patient: policyPatient, enteredBy: policyUser } = selectedInsurancePolicy ?? {};
+
+  const { isEditable: isPatientEditable } = policyPatient ?? {};
+
+  const { currentUser } = user ?? {};
+  const { id: currentUserId } = currentUser ?? {};
+  const { id: policyUserId } = policyUser ?? {};
+
+  return isPatientEditable || (currentUserId && policyUserId && currentUserId === policyUserId);
+};
```
joshxg, line 31: I'm confused why the user ID is being used here?

wlthomson: The use-case here is where you lookup a patient and add a new policy. That policy should remain editable, rather than the other policies, which should be uneditable. Does that make sense?

It's not a perfect solution; the assumption here is that if the policy was created by the current user, then the current user can edit it without causing sync issues (not a fool-proof solution, but the best compromise I can think of with existing data).

joshxg: ah I see - I imagined we would not allow creation of insurance policies for non-home store patients until there was a proper solution?

I suppose we never found a proper solution from this issue: https://github.com/sussol/msupply/issues/5491

joshxg: what is the behaviour now on desktop that you can have patients that aren't from the home store?

wlthomson: Yeah, it got a bit of a case of too many chefs! @mark-prins and I decided to allow creation of insurance policies but not editing of existing ones. I think you're right to lean on the conservative option, and that was our initial thinking too. The reason we came down on the side of allowing creation was that we considered it potentially a blocker to prescribing to that patient. 

I'm in the process of writing up docs for this, so this will be included. Have no problem with revisiting this decision!

joshxg: we can quite easily allow creation if we just add a `storeID` and make them a hybrid store data and patient data. This use of the user ID I think creates a worse outcome (for us, not the patient), where you have records editable in many stores?

wlthomson: > what is the behaviour now on desktop that you can have patients whose home store is another store?

Do you mean what is the insurance policy behaviour on desktop for non-local patients?

As I understand it, the patient API isn't implemented for desktop satellite sites, so you can't prescribe to non-local patients (as you can't seem them). My understanding is that the dispensing logic will be consistent with mobile, so the decision should probably be made here.

AFAIK, currently, you can't add insurance policies on desktop for non-local patients. However, as I understand it you also can't prescribe to them, so there's not much need.

Please correct me if any of the above is wrong! @mark-prins Have I missed anything?

joshxg: Was just checking as I thought I might have missed something and this was actually mimicing desktop behaviour (with the user stuff).

- Seems to be there are two worst case scenarios: A patient can't get their prescription as a user can't add an insurance policy for that patient or data inconsistencies due to being able to edit records in multiple stores.
- We maybe don't have time to fully solve these issues right now (?)

I think it is a lot easier to solve the former problem later (if we don't have time now), then having to deal and solve data issues (which IC have had a lot of, their data file is messsssed uuuup).

I'm sorry I'm happy to be told I'm wrong but I just think we need to make insurance policies uneditable and non-creatable(??) for non-local patients until there is a proper solution  ! obviously you don't, is it because the patient not getting a prescription worst case scenario is more important? It probably is.. 🤷‍♂

wlthomson: What you said is all correct, the reason for the workaround is to maintain pretty strong data consistency (the only case where a policy can be edited in two places is by the same user), while also avoiding the case where effectively a patient can't use their insurance because of constraints on foreign patients (if this occurs, the user is likely to create a duplicate patient and we will have data inconsistencies anyway).

So basically, it's an acknowledgement of both worst-cases you just mentioned, and an attempt to mitigate them both. Does that make sense?

wlthomson:
@joshxg @mark-prins Going to go ahead and merge into the feature branch, but open to making changes and have created an issue for further discussion (#2680).

src/selectors/insurance.js
mark-prins: sorry, bit late to the discussion. Ideally, should be locking down and preventing the addition of policies for 'foreign' patients. I think you've summarised the case well, Will. this solution isn't ideal but I think is the best of a bad bunch.

PR #2684, opened 2020-04-07
#2681 update lookup response logic

### LINKED ISSUE #2681: Add response code handling to API network helpers

The lookup API does not currently handle response codes.

Add logic for handling response codes.

## Implementation

Update lookup API helpers to handle response codes.

N/A.

## Additional context

See epic #2446.

wlthomson:
Fixes #2681.

## Change summary

Adds logic for handling:

- Connection errors.
- Timeout errors.
- Authentication errors.
- No matching records.

Basically just a spinner and a confirm modal. Not the nicest from a UI point of a view, but does the job!

Replicates a bit of sync, epic still has an issue for updating sync constants so I suggest using that to modularise some of the shared lookup/sync logic.

## Testing

- [ ] When a user searches for a patient on an offline server, a confirm modal displays the error: `"Unable to connect"`.
- [ ] When a user searches for a patient using invalid/expired credentials, a confirm modal displays the error: `"Invalid username or password"`.
- [ ] When a user searches for a patient but no response is received in 10 seconds, a confirm modal displays the error: `"Unable to connect"`.
- [ ] When a user searches for a non-existent patient, a confirm modal displays the error `"No matching records found"`.
- [ ] When a user searches for a prescriber on an offline server, a confirm modal displays the error: `"Unable to connect"`.
- [ ] When a user searches for a prescriber using invalid/expired credentials, a confirm modal displays the error: `"Invalid username or password"`.
- [ ] When a user searches for a prescriber but no response is received in 10 seconds, a confirm modal displays the error: `"Unable to connect"`.
- [ ] When a user searches for a non-existent prescriber, a confirm modal displays the error `"No matching records found"`.

N/A.

### DIFF
--- a/src/sync/lookupApiUtils.js
+++ b/src/sync/lookupApiUtils.js
@@ -215,7 +215,7 @@ export const queryPatientApi = async params => {
     }
     switch (status) {
       case 400:
-        throw new Error(ERROR_CODES.INVALID_URL);
+        throw new Error(ERROR_CODES.CONNECTION_FAILURE);
       case 401:
         throw new Error(ERROR_CODES.INVALID_PASSWORD);
       default:
@@ -245,7 +245,7 @@ export const queryPrescriberApi = async params => {
     }
     switch (status) {
       case 400:
-        throw new Error(ERROR_CODES.INVALID_URL);
+        throw new Error(ERROR_CODES.CONNECTION_FAILURE);
       case 401:
         throw new Error(ERROR_CODES.INVALID_PASSWORD);
       default:
--- a/src/sync/lookupApiUtils.js
+++ b/src/sync/lookupApiUtils.js
@@ -6,13 +6,20 @@
 /* eslint-disable camelcase */
 /* eslint-disable no-unused-expressions */
 
-import { getAuthHeader } from 'sussol-utilities';
+import { getAuthHeader, AUTH_ERROR_CODES } from 'sussol-utilities';
 
-import { SETTINGS_KEYS } from '../../settings';
-import { UIDatabase } from '../../database';
-import { createRecord, parseBoolean, parseDate, parseNumber } from '../../database/utilities';
+import { SETTINGS_KEYS } from '../settings';
+import { UIDatabase } from '../database';
+import { createRecord, parseBoolean, parseDate, parseNumber } from '../database/utilities';
 
 const { SYNC_URL, SYNC_SITE_NAME, SYNC_SITE_PASSWORD_HASH } = SETTINGS_KEYS;
+
+const CONNECTION_TIMEOUT_PERIOD = 10000;
+
+const ERROR_CODES = {
+  ...AUTH_ERROR_CODES,
+  NO_MATCHING_RECORDS: 'No matching records found',
+};
 
 const RESOURCES = {
   PATIENT: '/api/v4/patient',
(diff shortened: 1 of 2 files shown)


mark-prins:
@wlthomson is there some way to test on an emulator?
I'm using this branch but I don't get search buttons:

Screen Shot 2020-04-08 at 12 39 50 PM

src/widgets/modals/SearchForm.js
```diff
@@ -76,11 +91,40 @@ export const SearchFormComponent = ({
...
+
+  const lookupPatient = useCallback(params => {
+    setIsQuerying(true);
+    queryPatientApi(params).then(({ error: patientError, data: patientData }) => {
+      if (patientError) {
+        setIsError(true);
+        setError(patientError);
+      }
+      setData(patientData);
+      setIsQuerying(false);
```
joshxg, line 108: if you close the modal while querying is this component still mounted? I think it will either lead to memory leaks with trying to cause updates on unmounted components, or performance decreases and subtle bugs in the future if this component is just always mounted?

 I think you would need to create a new state or ref which would be set with a `useEffect` on mounting/unmounting which can be used to check if the component is still mounted before trying to update state. This gets pretty messy!

I think it's a lot easier to either use the redux store for state or alternatively use async/await rather than callbacks, although that's not a great solution

src/sync/lookupApiUtils.js
```diff
@@ -193,31 +200,75 @@ const processPrescriberResponse = response => {
...
+        throw new Error(ERROR_CODES.CONNECTION_FAILURE);
+      case 401:
+        throw new Error(ERROR_CODES.INVALID_PASSWORD);
+      default:
+        throw new Error(ERROR_CODES.CONNECTION_FAILURE);
+    }
   } catch (error) {
     // TODO: add bugsnag.
-    return [];
+    return { error: error.message };
```
joshxg, line 226: it's a TODO so I have to comment that I hate them 🤷‍♂

src/sync/lookupApiUtils.js
```diff
@@ -193,31 +200,75 @@ const processPrescriberResponse = response => {
...
+        : { error: ERROR_CODES.NO_MATCHING_RECORDS };
+    }
+    switch (status) {
+      case 400:
+        throw new Error(ERROR_CODES.CONNECTION_FAILURE);
+      case 401:
+        throw new Error(ERROR_CODES.INVALID_PASSWORD);
+      default:
+        throw new Error(ERROR_CODES.CONNECTION_FAILURE);
+    }
```
joshxg, line 253: this code looks eerily familiar! 🤣

wlthomson:
> @wlthomson is there some way to test on an emulator?
> I'm using this branch but I don't get search buttons:
> 
> Screen Shot 2020-04-08 at 12 39 50 PM

Hmm, looks like a styling issue! Will have a look with a few different emulators.

src/widgets/modals/SearchForm.js
wlthomson: Good point!

I don't think `await` will make a difference, but will try out a custom hook with `useEffect` as suggested!

src/sync/lookupApiUtils.js
wlthomson: I think its new? Is it already in the codebase?

wlthomson: Oh, you mean the method duplication :laughing:. Yeah, I don't like it either but trying to follow the rule of 2 (only abstract methods if used in more than two places).

wlthomson:
> @wlthomson is there some way to test on an emulator?
> I'm using this branch but I don't get a search button:
> 
> Screen Shot 2020-04-08 at 12 39 50 PM

e71809c.

wlthomson:
@joshxg Requested changes made in #2688 😃 .

PR #2685, opened 2020-04-07
#2591 New vaccine columns

### LINKED ISSUE #2591: Apply new columns to pages for vaccines

Need to add new columns to `CustomerInvoicePage`, `SupplierInvoicePage` and `StocktakeBatchModal`

Add new column definitions to the above pages under the vaccine and vaccine item conditions

## Implementation

Create a new `SupplierInvoiceWithVaccines` and `StocktakeBatchModalVaccineItem` or something constants, which have column definitions with the added columns for each as per below images

image

image

image

N/A

## Additional context

N/A

joshxg:
Fixes #2591 

## Change summary

- Supplier invoice: Adds Location, vvm status columns
- Customer invoice: Adds doses column
- Stocktake batch: adds doses, loation and vvm status columns

*Still need to add a breach column*

Thought this would be less code. Oops :( 

ci1
si1
stocktake1

## Testing

N/A

N/A


src/database/DataTypes/ItemBatch.js
```diff
@@ -21,6 +22,10 @@ import Realm from 'realm';
  * @property  {List.<TransactionBatch>}  transactionBatches
  */
 export class ItemBatch extends Realm.Object {
+  get isVaccine() {
+    return this.item?.isVaccine ?? false;
+  }
```
joshxg, line 27: same thing added to `StocktakeBatch`, `TransactionBatch`, `ItemBatch` `TransactionItem`, just all the things

src/database/DataTypes/ItemBatch.js
```diff
@@ -145,6 +150,20 @@ export class ItemBatch extends Realm.Object {
 
     return newLocationId !== currentLocationId;
   }
+
+  applyLocation(database, newLocation) {
+    if (!this.shouldApplyLocation(newLocation)) return null;
+
+    this.location = newLocation;
+
+    return createRecord(database, 'LocationMovement', this, newLocation);
```
joshxg, line 159: This is called after a stocktake or transaction are finalised - locking in the location movement. applying a location: Adds a `LocationMovement` record so there is a ledger of location movements and at what time. Also sets the location field to the current location (is already on item batch so just left it .. )

src/database/DataTypes/ItemBatch.js
```diff
@@ -145,6 +150,20 @@ export class ItemBatch extends Realm.Object {
...
+  }
+
+  applyVvmStatus(database, newVvmStatus) {
+    return this.shouldApplyVvmStatus(newVvmStatus)
+      ? createRecord(database, 'VaccineVialMonitorStatusLog', this, newVvmStatus)
+      : null;
```
joshxg, line 165: same as above. If the vvm status is not different than what it currently is, just do nothing

src/database/DataTypes/StocktakeBatch.js
```diff
@@ -211,7 +211,25 @@ export class StocktakeBatch extends Realm.Object {
    */
   set countedTotalQuantity(quantity) {
     // Handle packsize of 0.
-    this.countedNumberOfPacks = this.packSize ? quantity / this.packSize : 0;
+    const countedNumberOfPacks = this.packSize ? quantity / this.packSize : 0;
+
+    this.countedNumberOfPacks = countedNumberOfPacks;
+    this.doses = countedNumberOfPacks * this.dosesPerVial;
+  }
```
joshxg, line 218: when changing the counted number of packs, just adjust the number of doses issued to what it 'should' be.. i.e. if the counted number of vials is 10, and each vial has 2 doses, set it to 20. Basically makes the user enterign details have an easier time. They can adjust it more seperately

src/database/DataTypes/StocktakeBatch.js
```diff
@@ -211,7 +211,25 @@ export class StocktakeBatch extends Realm.Object {
...
+
+  get dosesPerVial() {
+    return this.isVaccine ? this.itemBatch?.item?.doses ?? 0 : 0;
+  }
+
+  setDoses(database, newValue) {
+    const maximumDosesPossible = this.dosesPerVial * this.countedTotalQuantity;
+    this.doses = Math.min(newValue, maximumDosesPossible);
+
+    database.save('StocktakeBatch', this);
```
joshxg, line 232: setting doses: just have a maximum rather than a minimum, otherwise entering details gets messy. I.e. if you enter quantity = 10 with number of doses = 2, the number of doses given should really be between 10 and 20. But you can't enter `1`.. so just enforce the max. Not sure if it should check on finalising. Probably should, I'll make an issue

src/database/DataTypes/StocktakeBatch.js
```diff
@@ -268,6 +286,9 @@ export class StocktakeBatch extends Realm.Object {
     this.itemBatch.sellPrice = this.sellPrice;
     this.itemBatch.supplier = this.supplier;
 
+    this.itemBatch.applyVvmStatus(database, this.vaccineVialMonitorStatus);
+    this.itemBatch.applyLocation(database, this.location);
```
joshxg, line 290: when finalising a stocktake, each stocktake batch handles it's own finalising. So when this stocktake batch is finalised, it adjusts the underlying item batches vvm status and location

src/database/DataTypes/StocktakeBatch.js
```diff
@@ -296,6 +317,8 @@ export class StocktakeBatch extends Realm.Object {
       // (i.e. always treat as positive).
       const snapshotDifference = Math.abs(this.difference);
       transactionBatch.setTotalQuantity(database, snapshotDifference);
+      transactionBatch.doses = this.doses;
+
```
joshxg, line 321: This is for making an inventory adjustment for a stocktake batch. Make sure the doses is correctly set on the transaction batch

src/database/DataTypes/StocktakeBatch.js
```diff
@@ -337,7 +360,7 @@ StocktakeBatch.schema = {
     option: { type: 'Options', optional: true },
     supplier: { type: 'Name', optional: true },
     location: { type: 'Location', optional: true },
-    doses: 'double?',
+    doses: { type: 'double', default: 0 },
```
joshxg, line 363: this really just makes UI stuff easier

src/database/DataTypes/Transaction.js
```diff
@@ -419,8 +419,14 @@ export class Transaction extends Realm.Object {
...
         costPrice,
         sellPrice,
+        doses,
+        location,
+        vaccineVialMonitorStatus,
       } = transactionBatch;
 
+      itemBatch.applyVvmStatus(database, vaccineVialMonitorStatus);
+      itemBatch.applyLocation(database, location);
+
```
joshxg, line 429: same as the stocktake, just setting the underlying item batches location adn vvm status when finalising a transaction

src/database/DataTypes/TransactionBatch.js
```diff
@@ -112,19 +112,43 @@ export class TransactionBatch extends Realm.Object {
...
+    // When transactions are confirmed (Customer invoices are always AT LEAST confirmed) adjustments
+    // to quantities effect stock levels. Adjust the underlying item batch with the difference.
     if (this.transaction.isConfirmed) {
+      // Incoming transactions increase stock levels, while outgoing decrease.
       const inventoryDifference = this.transaction.isIncoming ? difference : -difference;
+      const dosesAdjustment = this.transaction.isIncoming ? dosesDifference : -dosesDifference;
+
       this.itemBatch.totalQuantity += inventoryDifference;
+      this.itemBatch.doses += dosesAdjustment;
+
```
joshxg, line 137: added comments to this, hopefully it's clear!

src/database/DataTypes/TransactionItem.js
```diff
@@ -273,6 +273,43 @@ export class TransactionItem extends Realm.Object {
...
+  setDoses(database, value) {
+    const dosesToSet = Math.min(value, this.totalQuantity * this.dosesPerVial);
+    const dosesToAssignToEachBatch = dosesToSet / this.totalQuantity;
+
+    this.resetAllDoses(database);
+    this.batches.sorted('expiryDate', false).forEach(batch => {
+      const { totalQuantity: thisBatchesQuantity } = batch;
+      batch.setDoses(database, Math.floor(dosesToAssignToEachBatch * thisBatchesQuantity));
+    });
+  }
```
joshxg, line 292: this behaviour is from the old vaccines branch. This is for issuing in a customer invoice, which is a single line for multiple batches, which for normal quantity is dispensed first expiry, first out. For doses, the behaviour was intended to divide the number of doses evenly over the number of batches for that line

src/database/DataTypes/VaccineVialMonitorStatus.js
```diff
@@ -15,7 +15,7 @@ VaccineVialMonitorStatus.schema = {
     description: { type: 'string', optional: true },
     code: { type: 'string', optional: true },
     level: { type: 'double', default: 0 },
-    isActive: { type: 'string', default: false },
+    isActive: { type: 'bool', default: false },
```
joshxg, line 18: embares

src/pages/dataTableUtilities/getPageInfoColumns.js
```diff
@@ -85,19 +98,19 @@ const PAGE_INFO_ROWS = (pageObject, dispatch, route) => ({
   },
   numberOfAffectedBatches: {
     title: `${pageInfoStrings.number_of_affected_batches}:`,
-    info: pageObject.numberOfAffectedBatches(UIDatabase),
+    info: pageObject?.numberOfAffectedBatches?.(UIDatabase),
   },
   breachDuration: {
     title: `${pageInfoStrings.duration}:`,
-    info: formatTimeDifference(pageObject.duration),
+    info: pageObject.duration ? formatTimeDifference(pageObject.duration) : '',
```
joshxg, line 105: bug fixes :(

src/widgets/modals/StocktakeBatchModal.js
```diff
@@ -100,6 +100,18 @@ export const StocktakeBatchModalComponent = ({ stocktakeItem, page, dispatch: re
   const onEditSellPrice = (newValue, rowKey) =>
     dispatch(PageActions.editSellPrice(newValue, rowKey));
 
+  const onEditBatchDoses = (newValue, rowKey) =>
```
joshxg, line 103: this whole component is quite a mess. Will make an issue to refactor :(

src/pages/dataTableUtilities/actions/cellActions.js
```diff
@@ -36,6 +36,53 @@ export const refreshIndicatorRow = route => ({
...
+};
+
+export const editBatchVvmStatus = (vvmStatus, objectType, route) => (dispatch, getState) => {
+  const { modalValue, keyExtractor } = selectPageState(getState());
+
+  UIDatabase.write(() =>
+    UIDatabase.update(objectType, { ...modalValue, vaccineVialMonitorStatus: vvmStatus })
+  );
+
+  reduxBatch(() => {
```
wlthomson, line 57: I get the idea of the `reduxBatch` rename when dealing with vaccine batches, but something about it feels wrong to me! I tend to rename local variables rather than imported ones, as it maintains consistency across files (e.g. if I want to do a global search for all the times `batch` is used. I think if we rename like this, we should do this consistently across all files?

src/database/DataTypes/ItemBatch.js
```diff
@@ -145,6 +150,20 @@ export class ItemBatch extends Realm.Object {
 
     return newLocationId !== currentLocationId;
   }
+
+  applyLocation(database, newLocation) {
+    if (!this.shouldApplyLocation(newLocation)) return null;
```
wlthomson, line 155: Hm, feels like maybe this should be called by the caller rather than here? I feel like if the calling code tells the object to do something, the object shouldn't decide it knows better? I get that it probably enables cleaner calling code, but it feels like it might be a bit of an anti-pattern?

src/database/DataTypes/ItemBatch.js
```diff
@@ -145,6 +150,20 @@ export class ItemBatch extends Realm.Object {
...
+
+    this.location = newLocation;
+
+    return createRecord(database, 'LocationMovement', this, newLocation);
+  }
+
+  applyVvmStatus(database, newVvmStatus) {
+    return this.shouldApplyVvmStatus(newVvmStatus)
```
wlthomson, line 163: Same as above, feels odd to have a method called `applyX` which only sometimes actually does that?

src/database/DataTypes/StocktakeBatch.js
```diff
@@ -211,7 +211,25 @@ export class StocktakeBatch extends Realm.Object {
    */
   set countedTotalQuantity(quantity) {
     // Handle packsize of 0.
-    this.countedNumberOfPacks = this.packSize ? quantity / this.packSize : 0;
+    const countedNumberOfPacks = this.packSize ? quantity / this.packSize : 0;
+
+    this.countedNumberOfPacks = countedNumberOfPacks;
```
wlthomson, line 216:
```suggestion
    this.countedNumberOfPacks = this.packSize ? quantity / this.packSize : 0;
```

src/database/DataTypes/StocktakeBatch.js
```diff
@@ -211,7 +211,25 @@ export class StocktakeBatch extends Realm.Object {
...
+    this.doses = countedNumberOfPacks * this.dosesPerVial;
+  }
+
+  get isVaccine() {
+    return this.itemBatch?.item?.isVaccine ?? false;
```
wlthomson, line 221:
```suggestion
    return this.itemBatch?.isVaccine ?? false;
```

src/database/DataTypes/StocktakeBatch.js
```diff
@@ -211,7 +211,25 @@ export class StocktakeBatch extends Realm.Object {
...
+  }
+
+  get dosesPerVial() {
+    return this.isVaccine ? this.itemBatch?.item?.doses ?? 0 : 0;
```
wlthomson, line 225: Should this be:

```suggestion
    return this.isVaccine ? this.itemBatch?.doses ?? 0 : 0;
```

Also, think you can make it a bit more concise (but possibly a bit less explicit):

```suggestion
    return this.isVaccine && this.itemBatch?.doses || 0;
```

src/database/DataTypes/StocktakeBatch.js
wlthomson, line 232: Nit: maybe a helper method?

```suggestion
  get totalDoses() { 
    return this.dosesPerVial * countedTotalQuantity;
  }

  setDoses(database, newValue) {
    this.doses = Math.min(newValue, this.totalDoses);
    database.save('StocktakeBatch', this);
  }
```

wlthomson, line 232: Should the saving be done by the caller? Then could just make this a tidy setter?

src/database/DataTypes/StocktakeBatch.js
```diff
@@ -211,7 +211,25 @@ export class StocktakeBatch extends Realm.Object {
    */
   set countedTotalQuantity(quantity) {
     // Handle packsize of 0.
-    this.countedNumberOfPacks = this.packSize ? quantity / this.packSize : 0;
+    const countedNumberOfPacks = this.packSize ? quantity / this.packSize : 0;
+
+    this.countedNumberOfPacks = countedNumberOfPacks;
+    this.doses = countedNumberOfPacks * this.dosesPerVial;
```
wlthomson, line 217: Should this possibly be `countedNumberOfDoses`?

src/database/DataTypes/StocktakeBatch.js
```diff
@@ -211,7 +211,25 @@ export class StocktakeBatch extends Realm.Object {
...
+
+  get isVaccine() {
+    return this.itemBatch?.item?.isVaccine ?? false;
+  }
+
+  get dosesPerVial() {
+    return this.isVaccine ? this.itemBatch?.item?.doses ?? 0 : 0;
+  }
+
+  setDoses(database, newValue) {
```
wlthomson, line 228: `setCountedDoses`?

src/database/DataTypes/StocktakeBatch.js
wlthomson: Don't know much about this logic, but if `dosesPerVial` can be any significant number, couldn't this lead to stocktakes indicating a lot more doses are available than actually are? Could that cause inventory issues down the line?

src/database/DataTypes/Transaction.js
```diff
@@ -430,12 +436,17 @@ export class Transaction extends Realm.Object {
       const newNumberOfPacks = isIncomingInvoice
         ? itemBatch.numberOfPacks + packedToOneQuantity
         : itemBatch.numberOfPacks - packedToOneQuantity;
+
+      const newDoses = isIncomingInvoice ? itemBatch.doses + doses : itemBatch.doses - doses;
+
```
wlthomson, line 441: Suggestion:

```suggestion
      const dosesOffset = isIncomingInvoice ? doses : -doses;
      const newDoses = itemBatch.doses + dosesOffset;
```

src/database/DataTypes/TransactionBatch.js
```diff
@@ -112,19 +112,43 @@ export class TransactionBatch extends Realm.Object {
...
 
     database.save('TransactionBatch', this);
   }
 
+  get isVaccine() {
+    return this.itemBatch?.isVaccine ?? false;
+  }
+
+  get dosesPerVial() {
+    return this.isVaccine ? this.itemBatch?.item?.doses ?? 0 : 0;
```
wlthomson, line 149: Can do this?

```suggestion
    return this.isVaccine && this.doses || 0;
```

src/database/DataTypes/StocktakeBatch.js
joshxg: there are a few comments here regarding doses etc: 

- `item.doses` is the number of doses in a pack of an item. 
- Example item: `VaccineA`
- `item.doses` is the number of doses in a vial of `VaccineA` [Lost the fight on calling it `dosesPerVial` :( ..! ]. So if `item.doses == 2`, each vial of `VaccineA` has two doses in it. 
- Vials in general have a very short life span after being opened, so there is often open vial wastage, resulting in vials being only 'half used'
- `..Batch.doses` is tracking the actual number of doses. 
- The `ItemBatch`s for `VaccineA` have `doses` which is the number of doses currently available - so it should theoretically  always be `item.doses * numberOfPacks * packSize`, but with the wastage from a vial, so it could be less. 
- `TransactionBatch` have a `doses` field which is tracking how many doses were in some transaction.
- `StocktakeBatch` have a `doses` field which is tracking how many doses of a batch are left. There are vaccines which have 10 doses. If you gotta chuck it out with 9 doses left in it, you can make an inventory adjustment in the stocktake which reduces the number of doses by 9, if you want, rather than a full pack.

probably more confusing now, but this sets an upper limit on the `StocktakeBatch` for the number of doses. You're right though - the number of doses won't really ever be that accurate!

src/pages/dataTableUtilities/actions/cellActions.js
joshxg: actually it was already renamed ages ago - theres a method which has a param `batch` cause it can be a transactionbatch or stocktake batch and I DONT KNOW another name :(

src/database/DataTypes/ItemBatch.js
joshxg: 💯 1f2e2fed2c06fec822d90d19c9f791c6259616f7

src/database/DataTypes/ItemBatch.js
joshxg: 💯 1f2e2fed2c06fec822d90d19c9f791c6259616f7

src/database/DataTypes/StocktakeBatch.js
joshxg: that's probably a better name hmmmm

src/database/DataTypes/StocktakeBatch.js
joshxg: c015c7171c19ae8c19ef03372113daf2acb9dca9

src/database/DataTypes/StocktakeBatch.js
joshxg: your suggestion needs to have parenteitheris `(this.isVaccine && this.itemBatch?.doses) || 0`

But nah, need to use `item.doses` which is the number of doses per vial as above

src/database/DataTypes/StocktakeBatch.js
joshxg: i'll keep it as `setDoses` since the field is `doses` but will try and argue more to change `item` to `dosesPerVial` and will try go for `countedDoses` also

src/database/DataTypes/StocktakeBatch.js
joshxg: hmm, would be `get maximumDoses` - will do if I use the calculation another time - cheers

joshxg: I'm struggling - the negative side effects of not using `database.save` is pretty big but I know the setter is nicer. I don't knowwww

joshxg:
thanks @wlthomson that must have been an annoying one to review - will try keep them smaller :( Will get your feedback on all the new field names and if they make sense to you soon

PR #2690, opened 2020-04-08
#2689 fix dispensing read only constraints

### LINKED ISSUE #2689: Read-only constraints not being consistently applied

A few problems/inconsistencies with dispensing for non-local patients:

- the `supplyingStoreId` field is not being correctly set for patients created via the lookup API. 
- the `storeId` field is not being correctly set for patients created via the lookup API. 
- non-local patients/prescribers cannot be edited, but the edit form can still be accessed and the icon is not disabled as it is for policies.

- correctly set supplying store fields for patients/prescribers.
- disable edit icon in the dispensing window for non-local patients/prescribers

## Implementation

- update `PatientActions.patientUpdate`.
- update `PrescriberActions.updatePrescriber`
- update `PrescriptionInfo` component.

N/A.

## Additional context

See epic #2446.

wlthomson:
Fixes #2689.

## Change summary

- Update dispensary "edit" icons for read-only patients, prescribers, policies.
- Update policy editing logic to prevent editing read-only policies.
- Fix home store not being correctly set for patients and prescribers added via API.

## Screenshots

Example of dispensing for a read-only patient (prescriber is editable):

#2689-fix-dispensing-read-only-constraints-1

Example of rows in dispensary page (second patient is read-only):

#2689-fix-dispensing-read-only-constraints-2 png

## Testing

- [ ] Remote patients "edit" icon is a book in the dispensary page patient list. 
- [ ] Remote prescribers "edit" icon is a book in the dispensary page prescriber list.
- [ ] Remote patients "edit" icon is a book in the prescription wizard.
- [ ] Remote prescribers "edit" icon is a book in the prescription wizard.
- [ ] Policies belonging to remote patients "edit" icon is a bookin the prescription wizard.

Still need to resolve the discussion of when to allow editing of patients (#2544). For now have restricted editing to patient home store. This fixes immediate sync problems, but will need to make sure this is documented if we go with this... if a user makes a mistake creating a policy during user testing, they can't fix it without the server admin!


src/widgets/PrescriptionInfo.js
```diff
@@ -58,7 +65,11 @@ const PrescriptionInfoComponent = ({
 
           <FlexRow flex={1}>
             <CircleButton IconComponent={HistoryIcon} onPress={viewHistoryCallback} />
-            <CircleButton IconComponent={PencilIcon} onPress={editPatientCallback} />
+            <CircleButton
+              IconComponent={PencilIcon}
+              isDisabled={!canEditPatient}
+              onPress={editPatientCallback}
+            />
```
joshxg, line 72: I think this is the wrong way to go about this. I think you should still be able to press this button to bring up the patient details, but the form should be in read only. I thought it was already doing this?

That way you can always have access and see all of the patients details

src/actions/PatientActions.js
```diff
@@ -67,6 +67,7 @@ const patientUpdate = patientDetails => (dispatch, getState) => {
     addressOne: patientLine1,
     addressTwo: patientLine2,
     country: patientCountry,
+    supplyingStoreId: patientSupplyingStoreId,
```
joshxg, line 70: ahh ok I think I see whatsup!

The patient form isn't already disabled for patients being fetched from the API?

I Think that's because the default in `createRecord` for a patient is that the field `thisStoresPatient` is being set as true and we're not overriding it for a non-local patient (same for a prescriber?).

I think if we change this that should be all that is needed

wlthomson:
@josh updated logic and added some screenshots of the new icon.

src/widgets/FormControl.js
```diff
@@ -47,12 +55,17 @@ const FormControlComponent = ({
   // Save button state
   canSaveForm,
   saveButtonText,
+  // Confirm form state
+  confirmOnSave,
+  isConfirmFormOpen,
+  confirmText,
```
joshxg, line 61: do we need prop types for these

src/widgets/FormControl.js
```diff
@@ -223,23 +250,29 @@ const FormControlComponent = ({
...
   const form = selectForm(state);
   const canSaveForm = selectCanSaveForm(state);
+  const isConfirmFormOpen = selectIsConfirmFormOpen(state);
   const completedForm = selectCompletedForm(state);
-  return { form, canSaveForm, completedForm };
+  const { confirmOnSave } = ownProps;
+  console.log('FORM:', form);
+  console.log('CANSAVEFORM:', canSaveForm);
+  console.log('ISCONFIRMFORMOPEN:', isConfirmFormOpen);
+  console.log('COMPLETEDFORM:', completedForm);
```
joshxg, line 274: chef

src/utilities/formInputConfigs.js
```diff
@@ -221,7 +221,7 @@ const FORM_INPUT_CONFIGS = seedObject => ({
     type: FORM_INPUT_TYPES.TEXT,
     initialValue: '',
     key: 'policyNumberPerson',
-    validator: null,
+    validator: input => input.length > 0 && input.length < 50,
```
mark-prins, line 224: Yes, I know. we've been here before. You may safely ignore this comment.
In desktop, only the family number is mandatory now.

src/widgets/DataTable/DataTableRow.js
```diff
@@ -83,10 +84,13 @@ const DataTableRow = React.memo(
...
 
-          // This cell is disabled if the pageObject is finalised, the row has been explicitly set
-          // as disabled, or the rowData is disabled (i.e. data is an invoice),
-          const isDisabled =
-            isFinalised || (rowState && rowState.isDisabled) || rowData.isFinalised;
+          // This cell is disabled if:
+          // - the pageObject is finalised
+          // - the row has been explicitly set as disabled
+          // - the rowData is disabled (i.e. data is a finalised invoice)
+          const { isFinalised: isFinalisedPage = false } = rowData ?? {};
```
mark-prins, line 91: If I'm reading this correctly, the global `isFinalised` relates to the page and this one relates to the row

```suggestion
          const { isFinalised: isFinalisedRow= false } = rowData ?? {};
```

src/reducers/FormReducer.js
```diff
@@ -22,9 +22,13 @@ const initialState = config => {
     }),
     {}
   );
+  const formState = { isConfirmFormOpen: false };
+
+  return { formConfig, formState };
```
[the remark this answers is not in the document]
wlthomson: Yeah, initially I didn't use it, but I think it's a bit tricky/cludgey not to as it breaks a lot of logic which maps/iterates over all the fields to display the form/check if it is valid etc.  

I think it is relatively nice to separate out the input configs from the other form state, but right now it's a bit stupid looking as there's only one field. I do think it'll come in handy though.

I agree re. the nesting, but I think on balance its worth it to prevent workarounds throughout the rest of the code. Also could potentially tidy this up using `combineReducers`!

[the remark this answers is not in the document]
joshxg: Sorry I just don't know how having `{formConfig: { .. }, isConfirmModalOpen: true }` really messes with the iteration as you're always iterating over `formConfig`? Any additional fields which would then need to be added to `formState` object would then comfortably sit in the first level and be easier to access so they're seperate and easier with less nesting!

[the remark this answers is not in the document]
wlthomson: Sorry, I misunderstood (I thought you were suggesting having all fields as flat). Agree, just having them flat is good idea (although I still think there's something to be said for separating them and using `combineReducers`).

src/widgets/FormControl.js
wlthomson: Oops, forgot to remove my awesome new SNIPPET.

src/localization/dispensingStrings.json
```diff
@@ -9,6 +9,7 @@
     "choose_a_reason": "Choose a reason",
     "choose_a_payment_type": "Choose a payment type",
     "click_on_an_item_to_add_it": "Click on an item to add it",
+    "confirm_new_policy": "This policy is associated with a read-only patient record and will only be editable on this patient's home store. Please ensure that all details are correct before confirming.",
```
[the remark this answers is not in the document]
wlthomson: I think it's a pretty good amount of info? I think `"no one else will be able to edit"` without explaining why that's the case would probably confuse/frustrate users?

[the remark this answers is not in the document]
wlthomson: How about:

```
This policy will only be editable on this patient's home store. Please ensure that all details are correct before confirming.
```

src/widgets/FormControl.js
wlthomson: c68c5b681fdbcb9b89f58884d0a39893695f4dee.

src/widgets/FormControl.js
wlthomson: 6533a87bbf31d6eefcdcfeac7ab25515a99b4ac7.

src/widgets/DataTable/DataTableRow.js
wlthomson: 0ddf593d0ee261753ca1e7246acb131e7d749000.

PR #2696, opened 2020-04-14
#2382 remove reason column from stocktake edit page

### LINKED ISSUE #2382: Remove the 'Reason' column from the main stocktake screen

Place holder issue for Emily's comments from Super User Training from GTT:

Stocktake:   Remove the Reason column from the main stocktake screen. Reason should only   appear when you click into 'Lot' and specify the reason for each batch   individually.

wlthomson:
Fixes #2382.

## Change summary

Very basic, just removes the reason column from the stocktake edit page. 

Still a WIP.

@joshxg @mark-prins Trying to find a median between the UAT feedback and the requirements:

- Do we want the reason popup to still trigger? It doesn't seem to make sense to hide the column and still trigger a popup?
- Maybe quantities should only be enterable via batches? That way the reason column can be removed and only applied via the batch modal?

## Testing

TODO.

N/A.


src/pages/dataTableUtilities/actions/cellActions.js
```diff
@@ -286,12 +287,24 @@ export const enforceReasonChoice = (rowKey, route) => (dispatch, getState) => {
...
 
   const { validateReason } = objectToEdit;
+
   if (!validateReason) {
-    return dispatch(openModal(MODAL_KEYS.ENFORCE_STOCKTAKE_REASON, rowKey, route));
+    // If updated counted quantity in stocktake page, set batches with default
+    // adjustment reason.
+    if (route === ROUTES.STOCKTAKE_EDITOR) {
+      const [defaultReason] = UIDatabase.objects(
+        difference > 0 ? 'PositiveAdjustmentReason' : 'NegativeAdjustmentReason'
```
mark-prins, line 301: I was trying to test this part of the requirement: *allow opening the modal without a reason currently selected (currently you can't open the modal without one selected)*

and I can't because there is always the first item in the list selected 😜 

I would have expected that we'd have a blank here, with the user needing to select a reason. I've been told many times to not default a user input, such as an explanatory reason, as that would skew the values. Better to have nothing than a pre-populated value.

src/pages/dataTableUtilities/actions/cellActions.js
```diff
@@ -18,6 +18,7 @@ import {
 import { ACTIONS } from './constants';
 import { openModal, closeModal } from './pageActions';
 import { selectPageState } from '../../../selectors/pages';
+import { ROUTES } from '../../../navigation/index';
```
joshxg, line 21: nit: don't need index

src/pages/dataTableUtilities/actions/cellActions.js
joshxg: I agree - but when this happens because we aren't applying the default reason, the enforcement is going to start again. 

When using reasons the action changes : L205-207 in `StocktakeEditPage` - if we don't use the `withReasons` action, just the regular reason, then the enforcement won't happen. 

The enforcing still happens on a batch too - that's from a convoluted process of getting the actions from `getPageActions` which is called from `usePageReducer`. If we just pass the base page actions in `usePageReducer`, we can delete `getPageActions` completely, as well as the actions for editing a counted quantity with reasons and the `enforceReasonChoice` action! 

But this will mean a small adjustment to `DropDownCell`, to show the icon even when no value is set

joshxg:
Also my opinions to your q's

>  Do we want the reason popup to still trigger? It doesn't seem to make sense to hide the column and still trigger a popup?

Yeah also the popup is a UX nightmare and is hoorrrrible!

> Maybe quantities should only be enterable via batches? That way the reason column can be removed and only applied via the batch modal?

At that point we may as well just list all batches in the stocktake and not have a stocktake batch modal popup edit thing at all - I think leave it editable and we can iterate on the stocktake workflow, which I think could use some adjustments as a whole (why do you select items on a different page???)

src/pages/dataTableUtilities/actions/cellActions.js
mark-prins: > the enforcement is going to start again

I know! I was avoiding saying that because I thought it would create trouble again.
That all sounds good though. Sorry @wlthomson this may have blown the scope out a little - but it will be awesome if you can do it!

wlthomson:
@joshxg @mark-prins Just to confirm the scope here:

- Don't enforce reasons at all for stocktakes (including batches).
- Don't set a default reason, just `null`

Just to make sure, does that mean we saying we are OK with users choosing not to enter a reason at all?

joshxg:
@wlthomson ya. That's my understanding

mark-prins:
@wlthomson me too. I expect a client or two would prefer forcing selection: it's more problematic than useful though.

wlthomson:
@joshxg @mark-prins Did as asked, but I think there might be a few issues here which might need resolving.

By removing the reason enforcement logic, the user can now:

- set a reason when there is no difference between snapshot and counted quantities.
- set a positive reason for a negative difference (and vice versa).
- edit suppliers for batches which have `N/A` suppliers.

I have a feeling some (all?) of these might be things we don't want to allow?

joshxg:
> set a reason when there is no difference between snapshot and counted quantities.
> set a positive reason for a negative difference (and vice versa).

oh yup should definitely disable the cell when the difference = 0 and the reason selection should be based on the -/+ of the difference and maybe removing the reason when the difference = 0?

> edit suppliers for batches which have N/A suppliers.

Dunno what that means soz! I think it means if there is not supplier on a batch, you can edit it. Sounds like thats a good thing though and possibly an oversight on the original implementation

mark-prins:
oh. yes - what @joshxg said. I agree with the 'disable when 0' and 'set the +ve/-ve reasons' and with the 'don't understand what the suppliers problem is'. 

Sorry for the long winded requirements definition process on this one!

wlthomson:
> Dunno what that means soz! I think it means if there is not supplier on a batch, you can edit it. 

Yeah, previously you couldn't.

> Sounds like thats a good thing though and possibly an oversight on the original implementation

Sounds like it, just wanted to double check it wasn't intentional!

PR #2697, opened 2020-04-14
#2383 update expiry date input format to dd/mm/yyyy

### LINKED ISSUE #2383: Change the stocktake date format to show DD/MM/YYYY

Place holder issue for Emily's comments from Super User Training from GTT:

Stocktake:   Change the date format to DD/MM/YYYY (currently only month and year displayed)

wlthomson:
Fixes #2383.

## Change summary

Changes expiry date format from `MM/YYYY` to `DD/MM/YYYY`.

May need to put this behind a pref?

## Testing

- [ ] Expiry dates are formatted `DD/MM/YYYY`.
- [ ] Expiry dates are "autocorrected" with slashes, zeros.

N/A.

### DIFF
--- a/src/hooks/useExpiryDateMask.js
+++ b/src/hooks/useExpiryDateMask.js
@@ -3,7 +3,7 @@ import { useState } from 'react';
 
 /**
  * Simple custom hook which manages the state of input for an
- * expiry date in the form mm/yy[yy] such that the entered date
+ * expiry date in the form dd/mm/yy[yy] such that the entered date
  * always stays valid.
  *
  * @param {String} initialState Initial state of the expiry date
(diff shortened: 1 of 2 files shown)


src/hooks/useExpiryDateMask.js
```diff
@@ -12,49 +12,75 @@ export const useExpiryDateMask = initialState => {
...
-    if (date.length < 4) return setDate('');
+    // If the expiry date isn't in at least the form DD/MM/Y - reset.
+    if (date.length < 7) return setDate('');
 
-    const [month, year] = date.split('/');
+    const [day, month, year] = date.split('/');
     const prependedYear = `${'200'.substring(0, 4 - year.length)}${year}`;
-    return setDate(`${month}/${prependedYear}`);
+
+    return setDate(`${day}/${month}/${prependedYear}`);
```
joshxg, line 83: This function is predicated on `date` never being an invalid string, previously relying on the regex to ensure that. Since it's not ensuring that anymore, it's not really valid.

If you do type `"31/02/2020"` then un-focus the cell, the actual text within the cell is not reset, as this is a stateful component and the current state is the `date` string. The value that actually gets set in the realm is `null`, though. Easiest solution would probs be to just chuck a `moment(..).isValid()` in thiis `finaliseDate` which would `setDate('')` if false - that should keep things consistent

mark-prins:
We didn't come to a conclusion in the issue, and the original requirement out of the super user training wasn't clear to me. I think that expiry should be `mm/yyyy` as it was (sorry!) and we should just close the issue. 

It's another issue that came from super user training - and so isn't a formal request for change. It's something mentioned in the training and recorded by Emily; I don't see that we should be modifying functionality solely on that basis.

joshxg:
That sucks, sorry @wlthomson :( !

On the bright side I think we can learn that any issues we have listed in the release discussion need to be properly and fully discussed with a conclusion. Earlier rather than later! I think we had most of the issues that are in there listed for quite awhile, so next release, lets check that each and every issue has proper requirements and a solution decided from the start

mark-prins:
Indeed, on both counts!

wlthomson:
@joshxg @mark-prins No worries, I agree with all the above! The regex was a bit of a hack but was fun!

PR #2700, opened 2020-04-14
#2509 add user pref for hiding stocktake snapshot quantity

### LINKED ISSUE #2509: Stocktakes: The ability to hide the theoretical quantity column

Part of the Côte d'Ivoire phase two requirements.

Ability to hide the snapshot quantity when editing a stocktake.
Somehow, only for CI? Configurable by store?

wlthomson:
Fixes #2509.

A bit of a mess here, apologies!

- Adds hide stocktake snapshot user pref. 
- Adds hiding snapshot logic to batches.

Just for reference at this stage, another PR coming for adding the existing store preferences implementation to batches. At a latest stage, this will need to be merged with that and existing stocktake page logic.

WARNING: THIS PR IS NOT COMPATIBLE WITH THE LATEST MSUPPLY AND SHOULD NOT BE MERGED INTO DEVELOP AT THIS STAGE.

## Change summary

- Updated user schema.
- Migrated `isAdmin` user pref.
- Added `showStocktakeCurrentQuantity` user pref.

## Testing

TODO: TESTS TO BE ADDED AFTER DESKTOP CHANGES MERGED (AFTER NEXT RELEASE).

N/A.

### DIFF
--- a/src/database/DataTypes/User.js
+++ b/src/database/DataTypes/User.js
@@ -1,0 +1,75 @@
+/**
+ * mSupply Mobile
+ * Sustainable Solutions (NZ) Ltd. 2019
+ */
+
+import Realm from 'realm';
+
+/**
+ * A user.
+ *
+ * @property  {String}  id
+ * @property  {String}  username
+ * @property  {Date}    lastLogin
+ * @property  {String}  firstName
+ * @property  {String}  lastName
+ * @property  {String}  email
+ * @property  {String}  passwordHash
+ * @property  {String}  salt
+ * @property  {Object}  permissions
+ */
+export class User extends Realm.Object {
+  /* eslint-disable class-methods-use-this */
+  get defaultPermissions() {
+    return {
+      isAdmin: false,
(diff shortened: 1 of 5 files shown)


wlthomson:
Yeah, no need to review now. This PR just acting as a temporary pointer to the branch for now (just so it doesn't get lost or deleted in a tidy up). Not a good habit to get into, just a short term thing (sorry)!

src/database/DataTypes/User.js
```diff
@@ -0,0 +1,75 @@
+/**
+ * mSupply Mobile
+ * Sustainable Solutions (NZ) Ltd. 2019
```
mark-prins, line 3: If I was reviewing, I might annoyingly point out that the template should be updated to `2020` :)

wlthomson:
This has been hanging around for a bit, so I think consensus is that it's not really worth moving hide stocktake quantity to user prefs. Can reopen if we decide is something we want to do!

PR #2701, opened 2020-04-14
#2509 add hide snapshot pref logic to stocktake batch modal

### LINKED ISSUE #2509: Stocktakes: The ability to hide the theoretical quantity column

wlthomson:
Fixes #2509.

Really should have made a new issue, but a bit of time pressure on `v5`.

## Change summary

- Follows up on #2524 by adding column hiding logic to stocktake batches.

## Testing

- [ ] When the `usesHideSnapshotColumn` store pref is `true`, the snapshot quantity column is hidden for stocktakes, **including stocktake batches**.

Initial idea was to migrate this back to a user pref, but have bumped that to desktop `v11`.

For next steps, see desktop and mobile PRs: #2700 and https://github.com/sussol/msupply/pull/6023.

### DIFF
--- a/src/widgets/modals/StocktakeBatchModal.js
+++ b/src/widgets/modals/StocktakeBatchModal.js
@@ -12,6 +12,7 @@ import { connect, batch } from 'react-redux';
 import { MODAL_KEYS, getModalTitle } from '../../utilities';
 import { usePageReducer } from '../../hooks';
 import { getItemLayout } from '../../pages/dataTableUtilities';
+import { COLUMN_KEYS } from '../../pages/dataTableUtilities/constants';
 
 import { GenericChoiceList } from '../modalChildren/GenericChoiceList';
 import { PageInfo, DataTablePageView, PageButton } from '..';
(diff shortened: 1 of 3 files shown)


src/widgets/modals/StocktakeBatchModal.js
```diff
@@ -41,16 +42,27 @@ import { AutocompleteSelector } from '../modalChildren';
  * holding the state of a given row. Each object has the shape :
  * { isSelected, isFocused, isDisabled },
  *
- * @prop {Object} stocktakeItem The realm transaction object for this invoice.
+ * @prop {Object} stocktakeItem The realm transsingHaction object for this invoice.
```
joshxg, line 45: you're a transsingHaction

mark-prins:
> I just realised with the difference column, you can quite easily work out the snapshot quantity

Yes. we did have something of a discussion like this when hiding the column on desktop. End result - client insisted on hiding. I think we ignore the logical problem and do as requested 😃

PR #2703, opened 2020-04-14
#2451 add customer credits to customer invoices page

### LINKED ISSUE #2451: Mobile should list customer credits with customer invoices
## Describe the bug

- If a store creates a supplier credit where the supplier is a store on a mobile site, the customer credit will get lost!

### To reproduce

1. Create an internal requisition to a mobile store.
2. Finalise the Customer Requisition [This will create a CI]
3. Sync [This will create an SI on the desktop site]
4. Return goods from the SI [So, returning goods to the mobile store].
5. Sync. [This will create a CC for the mobile site]
6. See error [Mobile won't list the CC - so it can never be finalised!]

### Expected behaviour

Should list customer credits in customer invoices and allow them to be finalised. This is similar to supplier credits, which are listed with supplier invoices. Although I disagree with this approach as it's also mixing incoming v outgoing stock @craigdrown insists that a supplier credit is a type of supplier invoice and the same for customer credit and he is the boss

### Proposed Solution

Display customer invoices AND customer credits in the customer invoices page. When navigating to a customer credit, navigate to a customer invoice page [When navigating to a supplier credit page, different page info is used, so a different config has been set up. Although the columns are the same and it is always finalised, so you can't edit] - this may require different columns and/or different page info, not sure

### Version and device info

- App version: Forever
- Tablet model: N/A
- OS version: N/A

### Additional context

Technically could have done this forever(?) but the problem is worse now that you can create supplier credits on mobile - so potentially more supplier credits from being supplied from mobile stores.

wlthomson:
Fixes #2451.

Think this does the job! Not an expect on customer credits, so feedback appreciated from anyone who knows more than me... so just anyone :). 

## Change summary

- Adds customer credits to customer invoices page.
- Adds customer credit route/columns (same as customer invoice, but not editable).
- Adds customer credit logic to customer invoice page (same as customer invoice, but add item buttons disabled). 

## Testing

- [ ] Customer credits appear in the customer invoices page.
- [ ] Selecting a customer credit navigates to the customer invoice page.
- [ ] Customer credits cannot be edited.
- [ ] Customer credits can be finalised.

N/A.


src/database/UIDatabase.js
```diff
@@ -15,13 +15,14 @@ const { THIS_STORE_NAME_ID } = SETTINGS_KEYS;
...
     case 'CustomerInvoice':
+    case 'CustomerTransaction':
+    case 'Payment':
     case 'Prescription':
-    case 'SupplierInvoice':
     case 'Receipt':
-    case 'CustomerCredit':
-    case 'Payment':
-    case 'CashTransaction':
+    case 'SupplierInvoice':
```
wlthomson, line 25: Just a sneaky alphabetical ordering, pretend its not here.

giphy

src/database/UIDatabase.js
```diff
@@ -154,6 +159,37 @@ class UIDatabase {
...
+        const invoiceQueryString = 'type == $1 AND (linkedRequisition == $2 OR status == $3)';
+        const queryString = `${creditQueryString} OR ${invoiceQueryString}`;
+        return results.filtered(
+          queryString,
+          'customer_credit',
+          'customer_invoice',
+          null,
+          'finalised'
+        );
+      }
```
wlthomson, line 173: New bit here ^.

wlthomson:
> Add Type column in CustomerInvoicesPage: To say Customer Credit or Customer Invoice (could remove the status column to make room)

Yeah, that's a good idea. Assuming can justify removing status column as basically duplicated by the `"current/past"` toggle?

> The remove column is pressable, but the quantity column is not editable. I think this is possibly just a little in consistent. Either full editable or not? I say fully editable is probably the way to go

Hmm, good point, didn't realise the remove was pressable. is it OK to have the credits editable? I had the idea that they should be locked for editing? Am I wrong on that?

> At the moment the CC is synced with an invoice number -1 (Same as a SI - the server doesn't know what invoice number we at when being created). So we need to change the PostSyncProcessor to also include editing the invoice number for CC's as well as SI's. This part not 100% on: use the CUSTOMER_INVOICE sequence. I'm pretty sure that's right, but not 100%

Good point. I think the `CUSTOMER_INVOICE` makes the most sense.

joshxg:
> Yeah, that's a good idea. Assuming can justify removing status column as basically duplicated by the "current/past" toggle?

ya

> Hmm, good point, didn't realise the remove was pressable. is it OK to have the credits editable? I had the idea that they should be locked for editing? Am I wrong on that?

I dunno what the 'right' thing is. You can edit them in desktop afaik. Thinking something along the lines of: it's essentially the same thing as a supplier invoice: you have some stock turn up from some delivery service. You're essentially verifying at that point whether or not what the person who sent the stuff actually sent that amount- if they didn't, they've just messed up your stock levels and you have to make some inventory adjustments through a stocktake. Maybe we could get a third opinion on if they should be editable or not



-----------

it also occurs to me that I may have forgotten to check the page info for page when it's showing a customer credit. When you go to make the other changes - is the customer properly showing up?

Also - should probably change the title on the page from `invoice X` to `credit X`. This is in `navigation/actions`, just need to conditionally change the title there

src/pages/dataTableUtilities/getColumns.js
```diff
@@ -367,6 +374,14 @@ const COLUMNS = () => ({
...
     editable: false,
   },
+  [COLUMN_NAMES.INVOICE_TYPE]: {
+    type: COLUMN_TYPES.STRING,
+    key: COLUMN_KEYS.INVOICE_TYPE,
+    title: tableStrings.type,
+    alignText: 'left',
+    sortable: true,
+    editable: false,
+  },
```
joshxg, line 384: This is just below a `TRANSACTION_TYPE` column definition !

wlthomson: Yeah, I added this as the `TRANSACTION_TYPE` key doesn't seem to be pretty printed (e.g. `supplier_invoice`). I can see now there is a `formatType` helper which does this inside the data table row logic which substitutes the type for a localised string.

Will update to use the existing logic but I think this might be a small example of where the data table could be simplified. I think it's a bit of a potential smell that certain keys fire conditionals within the data table and are processed/formatted in a special way. My thinking here is that this would be better as a helper on the `Transaction` type (like I have here as `invoiceType`, admittedly not a great name). That way it would a be more explicit and wouldn't bind the view (the data table) as closely to the underlying model (the row object). I know there are probably other examples which make this only a small thing, but wherever it's possible to loosen the binding between the data table and the data types, I think it's probably a good idea?

src/database/DataTypes/Transaction.js
```diff
@@ -157,6 +157,15 @@ export class Transaction extends Realm.Object {
     return this.otherParty && this.otherParty.type === 'inventory_adjustment';
   }
 
+  /**
+   * Get invoice type: credit or invoice.
+   *
+   * @return  {string}
+   */
+  get invoiceType() {
+    return this.isCredit ? generalStrings.credit : generalStrings.invoice;
```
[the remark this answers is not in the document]
wlthomson: Oops, see comment below.

PR #2707, opened 2020-04-15
#2632 add bugsnagging to fetch hook

### LINKED ISSUE #2632: Add bugsnag to catch runtime API errors

The API network helpers are currently very basic and do not handle edge cases. For development, testing, etc. it will be useful to have bugsnag catch these exceptions and report on them.

Add bugsnag calls to `utilities/network/lookupApi`.

## Implementation

See above.

N/A.

## Additional context

See epic #2446.

wlthomson:
Fixes #2632.

## Change summary

Adds bugsnag calls to `useFetchHook`.

## Testing

N/A.

N/A.

### DIFF
--- a/src/hooks/useFetch.js
+++ b/src/hooks/useFetch.js
@@ -21,7 +21,7 @@ const ERROR_CODES = {
 const getAuthorizationHeader = () => {
   const username = UIDatabase.getSetting(SYNC_SITE_NAME);
   const password = UIDatabase.getSetting(SYNC_SITE_PASSWORD_HASH);
-  return { authorization: getAuthHeader(username, password) };
+  return getAuthHeader(username, password);
 };
 
 /**
@@ -37,12 +37,13 @@ export const useFetch = (url, options = {}, timeout = CONNECTION_TIMEOUT_PERIOD)
   React.useEffect(() => {
     if (url) {
       (async () => {
+        /* eslint-disable no-undef */
+        const abortController = new AbortController();
+        const bugsnagClient = new BugsnagClient();
+        const { signal } = abortController;
+        const headers = { authorization: getAuthorizationHeader() };
         try {
-          /* eslint-disable no-undef */
-          const abortController = new AbortController();
-          const { signal } = abortController;
-          const { authorization } = getAuthorizationHeader();
-          const response = await fetch(url, { headers: { authorization }, signal, ...options });
+          const response = await fetch(url, { headers, signal, ...options });
           const { ok, status } = response;
           if (ok) {
             const responseData = await response.json();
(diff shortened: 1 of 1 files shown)


src/hooks/useFetch.js
```diff
@@ -21,7 +21,7 @@ const ERROR_CODES = {
 const getAuthorizationHeader = () => {
   const username = UIDatabase.getSetting(SYNC_SITE_NAME);
   const password = UIDatabase.getSetting(SYNC_SITE_PASSWORD_HASH);
-  return { authorization: getAuthHeader(username, password) };
+  return getAuthHeader(username, password);
```
mark-prins, line 24: this doesn't look like bugsnagging... 🤔 
I thought it might be cross-contamination or offroad, but I see now after looking below. Go for it!

src/hooks/useFetch.js
```diff
@@ -51,15 +52,27 @@ export const useFetch = (url, options = {}, timeout = CONNECTION_TIMEOUT_PERIOD)
...
               case 401:
-                throw new Error(ERROR_CODES.INVALID_PASSWORD);
+                throw authorizationError;
+              default:
+                bugsnagClient.notify(connectionError, content => {
+                  content.requestUrl = url;
+                  content.requestHeaders = headers;
+                  content.responseStatus = status;
+                });
+                throw connectionError;
```
joshxg, line 68: sorry! im confused why we're snagging this one twice, one with the status one without. If we just throw the connection error and add the status into the catch snag, would that work?

src/hooks/useFetch.js
```diff
@@ -51,15 +52,27 @@ export const useFetch = (url, options = {}, timeout = CONNECTION_TIMEOUT_PERIOD)
...
             }
           }
         } catch (responseError) {
+          bugsnagClient.notify(responseError, content => {
+            content.requestUrl = url;
+            content.requestHeaders = headers;
```
joshxg, line 74: I'm not a bugsnag expert, so could be wrong but this one here that was added: 

https://github.com/openmsupply/mobile/blob/d00f1274640307c64fadd8a18df877ba20375954/src/sync/Synchroniser.js#L94-L100

Uses the `metadata` field, and I wouldn't have done it like this unless I basically had to, I imagine! 

This gives an error like this one in the bugsnag (hope this link works): https://app.bugsnag.com/sustainable-solutions-nz-ltd/msupply-mobile/errors/5e4ca18b2bd15d0019031ab6?filters[event.since][0]=2020-02-15T02%3A38%3A24.109Z&filters[search][0][type]=eq&filters[search][0][value]=%22sync%20error%22&filters[error.status][0]=fixed

if it doesn't:

image

Basically the `metaData` field is used to populate the `error` tab (which isn't there if you supply no `metaData` and the `context` is used for the .. 'name'?

The bugsnag docs suck imo! https://docs.bugsnag.com/platforms/react-native/react-native/#sending-diagnostic-data

wlthomson: Hmm, I was lazy and just referenced other bugsnags in the codebase. Looks like you're right though. 

I think there might be a few broken calls in other places (the places I looked at :laughing:)? Maybe I'm missing something... if not we might need to make an issue for this?

`outgoingSyncUtils`:

    bugsnagClient.notify(error, content => {
      content.syncSite = UIDatabase.getSetting(SETTINGS_KEYS.SYNC_SITE_NAME);
      content.record = syncOutRecord;
    });
    ...
    bugsnagClient.notify(error, content => {
      content.syncSite = UIDatabase.getSetting(SETTINGS_KEYS.SYNC_SITE_NAME);
      content.record = syncOutRecord;
    });

`outgoingSyncUtils`:

    bugsnagNotify('Failed parsing response from server', response);
    ...
    bugsnagNotify(responseJson.error, response);


`synchroniser`:

    bugsnagNotify('Failed parsing response from server', response);
      throw new Error('Unexpected response from sync server');
    }
    ...
    bugsnagNotify(responseJson.error, response);
    throw new Error('Server rejected pushed records');

joshxg: I mean it looks like yeah, those are broken. Benefit of the doubt: it was probably an old bugsnag version and the calls never got migrated. To confirm, would need to either trigger, or find snags of those in the big list of snagged errors and check if any of those details are there.

wlthomson: 63fcda17.

src/hooks/useFetch.js
wlthomson: Unfortunately the status is inside the `try`, so can't access it. It's not a biggie, removed in 63fcda1.

PR #2708, opened 2020-04-15
#2592 ItemDetail for requisitions

### LINKED ISSUE #2592: Enhanced item detail component

Require a way to view closed and open vial wastage for a requisition item. Previously an expansion was used. We don't have expansions anymore. Adding even more columns to a supplier requisition page is not the most ideal solution

Take the same/similar component which is used in `StockPage`, and have each row clickable in a requisition (initially only vaccine items), which will show the open and closed vial wastage.

## Implementation

Put the `ItemDetail` component at the root level and make it more generic, so that different pages can potentially show different details.

Make an expansion, use columns

## Additional context

N/A

joshxg:
Fixes #2592 

## Change summary

- Converts the current `ItemDetails` component into a more flexible component(s) which can be used elsewhere:
- this one:
image
- Made a generic `RowDetails` component which is a container component with the intention being that it will wrap other components - similar to the `DataTableModal`.
- Went with this, rather than a different pattern like the how the columns are organised as I think in general the use-cases for this are more unknown and it would be good to be more flexible - This is more of a "chuck info in it so we dont have to use up column space", so it is inherently unstructed and hard to plan a generic solution for
- Made a `RequisitionItemDetails` where the intention is to have it pop up when clicking a requisition item. Currently only going to show for a vaccine item, but I think we could easily put some of the current columns (of which there are quite a lot!) in here, and then have it popup for any row, and only show the vial wastage for vaccines, instead:

image

- Couple of off-road bug fixes which I'll comment on!

## Testing

N/A

- Although right now this is a soluytion for showing vial wastage, I think moving forward it would be a good solution for other problems we are having. For example: IC Want to show far more information than we have column room for. We should easily be able to make a `StocktakeItemDetails` component, and chuck a bunch of info in there - they want 'full row total price' - chuck it in there!
- So in saying that, any ideas on how to improve this for future proofing are definitely welcomed, although I am trying not to over complicate this right now as to let the use cases naturally come and improve the design in iterations
- Did you make it here? wow thanks :)

### DIFF
--- a/src/reducers/RowDetailReducer.js
+++ b/src/reducers/RowDetailReducer.js
@@ -1,6 +1,6 @@
 /**
  * mSupply Mobile
- * Sustainable Solutions (NZ) Ltd. 2019
+ * Sustainable Solutions (NZ) Ltd. 2020
  */
 
 import { ROW_DETAIL_KEYS, ROW_DETAIL_ACTIONS } from '../actions/RowDetailActions';
--- a/src/reducers/RowDetailReducer.js
+++ b/src/reducers/RowDetailReducer.js
@@ -3,12 +3,7 @@
  * Sustainable Solutions (NZ) Ltd. 2019
  */
 
-import { ROW_DETAIL_ACTIONS } from '../actions/RowDetailActions';
-
-const ROW_DETAIL_KEYS = {
-  ITEM_DETAIL: 'itemDetail',
-  REQUISITION_ITEM_DETAIL: 'requisitionItemDetail',
-};
+import { ROW_DETAIL_KEYS, ROW_DETAIL_ACTIONS } from '../actions/RowDetailActions';
 
 const initialState = () => ({
   rowData: null,
--- a/src/reducers/RowDetailReducer.js
+++ b/src/reducers/RowDetailReducer.js
@@ -1,0 +1,46 @@
+/**
+ * mSupply Mobile
+ * Sustainable Solutions (NZ) Ltd. 2019
+ */
+
+import { ROW_DETAIL_ACTIONS } from '../actions/RowDetailActions';
+
+const ROW_DETAIL_KEYS = {
+  ITEM_DETAIL: 'itemDetail',
+  REQUISITION_ITEM_DETAIL: 'requisitionItemDetail',
+};
+
+const initialState = () => ({
+  rowData: null,
+  detailKey: '',
+});
+
+export const RowDetailReducer = (state = initialState(), action) => {
+  const { type } = action;
+
+  switch (type) {
+    case ROW_DETAIL_ACTIONS.OPEN_ITEM_DETAIL: {
+      const { payload } = action;
+      const { rowData } = payload;
+
+      return { ...state, rowData, detailKey: ROW_DETAIL_KEYS.ITEM_DETAIL };
+    }
+
+    case ROW_DETAIL_ACTIONS.OPEN_REQUISITION_ITEM_DETAIL: {
+      const { payload } = action;
+      const { rowData } = payload;
+
+      return { ...state, rowData, detailKey: ROW_DETAIL_KEYS.REQUISITION_ITEM_DETAIL };
+    }
+
+    case 'Navigation/BACK':
+    case 'Navigation/REPLACE':
+    case 'Navigation/NAVIGATE':
+    case ROW_DETAIL_ACTIONS.CLOSE: {
+      return initialState();
+    }
+
+    default:
+      return state;
+  }
+};
--- a/src/reducers/TemperatureSyncReducer.js
+++ b/src/reducers/TemperatureSyncReducer.js
@@ -37,8 +37,8 @@ export const TemperatureSyncReducer = (state = initialState(), action) => {
   switch (type) {
     case REHYDRATE: {
       const { payload: previousState } = action;
-      const { temperatureSync: previousTemperatureSyncState } = previousState;
-      const { lastTemperatureSync } = previousTemperatureSyncState;
+      const { temperatureSync: previousTemperatureSyncState } = previousState ?? {};
+      const { lastTemperatureSync = null } = previousTemperatureSyncState ?? {};
 
       return { ...initialState(), lastTemperatureSync };
     }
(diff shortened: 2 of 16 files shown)


src/reducers/RowDetailReducer.js
```diff
@@ -0,0 +1,41 @@
...
+      const { payload } = action;
+      const { rowData } = payload;
+
+      return { ...state, rowData, detailKey: ROW_DETAIL_KEYS.REQUISITION_ITEM_DETAIL };
+    }
+
+    case 'Navigation/BACK':
+    case 'Navigation/REPLACE':
+    case 'Navigation/NAVIGATE':
+    case ROW_DETAIL_ACTIONS.CLOSE: {
```
joshxg, line 34: since this is at the root, whenever we navigate, just close it

src/reducers/TemperatureSyncReducer.js
```diff
@@ -37,8 +37,8 @@ export const TemperatureSyncReducer = (state = initialState(), action) => {
   switch (type) {
     case REHYDRATE: {
       const { payload: previousState } = action;
-      const { temperatureSync: previousTemperatureSyncState } = previousState;
-      const { lastTemperatureSync } = previousTemperatureSyncState;
+      const { temperatureSync: previousTemperatureSyncState } = previousState ?? {};
+      const { lastTemperatureSync = null } = previousTemperatureSyncState ?? {};
```
joshxg, line 41: offroad: bug fix for when there is no previous state to rehydrate

src/reducers/index.js
```diff
@@ -44,4 +45,5 @@ export default combineReducers({
   wizard: WizardReducer,
   fridge: FridgeReducer,
   breach: BreachReducer,
+  rowDetail: RowDetailReducer,
```
joshxg, line 48: do we like this name?

src/widgets/RequisitionItemDetail.js
```diff
@@ -0,0 +1,39 @@
...
+import { DARKER_GREY, WHITE, SUSSOL_ORANGE } from '../globalStyles';
+
+const RequisitionItemDetailsComponent = ({ item }) => (
+  <View style={localStyles.container}>
+    <PageInfo
+      titleColor={SUSSOL_ORANGE}
+      infoColor={WHITE}
+      columns={getPageInfoColumns('requisitionItemDetail')(item)}
+    />
+  </View>
```
joshxg, line 24: this could be a bit nicer memoize the columns and stuff, but with no interactions, it's minimal chance of there being a re-render is minimal so just kept it simple for now

src/widgets/RowDetail.js
```diff
@@ -0,0 +1,150 @@
...
+  headerRowStyle,
+  containerStyle,
+  dismissKeyboardOnOpen,
+  dismissOnKeyboardOpen,
+}) => {
+  // This component renders at the bottom of the screen. Dismiss the keyboard when rendering
+  // so the full screen isn't covered by keyboard and this modal.
+  React.useEffect(() => {
+    if (isOpen && dismissKeyboardOnOpen) Keyboard.dismiss();
+  }, [isOpen]);
```
joshxg, line 44: basically: When you click a row and the keyboard is open, close it

src/widgets/RowDetail.js
```diff
@@ -0,0 +1,150 @@
...
+
+  // Add/removing a listener on mounting/unmounting which will hide this component when the
+  // keyboard shows.
+  React.useEffect(() => {
+    if (dismissOnKeyboardOpen) {
+      Keyboard.addListener('keyboardDidShow', onClose);
+      return () => Keyboard.removeListener('keyboardDidShow', onClose);
+    }
+    return null;
```
joshxg, line 53: basically: When you open the keyboard, close this modal

src/widgets/RowDetail.js
```diff
@@ -0,0 +1,150 @@
...
+  }, []);
+
+  const getDetailComponent = () => {
+    switch (detailKey) {
+      case ROW_DETAIL_KEYS.ITEM_DETAIL:
+        return <ItemDetails item={rowData} />;
+      case ROW_DETAIL_KEYS.REQUISITION_ITEM_DETAIL:
+        return <RequisitionItemDetails item={rowData} />;
+      default:
+        return null;
```
joshxg, line 63: so plan would be to just render a different component for different rows. Seems nicer than an even more generic solution

src/widgets/bottomModals/ItemDetails.js
```diff
@@ -1,36 +1,24 @@
...
- * Modal like component. Opens a small pop-over at the bottom of the screen
- * displaying item details.
- *
- * @param {Bool}   isOpen     Indicator if this modal is open.
- * @param {Object} item       Realm Item object to display details for.
- * @param {Func}   onClose    Callback for closing the modal.
- * @param {Any}    modalProps Any additional props for the modal component.
- */
-export const ItemDetailsComponent = ({ isOpen, item, onClose, ...modalProps }) => {
+export const ItemDetailsComponent = ({ item }) => {
```
joshxg, line 19: just removing the modal container and making this a child component

src/reducers/RowDetailReducer.js
```diff
@@ -0,0 +1,41 @@
+/**
+ * mSupply Mobile
+ * Sustainable Solutions (NZ) Ltd. 2019
```
mark-prins, line 3: :nit 2020

src/reducers/index.js
mark-prins: we think we do. I can't think of a better  at the moment, and that's good enough for me.

src/selectors/rowDetail.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * mSupply Mobile
+ * Sustainable Solutions (NZ) Ltd. 2019
```
mark-prins, line 3: :nit (sorry)

src/widgets/RequisitionItemDetail.js
mark-prins: agree - there's no real need to memoize this one

PR #2727, opened 2020-04-20
#2596 VacicneModuleAdmin page

### LINKED ISSUE #2596: Vaccine Module Admin page

Need a way to admin sensors and their locations

A vaccine module admin page.

## Implementation

Page should have a Table listing all of the current sensors with the ability to search for new sensors.

Old implementation: https://github.com/openmsupply/mobile/blob/vaccine-development/src/pages/VaccineModuleAdminPage.js

N/A

## Additional context

Not sure where to navigate to this from.. I am thinking we can tie it in with the current `Settings` page somehow, or at least use the `isAdmin` field for a user from that

joshxg:
Fixes #2596 

## Change summary

- Adds an admin page for vaccines to be able to edit locations and sensors, as well as scan for new sensors (next PR).
- This PR adds basic functionality (stopped to PR to have it be a bit smaller, so not fully complete).
- This is just adding a table with a toggle for sensors/locations, where the table will be editable.

admin1

admin2

## Testing

N/A

- Next PR will include: editable sensors. 
- A button to create a `Location`. [Right of the search bar]
- A button to scan for new sensors. [Right of the search bar]
- Conditional logic to only allow admin access with admin preference on.

### DIFF
--- a/src/database/DataTypes/Sensor.js
+++ b/src/database/DataTypes/Sensor.js
@@ -42,6 +42,10 @@ export class Sensor extends Realm.Object {
   get mostRecentTemperatureBreach() {
     return this.location?.mostRecentTemperatureBreach;
   }
+
+  get batteryLevelString() {
+    return `${this.batteryLevel}%`;
+  }
 }
 
 Sensor.schema = {
@@ -50,7 +54,7 @@ Sensor.schema = {
   properties: {
     id: 'string',
     macAddress: { type: 'string', optional: true },
-    name: { type: 'string', optional: true },
+    name: { type: 'string', default: '' },
     location: { type: 'Location', optional: true },
     batteryLevel: { type: 'double', default: 0 },
     sensorLogs: { type: 'linkingObjects', objectType: 'SensorLog', property: 'sensor' },
--- a/src/pages/VaccineAdminPage.js
+++ b/src/pages/VaccineAdminPage.js
@@ -19,7 +19,7 @@ import { ROUTES } from '../navigation';
 import { generalStrings, vaccineStrings } from '../localization';
 import globalStyles from '../globalStyles';
 
-export const VaccineAdminPageComponent = ({
+const VaccineAdminPageComponent = ({
   data,
   sortKey,
   isAscending,
--- a/src/pages/VaccineAdminPage.js
+++ b/src/pages/VaccineAdminPage.js
@@ -1,0 +1,140 @@
+/* eslint-disable no-unused-vars */
+/* eslint-disable react/forbid-prop-types */
+/**
+ * mSupply Mobile
+ * Sustainable Solutions (NZ) Ltd. 2020
+ */
+
+import React from 'react';
+import { View } from 'react-native';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+
+import { ToggleBar, DataTablePageView, SearchBar, PageButton } from '../widgets';
+import { DataTable, DataTableRow, DataTableHeaderRow } from '../widgets/DataTable';
+
+import { recordKeyExtractor, getItemLayout, getPageDispatchers } from './dataTableUtilities';
+
+import { ROUTES } from '../navigation';
+import { generalStrings, vaccineStrings } from '../localization';
+import globalStyles from '../globalStyles';
+
+export const VaccineAdminPageComponent = ({
+  data,
+  sortKey,
+  isAscending,
+  columns,
+  dataSet,
+  modalKey,
+  dataState,
+  searchTerm,
+  onFilterData,
+  onEditLocationCode,
+  onEditLocationDescription,
+  onSortColumn,
+  onToggleFridges,
+  onToggleSensors,
+}) => {
+  const getCallback = colKey => {
+    switch (colKey) {
+      case 'description':
+        return onEditLocationDescription;
+      case 'code':
+        return onEditLocationCode;
+      default:
+        return null;
+    }
+  };
+
+  const renderHeader = React.useCallback(
+    () => (
+      <DataTableHeaderRow
+        columns={columns}
+        onPress={onSortColumn}
+        isAscending={isAscending}
+        sortKey={sortKey}
+      />
+    ),
+    [sortKey, isAscending, columns]
+  );
+
+  const renderRow = React.useCallback(
+    listItem => {
+      const { item, index } = listItem;
+
+      const rowKey = recordKeyExtractor(item);
+      return (
+        <DataTableRow
+          rowData={data[index]}
+          rowState={dataState.get(rowKey)}
+          rowKey={rowKey}
+          columns={columns}
+          getCallback={getCallback}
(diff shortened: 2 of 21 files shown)


src/database/DataTypes/Sensor.js
```diff
@@ -42,6 +42,10 @@ export class Sensor extends Realm.Object {
   get mostRecentTemperatureBreach() {
     return this.location?.mostRecentTemperatureBreach;
   }
+
+  get batteryLevelString() {
+    return `${this.batteryLevel}%`;
+  }
```
joshxg, line 48: just to make the UI easier. Maybe shouldn't do this, I dunno

src/database/DataTypes/Sensor.js
```diff
@@ -50,7 +54,7 @@ Sensor.schema = {
   properties: {
     id: 'string',
     macAddress: { type: 'string', optional: true },
-    name: { type: 'string', optional: true },
+    name: { type: 'string', default: '' },
```
joshxg, line 57: Makes the UI easier - renders nothing rather than `null`, and.. why not

src/pages/VaccineAdminPage.js
```diff
@@ -0,0 +1,140 @@
+/* eslint-disable no-unused-vars */
```
joshxg, line 1: just props that'll be used in next PR

src/pages/VaccineAdminPage.js
```diff
@@ -0,0 +1,140 @@
...
+        />
+      );
+    },
+    [data, dataState]
+  );
+
+  const toggles = React.useMemo(
+    () => [
+      { text: vaccineStrings.fridges, onPress: onToggleFridges, isOn: dataSet === 'fridges' },
+      { text: vaccineStrings.sensors, onPress: onToggleSensors, isOn: dataSet === 'sensors' },
```
joshxg, line 88: Would like these to be constants, but a little bit .. whatever.. was considering making a `DataSet` or something reducer to manage this but was a bit lazy

src/pages/VaccineAdminPage.js
```diff
@@ -0,0 +1,140 @@
...
+  return (
+    <DataTablePageView>
+      <View style={globalStyles.pageTopSectionContainer}>
+        <ToggleBar toggles={toggles} />
+        <SearchBar
+          onChangeText={onFilterData}
+          value={searchTerm}
+          placeholder={`${generalStrings.search_by} ${generalStrings.code} ${generalStrings.or} ${generalStrings.description}`}
+        />
+        <PageButton text="" />
```
joshxg, line 98: this will be for the scan for sensors / new location button

src/pages/dataTableUtilities/actions/cellActions.js
```diff
@@ -36,6 +36,30 @@ export const refreshIndicatorRow = route => ({
...
+export const editLocationCode = (newValue, rowKey, route) => (dispatch, getState) => {
+  const { data, keyExtractor } = selectPageState(getState());
+
+  const objectToEdit = data.find(row => keyExtractor(row) === rowKey);
+
+  if (objectToEdit) {
+    UIDatabase.write(() => UIDatabase.update('Location', { ...objectToEdit, code: newValue }));
+    dispatch(refreshRow(rowKey, route));
+  }
+};
```
joshxg, line 61: I find these interesting. A lot you could do to make this more dry. Extra prop for the field? Also a prop for the data type? Then wrap it in another thunk ?

🤷

src/pages/dataTableUtilities/getColumns.js
```diff
@@ -41,11 +41,20 @@ const PAGE_COLUMN_WIDTHS = {
   [ROUTES.SUPPLIER_REQUISITIONS]: [1.5, 2, 1, 1, 1, 1],
   [ROUTES.PRESCRIPTION]: [2, 4, 2, 2, 1],
   [ROUTES.PRESCRIPTIONS]: [1.5, 2.5, 2, 1.5, 3, 1],
+  fridges: [1, 3],
+  sensors: [1, 3, 2, 2],
```
joshxg, line 45: happy to get ideas here - technically not a route/modal/tab.. maybe I should use tab? Or data set? I dunno :(

src/pages/dataTableUtilities/getPageInitialiser.js
```diff
@@ -14,6 +14,27 @@ import { COLUMN_KEYS } from './constants';
...
+    isAscending: true,
+    keyExtractor: recordKeyExtractor,
+    searchTerm: '',
+    modalKey: '',
+    columns: getColumns('fridges'),
+    route: ROUTES.VACCINES_ADMIN,
+    dataSet: 'fridges',
+    filterDataKeys: ['description', 'code'],
+  };
+};
```
joshxg, line 36: wanted to use selectors and things to get data but just went with this as it's consistent/easier 🤷

src/database/DataTypes/Sensor.js
wlthomson: I think does the job for now. I think longer term, maybe suitable to add as part of parsing and formatting utilities refactoring?

src/pages/VaccinePage.js
```diff
@@ -55,13 +57,21 @@ export const VaccinePageComponent = ({
     []
   );
 
-  return fridges.length ? <FlatList data={fridges} renderItem={Fridge} /> : <BlankComponent />;
+  return (
+    <DataTablePageView>
+      <FlexView flex={0} alignItems="flex-end" style={{ margin: 10 }}>
+        <PageButton text={generalStrings.admin} style={{}} onPress={onViewAdminPage} />
```
wlthomson, line 63: I think `style` redundant here?

```suggestion
        <PageButton text={generalStrings.admin} onPress={onViewAdminPage} />
```

src/pages/dataTableUtilities/getColumns.js
wlthomson: Hmm, maybe slightly breaking the pattern, but could get away with `VACCINES`, e.g. `VACCINES.FRIDGES`?

src/pages/dataTableUtilities/actions/cellActions.js
wlthomson: Or could pass an object with key constants? That way could update multiple values (if ever needed) and lookup the type in the reducer using the key. Also keeps the number of parameters at the same length, e.g `(newValues, rowKey, route)`.

joshxg: yeah a lot you can do - just not sure any of it is worth making the code more confusing/convoluted by making it more generic - does it really bring any benefit, I don't know if it does!

src/pages/dataTableUtilities/getColumns.js
joshxg: I think something more generic is a bit nicer? `DATA_SET.FRIDGES` or something? Seems to be a pattern we're using more and more

PR #1943, opened 2020-01-16
#1781 replace confirm modal with new confirm modal

Rebecca367:
Fixes #1781

**Change summary**
replace confirmModal with NewConfirmModal in related files and delete confirmModal.js accordingly.

**Testing**
recommend to test it on the latest structure and data file
[mSupplyV4-08-00-RC18.zip](https://github.com/openmsupply/mobile/files/4088908/mSupplyV4-08-00-RC18.zip)

[ ] log-out and then log in to see if the app/mobile still running properly

[ ] click on search bar on customer invoice to see if customerInvoicepage working 
Screen Shot 2020-01-21 at 3 15 53 PM

[ ] click on search bar on customer requisition page to see if the page is working 
Screen Shot 2020-01-21 at 3 18 01 PM

[ ] click on search bar on supplier requisition page to see if the page is working 
Screen Shot 2020-01-21 at 3 18 20 PM

[ ] click on search bar on supplier invoice to see if supplierInvoicepage working 
Screen Shot 2020-01-21 at 3 17 03 PM

[ ] click on search bar on stocktake to see if stocktake page working 
Screen Shot 2020-01-21 at 3 38 05 PM

**Related areas to think about**
If there are files that haven't been replaced with newConfirmModal.

PR #2729, opened 2020-04-20
#2717 fix confirm modal cancellation

### LINKED ISSUE #2717: Cannot cancel when confirming an insurance policy for a non-local patient
## Describe the bug

When confirming a prescription for a non-local patient, the user cannot cancel and return to editing.

### To reproduce

1. Lookup a non-local patient using the lookup modal.
2. Dispense to the patient.
3. Create an insurance policy at the payment step.
4. Save the insurance policy.
5. See error.

### Expected behaviour

The modal should include a cancel button (or pressable`X` in the top right).

### Proposed Solution

Add a cancel button to the popup confirm modal.

### Version and device info

- App version: `feature/patient-api` branch.
- Tablet model: Emulator (Nexus 10).
- OS version: API29.

### Additional context

Found during dispensary demo (17/04/20).

wlthomson:
Fixes #2717,

## Change summary

Fixes the insurance confirmation modal popup not cancelling.

## Testing

- [ ] When saving an insurance policy for a non-local patient, a confirmation modal is displayed.
- [ ] When the confirmation modal is cancelled, the modal is hidden and the insurance policy details are NOT reset.
- [ ] When the confirmation modal is confirmed, the modal and form are closed and the insurance policy saved and selected.

N/A.

### DIFF
--- a/src/widgets/FormControl.js
+++ b/src/widgets/FormControl.js
@@ -24,6 +24,8 @@ import {
   selectIsConfirmFormOpen,
 } from '../selectors/form';
 import { FormActions } from '../actions/FormActions';
+
+import { ModalContainer } from './modals';
 import { ConfirmForm } from './modalChildren';
 
 /**
@@ -223,30 +225,16 @@ const FormControlComponent = ({
   );
 
   return (
-    <>
-      <View
-        style={
-          isConfirmFormOpen
-            ? { ...localStyles.flexOne, ...localStyles.hidden }
-            : localStyles.flexOne
-        }
-      >
-        <ScrollView style={localStyles.whiteBackground}>
-          <View style={localStyles.flexRow}>
-            <View style={localStyles.flexOne} />
-            <View style={localStyles.flexTen}>{formInputs()}</View>
-            <View style={localStyles.flexOne} />
-          </View>
-        </ScrollView>
-        <Buttons />
-      </View>
-      <View
-        style={
-          isConfirmFormOpen
-            ? localStyles.flexOne
-            : { ...localStyles.flexOne, ...localStyles.hidden }
-        }
-      >
+    <View style={localStyles.flexOne}>
+      <ScrollView style={localStyles.whiteBackground}>
+        <View style={localStyles.flexRow}>
+          <View style={localStyles.flexOne} />
+          <View style={localStyles.flexTen}>{formInputs()}</View>
+          <View style={localStyles.flexOne} />
+        </View>
+      </ScrollView>
+      <Buttons />
+      <ModalContainer fullScreen isVisible={isConfirmFormOpen} noCancel>
         <ConfirmForm
           isOpen={isConfirmFormOpen}
           questionText={confirmText}
@@ -254,8 +242,8 @@ const FormControlComponent = ({
           onCancel={onCancelForm}
           confirmText={modalStrings.confirm}
         />
-      </View>
-    </>
+      </ModalContainer>
+    </View>
   );
 };
 
@@ -373,5 +361,4 @@ const localStyles = StyleSheet.create({
   flexRow: { flex: 1, flexDirection: 'row' },
   buttonsRow: { marginTop: 10, flexDirection: 'row-reverse' },
   whiteBackground: { backgroundColor: WHITE },
-  hidden: { display: 'none' },
 });
--- a/src/widgets/FormControl.js
+++ b/src/widgets/FormControl.js
@@ -56,25 +56,23 @@ const FormControlComponent = ({
   canSaveForm,
   saveButtonText,
   // Confirm form state
-  confirmOnSave,
   isConfirmFormOpen,
   confirmText,
   // Cancel button state
   showCancelButton,
   cancelButtonText,
   // Form callbacks
-  initialiseForm,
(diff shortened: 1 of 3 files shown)


src/widgets/FormControl.js
```diff
@@ -229,54 +222,102 @@ const FormControlComponent = ({
...
+      >
+        <ConfirmForm
+          isOpen={isConfirmFormOpen}
+          questionText={confirmText}
+          onConfirm={onSaveForm}
+          onCancel={onCancelForm}
+          confirmText={modalStrings.confirm}
+        />
+      </View>
+    </>
```
wlthomson, line 258: A bit of an antipattern here, I think. In order to prevent the form cells from being unmounted (and the refs being nulled, and form state getting all messed up), the form style is set to `display: 'none'`. It's not great at all, but I couldn't get the `ConfirmForm` to hide otherwise (even when `isOpen={false}`, it would still render half the screen).

As it's working, it's probably better to leave as is and open an issue (maybe for `5.1.0`?) if it's worth it!

src/widgets/FormControl.js
```diff
@@ -229,54 +222,102 @@ const FormControlComponent = ({
...
   );
+};
 
-  return isConfirmFormOpen ? InputConfirm : InputDetails;
+const mergeProps = (stateProps, dispatchProps, ownProps) => {
```
wlthomson, line 262: giphy

src/widgets/FormControl.js
joshxg: `ConfirmForm` needs to be wrapped in a `ModalContainer` - looks like you were just rendering in within the last modal container which because of flex would take up space

Can make it work doing:

```
    <>
      <ScrollView style={localStyles.whiteBackground}>
        <View style={localStyles.flexRow}>
          <View style={localStyles.flexOne} />
          <View style={localStyles.flexTen}>{formInputs()}</View>
          <View style={localStyles.flexOne} />
        </View>
      </ScrollView>

      <Buttons />

      <ModalContainer fullScreen isVisible={isConfirmFormOpen} noCancel>
        <ConfirmForm
          questionText={confirmText}
          onConfirm={onSaveForm}
          onCancel={onCancelForm}
          confirmText={modalStrings.confirm}
        />
      </ModalContainer>
    </>
```

just a couple lines, probs worth it

src/widgets/FormControl.js
joshxg: I think it just means code for the component is harder to read, looking in quite a few different places. Think it would be good (not now) idea to use only `mergeProps` or `mapDispatchToProps` - the dispatch logic is just too scattered and hard to read

mark-prins: not sure that I follow you @joshxg - I thought the `mergeProps` was ok here. makes the wiring section at the bottom of the component much more involved, sure, but doesn't it mean that the declarative part of the component is slightly neater?

joshxg: ya the declarative part is way better and I do like the use of `mergeProps`, not saying to change - just it was a little harder to read dispatch code when it was over two different functions. If you didn't use a `mapDispatchToProps`, `dispatchProps` would be just the `dispatch` function in `mergeProps`, so could just have all the code there - a bit easier to scan maybe (but at that point maybe you should just return state from `mapStateToProps` and have all of it in `mergeProps` 🤷!

src/widgets/FormControl.js
wlthomson: Ah yup, makes sense. Maybe `ConfirmForm` should be called `ConfirmModal` or something similar, just to make that a bit clearer?

joshxg: https://github.com/openmsupply/mobile/pull/1943#discussion_r379949015

🤷

src/widgets/FormControl.js
wlthomson: @joshxg Yup, that was my thought, too. But my thinking was that it's actually tidier to have all the selectors in the `mapState`, all the dispatches in `mapDispatch` and then all the conditional logic in `mergeProps`. With components with more complex prop logic, I think it's actually tidier to split it up like that and isn't really that much of a difficult pattern to follow if it's applied consistently?

src/widgets/FormControl.js
wlthomson: :smiley: maybe something for a new issue as it looks from that checklist like there is some migration remaining to do also?

joshxg: yeah heh

src/widgets/FormControl.js
joshxg: I'm happy either way, it's really just styling which really we're all happy with! Last point but leaving it here 🤣 

I believe `mergeProps` return isn't cached, where as `mapDispatchToProps` is, so the new callbacks will always have new references. So you come back to using (when needed!) to `useCallback`, anyway

mark-prins: got it, thanks for clarifying.
I rather like the split of having `mapDispatch` handle callbacks and `mergeProps` handle the conditional logic. Though if it isn't cached there is another problem and more code. as long as this pattern is an edge case for us, only when necessary..

PR #2733, opened 2020-04-21
#2728 admin page features

### LINKED ISSUE #2728: Further vaccine admin features

`VaccineAdminPage` was added in #2727 - it's missing a scan for sensors and add new location buttons

- A button which will scan for new sensors and add them to the table
- A button which will open a form to create a new location

## Implementation

N/A

N/A

## Additional context

N/A

joshxg:
Fixes #2728 

## Change summary

- Adds sensor editing logic
- Adds code to scan for sensors
- Adds the ability to add a new fridge
- Adds permission checking on the admin navigation button
- Adds ability to delete sensors/locations

## Testing

N/A

N/A

### DIFF
--- a/src/actions/TemperatureSyncActions.js
+++ b/src/actions/TemperatureSyncActions.js
@@ -4,6 +4,7 @@
  * Sustainable Solutions (NZ) Ltd. 2020
  */
 
+import { ToastAndroid } from 'react-native';
 import { generateUUID } from 'react-native-database';
 
 import { UIDatabase } from '../database';
@@ -11,6 +12,8 @@ import { Sensor } from '../database/DataTypes';
 import { createRecord } from '../database/utilities';
 import { MILLISECONDS } from '../utilities';
 import { chunk } from '../utilities/chunk';
+
+import { vaccineStrings } from '../localization';
 
 export const TEMPERATURE_SYNC_ACTIONS = {
   OPEN_MODAL: 'TemperatureSync/openModal',
@@ -77,7 +80,15 @@ const scanForSensors = () => async dispatch => {
 
   dispatch(updateSensors(data));
 
-  if (success) dispatch(scanComplete());
+  if (success) {
+    ToastAndroid.show(
+      vaccineStrings.formatString(vaccineStrings.found_sensors, data.length),
+      ToastAndroid.LONG
+    );
+    dispatch(scanComplete());
+  } else {
+    ToastAndroid.show(vaccineStrings.could_not_find_sensors, ToastAndroid.LONG);
+  }
 };
 
 const downloadLogsError = () => ({ type: TEMPERATURE_SYNC_ACTIONS.DOWNLOAD_LOGS_ERROR });
--- a/src/database/DataTypes/Location.js
+++ b/src/database/DataTypes/Location.js
@@ -61,8 +61,8 @@ Location.schema = {
   primaryKey: 'id',
   properties: {
     id: 'string',
-    description: { type: 'string', optional: true },
-    code: { type: 'string', optional: true },
+    description: { type: 'string', default: '' },
+    code: { type: 'string', default: '' },
     locationType: { type: 'LocationType', optional: true },
     locationMovements: {
       type: 'linkingObjects',
--- a/src/database/DataTypes/Sensor.js
+++ b/src/database/DataTypes/Sensor.js
@@ -46,6 +46,10 @@ export class Sensor extends Realm.Object {
   get batteryLevelString() {
     return `${this.batteryLevel}%`;
   }
+
+  get currentLocationName() {
+    return this.location?.description ?? '';
+  }
 }
 
 Sensor.schema = {
--- a/src/database/utilities/createRecord.js
+++ b/src/database/utilities/createRecord.js
@@ -750,7 +750,7 @@ const createTransactionItem = (database, transaction, item, initialQuantity = 0)
   return transactionItem;
 };
 
-const createLocation = (database, description, code, locationType) => {
+const createLocation = (database, locationType, description, code) => {
   const location = database.create('Location', {
     id: generateUUID(),
     description,
--- a/src/pages/VaccineAdminPage.js
+++ b/src/pages/VaccineAdminPage.js
@@ -1,4 +1,3 @@
-/* eslint-disable no-unused-vars */
 /* eslint-disable react/forbid-prop-types */
 /**
  * mSupply Mobile
@@ -6,18 +5,25 @@
  */
 
 import React from 'react';
-import { View } from 'react-native';
+import { View, StyleSheet, ActivityIndicator } from 'react-native';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 
 import { ToggleBar, DataTablePageView, SearchBar, PageButton } from '../widgets';
 import { DataTable, DataTableRow, DataTableHeaderRow } from '../widgets/DataTable';
+import { DataTablePageModal } from '../widgets/modals';
+import { BottomConfirmModal } from '../widgets/bottomModals';
 
 import { recordKeyExtractor, getItemLayout, getPageDispatchers } from './dataTableUtilities';
 
+import { TemperatureSyncActions } from '../actions/TemperatureSyncActions';
+import { selectIsSyncingTemperatures } from '../selectors/temperatureSync';
+
 import { ROUTES } from '../navigation';
-import { generalStrings, vaccineStrings } from '../localization';
-import globalStyles from '../globalStyles';
+import { MODAL_KEYS } from '../utilities';
+
+import { modalStrings, buttonStrings, generalStrings, vaccineStrings } from '../localization';
+import globalStyles, { SUSSOL_ORANGE } from '../globalStyles';
 
 const VaccineAdminPageComponent = ({
   data,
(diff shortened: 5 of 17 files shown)


src/actions/TemperatureSyncActions.js
```diff
@@ -77,7 +80,15 @@ const scanForSensors = () => async dispatch => {
...
 
-  if (success) dispatch(scanComplete());
+  if (success) {
+    ToastAndroid.show(
+      vaccineStrings.formatString(vaccineStrings.found_sensors, data.length),
+      ToastAndroid.LONG
+    );
+    dispatch(scanComplete());
+  } else {
+    ToastAndroid.show(vaccineStrings.could_not_find_sensors, ToastAndroid.LONG);
```
joshxg, line 90: I don't like the toast here. I really want to create a `<Toast/>` component which sits at the root which can be decoratively show a toast - but not 100% sure how I should do it and didn't want to do it here.

src/database/DataTypes/Location.js
```diff
@@ -61,8 +61,8 @@ Location.schema = {
   primaryKey: 'id',
   properties: {
     id: 'string',
-    description: { type: 'string', optional: true },
-    code: { type: 'string', optional: true },
+    description: { type: 'string', default: '' },
+    code: { type: 'string', default: '' },
```
joshxg, line 65: more simpler ui changes

src/database/DataTypes/Sensor.js
```diff
@@ -46,6 +46,10 @@ export class Sensor extends Realm.Object {
   get batteryLevelString() {
     return `${this.batteryLevel}%`;
   }
+
+  get currentLocationName() {
+    return this.location?.description ?? '';
+  }
```
joshxg, line 52: consistent with other location editing getters

src/database/utilities/createRecord.js
```diff
@@ -750,7 +750,7 @@ const createTransactionItem = (database, transaction, item, initialQuantity = 0)
   return transactionItem;
 };
 
-const createLocation = (database, description, code, locationType) => {
+const createLocation = (database, locationType, description, code) => {
```
joshxg, line 753: just changing to be easier to call

src/pages/VaccineAdminPage.js
```diff
@@ -94,29 +130,84 @@ const VaccineAdminPageComponent = ({
...
+          placeholder={placeholderString}
+        />
+        {isScanning ? (
+          <ActivityIndicator
+            style={localStyles.indicatorContainer}
+            color={SUSSOL_ORANGE}
+            size="small"
+          />
+        ) : (
+          <PageButton text={buttonText} onPress={onPress} />
```
joshxg, line 146: Just a bit of a lazy way to indicate loading, replacing the button with an activity indicator: https://reactnative.dev/docs/activityindicator. Would be cool to also make a `ButtonWithSpinnerThing` which did this out the box - combined with the `Toast` component mentioned above

src/pages/dataTableUtilities/actions/tableActions.js
```diff
@@ -161,6 +161,19 @@ export const addItem = (item, addedItemType, route) => (dispatch, getState) => {
   }
 };
 
+export const addFridge = route => dispatch => {
+  UIDatabase.write(() => {
+    const fridgeLocationType = UIDatabase.objects('LocationType').filtered(
+      'description == $0',
+      'fridge'
+    )[0];
```
joshxg, line 169: this is not great, but I am thinking to add a `LocationType` to desktop update/upgrade/startup code which ensures there is always a `LocationType` with a type.. I was thinking of using something more like `@sussol/fridge` - similar to `inventory_adjustment` name record

src/pages/dataTableUtilities/getColumns.js
```diff
@@ -41,20 +41,22 @@ const PAGE_COLUMN_WIDTHS = {
...
   sensors: [
-    COLUMN_NAMES.LOCATION,
     COLUMN_NAMES.EDITABLE_NAME,
+    COLUMN_NAMES.LOCATION,
     COLUMN_NAMES.BATTERY_LEVEL,
     COLUMN_NAMES.MAC_ADDRESS,
+    COLUMN_NAMES.REMOVE,
   ],
-  fridges: [COLUMN_NAMES.CODE, COLUMN_NAMES.DESCRIPTION],
+  fridges: [COLUMN_NAMES.EDITABLE_CODE, COLUMN_NAMES.DESCRIPTION, COLUMN_NAMES.REMOVE],
```
joshxg, line 58: adding pretty basic remove logic to sensors and fridges for now. I think some more detailed logic might be needed, given this could leave orphan records. Since it's store data, I think dissallowing deletion when there are no related records would be all good

src/actions/TemperatureSyncActions.js
joshxg: but this action was created previously - just uses `Sensor.scanForSensors` which initializes the bluetooth module and scans for any advertisement frequencies from sensors with the right manufacturer ID and characteristics and returns an object for each found sensor with all the info from its advertisement - battery level, mac address etc. Uses another action `updateSensors` with that data to actually update sensors. Toasts when finding sensors with how many and if none were found, says that. The `startScan` and `completeScan` set the state to scanning: true/false, which will also stop the `TemperatureSyncModal` from being enabled and trying to start another scan - also updates the modal state so if you open the modal it will say "scanning for sensors"

src/pages/VaccineAdminPage.js
```diff
@@ -34,13 +40,32 @@ const VaccineAdminPageComponent = ({
...
+  onCloseModal,
+  modalValue,
+  dispatch,
+  onAddFridge,
+  hasSelection,
+  onDeselectAll,
+  onDeleteRecords,
+  onCheck,
+  scanForSensors,
+  isScanning,
```
joshxg, line 55: whole bunch of callbacks, mainly reused `pages` actions with a couple more for sensor/location specific

src/pages/VaccineAdminPage.js
```diff
@@ -94,29 +130,84 @@ const VaccineAdminPageComponent = ({
...
+        dispatch={dispatch}
+        currentValue={modalValue}
+      />
+
+      <BottomConfirmModal
+        isOpen={hasSelection}
+        questionText={modalStrings.remove_these_items}
+        onCancel={onDeselectAll}
+        onConfirm={onDeleteRecords}
+        confirmText={modalStrings.remove}
```
joshxg, line 175: more copy paste code from other pages!

src/pages/VaccinePage.js
```diff
@@ -72,10 +76,12 @@ const mapStateToProps = state => {
   const { fridge } = state;
   const { fridges, selectedFridge } = fridge;
   const { breaches } = selectBreaches(state);
+  const isAdmin = selectCurrentUserIsAdmin(state);
```
joshxg, line 79: just using the admin pref to conditionally load the admin button

src/actions/TemperatureSyncActions.js
mark-prins: I like the idea of dispatching a `showToast` action and showing the toast from a reducer just because that's a similar process to other UI display tasks. Fair enough to leave the 🍞 changes to a separate issue though, and this is fine like it is.

src/pages/VaccineAdminPage.js
mark-prins: Was thinking the same thing 😃

src/pages/dataTableUtilities/actions/cellActions.js
```diff
@@ -36,6 +36,28 @@ export const refreshIndicatorRow = route => ({
   payload: { route },
 });
 
+export const editSensorLocation = (location, route) => (dispatch, getState) => {
```
mark-prins, line 39: this file is getting pretty big! may need to split the cellActions into different files at some point.

src/pages/dataTableUtilities/actions/tableActions.js
mark-prins: yes, was wondering the same thing, what happens when there isn't a fridge `LocationType`?
wouldn't there be a `Location` with an undefined `LocationType`? 
I think better to check for a `LocationType` - Toast if none exist and add if some do.

src/pages/dataTableUtilities/actions/cellActions.js
joshxg: definitely .. ! I don't really like 'cell' actions at all to be honest.. I would prefer not having the 'pages' reducer which is really confusing and think refactoring that into more specific `'requisitionReducer'`, `stocktakeReducer` would probably be a lot nicer!

src/pages/dataTableUtilities/actions/tableActions.js
joshxg: I am thinking actually that using a `Form` to create `Locations` might be better. That way can just have a drop down for selecting a location type. But actually, maybe just adding a location type cell and instead calling this 'dataset' 'locations'? And just leave it up to the user to make sure they have selected the right location type..

mark-prins: yeah, sounds interesting. just keep it flexible - the user picks the location type, and in this case chooses a 'fridge' (hopefully). +1 from me.

PR #2772, opened 2020-04-24
#2756 update post sync to ensure all incoming customer credits are finalised

### LINKED ISSUE #2756: Customer credits allow for setting negative quantities
## Describe the bug

This is a big oversight on my side! Customer credits are automatically confirmed, and therefore will automatically adjust real stock values instantly. This means you're able to reduce the amount of stock coming in, which may have been used elsewhere.

### To reproduce

*Create a supplier credit to your mobile store, which will create a customer credit on the mobile side*
1. Go to the customer credit and adjust the values so that you have 100 total stock for that item (or 1000 if you have more than 100 etc..)
2. Click on back, and navigate to a customer invoice (unfinalized).
3. Add the same item that you edited in the customer credit, editing the quantity to be the maximum allowed (i.e. 100 or 1000).
4. Now, go back to the customer credit and reduce the amount to 10.
5. See error

### Expected behaviour

The app shouldn't crash and you shouldn't be able to set negative values.

### Proposed Solution

This was unexpected as I didn't know that customer credits were automatically confirmed! A bit annoying - similar problem as we have with stocktakes and customer invoices. Example: Snapshot is the quantity at the time of finalizing, not from the time of creation. So, when you make a stocktake, then go edit quantities and go back to the stocktake, the snapshot might be wrong - so we auto adjust on navigating. That's less ideal to do for a CI, I think. But possible.

Desktop solution is to:

image
When trying to edit the customer credit line

Then shows this:
image

I am thinking for the easiest solution, some method of restricting the editable quantity of the customer credit so you can't reduce it low enough, and it just adjusts to the lowest amount when editing - similar to how a CI editing works - if you only have 100, and you try issue 101, it goes to 0 (not the greatest UX, but easy solution for now?). Something along the lines of having the line be disabled when there is an unfinalized CI line with that particular item. It's not great because it offers 0 feedback for the user,  but it's a start of a solution, at least 🤣 Possibly allow for editing but default back if `hasStockIssued` is true, or something, and use a toast to explain why..?

### Version and device info

- App version: 5.0.0-RC2
- Tablet model: N/A
- OS version: N/A

### Additional context

@wlthomson thoughts?

### LINKED ISSUE #2758: Crash on customer requisition use required/suggested quantity
## Describe the bug

The app crashes when working on a `"Customer Requisition"` and pressing `"Use Required Quantity"` and  `"Use Suggested Quantity"` . In two different attempts, these following messages showed up (2 different errors, check below):

Customer requisition before clicking on `"Use Required Quantity"`:
Screen Shot 2020-04-23 at 4 08 49 PM

  - 1st attempt:
Screen Shot 2020-04-23 at 4 09 06 PM

  - 2nd attempt:
Trying again, I didn't get that error message, but the app freezes and when trying to reload, this is the error:

Screen Shot 2020-04-23 at 4 14 19 PM

- Pressing the `"Use Suggested Quantity"` button. This is the error message retrieved after trying to reload the page:

Screen Shot 2020-04-23 at 4 20 00 PM

### To reproduce

1. Go to 'Customer Requisitions' -> Select a 'Customer Requisition'
2. Click on the button: `"Use Required Quantity"`.

Repeat for `"Use Suggested Quantity"`.

### Expected behaviour

The app shouldn't crash and update the Supplied quantity (if necessary).

### Proposed Solution

N/A.

### Version and device info

- App version: v5.0.0-rc1
- Tablet model: AVD
- OS version: Android 9.0 API 28

### Additional context

This happened while testing #1244.

wlthomson:
Fixes #2756.

## Change summary

Auto-finalises customer credit on incoming sync.

## Testing

### Setup 

- Create a supplier credit for mobile store.
- Manually sync mobile.
- Navigate to customer invoices and toggle to `"Past"`.

### Test cases

- [ ] The customer credit which corresponds to the supplier credit is displayed
- [ ] The customer credit is finalised and cannot be edited

Not too sure about existing finalisation logic, so played it safe and just replicated the existing logic for manually finalising a customer credit.

### DIFF
--- a/src/sync/PostSyncProcessor.js
+++ b/src/sync/PostSyncProcessor.js
@@ -205,6 +205,12 @@ export class PostSyncProcessor {
       }
     }
 
+    if (record.type === 'customer_credit' && !record.isFinalised) {
+      funcs.push(() => {
+        record.finalise();
+      });
+    }
+
     // If any changes, add database update for record.
     if (funcs.length > 0) {
       funcs.push(() => {
--- a/src/sync/PostSyncProcessor.js
+++ b/src/sync/PostSyncProcessor.js
@@ -205,6 +205,7 @@ export class PostSyncProcessor {
       }
     }
 
+    // Auto-finalise all incoming consumer credits.
     if (record.type === 'customer_credit' && !record.isFinalised) {
       funcs.push(() => {
         record.finalise();


src/sync/PostSyncProcessor.js
```diff
@@ -205,6 +205,12 @@ export class PostSyncProcessor {
       }
     }
 
+    if (record.type === 'customer_credit' && !record.isFinalised) {
+      funcs.push(() => {
+        record.finalise();
+      });
```
joshxg, line 211: Just a comment I reckon, it's a pretty odd thing ya know

wlthomson: Yup, good point. 252aa6c.

PR #2781, opened 2020-04-25
#2780 reset finalise modal item on go back navigation

### LINKED ISSUE #2780: Crash on deleting new records
## Describe the bug

mSupply crash (close) when I delete new un-finalised customer invoice with lines or no lines. It is working fine when deleting existing customer invoices. Seems also to work fine when creating new customer invoice, get out of mSupply and reopen it again and then delete the invoice

### To reproduce

1. Go to Customer invoice
2. Click on New customer invoice and add lines or not. 
3. From the customer invoices list, delete the customer invoice

### Expected behaviour

mSupply should not crash

### Proposed Solution

### Version and device info

- App version: v5.0.0RC1
- Tablet model: Galaxy tab A (SM-T510)
- OS version: Android

### Additional context

You might have catch that with bugsnag :)

wlthomson:
Fixes #2780.

## Change summary

- Adds `resetFinaliseItem` to `FinaliseActions`.
- Dispatches `resetFinaliseItem` on `goBack` from a finalisable page. 

## Testing

- [ ] Create a new customer invoice and navigate back to customer invoices page. No error is thrown on deleting the new invoice.

Not sure if piggybacking onto the `goBack` action is the right way to do this, but seems consistent with other navigation actions. Might need to add some checks to ensure the item is only reset when navigating to `CustomerInvoicesPage`, etc?

### DIFF
--- a/src/navigation/actions.js
+++ b/src/navigation/actions.js
@@ -34,6 +34,11 @@ import { PREFERENCE_KEYS } from '../database/utilities/constants';
  *
  */
 
+/**
+ * Action creator for handling back navigation.
+ *
+ * Triggers back navigation and cleans up current state.
+ */
 export const goBack = () => dispatch => {
   if (!RootNavigator.canGoBack()) BackHandler.exitApp();
   else {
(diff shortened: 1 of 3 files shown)


src/navigation/actions.js
```diff
@@ -34,21 +34,37 @@ import { PREFERENCE_KEYS } from '../database/utilities/constants';
...
           ...NavigationActions.back(),
-          payload: {
-            prevRouteName: RootNavigator.getPrevRouteName(),
-          },
+          payload: { prevRouteName },
         });
+
+      const cleanUp = () => {
         dispatch(PrescriptionActions.deletePrescription());
+        if (FINALISABLE_PAGES.includes(currRouteName)) {
```
joshxg, line 60: I think you could just do this unconditionally, but no damage in a check right

wlthomson: Yeah, think you're right, but no added benefit I can see... except for maybe reducing the branching complexity? But on the other hand dispatching unnecessary actions might make the code a bit less clear? On balance I think no harm in a little conditional :rofl:.

wlthomson:
> Nice. Yeah the navigation logic isn't ideal. What do you think: ideally you would not duplicate the state and not have a `finaliseItem` - and just not have a `FinaliseReducer` would be nice but then dynamically handling which finalise algo is run is a bit ugly

Hmm, prob missing something, but I think we have a bit of a template method thing going on anyway (with the reducer being the skeleton and the algo just being the `finalise` method which is implemented by the `finaliseItem` object)? So I think we might have that handled already?

PR #2782, opened 2020-04-25
#2755 fix lookup form buggy behaviour on no matching records

### LINKED ISSUE #2755: Dispensary lookup results not resetting on empty search
## Describe the bug

Trying to use a filter to search for Prescribers one first time works.
The problem is making a search with a filter after a search without filters is done (which results in a query with all prescribers.

lookupFilterBug

The same problem might be happening on the Lookup for patients too. 

### To reproduce

1. Go to Dispensary
2. Click on the Prescribers tab
3. Click on Lookup prescriber
4. Leave all fields blank and "Search"
5. The result should show all Prescribers
6. Enter a name on "First name" and "Search"
7. The result is not filtering only prescribers with the same "First name"

### Expected behaviour

The filters should be included in the result when the "Search" is clicked refreshing the result list (after a search for all prescribers was done without filters). 

### Proposed Solution

### Version and device info

- App version: 5.0.0-RC1
- Tablet model: Emulator of Nexus 10
- OS version: Android 4.4.4

### Additional context

N/A

wlthomson:
Fixes #2755.

## Change summary

- Fixes error code typo preventing error modal on no matching records.
- Fixes data not reset on error.

## Testing

- [ ] Searching for a non-existent patient displays the error modal (`"No matching records found"`)
- [ ] Searching for a non-existent prescriber displays the error modal (`"No matching records found"`)
- [ ] After searching for a non-existent patient, no records are displayed from previous queries.
- [ ] After searching for a non-existent prescriber, no records are displayed from previous queries.

Just a quick fix, reminded me of a lot of stuff I don't like about how I've done the current form/hook... but forcing myself to not do any refactoring until after release :cry:.

### DIFF
--- a/src/widgets/modals/SearchForm.js
+++ b/src/widgets/modals/SearchForm.js
@@ -102,9 +102,11 @@ export const SearchFormComponent = ({
 
   const onError = useCallback(responseError => {
     setError(responseError);
+    resetData();
     resetQueryUrl();
   }, []);
 
+  const resetData = useCallback(() => setData(''), []);
   const resetError = useCallback(() => setError(''), []);
   const resetQueryUrl = useCallback(() => setQueryUrl(''), []);
 
--- a/src/hooks/useFetch.js
+++ b/src/hooks/useFetch.js
@@ -52,7 +52,7 @@ export const useFetch = (url, options = {}, timeout = CONNECTION_TIMEOUT_PERIOD)
             const responseData = await response.json();
             const { error: responseError } = responseData;
             if (responseError) throw new Error(responseError);
-            if (!responseData.length) throw new Error(ERROR_CODES.RESPONSE_NO_RECORDS);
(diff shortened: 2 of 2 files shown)


src/widgets/modals/SearchForm.js
```diff
@@ -102,9 +102,11 @@ export const SearchFormComponent = ({
 
   const onError = useCallback(responseError => {
     setError(responseError);
+    resetData();
     resetQueryUrl();
   }, []);
 
+  const resetData = useCallback(() => setData(''), []);
   const resetError = useCallback(() => setError(''), []);
   const resetQueryUrl = useCallback(() => setQueryUrl(''), []);
```
wlthomson, line 111: Revisiting this, seems a bit OTT :laughing:. Decided to maintain consistency for now, but I think these methods are pretty redundant (as @joshxg and/or @mark-prins probably already have said)!

PR #2787, opened 2020-04-27
#2768 update stocktake manage page confirm modal conditionals

### LINKED ISSUE #2768: Stocktake manage page confirm modal does not open/close on selection/deselection
## Describe the bug

Issues found while testing #1241. Section: COLUMNS -> SELECTED.
When opening the Stocktake Manage Page, the confirmation modal opens up right immediately (not waiting for me to select and item), and then when unselecting an item, it doesn't disappears. (as expected). 
**NOTE**: Same is happening with program stocktakes.
See example below:

confirmModalStocktake

### To reproduce

1. Go to 'StocktakeManagePage' and create a new stocktake or open a current one.
4. See confirmation modal opens immediately (error).

### Expected behaviour

The confirmation modal should open when selecting an item and close when unselecting all.

### Proposed Solution

N/A

### Version and device info

- App version: v5.0.0-rc1
- Tablet model: AVD
- OS version: Android 9.0 - API 28

### Additional context

While testing #1241

wlthomson:
Fixes #2768.

May be the case that this fix is not needed, but made a fix while investigating the issue so may as well put it here :rofl:!

THIS IS ONLY FOR REFERENCE, DO NOT REVIEW (will remove this if issue is confirmed).

## Change summary

Stocktake bottom modal is hidden if no items selected.

## Testing

- [ ] If no items selected, confirm modal is closed.
- [ ] If at least one item selected, confirm modal is open.

N/A.

### DIFF
--- a/src/pages/StocktakeManagePage.js
+++ b/src/pages/StocktakeManagePage.js
@@ -46,10 +46,13 @@ export const StocktakeManage = ({
   route,
 }) => {
   const runWithLoadingIndicator = useLoadingIndicator();
+
   // On navigating to this screen, if a stocktake is passed through, update the selection with
   // the items already in the stocktake.
   useEffect(() => {
-    if (pageObject) dispatch(PageActions.selectItems(pageObject.itemsInStocktake, route));
+    const { itemsInStocktake = [] } = pageObject ?? {};
+    const stocktakeHasItems = !!itemsInStocktake.length;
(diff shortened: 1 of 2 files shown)


wlthomson:
Have a feeling this isn't going to be resolved quickly, so going to close.

Just to be safe, don't delete the branch :smile:.

PR #2792, opened 2020-04-28
#2735 fix prescription undefined title

### LINKED ISSUE #2735: Prescription form shows Label as "Invoice undefined" in header title
## Describe the bug

Prescription form shows Label as "Invoice undefined" in header title

### To reproduce

1. Click on Dispensary
2. From patient list, click on dispense option
3. See error

Screen Shot 2020-04-21 at 11 11 59 AM

### Expected behaviour

Header title should not labelled as "Invoice undefined"

### Proposed Solution

### Version and device info

- App version: 5.0.0 RC1
- Tablet model: Nexus, API 25
- OS version:

### Additional context

arises while testing #2720

wlthomson:
Fixes #2735.

## Change summary

Prescription page shape is shallower than, e.g. customer invoice, so `pageObject` was not being passed to `getTitle`

## Testing

- [ ] Prescription title is `"Prescription X"`, where `X` is the serial number of the invoice.

May be that better option is to update the prescription page shape to be consistent with the other pages? Would require adding to initialisers, which I felt might be a bit OTT. Open to suggestions!

### DIFF
--- a/src/reducers/PagesReducer.js
+++ b/src/reducers/PagesReducer.js
@@ -28,7 +28,7 @@ export const PagesReducer = (state = {}, action) => {
       const { pageObject } = params;
 
       const pageInitialiser = getPageInitialiser(routeName);
-      const pageInitialState = pageInitialiser ? pageInitialiser(pageObject) : pageObject;
+      const pageInitialState = pageInitialiser(pageObject);
 
       return { ...state, [routeName]: pageInitialState, currentRoute: routeName };
     }
--- a/src/reducers/PagesReducer.js
+++ b/src/reducers/PagesReducer.js
@@ -8,6 +8,7 @@ import getPageInitialiser from '../pages/dataTableUtilities/getPageInitialiser';
 
 /**
  * Redux reducer controlling the `pages` field.
+ *
  * @param {Object} state  Redux state, `pages` field.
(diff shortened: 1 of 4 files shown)


wlthomson:
@Chris-Petty 

notify

src/reducers/PagesReducer.js
```diff
@@ -8,6 +8,7 @@ import getPageInitialiser from '../pages/dataTableUtilities/getPageInitialiser';
 
 /**
  * Redux reducer controlling the `pages` field.
+ *
```
mark-prins, line 11: 👍

wlthomson:
> it works fine.
> I think it would be more consistent to create a very simple initialiser in `getPageinitialiser` for prescriptions which just returned the pageObject, so that all pages had the same shape.
> Am I missing something though?
> 
> ```
> const prescriptionInitialiser = prescription => ({
>   pageObject: prescription,
> });
> ```
> 
> and
> 
> ```
>   dispensary: prescriptionInitialiser,
>   prescription: prescriptionInitialiser,
> ```

Yup, makes sense to me, that's what I was referring to in description related areas. I think the "page initialisation" stuff is related to the old page reducer which we are moving away from, but I agree that it's a more elegant solution.

Issue #2761, opened 2020-04-23
Lookup patient with DOB filter result in patients with 'Invalid dates'


nmadruga:
## Describe the bug

Trying to search for patients with data of birth in a database full of patients with 'Invalid date' is not showing the result properly. Either the results for the correct day are in the bottom and I can't reach it or not even filtered in the result.

mSupply datafile used: Ivory Coast mSupply Master. Sync store: CSRDuokro

### To reproduce

1. Go to Dispensary
2. Click on the Patients tab
3. Click on Lookup patients
4. Add one existing date of birth and click "Search"
5. The result is showing many patients with Invalid date

### Expected behaviour

The result should include the filtered ones

### Proposed Solution

Maybe not list users with Invalid date, or change the order so the ones matching the filter appear first, in case showing the ones with Invalid date is required.

### Version and device info

App version: 5.0.0-RC1
Tablet model: Emulator of Nexus 10
OS version: Android 5.1 (API 22)

### Additional context

N/A

wlthomson:
Fixed by #2818.

PR #2818, opened 2020-05-06
#2817 fix missing lookup query string separator

### LINKED ISSUE #2817: Patient lookup not matching first and last names correctly
## Describe the bug

See title. 

### To reproduce

See https://github.com/openmsupply/mobile/issues/2713#issuecomment-624360599.

### Expected behaviour

Patients should be matched by first and last name, and a combination of both.

### Proposed Solution

First step is to determine if this is a mobile or desktop side problem. If desktop, will need to get a fix into the next `v10` patch.

### Version and device info

- App version: `v5.0.0-rc3`
- Tablet model: Android emulator.
- OS version: ~API28.

### Additional context

Raised by @diegoevangelisti.

wlthomson:
Fixes #2817.

## Change summary

Lookup queries for more than one field were returning no records due to typo in query string `"&"` constant.

## Testing

- [ ]  Lookup queries involving multiple fields work correctly (e.g. lookup patient with given first name and last name).

N/A.

### DIFF
--- a/src/sync/lookupApiUtils.js
+++ b/src/sync/lookupApiUtils.js
@@ -42,7 +42,8 @@ const getQueryString = params =>
     const [[key, value]] = Object.entries(param);
     if (!value) return queryString;
     const paramString = `${key}=${SEPARATORS.QUERY_WILDCARD}${value}${SEPARATORS.QUERY_WILDCARD}`;
-    const paramSeparator = queryString.length > 0 ? SEPARATORS.PARAMETERS : SEPARATORS.QUERY_STRING;
+    const paramSeparator =
+      queryString.length > 0 ? SEPARATORS.QUERY_PARAMETERS : SEPARATORS.QUERY_STRING;
     return queryString + paramSeparator + paramString;
   }, '');
 


diegoevangelisti:
Looking up for `First Name: Dan` and `Last Name: Kou` retrieves some records - working ok:

Screen Shot 2020-05-07 at 8 36 29 AM

But then nothing is retrieved when adding the DOB (29/10/19876) to the search:

lookupPatient

**NOTE**: Pay attention that when removing the DOB the `Save` button gets disable (I'll create a separate issue)

mark-prins:
@wlthomson @diegoevangelisti unfortunately, we have DOB format in the query in mSupply (at least, in my current copy) as `DD/MM/YY` i.e. the year is two digits.

the query param is sending the dob as a date, so this would require a fix in mSupply.

That would also give the opportunity to do this:
```
				$obPatients:=$obPatients.query("date_of_birth=:1 && date_of_birth!=''";$tDateOfBirth)

```

( at the moment, if you search for a date you get a TON of dob = `N/A` which, I think, shouldn't be returned if you send a date

diegoevangelisti:
> ( at the moment, if you search for a date you get a TON of dob = `N/A` which, I think, shouldn't be returned if you send a date )

Thanks @mark-prins for the clarification! And yes, I've experimented the same recently, lots of `N/A` DOBs but nothing matching with the typed date.
I'm happy to help with that but I will wait for @wlthomson to see what is the idea regarding this fixes.

wlthomson:
@diegoevangelisti @mark-prins Thanks for looking into this! Looks like the date bug is a duplicate of #2761.

I agree that this needs to be a desktop fix, not sure what changes to the date format broke this! Will also need to be careful that 4D isn't being clever and applying time zone offsets which offset the date by a day.

I think we can consider that a separate issue, though (so if that's the only problem, this can be approved :tada:)!

PR #2827, opened 2020-05-07
#2823 update form date input date parsing

### LINKED ISSUE #2823: Save button gets disabled when DOB is removed after searching in lookup patient/prescriber
## Describe the bug

In Dispensary module, the `Save` button in `Lookup patient` gets disable when removing the entered D.O.B (filter) after searching for a patient/prescriber.

lookupPatient

### To reproduce

1. Go to 'Dispensary'
2. Click on 'Lookup patient' and type a DOB
3. Click on `Save` and then delete the DOB

### Expected behaviour

The button should remain enable.

### Proposed Solution

N/A.

### Version and device info

- App version: v5.0.0-rc3
- Tablet model: AVD
- OS version: Android 9.0 - API 28

### Additional context

Raised while testing #2818

wlthomson:
Fixes #2823.

## Change summary

Fixes bug in validation logic which was causing the search button to be disabled when the date of birth field was reset.

## Testing

- [ ] The patient lookup search button is not disabled when the modal is opened.
- [ ] The patient lookup search button is not disabled when the date of birth field is updated.
- [ ] The patient lookup search button is not disabled when the date of birth field is reset to blank.

N/A.


mark-prins:
when I enter a date and search, I get this every time. Checking now - just making a note to alert other potential reviewers

Screen Shot 2020-05-08 at 3 44 55 PM

wlthomson:
@mark-prins That's odd, I thought that was the case and changed it after getting that error. I must have got it backwards!

Update:

The `onUpdate` date formatting wasn't the issue (it works either way). The problem was the server rejecting the request with a date in the query string, as the 4D routing doesn't handle backslashes in query strings. The routing was splitting on all `/` characters to split the URL into components, even if the `/` was escaped and came after the `&` separator. 

I've fixed the query string constructor, so now date queries are working fine (for me). 

Note: if you're getting `Invalid date` records that don't match, make sure your structure has this fix: https://github.com/sussol/msupply/commit/c4d5bf2f288588f2252341552e6f392a15a8b35a.

diegoevangelisti:
> when I enter a date and search, I get this every time. Checking now - just making a note to alert other potential reviewers
> 
> Screen Shot 2020-05-08 at 3 44 55 PM

@wlthomson @mark-prins I am getting the same here - `No records found` when trying to match an existing Patient entering a First Name, Last Name and DOB. (It does work with just First and Last Name).

src/sync/lookupApiUtils.js
```diff
@@ -37,20 +45,22 @@ export const createPolicyRecord = policy => {
...
-  }, '');
+const getQueryString = params => {
+  const object = params.reduce((queryObject, param) => {
+    const [[key, value], [, type]] = Object.entries(param);
+    if (!value) return queryObject;
+    const paramValue = type !== TYPES.DATE ? `@${value}@` : moment(value).format('YYYYMMDD');
+    return { ...queryObject, [key]: paramValue };
+  }, {});
+  const queryString = `?${querystring.stringify(object)}`;
+  return queryString;
```
mark-prins, line 56: mainly because the mix of 'querystring' and 'queryString' tripped me up.
```suggestion
  return `?${querystring.stringify(object)}`;
```

src/sync/lookupApiUtils.js
```diff
@@ -37,20 +45,22 @@ export const createPolicyRecord = policy => {
...
-    if (!value) return queryString;
-    const paramString = `${key}=${SEPARATORS.QUERY_WILDCARD}${value}${SEPARATORS.QUERY_WILDCARD}`;
-    const paramSeparator = queryString.length > 0 ? SEPARATORS.PARAMETERS : SEPARATORS.QUERY_STRING;
-    return queryString + paramSeparator + paramString;
-  }, '');
+const getQueryString = params => {
+  const object = params.reduce((queryObject, param) => {
+    const [[key, value], [, type]] = Object.entries(param);
+    if (!value) return queryObject;
+    const paramValue = type !== TYPES.DATE ? `@${value}@` : moment(value).format('YYYYMMDD');
```
mark-prins, line 52: The date format should be `DDMMYYYY`

Screen Shot 2020-05-11 at 1 23 06 PM

wlthomson: oops

49ca9ccc1f65540776b674b0eeeb5d392fc27b46.

src/sync/lookupApiUtils.js
wlthomson: 49ca9cc.

wlthomson:
@diegoevangelisti Date not matching would have been due to the format inconsistency in @mark-prins review. Should work now.

src/sync/lookupApiUtils.js
mark-prins: dobby-finger-snap

( works a treat! thanks )

PR #2862, opened 2020-05-26
#2738 locations through form control

### LINKED ISSUE #2738: Locations should be created through `FormControl`

Currently adding a `Location` just adds a plain location with a default `location_type` of *some* `location_type` which has the description 'fridge'.

When creating a program, have a `FormControl` modal open for `Location`, which will allow a user to create a `Location`, typing the `code`, `description` and selecting a `LocationType`.

## Implementation

- Add a `FormInputConfig` for `Location`
- Add the `FormControl` component to `VaccineAdminPage`, which opens when pressing `Add Location`.

N/A

## Additional context

N/A

joshxg:
Fixes #2738 

## Change summary

- Adds a `FormControl` for `Location` editing. This only has 3 fields - location type, code and description, but if you're just adding a record straight to the list box, it is a lot harder to enforce things - i.e. code, description.
- Adds a column into the Location list box to have a pencil for editing a location.

## Testing

N/A

It's a little naff here because location types are user defined. So you can't really have a list of 'fridges' because the fridge location type would be different for every data file. Also what if you have a type 'large fridge' and 'small fridge' or 'cool room' and 'vaccine fridge'. So you have to list all locations. But then it's in the 'vaccine module' area which is a little odd. 🤷


src/localization/formInputStrings.json
```diff
@@ -7,12 +7,14 @@
     "code": "Code",
     "country": "Country",
     "date_of_birth": "Date of birth",
-    "discount_rate": "Discount rate",
+    "description": "Description",
+    "Discount_rate": "Discount rate",
```
mark-prins, line 11: the other ones are still lowercase. I think this one should be too

mark-prins:
Am checking now. I tried to add a location and then edit and I got told not to:

Screen Shot 2020-05-27 at 12 29 51 PM

src/utilities/formInputConfigs.js
```diff
@@ -327,6 +348,7 @@ const FORM_CONFIGS = {
     FORM_INPUT_KEYS.SEARCH_FIRST_NAME,
     FORM_INPUT_KEYS.SEARCH_REGISTRATION_CODE,
   ],
+  location: ['code', 'description', 'locationType'],
```
mark-prins, line 351:
```suggestion
  location: [FORM_INPUT_KEYS.CODE, FORM_INPUT_KEYS.DESCRIPTION, FORM_INPUT_KEYS.LOCATION_TYPE],
```

joshxg:
Silly silly me :(

Didn't actually get your error but I am guessing it is because of a lack of `LocationType`s. Have added a few commits - there will be a Toast now when you try to add or edit a location when you don't have any location types. I was thinking of having code in msupply to always ensure there is at least one location type (with like.. SUSSOL_FRIDGE?) or something could be an idea.. but I'm not sure. This is difficult with it being user defined !

joshxg:
Dumb, I can't reply to your comments  ! :(

Removed from UIDatabase: 953eaef46180676008ba8c1d4192907fc75c3d8d
Reverted the capitals mishap: fb6c8062f3572a7fb748a79ea7d80ae681bc37f7
use constants: 28447732abaf484754cf3693ff64761b94f5dcc5

Thanks!!

PR #2864, opened 2020-05-28
#2739 date pickers on charts

### LINKED ISSUE #2739: Add date pickers to `VaccinePage`

Currently you can only view the last 30 days of temperatures. It would be great if you could view any date!

Add date pickers, which set a start and end date for viewing temperatures

## Implementation

- Add a date picker for the start and end date, which determine which dates to select temperature logs from.
- Potentially it would be better that rather than having a list of fridges, have a single component which has a drop down to select a fridge to show temperatures for?

N/A

## Additional context

There may be an issue here with potentially showing a years worth of temperatures. That's a lot of records (> 15000). Possibly need to have a limit on the number of days - but could just have a longer 'loading' time rather than restricting like this

joshxg:
Fixes #2739 

## Change summary

- Adds date pickers to choose from a range of dates. Have tested with 10,000 logs and the performance is fine (7 months worth).
- Fixed the bug when the SVG has no data to render

image

image

## Testing

N/A

I wonder if rather than having a list of fridges a drop down would be nicer:

image

### DIFF
--- a/src/hooks/useDatePicker.js
+++ b/src/hooks/useDatePicker.js
@@ -1,0 +1,42 @@
+/* eslint-disable import/prefer-default-export */
+import React from 'react';
+
+/**
+ * Hook to simplify the use of the native date picker component by unwrapping
+ * the callback event and firing the callback only when the date is changing,
+ * providing a value to conditionally render the native picker and a callback
+ * to set the render value.
+ *
+ * Pass an onChange callback which will be called when a new date is selected
+ * from the datepicker and not when cancelled or dismissed, with a JS date object
+ * rather than a nativeEvent.
+ *
+ * @param  {Func}  callback callback
+ * @return {Array} [
+ * datePickerIsOpen: boolean indicator whether the date picker is open.
+ * openDatePicker: Callback function to set the datePickerIsOpen to true.
+ * datePickerCallback: Wrapped callback, triggered on changing the date in the picker,
+ * returning a JS date object.
+ * ]
+ */
+export const useDatePicker = onChangeCallback => {
+  const [datePickerIsOpen, setDatePickerIsOpen] = React.useState(false);
+
+  const datePickerCallback = React.useCallback(
+    event => {
+      setDatePickerIsOpen(false);
+      const { type, nativeEvent } = event;
+      if (type === 'set' && onChangeCallback) {
+        const { timestamp } = nativeEvent;
+        onChangeCallback(new Date(timestamp));
+      }
+    },
+    [onChangeCallback]
+  );
+
+  const openDatePicker = React.useCallback(() => {
+    setDatePickerIsOpen(true);
+  }, []);
+
+  return [datePickerIsOpen, openDatePicker, datePickerCallback];
+};
--- a/src/localization/vaccineStrings.json
+++ b/src/localization/vaccineStrings.json
@@ -8,7 +8,8 @@
     "no_fridges": "No Fridges",
     "temperature_breaches_for": "Temperature breaches for",
     "fridges": "Fridges",
-    "sensors": "Sensors"
+    "sensors": "Sensors",
+    "oops_no_temperatures": "Oops! There are no temperatures recorded during this time"
   },
   "fr": {},
   "gil": {},
--- a/src/selectors/fridge.js
+++ b/src/selectors/fridge.js
@@ -57,7 +57,7 @@ export const selectChunkedTemperatureLogs = createSelector(
   }
 );
 
-export const selectMinAndMaxLogs = createSelector(selectChunkedTemperatureLogs, logs =>
+export const selectMinAndMaxLogs = createSelector([selectChunkedTemperatureLogs], logs =>
   logs.reduce(
     (acc, logGroup) => {
       const temperatures = logGroup.map(({ temperature }) => temperature);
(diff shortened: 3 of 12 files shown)


src/hooks/useDatePicker.js
```diff
@@ -0,0 +1,42 @@
...
+ * from the datepicker and not when cancelled or dismissed, with a JS date object
+ * rather than a nativeEvent.
+ *
+ * @param  {Func}  callback callback
+ * @return {Array} [
+ * datePickerIsOpen: boolean indicator whether the date picker is open.
+ * openDatePicker: Callback function to set the datePickerIsOpen to true.
+ * datePickerCallback: Wrapped callback, triggered on changing the date in the picker,
+ * returning a JS date object.
+ * ]
```
joshxg, line 20: I for some reason found this comment hard to write. Struggled to make it short and concise but helpful

src/selectors/fridge.js
```diff
@@ -57,15 +57,14 @@ export const selectChunkedTemperatureLogs = createSelector(
   }
 );
 
-export const selectMinAndMaxLogs = createSelector(selectChunkedTemperatureLogs, logs =>
+export const selectMinAndMaxLogs = createSelector([selectChunkedTemperatureLogs], logs =>
```
joshxg, line 60: bug fix

src/selectors/fridge.js
```diff
@@ -79,7 +78,7 @@ export const selectMinAndMaxLogs = createSelector(selectChunkedTemperatureLogs,
   )
 );
 
-export const selectMinAndMaxDomains = createSelector(selectMinAndMaxLogs, minAndMaxLogs => {
+export const selectMinAndMaxDomains = createSelector([selectMinAndMaxLogs], minAndMaxLogs => {
```
joshxg, line 81: bug fix

src/selectors/fridge.js
```diff
@@ -88,8 +87,16 @@ export const selectMinAndMaxDomains = createSelector(selectMinAndMaxLogs, minAnd
...
+  [selectTemperatureLogsFromDate, selectTemperatureLogsToDate, selectSelectedFridge],
+  (fromDate, toDate, fridge) => {
+    const { breaches } = fridge ?? {};
+    const breachesInDateRange = breaches.filtered(
+      'startTimestamp >= $0 && endTimestamp <= $1',
+      fromDate,
+      toDate
+    );
+
+    return breachesInDateRange;
```
joshxg, line 99: maybe not the best - only show a breach if it started within the range. Would be better to have this show a breach if it was anywhere within the range, but if it started before hand, adjust the breach values to show it beginning at the first x value in the chart. Could make an issue for this as it would take a little more work - just ensuring that the X value is at least as large as the smallest X value of the temperature logs (otherwise you get a funny chart with the breach icon stretching the X domain - i.e. breach starts on the 1st, where as temperature logs start on the 10th, funny chart

src/widgets/DatePickerButton.js
```diff
@@ -0,0 +1,44 @@
...
+          minimumDate={minimumDate}
+          onChange={datePickerCallback}
+          mode="date"
+          display="spinner"
+          value={initialValue}
+        />
+      )}
+    </>
+  );
+};
```
joshxg, line 32: should be doing this more - i.e. when I used the date picker for picking date of births in the `FormControl`/`FormDateInput` component. Try to break components down into the smallest forms and compose them

src/widgets/DateRangeSelector.js
```diff
@@ -0,0 +1,76 @@
...
+import moment from 'moment';
+
+import { DatePickerButton } from './DatePickerButton';
+import { FlexRow } from './FlexRow';
+import { FlexColumn } from './FlexColumn';
+
+import { textStyles, SUSSOL_ORANGE } from '../globalStyles';
+import { generalStrings } from '../localization/index';
+
+export const DateRangeSelector = ({
```
joshxg, line 18: tried to make this so that it is able to be used in for example the requisitions page to select a range of entry dates for requisitions to show

src/widgets/FridgeDisplay.js
```diff
@@ -31,9 +39,20 @@ export const FridgeDisplay = ({
   const [render, setRender] = React.useState(false);
 
   React.useEffect(() => {
-    setTimeout(() => setRender(isActive), 1000);
+    setTimeout(() => setRender(isActive), 500);
```
joshxg, line 43: just reduced this a little, it was a little long

src/widgets/FridgeDisplay.js
```diff
@@ -50,14 +69,22 @@ export const FridgeDisplay = ({
           <ActivityIndicator size="small" color={SUSSOL_ORANGE} />
         </FlexView>
       ),
-    [render]
+    [render, minLine, maxDomain, minDomain, maxLine, breaches, onOpenBreachModal]
```
joshxg, line 71: every bug ever is due to silly dependency list  values 🤣

src/localization/vaccineStrings.json
```diff
@@ -8,7 +8,8 @@
     "no_fridges": "No Fridges",
     "temperature_breaches_for": "Temperature breaches for",
     "fridges": "Fridges",
-    "sensors": "Sensors"
+    "sensors": "Sensors",
+    "oops_no_temperatures": "Oops! There are no temperatures recorded during this time"
```
joshxg, line 12: is this.. ok?

joshxg:
https://github.com/openmsupply/mobile/pull/2864/commits/46d2ce6fd9f8976951677b7cd772420e1a8d5c9d

embares

Issue #2767, opened 2020-04-23
Syncing temperatures should be a single transaction


joshxg:
## Describe the bug

When syncing temperatures, the sensor needs to be reset. This should be done within a single transaction to ensure the data is valid.

### To reproduce

1. Go to `TemperatureSyncModal`
2. Click on `Manual Sync`
3. Try to interrupt the blue tooth connection between syncing temperatures and resetting the sensor.

### Expected behaviour

You should not save temperature logs if the sensor didn't reset.

### Proposed Solution

1. Make all of the dispatches in a single write transaction. This will lock the db for too long so is less ideal. ❌ 
2. Rather than saving temperatures instantly, return them and save them only after resetting the sensor. ✅ 

### Version and device info

- App version: 5.0.0+
- Tablet model: All
- OS version: All

### Additional context

N/A

joshxg:
Closed by #2865

PR #2865, opened 2020-05-29
#2767 single transaction sensor logs

joshxg:
Fixes #?

## Change summary

- extracts the saving of sensorlogs and only saves thema fter having reset a sensor. This is because if we don't reset the sensor, the logs aren't cleared so next time you read it you would have double the sensor logs.
- This is pretty tricky and there's no real solution to being able to do this atomically without having timestamps on the logs on the sensor, so going with this for now (for example, what if you reset the sensor and dont write the logs. So ok, switch around the reset/write - what if you write before resetting...).
- The other option is to try and wrap this all in a `write`, however that means if its running in the background you might have issues with starting multiple writes with normal user tasks

## Testing

N/A

N/A

### DIFF
--- a/src/actions/TemperatureSyncActions.js
+++ b/src/actions/TemperatureSyncActions.js
@@ -104,6 +104,18 @@ const downloadLogsComplete = () => ({
   type: TEMPERATURE_SYNC_ACTIONS.DOWNLOAD_LOGS_COMPLETE,
 });
 
+const saveLogs = (logData, sensor) => async dispatch => {
+  dispatch(startSavingTemperatureLogs(logData?.length));
+  const timeNow = new Date().getTime();
+
+  UIDatabase.write(() =>
+    logData.forEach(({ temperature }, i) => {
+      const timestamp = timeNow - (i + 1) * MILLISECONDS.FIVE_MINUTES;
+      createRecord(UIDatabase, 'SensorLog', temperature, new Date(timestamp), sensor);
+    })
+  );
+};
+
 const downloadLogs = sensor => async dispatch => {
   dispatch(downloadLogsStart());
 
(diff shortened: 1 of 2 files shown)


src/actions/TemperatureSyncActions.js
```diff
@@ -104,31 +104,30 @@ const downloadLogsComplete = () => ({
   type: TEMPERATURE_SYNC_ACTIONS.DOWNLOAD_LOGS_COMPLETE,
 });
 
+const saveLogs = (logData, sensor) => async dispatch => {
+  dispatch(startSavingTemperatureLogs(logData?.length));
+  const timeNow = new Date().getTime();
+
+  UIDatabase.write(() =>
+    logData.forEach(({ temperature }, i) => {
+      const timestamp = timeNow - (i + 1) * MILLISECONDS.FIVE_MINUTES;
```
wlthomson, line 113: Nit: I think specifying the base unit of milliseconds in the constant is a bit redundant since the key specifies minutes? Could have something like `5 * TIME.MINUTE` to make it a bit more flexible.

src/actions/TemperatureSyncActions.js
```diff
@@ -184,7 +183,7 @@ const completeSavingTemperatureLogs = () => ({
   type: TEMPERATURE_SYNC_ACTIONS.COMPLETE_SAVING_TEMPERATURE_LOGS,
 });
 
-const createTemperatureLogs = sensor => dispatch => {
+const createTemperatureLogs = sensor => async dispatch => {
```
wlthomson, line 186: :+1:.

src/selectors/fridge.js
```diff
@@ -91,7 +91,7 @@ export const selectBreaches = createSelector(
   [selectTemperatureLogsFromDate, selectTemperatureLogsToDate, selectSelectedFridge],
   (fromDate, toDate, fridge) => {
     const { breaches } = fridge ?? {};
-    const breachesInDateRange = breaches.filtered(
+    const breachesInDateRange = breaches?.filtered(
```
wlthomson, line 94: REFACTOR!? :laughing:

Issue #1949, opened 2020-01-16
Shouldn't be able to add items with 0 stock to a CI


josh-griffin:
## Describe the bug

You can add items with 0 quantity to a CI. You can't issue any of this item. The item line gets pruned from the transaction when finalised.

### To reproduce

1. Create a new CI
2. Add two items, one with stock one without
3. Issue some of the item with stock and finalise

### Expected behaviour

Shouldn't be able to add items with 0 stock to a CI. Either a toast or just a disabled row.

### Proposed Solution

N/A

### Version and device info

- App version: 3.1.1+
- Tablet model: N/A
- OS version: N/A

### Additional context

N/A

PR #2873, opened 2020-06-01
#2872 Doses column display values

### LINKED ISSUE #2872: Display N/A rather than 0 for non-vaccine-item doses
## Describe the bug

The doses column on a CI/Stocktake is a little odd in that it is a column which should only be active for subset of rows. This makes data entry a little annoying and handling of the `doses` cell a little strange for non-vaccine items. i.e. what do you do when a user tries to edit the doses column of a non-vaccine item?

Further, the focusing logic either would need quite a big change to handle being able to focus normal quantity cells, then the cells which are focusable on a row condition - even more - the point where the focusing is handled means the entire table would need to re-render more frequently. All bad. No good.

### To reproduce

1. With vaccine module enabled, create a CI with a vaccine item and a non-vaccine item.
2. See the odd behaviour with a doses column for a non-vaccine item.

### Expected behaviour

Should not be really strange

### Proposed Solution

Disabled the doses cell for a non-vaccine item. Display "N/A" in the cell for non-vaccine items. Allow it to be editable for vaccine items. Remove focusing of the doses column completely.

### Version and device info

- Tablet model: N/A
- OS version: N/A

### Additional context

N/A

joshxg:
Fixes #2872

## Change summary

- Now displays N/A for non-vaccine items in the doses column
- Couple offroad bugs which i'll point out!

## Testing

N/A

N/A

### DIFF
--- a/src/database/DataTypes/ItemBatch.js
+++ b/src/database/DataTypes/ItemBatch.js
@@ -27,7 +27,7 @@ export class ItemBatch extends Realm.Object {
   }
 
   get hasBreached() {
-    return this.location?.hasBreached() ?? false;
+    return this.location?.hasBreached ?? false;
   }
 
   get breaches() {
--- a/src/database/DataTypes/Stocktake.js
+++ b/src/database/DataTypes/Stocktake.js
@@ -250,7 +250,8 @@ export class Stocktake extends Realm.Object {
 
   get canFinalise() {
     const finaliseStatus = { success: true, message: modalStrings.finalise_stocktake };
-    if (!this.hasSomeCountedItems || !this.hasValidDoses) {
+
+    if (!this.hasSomeCountedItems) {
       finaliseStatus.success = false;
       finaliseStatus.errorMessage = modalStrings.stocktake_no_counted_items;
     }
--- a/src/database/utilities/createRecord.js
+++ b/src/database/utilities/createRecord.js
@@ -661,10 +661,14 @@ const createItemBatch = (database, item, batchString, supplier) => {
       .filtered('itemId == $0 && joinsThisStore == true', itemId)[0];
     const { restrictedLocationType } = itemStoreJoin ?? {};
     const { id: locationTypeId = '' } = restrictedLocationType ?? {};
-    const location = database.objects('Location').filtered('locationType.id == $0', locationTypeId);
+    const location = database
+      .objects('Location')
+      .filtered('locationType.id == $0', locationTypeId)[0];
 
-    itemBatch.applyLocation(database, location);
-    itemBatch.applyVvmStatus(database, vaccineVialMonitorStatus);
+    if (itemBatch.shouldApplyLocation(location)) itemBatch.applyLocation(database, location);
+    if (itemBatch.shouldApplyVvmStatus(vaccineVialMonitorStatus)) {
+      itemBatch.applyVvmStatus(database, vaccineVialMonitorStatus);
+    }
   }
 
   return itemBatch;
--- a/src/pages/dataTableUtilities/getColumns.js
+++ b/src/pages/dataTableUtilities/getColumns.js
@@ -744,7 +744,7 @@ const COLUMNS = () => ({
     title: tableStrings.doses,
     alignText: 'right',
     sortable: true,
-    editable: true,
+    editable: false,
   },
 
   [COLUMN_NAMES.RETURN_AMOUNT]: {
--- a/src/widgets/DataTable/DataTableRow.js
+++ b/src/widgets/DataTable/DataTableRow.js
@@ -102,22 +102,45 @@ const DataTableRow = React.memo(
           switch (type) {
             case COLUMN_TYPES.EDITABLE_STRING:
             case COLUMN_TYPES.EDITABLE_NUMERIC: {
-              // Special condition for stocktake counted total quantity cells.
-              // Use the placeholder 'Not counted' when a stocktake item or batch
-              // has not been counted yet.
-              let placeholder = '';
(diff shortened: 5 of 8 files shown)


src/database/DataTypes/ItemBatch.js
```diff
@@ -27,7 +27,7 @@ export class ItemBatch extends Realm.Object {
   }
 
   get hasBreached() {
-    return this.location?.hasBreached() ?? false;
+    return this.location?.hasBreached ?? false;
```
joshxg, line 30: not a function! 🐛

src/database/DataTypes/Stocktake.js
```diff
@@ -250,7 +250,8 @@ export class Stocktake extends Realm.Object {
 
   get canFinalise() {
     const finaliseStatus = { success: true, message: modalStrings.finalise_stocktake };
-    if (!this.hasSomeCountedItems || !this.hasValidDoses) {
+
+    if (!this.hasSomeCountedItems) {
```
joshxg, line 254: I have removed the check for valid doses as it was introduced initially when I was restricting the doses value to be within `totalQuantity >= doses <= totalQuantity * item.doses` - now the only restriction is `doses <= totalQuantity * item.doses`, so I'm not sure where to go with this. Could change the validity to be just that `doses <= totalQuantity * item.doses`, but also it is a pretty bad UI experience to go to finalise only to be told no and then have to find the batch. Maybe that's a better solution though. I will make an issue for this as it has to change either way!

src/database/utilities/createRecord.js
```diff
@@ -661,10 +661,14 @@ const createItemBatch = (database, item, batchString, supplier) => {
...
-
-    itemBatch.applyLocation(database, location);
-    itemBatch.applyVvmStatus(database, vaccineVialMonitorStatus);
+    const location = database
+      .objects('Location')
+      .filtered('locationType.id == $0', locationTypeId)[0];
+
+    if (itemBatch.shouldApplyLocation(location)) itemBatch.applyLocation(database, location);
+    if (itemBatch.shouldApplyVvmStatus(vaccineVialMonitorStatus)) {
+      itemBatch.applyVvmStatus(database, vaccineVialMonitorStatus);
```
joshxg, line 670: added these checks to check if it should apply or not - shouldv'e been added originally, oops

src/pages/dataTableUtilities/getColumns.js
```diff
@@ -744,7 +744,7 @@ const COLUMNS = () => ({
     title: tableStrings.doses,
     alignText: 'right',
     sortable: true,
-    editable: true,
+    editable: false,
```
joshxg, line 747: this is weird and a bit hacky. This field is only used for focusing - I wish I called it focusable :(. Maybe I should refactor and change the field name? I dunno. But making this column not focusable - the doses gets auto-updated according to the total quantity, and if it happens to have a mismatch, I think it will be a better experience for data entry to click on the cell and edit it, only when there is a discrepancy. And saves all the focus logic drama! 🤣

src/widgets/DataTable/DataTableRow.js
```diff
@@ -102,22 +102,45 @@ const DataTableRow = React.memo(
...
+              // cell is disabled. For example display N/A on a row for the doses column rather
+              // than the underlying value "0".
+              const disabledText = {
+                [COLUMN_KEYS.DOSES]: isVaccine ? '' : generalStrings.not_available,
+              };
+
+              const value = disabledText[columnKey] ? disabledText[columnKey] : rowData[columnKey];
+              const placeholder = extraPlaceholders[columnKey] ?? '';
+
+              const inputIsDisabled = isDisabled || !!disabledCondition[columnKey];
```
joshxg, line 134: A little weird and not very nice to be in this component, really. But here it is, open to suggestions. Basically want to add a row-level disable condition, different text when disabled (to display N/A on a doses column rather than the underlying value of 0). Thought this was a niceish implementation for extending?

src/database/DataTypes/Stocktake.js
mark-prins: would be nice to have the validation earlier, and some sort of notification to the user. I'm currently typing in a number for `DOSES` in a CI and it is just changing back to `0` for some reason

joshxg: Do you have 0 quantity for that item? That happens for normal quantity too :( https://github.com/openmsupply/mobile/issues/1949 unless it's something different, of course

src/widgets/DataTable/DataTableRow.js
mark-prins: agree, not ideal to have in the component. I wondered about passing in functions to do the placeholder, disabled condition and disabled text - ones which accept the `rowData` as a parameter. They could be passed in as props?

mark-prins: is fine as it is, just wondering about alternatives to keep some of this specificity out of the more general DataTable code

src/pages/dataTableUtilities/getColumns.js
mark-prins: perhaps a bit hacky, but for a good reason!

src/database/DataTypes/Stocktake.js
mark-prins: yes, it is a zero stock issue. I'm having a separate problem, in that I cannot finalise a stocktake with a vaccine in it, so I couldn't quickly give myself stock

joshxg:
I like that idea a lot. It can just have some default behaviour so that refactoring won't really change the current interface.

PR #2890, opened 2020-06-06
#2882 datepicker returning times

### LINKED ISSUE #2882: Date pickers returning dates with times
## Describe the bug

When selecting dates with the `useDatePicker` custom hook, the returned date has a time

### To reproduce

1. Go to VaccinePage
2. Click on to and from dates on the same day

### Expected behaviour

when no time is selected, there should be no times on the timestamp

### Proposed Solution

in the `useDatePicker` hook, remove the time using `.toDateString()` (I think), conditional on if times are being selected, or not

### Version and device info

- Tablet model: N/A
- OS version: N/A

### Additional context

N/A

joshxg:
Fixes #2882 

## Change summary
- My solution in the issue I think was flawed. Instead, changed it to make the custom hook just pass the timestamp, and made the alterations to the date where I needed it for this particular use case.
- the fromDate is from the start of the day, where as the toDate is the end of the day

## Testing

N/A

Not 100% on dates, so if there's a better way more then happy to hear it!

### DIFF
--- a/src/reducers/FridgeReducer.js
+++ b/src/reducers/FridgeReducer.js
@@ -35,14 +35,20 @@ export const FridgeReducer = (state = initialState(), action) => {
       const { payload } = action;
       const { date } = payload;
 
-      return { ...state, fromDate: date };
+      const fromDate = new Date(date);
+      fromDate.setUTCHours(0, 0, 0, 0);
+
+      return { ...state, fromDate };
     }
 
     case FRIDGE_ACTIONS.CHANGE_TO_DATE: {
       const { payload } = action;
       const { date } = payload;
 
-      return { ...state, toDate: date };
+      const toDate = new Date(date);
+      toDate.setUTCHours(23, 59, 59, 999);
+
+      return { ...state, toDate };
     }
 
     default:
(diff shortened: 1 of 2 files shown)


src/reducers/FridgeReducer.js
```diff
@@ -35,14 +35,20 @@ export const FridgeReducer = (state = initialState(), action) => {
...
+      return { ...state, fromDate };
     }
 
     case FRIDGE_ACTIONS.CHANGE_TO_DATE: {
       const { payload } = action;
       const { date } = payload;
 
-      return { ...state, toDate: date };
+      const toDate = new Date(date);
+      toDate.setUTCHours(23, 59, 59, 999);
```
wlthomson, line 49: Nit: I think incrementing the day and decrementing the ms is a bit clearer/tidier:

```suggestion
      toDate.setUTCDate(toDate.getUTCDate() + 1);
      toDate.setUTCHours(0, 0, 0, -1);
```

Could also do as a one-liner... but I think it's less clear than what you already have :laughing:.

```suggestion
      toDate.setUTCHours(24, 0, 0, -1);
```

joshxg: Maybe my lack of ARITHMETIC skills, but the 23,59,59,9999 is easier to grok then seeing the -1 and having to comprehend

PR #2892, opened 2020-06-06
#2888 VVMLogic [Branched from: #2882]

### LINKED ISSUE #2888: Incorrect logic when creating vvm logs
## Describe the bug

There is a logic error on the check determining if a new VVM status log should be created

### To reproduce

N/A

### Expected behaviour

No logic error!

### Proposed Solution

The method `shouldApplyVvmStatus` is logically incorrect and uses `===` rather than `!==` - fix it!

### Version and device info

- App version: Vaccine feature branch
- Tablet model: N/A
- OS version: N/A

### Additional context

N/A

joshxg:
Fixes #2888 

## Change summary

- Adds corrected operator so that when checking if a VVM Status should be applied, it will be applied if the 'new one' is different - i.e. going from two to three

## Testing
N/A

N/A


src/database/DataTypes/ItemBatch.js
```diff
@@ -138,7 +138,7 @@ export class ItemBatch extends Realm.Object {
    * @return {Bool} Indicator whether the new vvm status should be applied to this batch.
    */
   shouldApplyVvmStatus(newVvmStatus = {}) {
-    return newVvmStatus?.id === this.currentVvmStatus?.id;
+    return newVvmStatus?.id !== this.currentVvmStatus?.id;
```
wlthomson, line 141: I think this will be truthy if `newVvmStatus.id === undefined` and `currentVvmStatus.id` is defined? Intended behaviour?

joshxg: Yup - intended. Open to different ideas, but basically, if they're not the same then applying it should be OK

PR #2894, opened 2020-06-06
#2885 TemperatureSyncStateMessage incorrect

### LINKED ISSUE #2885: TemperatureSyncState incorrect when errored
## Describe the bug

When an error is thrown during sync, the sync state (on the navigation bar) incorrectly continues to state "Sync in progress".

### To reproduce

1. Go to `TemperatureSyncModal`
2. Click on manual sync (while doing something to ensure there is an error thrown during sync. I use a pillow)
3. After an error is thrown, close the modal
4. See error (on navigation bar, temperature sync state still says "sync in progress".

### Expected behaviour

Should state "SYNC ERROR. Last sync d/m/yyyy" - same as regular sync

### Proposed Solution

There is a syntax error in the selector

### Version and device info

- Tablet model: N/A 
- OS version: N/A

### Additional context

N/A

joshxg:
Fixes #2885 

## Change summary

- Corrected logic in the short circuit
- Added corrected message to match the same sync message of normal sync

## Testing

N/A

N/A

### DIFF
--- a/src/selectors/temperatureSync.js
+++ b/src/selectors/temperatureSync.js
@@ -50,10 +50,9 @@ export const selectTemperatureSyncStateMessage = ({ temperatureSync }) => {
   const { syncState, syncError, lastTemperatureSync } = temperatureSync;
   const formattedDate = moment(lastTemperatureSync).format('D.M.YYYY');
 
-  return (
-    (syncError && `${syncStrings.sync_error}. ${syncStrings.last_sync} ${formattedDate}`) ||
-    (syncState ? syncStrings.sync_in_progress : syncStrings.sync_enabled)
-  );
+  if (syncError) return `${syncStrings.sync_error}. ${syncStrings.last_sync} ${formattedDate}`;
+
+  return syncState ? `${syncStrings.sync_in_progress}` : `${syncStrings.sync_enabled}`;
 };
 
 export const selectTemperatureModalIsOpen = ({ temperatureSync }) => {
--- a/src/selectors/temperatureSync.js
+++ b/src/selectors/temperatureSync.js
@@ -47,11 +47,13 @@ export const selectTemperatureSyncIsComplete = ({ temperatureSync }) => {
 };
 
 export const selectTemperatureSyncStateMessage = ({ temperatureSync }) => {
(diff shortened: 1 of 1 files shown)


src/selectors/temperatureSync.js
```diff
@@ -47,11 +47,13 @@ export const selectTemperatureSyncIsComplete = ({ temperatureSync }) => {
 };
 
 export const selectTemperatureSyncStateMessage = ({ temperatureSync }) => {
-  const { syncState, syncError } = temperatureSync;
+  const { syncState, syncError, lastTemperatureSync } = temperatureSync;
+  const formattedDate = moment(lastTemperatureSync).format('D.M.YYYY');
```
wlthomson, line 51: Any reason for `D.M.YYYY` (as opposed to, e.g. `DD.MM.YYYY`)?

src/selectors/temperatureSync.js
```diff
@@ -47,11 +47,13 @@ export const selectTemperatureSyncIsComplete = ({ temperatureSync }) => {
...
 
-  return (syncError && syncStrings.sync_error) || syncState
-    ? syncStrings.sync_in_progress
-    : syncStrings.sync_enabled;
+  return (
+    (syncError && `${syncStrings.sync_error}. ${syncStrings.last_sync} ${formattedDate}`) ||
+    (syncState ? syncStrings.sync_in_progress : syncStrings.sync_enabled)
```
wlthomson, line 55: Hm, feel like this would be a bit clearer not as a one-liner?
 
```suggestion
    if (syncError) return `${syncStrings.sync_error}. ${syncStrings.last_sync} ${formattedDate}`;
    return syncState ? `${syncStrings.sync_error}. ${syncStrings.last_sync} ${formattedDate}`;
```

src/selectors/temperatureSync.js
joshxg: This is whats used for normal sync, and `.getDate()` returns for example, `9`, for today (So trying to match existing stuffs)

src/selectors/temperatureSync.js
joshxg: 9e1ad53

Issue #2874, opened 2020-06-02
Batch in a breach determination is incorrect


joshxg:
## The suggestion

Determining if a batch is in a breach is not quite correct:

https://github.com/openmsupply/mobile/blob/550d6343dd01080b502529f11954ec3823404446/src/database/DataTypes/LocationMovement.js#L11-L15

This checks if the start of the breach was before the batch moved to the location and that the end of the breach was either after leaving the location or if it has not finished yet. However the conditions should be more:

- The breach started before the location movement, then it should have an end time of null, or any end time after the start time
- The breach started after the location movement, then the start time is also before the exit time (or now)

**Why?**
This will more accurately capture that the batch was in a breach 

## Describe alternatives you've considered

N/A

## Additional context

N/A

joshxg:
Closed by #2901

PR #2902, opened 2020-06-07
#2900 TemperatureBreach stats incorrect [Branched from: #2874] 

### LINKED ISSUE #2900: Breach stats logically incorrect
## Describe the bug

The stats for a breach: Total stock, number of affected batches are not being determined correctly.

The `Location` is determining if item batches were in a breach using a query on it's `LocationMovements` like:
- `(enterTimestamp <= $0 && (exitTimestamp > $0 || exitTimestamp == null))`

But it should be using a query like:

- `'enterTimestamp <= $0 && (exitTimestamp > $0 || exitTimestamp == null) ||' + 'enterTimestamp >= $0 && enterTimestamp <= $1',`

This will more accurately capture if an `ItemBatch` was in a Location during the time of a breach.

- The `totalStock()` can be more easily calculated by adding a `linkingObjects` on `Location` for `itemBatches`, and calculating the number of packs, from there.

### To reproduce

code logic bug

### Expected behaviour

As above

### Proposed Solution

As above

### Version and device info

- Tablet model: N/A
- OS version: N/A

### Additional context

N/A

joshxg:
Fixes #2900 

## Change summary

- Changed `batchesAtTime` to `batchesDuringTime`, having it receive a range rather than a single moment in time. Then changed the query such that the batch should be in the location either before the initial start time, and having left it either never, or some time after the starting time OR entered it after the start time but before the end time. This is because the old query was not capturing those batches which had entered after a breach had started, but before it had finished (primarily, for very long breaches! Some fridges in Vannuatu are just permanently breached :( )
- simplified `totalStock`

## Testing
N/A

N/A

### DIFF
--- a/src/database/DataTypes/TemperatureBreach.js
+++ b/src/database/DataTypes/TemperatureBreach.js
@@ -44,7 +44,7 @@ export class TemperatureBreach extends Realm.Object {
   }
 
   affectedBatches(database) {
-    return this.location.batchesAtTime(database, this.endTimestamp);
+    return this.location.batchesDuringTime(database, this.startTimestamp, this.endTimestamp);
   }
 
   numberOfAffectedBatches(database) {
--- a/src/database/DataTypes/TemperatureBreach.js
+++ b/src/database/DataTypes/TemperatureBreach.js
@@ -8,7 +8,7 @@ import moment from 'moment';
 
 export class TemperatureBreach extends Realm.Object {
   get colour() {
-    return this.temperatureBreachConfiguration.colour ?? null;
+    return this.temperatureBreachConfiguration?.colour ?? null;
   }
 
   get maximumTemperature() {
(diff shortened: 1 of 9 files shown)


src/database/DataTypes/TemperatureBreach.js
```diff
@@ -44,7 +44,7 @@ export class TemperatureBreach extends Realm.Object {
   }
 
   affectedBatches(database) {
-    return this.location.batchesAtTime(database, this.endTimestamp);
+    return this.location.batchesDuringTime(database, this.startTimestamp, this.endTimestamp);
```
wlthomson, line 47: I think `location` is optional? Maybe some defensive code here?

joshxg: Added here. 
31b2b08 

Note I've merged a whole bunch of PRs into this one so just look at that commit 🤣

Issue #2887, opened 2020-06-06
Selector for breaches logically incorrect


joshxg:
## Describe the bug

This function is incorrect 

https://github.com/openmsupply/mobile/blob/550d6343dd01080b502529f11954ec3823404446/src/selectors/fridge.js#L90-L97

### To reproduce

N/A

### Expected behaviour

N/A

### Proposed Solution

This should find all the breaches which started before the `fromDate` and either ended before the `toDate`, or not ended yet OR which started after the `fromDate`, but before the `fromDate`.

There will also need to be some mechanism to adjust the `temperatureBreach` X value on the chart, such that it fits within the X range of the `temperatureLogs`. I think:

- `if breach.startTimestamp < min(temperatureLogs.timestamp) : use `min(temperatureLogs.timestamp)

### Version and device info

- App version: Vaccine development
- Tablet model: N/A
- OS version: N/A

### Additional context

N/A

joshxg:
Closed by #2904

PR #2905, opened 2020-06-07
#2903 Show times on charts [Branched from: #2887]

### LINKED ISSUE #2903: Show times on charts

`BreachChart` and `VaccineChart` should show times

Change the `tickFormat` callback to add a time. I think a good format would be something like `HH:MM - DD/MM/YYYY`

## Implementation

In `VaccineChart` and `BreachChart` there is a `xTickFormat` prop - change the formatting

Dynamically change the formatting given the amount of time in the range. For example, if it is one day, only show times. More than one day, show times + dates, more than 3(?) days, show just dates. Maybe a good idea

## Additional context

N/A

joshxg:
Fixes #2903

## Change summary

- Adds dyanmic tick formatter - less then a day, show times, less than 3, times+dates, otherwise dates

## Testing

N/A

- Consider the case where it is times over two different days. Is it weird to not have the date?
- Dates are 24 hour - is it weird to use a colon? i.e. `19:06` not `1906`? Would it be better to do `07:06 pm`? 🤷

### DIFF
--- a/src/selectors/fridge.js
+++ b/src/selectors/fridge.js
@@ -42,7 +42,7 @@ export const selectFridgeTemperatureLogsFromDate = createSelector(
   [selectFridgeTemperatureLogs, selectTemperatureLogsFromDate, selectTemperatureLogsToDate],
   (logs, fromDate, toDate) => {
     const temperatureLogs =
-      logs?.filtered?.('timestamp > $0 && timestamp < $1', fromDate, toDate) ?? [];
+      logs?.filtered?.('timestamp >= $0 && timestamp =< $1', fromDate, toDate) ?? [];
 
     return temperatureLogs;
   }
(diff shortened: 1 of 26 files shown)


src/selectors/fridge.js
```diff
@@ -106,3 +106,21 @@ export const selectBreaches = createSelector(
...
 );
+
+export const selectTimestampFormatter = createSelector(
+  [selectTemperatureLogsFromDate, selectTemperatureLogsToDate],
+  (fromDate, toDate) => {
+    const durationInDays = moment(toDate).diff(moment(fromDate), 'days', true);
+
+    const justTime = 'HH:MM';
+    const dateAndTime = 'HH:MM - DD/MM';
+    const justDate = 'DD/MM';
```
wlthomson, line 117:
```suggestion
    const time = 'HH:MM';
    const date = 'DD/MM';
    const dateTime = `${time} - ${date}`;
```

joshxg: Nice - wanted to commit the suggestion but it wasn't quite complete 🤣 


4d6005f

PR #2906, opened 2020-06-07
#2891 2D array creation fix [Branched from: #2903]

### LINKED ISSUE #2891: Selecting logs is incorrect when not many logs
## Describe the bug

the method `selectMinAndMaxLogs` in `fridge.js` wrongly assumes an array, when the input could be an object.

### To reproduce

N/A

### Expected behaviour

no runtime errors!

### Proposed Solution

In `selectChunkedTemperatureLogs`, if there are more than 30 logs, we chunk the logs into a 2d array but if there are less, we leave it as a 1d array. I think the 1d array should still be turned into a 2d array, just with each subarray having a singular element. Not a huge cost since it's such a small array, anyway!

### Version and device info

- App version: Vaccine-development
- Tablet model: N/A
- OS version: N/A

### Additional context

N/A

joshxg:
Fixes #2891 

## Change summary

- Updates the selector to create a 2D array in all cases

## Testing

N/A

N/A

### DIFF
--- a/src/selectors/fridge.js
+++ b/src/selectors/fridge.js
@@ -52,7 +52,7 @@ export const selectChunkedTemperatureLogs = createSelector(
   [selectFridgeTemperatureLogsFromDate],
   logs => {
     const { length: numberOfLogs } = logs;
-    const MAX_NUMBER_OF_DATA_POINTS = 30;
+    const MAX_DATA_POINTS = 30;
 
     // If the number of temperature logs is less than the maximum number of data points,
     // then the array does not need to be chunked together - however still need to create
--- a/src/selectors/fridge.js
+++ b/src/selectors/fridge.js
@@ -52,7 +52,12 @@ export const selectChunkedTemperatureLogs = createSelector(
   [selectFridgeTemperatureLogsFromDate],
   logs => {
     const { length: numberOfLogs } = logs;
-    return numberOfLogs < 30 ? logs : chunk(logs, Math.ceil(numberOfLogs / 30));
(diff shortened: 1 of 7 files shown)


src/selectors/fridge.js
```diff
@@ -52,7 +52,12 @@ export const selectChunkedTemperatureLogs = createSelector(
   [selectFridgeTemperatureLogsFromDate],
   logs => {
     const { length: numberOfLogs } = logs;
-    return numberOfLogs < 30 ? logs : chunk(logs, Math.ceil(numberOfLogs / 30));
+    const MAX_NUMBER_OF_DATA_POINTS = 30;
```
wlthomson, line 55:
```suggestion
    const MAX_DATA_POINTS = 30;
```

src/selectors/fridge.js
```diff
@@ -52,7 +52,12 @@ export const selectChunkedTemperatureLogs = createSelector(
...
+
+    // If the number of temperature logs is less than the maximum number of data points,
+    // then the array does not need to be chunked together - however still need to create
+    // a 2D array, so chunk with a chunk size of 1.
+    return chunk(logs, numberOfLogs < MAX_NUMBER_OF_DATA_POINTS ? 1 : Math.ceil(numberOfLogs / 30));
```
wlthomson, line 60:
```suggestion
    return chunk(logs, numberOfLogs < MAX_DATA_POINTS ? 1 : Math.ceil(numberOfLogs / MAX_DATA_POINTS));
```

joshxg:
B-e-a-utiful, thanks

PR #2917, opened 2020-06-08
#2916 fix patient lookup insurance provider logic

### LINKED ISSUE #2916: Sync error on lookup patient with insurance policy
## Describe the bug

When a patient with one or more insurance policies is added via the lookup modal, the following sync error occurs:

### To reproduce

1. Create a patient on desktop with an associated insurance policy.
2. Add the patient on mobile using the lookup modal. 
3. See error.

### Expected behaviour

No error should occur.

### Proposed Solution

Looks like a bug in `InsuranceActions.update` (reminder to self that this logic pretty badly needs to be refactored!).

Quick fix is to add a `getOrCreate('InsuranceProvider', `policyDetails.insuranceProviderId`).

### Version and device info

- App version: v5.0.2.
- Tablet model: Android emulator (Nexus 10).
- OS version: API 28.

### Additional context

Raised by @JonnaQuismundo.

wlthomson:
Fixes #2916.

## Change summary

Fixes insurance provider not being created for patients added via lookup API.

## Testing

- [ ] No sync error occurs when adding a new patient with one or more insurance policies.
  - [ ] Above error does not occur if the policy provider already exists on mobile.
  - [ ] Above error does not occur if the policy provider does not exist on mobile.

### DIFF
--- a/src/actions/InsuranceActions.js
+++ b/src/actions/InsuranceActions.js
@@ -58,6 +58,7 @@ const update = policyDetails => (dispatch, getState) => {
     policyNumberFamily,
     discountRate: policyDiscountRate,
     insuranceProviderId: policyProviderId,
+    insuranceProvider: policyProvider,
     isActive: policyIsActive,
     expiryDate: policyExpiryDate,
     type: policyType,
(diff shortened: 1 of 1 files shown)


src/actions/InsuranceActions.js
```diff
@@ -57,7 +57,7 @@ const update = policyDetails => (dispatch, getState) => {
     policyNumberPerson,
     policyNumberFamily,
     discountRate: policyDiscountRate,
-    insuranceProvider: policyProvider,
+    insuranceProviderId: policyProviderId,
```
joshxg, line 60: Does this break regular updating of insurance policies through dispensing? 

The regular insurance policy updating passes back the 'form object' from the `FormControl` which uses an insurance provider proper?

wlthomson: Good spot. This whole logic needs to be refactored, it's a bad design and an example of too much DRY breaking SRP and SOC all over the place (I can say this because I wrote it!). 

b94a161585a67ab9b62ef93a657703fc1ed34f61.

PR #2921, opened 2020-06-09
#2918 update insurance policy number config

### LINKED ISSUE #2918: Insurance policy family number is not a required field

Desktop logic has been updated such that family policy number is not a required field.

Update mobile insurance policy form for consistency with desktop.

## Implementation

Update `FORM_INPUT_KEYS.]POLICY_NUMBER_FAMILY].isRequired = false`.

N/A.

## Additional context

Raised by @JonnaQuismundo.

wlthomson:
Fixes #2918.

## Change summary

Update family policy number to be optional.

## Testing

- [ ] Can create an insurance policy with a blank family policy number.

N/A.

### DIFF
--- a/src/utilities/formInputConfigs.js
+++ b/src/utilities/formInputConfigs.js
@@ -185,7 +185,7 @@ const FORM_INPUT_CONFIGS = seedObject => ({
     type: FORM_INPUT_TYPES.TEXT,
     initialValue: '',
     key: 'policyNumberFamily',
-    validator: input => input.length > 0 && input.length < 50,
+    validator: input => input.length >= 0 && input.length < 50,
     isRequired: false,
     invalidMessage: formInputStrings.unique_policy,
     label: formInputStrings.family_policy_number,
--- a/src/utilities/formInputConfigs.js
+++ b/src/utilities/formInputConfigs.js
@@ -186,7 +186,7 @@ const FORM_INPUT_CONFIGS = seedObject => ({
     initialValue: '',
     key: 'policyNumberFamily',
     validator: input => input.length > 0 && input.length < 50,
-    isRequired: true,
+    isRequired: false,
     invalidMessage: formInputStrings.unique_policy,
     label: formInputStrings.family_policy_number,
     isEditable: !seedObject,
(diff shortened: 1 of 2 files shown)


src/utilities/formInputConfigs.js
```diff
@@ -186,7 +186,7 @@ const FORM_INPUT_CONFIGS = seedObject => ({
     initialValue: '',
     key: 'policyNumberFamily',
     validator: input => input.length > 0 && input.length < 50,
```
joshxg, line 188: I think the `validator` would need to be changed to be `input.length >= 0 ..`?

wlthomson: Good spot, again :smiley:. This is a sneaky one, because only pops up when you enter something into the field and then clear it.

a68d1dfc.

On the topic, this logic needs to be refactored anyway, I think, as it's currently in a bit of a cludge state (my fault, the need to have a validation function which takes two fields as parameters made me resort to using the reducer for policy fields, but I didn't migrate other logic). 

I think perhaps moving the validation to the `FormReducer` rather than passing the config function around might be more consistent/flexible? The reason I'm putting it off a bit is that it's not a trivial amount of work, and it doesn't really add too much except for tidy things up. Thoughts?

joshxg: 100% needs to be refactored, although I'm not sure about it being in the reducer, personally. 

I really like not having conditional logic in the reducer - i.e. if the action is "UPDATE_FORM_FIELD" then it should update the form field - not check validation and return a state that is `{ error }`, if that makes sense. It probably doesn't really make a difference for our application in the current state, but in the imaginery world where we have a log of actions taken then it would be a lot easier to track bugs. 

Maybe it is the best place to put it, I'm just not 💯 at the mo!

wlthomson: Yeah, I get what you mean... I'm wondering if something like a selector would be more appropriate. Selector can validate the field and could also be used to compose a secondary selector for setting the field error message. It probably wouldn't be trivial in the case of error states dependent on multiple fields, but I think it's probably the most "redux" way of doing it (i.e. validation as derived state).

joshxg: Yeah I think that's a good solution. Just wonder a little on the implementation. I kind of like having sort of.. 'all of the config' in one place - although it is quite restrictive! 

Initial thought it something like having the validation in the same config and used somehow from the selector... bad idea but initial thoughts!

PR #2948, opened 2020-06-15
#2946 update outgoing sync date format

### LINKED ISSUE #2946: Dates are synced to desktop without timezone offset
## Describe the bug

When syncing a transaction from mobile to desktop, the `entry_date` (and other date fields) are incorrectly offset if the mobile device and server are not on the same timezones. Unlikely configuration in practice, but something that could occur (and occur often in testing).

### To reproduce

1. Setup mobile device/emulator on different timezone to server (such that the current day differs).
2. Create and finalise a customer invoice (or other record with date fields).
3. Sync to desktop.
4. View the transaction in desktop. Note that the date has been incorrectly offset. 

### Expected behaviour

Dates should be sent to desktop in ISO format (including timezone offset). 

Dates are synced as strings and parsed on incoming sync using the 4D date constructor. 4D documentation recommends formatting dates in ISO with timezone offset: 

> Note : Starting with 4D v14, it is recommended to use the "YYYY-MM-DDTHH:MM:SSZ" format, compliant with the ISO standard and allowing to express the time zone.

See https://doc.4d.com/4Dv16/4D/16.6/Date.301-4444819.en.html

### Proposed Solution

Update outgoing sync helpers to format dates in ISO format. 

### Version and device info

- App version: v5.0.3
- Tablet model: Android emulator (Nexus 10)
- OS version: API28.

### Additional context

Bug raised by @JonnaQuismundo.

wlthomson:
Fixes #2946.

## Change summary

Updates outgoing date string format to ISO + UTC.

### Setup

- [ ] Setup mobile and desktop devices on different timezones such that the current day of the week differs. 
- [ ] Create a new customer invoice on mobile, finalise and sync to desktop.

### Test cases

- [ ] When the invoice is viewed on desktop, the dates are correct (from desktops point of view). I.e. if the transaction was made a few minutes ago on mobile, the transaction `entry_date` should be only a few minutes ago in desktop time (not a day ahead/behind). 

N/A.

### DIFF
--- a/src/sync/outgoingSyncUtils.js
+++ b/src/sync/outgoingSyncUtils.js
@@ -26,9 +26,8 @@ const bugsnagClient = new BugsnagClient();
 
 const getDateString = date => {
   let returnDate = '0000-00-00';
-  if (date && typeof date === 'object') returnDate = date.toISOString().slice(0, 10);
-
-  return `${returnDate}T00:00:00`;
+  if (date && typeof date === 'object') returnDate = moment(date).toISOString();
+  return returnDate;
 };
 
 function getTimeString(date) {


wlthomson:
> I'm concerned that there was a reason for the original code, which we are now ignoring.
> and the method could be a one-liner 😉
> but should work ok!

Yeah, I couldn't find one, but if this is for a reason then I'm not sure of another way around it (except accepting that dates can get munged on sync) 🤷‍♂️ .

In any case, I think the only way this could cause regressions would be if this was replicable on an actual deployment. For that to happen, the server site and mobile site would have to be far enough apart to be on different timezones (and for the date to change the data would have to be entered within a window where each site was on a different day). I think quite unlikely in practice... which might be why this hasn't come up before?

mark-prins:
yep, couldn't see how it would cause a regression. only if mSupply doesn't cope with ISO formatted dates, and that would be an issue to address rather than working around it in a potentially complicated way.

wlthomson:
> yep, couldn't see how it would cause a regression. only if mSupply doesn't cope with ISO formatted dates, and that would be an issue to address rather than working around it in a potentially complicated way.

Yup, tested working. Desktop directly parses the string using `Date`, which accepts ISO strings (see issue for details).

PR #2986, opened 2020-06-22
#2984 Respect LocationType restrictions

### LINKED ISSUE #2984: Item selection should respect restricted location type
## Describe the bug

When selecting a location for an item batch, the restricted location type is not respected.

### To reproduce
1. Go to anywhere to select a location
2. Click on select location

### Expected behaviour

The locations to select from should be filtered by restricted location type

### Proposed Solution

Filter the list of locations by restricted location type

### Version and device info

- App version: Vaccines development
- Tablet model: N/A
- OS version: N/A

### Additional context

N/A

joshxg:
Fixes #2984 

## Change summary

- Adds restriction of locations for items with a restricted location type

## Testing

N/A

N/A

### DIFF
--- a/src/localization/generalStrings.json
+++ b/src/localization/generalStrings.json
@@ -51,7 +51,7 @@
     "minutes": "minutes",
     "no_location_types": "You don't have any location types. Contact your administrator.",
     "from": "from",
-    "no_locations_for_batch": "There are no locations available for this batch. Contact your administrator."
+    "no_locations_for_batch": "This item is restricted to locations of type {0} and none are available. Contact your administrator."
   },
   "fr": {
     "must_be_a_number_between": "Doit être un nombre entre",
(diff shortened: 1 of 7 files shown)


diegoevangelisti:
@joshxg I am taking a look and testing it right now - Will keep you updated

src/localization/generalStrings.json
```diff
@@ -50,7 +50,8 @@
     "hours": "hours",
     "minutes": "minutes",
     "no_location_types": "You don't have any location types. Contact your administrator.",
-    "from": "from"
+    "from": "from",
+    "no_locations_for_batch": "There are no locations available for this batch. Contact your administrator."
```
diegoevangelisti, line 54: I think this message could be different in those cases in which there are location types but are not showing up because there is restriction regarding the location_type? Mainly to give a better feedback to the user about what is actually happening 😀

joshxg: great suggestion - 
a58dd0b 

Hows that?

diegoevangelisti: Thanks! It looks much better indeed - at least for me

Screen Shot 2020-06-24 at 9 45 27 AM

PR #3002, opened 2020-07-02
#2669 Temperature localizing

### LINKED ISSUE #2669: Create Temperature class for localized temperature formatting

joshxg:
Fixes #2669 

## Change summary

- Adds a Temperature class (exported as a singleton function as that seems to be what the big JS packages seem to do and the JS way to do things..! i.e. `moment`, `axios` etc
- Adds use of `temperature` where temps are used.

## Testing

I think the easiest thing to do is add some code to `componentDidMount` in `mSupplyMobileApp.js`!

import: 

`import temperature from './utilities/temperature';`

```
componentDidMount = () => {
    console.log(
      temperature(4)
        .setMetric('f')
        .format()
    );
  };
```

Doesn't lend itself too well to blackbox user testing!

- Probably going to use something at least similar in CCE app - happy for improvement suggestions!
- Not adding 'localizing' currently (kind of the point of the issue?!) as all our customers seemingly use celsius. This was more of a 'be ready for future requirements' kind of thing (also a bit of fun!) as well as going to need it for CCE so why not do it now. To add changing, would add a `setDefaults` function which would just change the default options to use `f` as a default or whatever!
- Wouldv'e preferred to just use a package, but there doesn't seem to be any for this need - i.e. like `currency.js`/`moment` - maybe I should make it :-)
- Surely other metrics aren't really needed.. right?! Kelvin etc?


wlthomson:
@joshxg I think you can merge directly into `develop` now? 

`v5.1.0` is released and has been merged back into `develop`, so `develop` is now `v6` and `master` is `v5.1`. Would probably keep things simpler, after I made things unnecessarily complex!

joshxg:
@wlthomson cool - just discussing some possible testing with alain at the mo - but will do some mergin' today!

src/utilities/temperature.js
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  format(asMetric) {
+    const { pattern, metric, symbol, negativePattern, formatWithSymbol } = this.options;
+
+    const formatMetric = asMetric || metric;
+    const temperatureValue = this.temperature(formatMetric);
+    const patternToReplace = temperatureValue >= 0 ? pattern : negativePattern;
+
+    if (!Number.isFinite(Number(temperatureValue))) return 'N/A';
```
mark-prins, line 63: this might be overkill: the `N/A` could be either part of the options or set as a const at the top of file.

joshxg: nice! Added as an `invalidPattern` option.. maybe a bit strange? But gives more flexibility if ever for some reason needed - 691bf4f

PR #3007, opened 2020-07-03
#3006 fix insurance policy lookup is active field

### LINKED ISSUE #3006: Insurance policies from lookup API are inactive (v6)
## Describe the bug

All insurance policies from lookup API are inactive.

### To reproduce

1. Add a patient with at least one active policy using lookup API.
2. Dispense to the patient. Navigate to the payment summary page.
3. Note that the active policy is listed as inactive in the policy dropdown.

### Expected behaviour

Active policies should remain active when added to mobile via the lookup API.

### Proposed Solution

Will need to do a fix for `v4` and `v5` majors.

Problem looks like `parseBoolean` is being called on a value which is a boolean (as opposed to a string). Most defensive solution (although a bit hacky), is to make a small fix to `parseBoolean` to ensure booleans are correctly parsed (currently `parseBoolean(true)` returns `false`).

### Version and device info

- App version: v5.1.0.
- Tablet model: Nexus 10 (emulated).
- OS version: API 28.

### Additional context

N/A.

wlthomson:
Fixes #3006.

## Change summary

Minor fix to boolean parse helper.

## Testing

- [ ] Active policies for patients looked up via lookup API are displayed as active.
- [ ] Inactive policies for patients looked up via lookup API are displayed as inactive.

N/A.

### DIFF
--- a/src/database/utilities/parsers.js
+++ b/src/database/utilities/parsers.js
@@ -5,13 +5,11 @@
 
 /**
  * Returns the boolean string as a boolean (false if none passed)
- * @param  {string} numberString The string to convert to a boolean
+ *
+ * @param  {string}  numberString  The string to convert to a boolean
  * @return {boolean}               The boolean representation of the string
  */
-export const parseBoolean = booleanString => {
-  const trueStrings = ['true', 'True', 'TRUE'];
-  return trueStrings.includes(booleanString);
-};
+export const parseBoolean = booleanString => booleanString.toString().toLowerCase() === 'true';
 
 /**
  * Return a Date object representing the given date, time.
--- a/src/sync/lookupApiUtils.js
+++ b/src/sync/lookupApiUtils.js
@@ -120,8 +120,8 @@ const processInsuranceResponse = response =>
     })
   );
 
-export const processPatientResponse = response => {
-  const patientData = response.map(
+export const processPatientResponse = response =>
+  response.map(
     ({
       ID: id,
       name,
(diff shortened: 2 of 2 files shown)


src/database/utilities/parsers.js
```diff
@@ -5,13 +5,11 @@
...
- * @param  {string} numberString The string to convert to a boolean
+ *
+ * @param  {string}  numberString  The string to convert to a boolean
  * @return {boolean}               The boolean representation of the string
  */
-export const parseBoolean = booleanString => {
-  const trueStrings = ['true', 'True', 'TRUE'];
-  return trueStrings.includes(booleanString);
-};
+export const parseBoolean = booleanString => booleanString.toString().toLowerCase() === 'true';
```
joshxg, line 12: Looks like this might be getting called with `undefined` or `null`s some might be best to be even more defensive : `booleanString?.toString?.()...`

src/sync/lookupApiUtils.js
```diff
@@ -120,8 +120,8 @@ const processInsuranceResponse = response =>
     })
   );
 
-export const processPatientResponse = response => {
-  const patientData = response.map(
+export const processPatientResponse = response =>
```
joshxg, line 123: refactor???? 😠

src/database/utilities/parsers.js
wlthomson: 4f4a5bd17844653800c02cdf9e2d140a4c308c5a.

src/sync/lookupApiUtils.js
wlthomson: run

wlthomson:
> it breaks still if you pass `{toString: "AHHHH"}` but I that's not going to happen 🤣

why

PR #3013, opened 2020-07-07
#3011 temperature syncing and saving

### LINKED ISSUE #3011: New temperatures should be recorded from the last temperature synced
## Describe the bug

When syncing temperatures, the time set should be counted from the last temperature which was synced

### To reproduce

Internal code error here: 

https://github.com/openmsupply/mobile/blob/9f5deeaa04f9084958114525607c2850e3c61258/src/actions/TemperatureSyncActions.js#L108-L118

- Rather than counting back from 'now' find either the most recent `SensorLog` or `TemperatureLog` record (which ever is most recent).
- Then, calculate how many records should be read from the sensor from the time now, until that point and only save those records.

### Expected behaviour

As above

### Proposed Solution

As above

### Version and device info

- App version: Vaccine feature branch
- Tablet model: N/A
- OS version: N/A

### Additional context

Reasoning: If there is a crash between syncing temperatures and resetting the sensor, we might still have logs on a sensor. If we then sync again, we will save all the records.

i.e: 
- Sync 1000 logs, save them, crash before resetting (or fail to reset)
- Next sync, sync 1010 logs, save them. We're now well into the future!

If we change to the above behaviour:
- Sync 1000 logs, save them, crash before resetting (or fail to reset)
- Next sync, sync 1010 logs, save only ~10. Better!

Of course there is still the problem where the sensor itself has some undefined/incorrect behaviour - i.e.
- Sync 1000 logs, save them, crash before resetting (or fail to reset)
- Sensor doesn't keep recording temps
- Next sync, sync 1010 logs, save only ~10. But they're not the right times!

Not ideal but much better.

joshxg:
Fixes #3011 

## Change summary
- Adds corrected logic to 'count' from the most recent log synced, rather than counting back from the current time 

## Testing

N/A

N/A

### DIFF
--- a/src/actions/PermissionActions.js
+++ b/src/actions/PermissionActions.js
@@ -23,7 +23,8 @@ const setWriteStorage = status => ({
 });
 
 const requestBluetooth = () => async dispatch => {
-  const result = await BluetoothStatus.enable();
+  BluetoothStatus.enable();
+  const result = await BluetoothStatus.state();
 
   dispatch(setBluetooth(result));
 };
--- a/src/actions/TemperatureSyncActions.js
+++ b/src/actions/TemperatureSyncActions.js
@@ -138,10 +138,6 @@ const saveLogs = (logData, sensor) => async dispatch => {
   const tempLogIsEarlier = mostRecentTempLogTime.isBefore(mostRecentSensorLogTime);
   const mostRecentLogTime = tempLogIsEarlier ? mostRecentSensorLogTime : mostRecentTempLogTime;
 
-  // let lookbackTime = moment(new Date(null));
-  // if (sensorLogs.length) lookbackTime = moment(sensorLogs.max('timestamp'));
-  // else if (tempLogs.length) lookbackTime = moment(tempLogs.max('timestamp'));
-
   // Calculate the number of LogIntervals it has been since the last sync/last log was
   // created and now. Then save that number of data points that were synced from
   // the bluetooth sensor. This ensures we are using the correct logs from the sensor
--- a/src/actions/TemperatureSyncActions.js
+++ b/src/actions/TemperatureSyncActions.js
@@ -6,11 +6,12 @@
 
 import { ToastAndroid } from 'react-native';
 import { generateUUID } from 'react-native-database';
+import moment from 'moment';
 
 import { UIDatabase } from '../database';
 import { Sensor } from '../database/DataTypes';
 import { createRecord } from '../database/utilities';
-import { MILLISECONDS } from '../utilities';
+
 import { chunk } from '../utilities/chunk';
 
 import { syncStrings, vaccineStrings } from '../localization';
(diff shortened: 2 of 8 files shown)


src/actions/PermissionActions.js
```diff
@@ -23,7 +23,8 @@ const setWriteStorage = status => ({
 });
 
 const requestBluetooth = () => async dispatch => {
-  const result = await BluetoothStatus.enable();
+  BluetoothStatus.enable();
+  const result = await BluetoothStatus.state();
```
joshxg, line 27: bug fix

src/actions/TemperatureSyncActions.js
```diff
@@ -119,13 +120,40 @@ const downloadLogsComplete = () => ({
 });
 
 const saveLogs = (logData, sensor) => async dispatch => {
+  const { length: numberOfLogs = 0 } = logData;
```
joshxg, line 123: hooopefully my comments help with this. Essentially:

- find the most recent log that has been saved. 
- work out how long it has been since then and now in terms of the number of log intervals of the sensor. 
- count back that many logs from the the overall logs downloaded from the sensor

We were: syncing the logs and counting back from the current time, then deleting the logs. But, if the delete fails - 💣 oops

src/actions/TemperatureSyncActions.js
```diff
@@ -303,17 +332,12 @@ const syncTemperatures = () => async (dispatch, getState) => {
     const downloadedLogsResult = await dispatch(downloadLogs(sensor));
 
     if (downloadedLogsResult) {
+      await dispatch(saveLogs(downloadedLogsResult, sensor));
+      await dispatch(createTemperatureLogs(sensor));
```
joshxg, line 336: changing the order - just save before deleting now

src/selectors/fridge.js
```diff
@@ -5,7 +5,7 @@
 import moment from 'moment';
 import { createSelector } from 'reselect';
 import { UIDatabase } from '../database';
-import chunk from '../utilities';
+import { chunk } from '../utilities';
```
joshxg, line 8: bug fix..

src/widgets/modals/DataTablePageModal.js
```diff
@@ -227,7 +227,8 @@ const DataTablePageModalComponent = ({
         );
       case MODAL_KEYS.SELECT_LOCATION: {
         const { currentLocationName } = currentValue;
-        const { id, description } = currentValue.restrictedLocationType(UIDatabase) ?? {};
+        const { id = '', description = '' } =
+          currentValue?.restrictedLocationType?.(UIDatabase) ?? {};
```
joshxg, line 231: more bug fixes

src/actions/TemperatureSyncActions.js
```diff
@@ -119,13 +120,40 @@ const downloadLogsComplete = () => ({
...
+  const { location, logInterval } = sensor;
+  const { id: locationID } = location;
+
+  // SensorLog are saved directly from a sensor as five-minute logs. A TemperatureLog is 30 minutes
+  // worth of SensorLog records. If there isn't enough SensorLogs to aggregate into a
+  // TemperatureLog, there may be some 'left over'. Start counting from whichever is sooner,
+  // the most recent SensorLog or the most recent TemperatureLog.
+  const sensorLogs = UIDatabase.objects('SensorLog').filtered('location.id == $0', locationID);
+  const tempLogs = UIDatabase.objects('TemperatureLog').filtered('location.id == $0', locationID);
+  const mostRecentSensorLogTime = moment(sensorLogs.max('timestamp') ?? timeNow);
```
joshxg, line 136: `max` returns undefined if there are no logs

PR #3030, opened 2020-07-15
#3015 fix insurance policy outgoing sync formatting

### LINKED ISSUE #3015: Insurance: "-" should not be added in absence of Family Policy Number
## Describe the bug

"-" should not be added in absence of Family Policy Number

### To reproduce

1. Go to Mobile apk -> create a Prescription
2. Add an insurance such that Family Policy number is blank/empty
3. Finalize and sync
4. GO to Desktop store. Check the following prescription
4. See error. "-" is shown as prefix to Personal Policy Number
Screen Shot 2020-07-10 at 12 02 12 PM

### Expected behaviour

"-" should be removed in absence of Family Policy Number

### Proposed Solution

### Version and device info

- App version: 5.1.1
- Tablet model:
- OS version:

### Additional context

arises while testing #2918

wlthomson:
Fixes #3015.

## Change summary

Fixes formatting of policy numbers synced from mobile to desktop.

## Testing

- [ ] Policies with personal numbers are correctly formatted after being synced from mobile to desktop  (`familyNumber-personalNumber`).
- [ ] Policies without personal numbers are correctly formatted after being synced from mobile to desktop  (`familyNumber`).

N/A.

### DIFF
--- a/src/database/DataTypes/InsurancePolicy.js
+++ b/src/database/DataTypes/InsurancePolicy.js
@@ -2,7 +2,8 @@ import Realm from 'realm';
 
 export class InsurancePolicy extends Realm.Object {
   get policyNumber() {
-    return `${this.policyNumberPerson} ${this.policyNumberFamily}`;
+    if (!this.policyNumberPerson) return this.policyNumberFamily;
+    return `${this.policyNumberFamily}-${this.policyNumberPerson}`;
   }
 }
 
--- a/src/database/DataTypes/InsurancePolicy.js
+++ b/src/database/DataTypes/InsurancePolicy.js
@@ -13,7 +13,7 @@ InsurancePolicy.schema = {
   properties: {
     id: 'string',
     policyNumberFamily: 'string',
-    policyNumberPerson: 'string',
+    policyNumberPerson: { type: 'string', default: '' },
     type: 'string',
     discountRate: 'double',
     isActive: { type: 'bool', default: true },
--- a/src/database/schema.js
+++ b/src/database/schema.js
@@ -190,7 +190,7 @@ export const schema = {
     User,
     Currency,
   ],
-  schemaVersion: 16,
+  schemaVersion: 17,
 };
 
 export default schema;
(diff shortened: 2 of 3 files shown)


src/database/DataTypes/InsurancePolicy.js
```diff
@@ -2,7 +2,8 @@ import Realm from 'realm';
 
 export class InsurancePolicy extends Realm.Object {
   get policyNumber() {
-    return `${this.policyNumberPerson} ${this.policyNumberFamily}`;
+    if (!this.policyNumberPerson) return this.policyNumberFamily;
```
joshxg, line 5: both values are required in the schema aren't they? 

https://github.com/openmsupply/mobile/blob/d515a274516ae0b1b434add43eb967e4d94f8ad4/src/database/DataTypes/InsurancePolicy.js#L15-L16

🤔

wlthomson: Yeah, I believe that just means it can't be set to `null` or `undefined`? As it's a primitive I think it makes more sense to have it initialised to a falsey but non-null value? 

To make it more explicit can do:

```
policyNumberPerson: { type: 'string', default: '' };
```

but I think the above is equivalent to the existing declaration?

joshxg: I didn't know so I did a quick test with:

```
UIDatabase.write(() => {
      const insuranceProvider = UIDatabase.update('InsuranceProvider', {
        id: '1',
      });

      UIDatabase.update('InsurancePolicy', { id: '1', insuranceProvider });
    });
```

And got the error "Missing value for property `InsurancePolicy.policyNumberFamily`" - so I believe you need to explicitly set the default. Also doesn't say anything in the docs about defaults being set for primitives automatically: https://realm.io/docs/javascript/latest/#default-property-values

wlthomson: Oops, looks like I misinterpreted the examples in the [docs](https://realm.io/docs/javascript/latest/#models) (I guess that includes implicitly on construction).

eeb4b8c36cbd900662740f69499455cbffe7d945.

PR #3036, opened 2020-07-21
#3034 update period date formatting

### LINKED ISSUE #3034: Format and value of period dates
## Describe the bug

The start and end date of periods, shown when creating a new requisition, are both incorrect and in US date format.

### To reproduce

1. Change regional settings on the tablet, so that the timezone is non-NZ ( I selected the region of Côte d'Ivoire => GMT )
2. Create a new program-based supplier requisition
3. When you reach the `period` selection, observe the dates and compare to mSupply

### Expected behaviour

- Date format should be consistent. Currently, mSupply shows DD/MM/YYYY and mobile shows MM/DD/YYYY
- Dates are affected by the timezone difference between server and mobile

### Additional context

Screen Shot 2020-07-20 at 5 29 13 PM

wlthomson:
Fixes #3034.

## Change summary

Update period date formatting for consistency with desktop.

## Testing

- [ ] Period start and end date is formatted as `DD/MM/YY` when creating a new program requisition.
- [ ] Period start and end date is formatted as `DD/MM/YY` when viewing a program requisition.

N/A.

### DIFF
--- a/src/database/DataTypes/Period.js
+++ b/src/database/DataTypes/Period.js
@@ -4,6 +4,7 @@
  */
 
 import Realm from 'realm';
+import moment from 'moment';
 import { getIndicatorValuesByPeriod, getIndicatorsByValues } from '../utilities/getIndicatorData';
 
 /**
@@ -48,7 +49,9 @@ export class Period extends Realm.Object {
   }
 
   toString() {
-    return `${this.startDate.toLocaleDateString()} - ${this.endDate.toLocaleDateString()}`;
+    const startDate = moment(this.startDate).format('DD/MM/YY');
+    const endDate = moment(this.endDate).format('DD/MM/YY');
+    return `${startDate} - ${endDate}`;
   }
 }
 


src/database/DataTypes/Period.js
```diff
@@ -48,7 +49,9 @@ export class Period extends Realm.Object {
   }
 
   toString() {
-    return `${this.startDate.toLocaleDateString()} - ${this.endDate.toLocaleDateString()}`;
+    const startDate = moment(this.startDate).format('DD/MM/YY');
```
mark-prins, line 52: I went with a 
```
const DATE_FORMAT = 'DD/MM/YY';

export const formatDate = date => moment(date).format(DATE_FORMAT);
```

in `localization/utilities` and used that. Possibly overkill, but we may need date formatting elsewhere and may need to have a localisation friendly format?

wlthomson: Fair enough. There's a few other hardcoded formatting strings so I'll make a new issue for this 👍 .

PR #3054, opened 2020-07-28
#3019 update use fetch hook

### LINKED ISSUE #3019: Update useFetch hook

Update `useFetch` hook to improve existing lookup logic.

Refactor `useFetch` and calling code according to SOLID principles.

## Implementation

- Update hook to return function for fetching remote data.
- Improve modularity by using DI for error handling and response mappers.

Leave as is.

## Additional context

N/A.

wlthomson:
Fixes #3019.

## Change summary

Long overdue tidy up of the `SearchForm` component 😨, which mainly involves a refactor of the `useFetch` hook. 

## Testing

- [ ] Loading animation is displayed on search button press.
- [ ] Data is populated correctly with data on successful fetch.
- [ ] Error modal is displayed on error (timeout, network error, no results found).
- [ ] If fetch is in progress, pressing the search button does not have any effect (no new request, no errors!).

Revisiting the UI has reminded there is still quite a bit of work which could be done to improve it (and a few design decisions I would like to change!), REVIEWER: forgive the UI, comments on functionality most relevant here!


src/localization/vaccineStrings.json
```diff
@@ -14,6 +14,14 @@
   "gil": {},
   "tl": {},
   "la": {},
+  "my": {
```
wlthomson, line 17: Hmm, not sure what's going on here! (see https://github.com/openmsupply/mobile/pull/2993/commits/605b417c82a1a13840ddddaa40f5fdeffe0597a5).

src/hooks/useFetch.js
```diff
@@ -4,103 +4,91 @@
...
-}
-
-const getAuthorizationHeader = () => {
-  const username = UIDatabase.getSetting(SYNC_SITE_NAME);
-  const password = UIDatabase.getSetting(SYNC_SITE_PASSWORD_HASH);
-  return getAuthHeader(username, password);
-};
-
-const isValidError = error => Object.values(ERROR_CODES).includes(error?.message);
+const doIf = (ifThis, doThis, ...withThese) => (ifThis() ? doThis(...withThese) : null);
```
wlthomson, line 8: shot

src/hooks/useFetch.js
```diff
@@ -4,103 +4,91 @@
...
+      const processedError = await errorHandler(err);
+      doIf(isMounted, setError, processedError);
+    };
+
+    const beforeFetch = () => {
+      _isBlocked.current = true;
+      // eslint-disable-next-line no-undef
+      _controller.current = new AbortController();
+      _timer.current = setTimeout(() => {
+        if (_controller.current) _controller.current.abort();
```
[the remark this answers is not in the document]
wlthomson: `reset` will clear the timeout, which should hopefully prevent this (but still potential race conditions everywhere!)

src/hooks/useFetch.js
```diff
@@ -4,103 +4,91 @@
...
+    const afterFetch = () => {
+      _isBlocked.current = false;
+      clearTimeout(_timer.current);
+      setIsLoading(false);
+    };
+
+    const doFetch = async () => {
+      try {
+        doIf(isMountedAndUnblocked, beforeFetch);
+        const res = await doIf(isMounted, tryFetch);
```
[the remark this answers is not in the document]
wlthomson: As `beforeFetch` sets `isBlocked`, this can't be wrapped in that conditional, too! The idea would be that we shouldn't get to this point if its blocked (as `doFetch` shouldn't be executed, the extra conditional on `beforeFetch` is just added protection against race conditions!). In any case, on revisiting this it's quite clear that the `isBlocked` logic has become really convoluted, so I think it might be better to loosen some of the protections against the more improbable race conditions!

src/hooks/useFetch.js
```diff
@@ -4,103 +4,91 @@
...
+        await doIf(isMounted, onResponse, res);
```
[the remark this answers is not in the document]
wlthomson: Another race condition thing (if this code gets execute before the `fetch` request registers the `abort()`).

src/hooks/useFetch.js
```diff
@@ -4,103 +4,91 @@
...
+        await doIf(isMounted, onError(err));
+      } finally {
+        doIf(isMounted, afterFetch);
+      }
+    };
+
+    doIf(isMountedAndUnblocked, doFetch);
+  };
+
+  return { fetch: _fetch, reset, isLoading, response, error };
```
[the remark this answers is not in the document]
wlthomson: Yeah, the user needs some way of "clearing" the error (or response), so `reset` is exposed for that purpose.

wlthomson:
@joshxg Updated with relaxed race condition conditionals, should be a bit clearer now!

src/hooks/useFetch.js
```diff
@@ -4,103 +4,91 @@
...
-        isMounted.current = false;
-        clearTimeout(timer);
-        abortController.abort();
-      };
-    }
+    _isMounted.current = true;
     return () => {
-      isMounted.current = false;
+      _isMounted.current = false;
+      reset();
```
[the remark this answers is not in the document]
wlthomson: Resolving this so nobody can see evidence of how stupid that was...

src/hooks/useFetch.js
```diff
@@ -4,103 +4,91 @@
...
+  const _isBlocked = React.useRef(false);
+
+  const [isLoading, setIsLoading] = React.useState(false);
+  const [response, setResponse] = React.useState(null);
   const [error, setError] = React.useState(null);
-  const isMounted = React.useRef(true);
 
-  /* eslint-disable no-undef */
-  const abortController = new AbortController();
+  const reset = () => {
```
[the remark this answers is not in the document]
wlthomson: This should be clearer now. Idea for `isBlocked` was to basically prevent race conditions or multiple fetch calls from causing errors/unexpected behaviour. On revisiting the code, I think I went a bit OTT with this, to the extent that it was duplicating some stuff that was already handled by `isMounted` and became quite confusing and probably risked introducing errors as much as preventing them. 

After refactor, `isBlocked` has been reduced to a simple semaphore to prevent multiple fetches being dispatched at once. It has a similar function to `isLoading`, but at the logic rather than UI level (can think of it sort of like getting a lock on fetching data). Main reason for it being a ref is to make it independent from the render cycle, just in case of race conditions or anything like that.

src/hooks/useFetch.js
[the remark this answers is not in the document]
wlthomson: Update: after refactor, this behaviour is pretty much the same, but might not be instantly obvious. On the event that a fetch is in progress and the component is unmounted:

1. `_isMounted.current` is set to `false`.
2. `_controller.current.abort()` is called.
3. `doFetch` catches an error (as the `fetch` throws an error due to the abort signal).
4. `onError` is run, but no state is updated as `setError` is wrapped in a `doIf(isMounted, ...)`.
5. `afterFetch` is executed, which sets `_isBlocked` to `false` and clears the timeout (the `isLoading` state is not updated for same reason as `setError` above).

Hopefully that clarifies, feedback appreciated though!

src/hooks/useFetch.js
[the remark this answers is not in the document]
joshxg: nit: it's now `resetState` 🤣

src/hooks/useFetch.js
```diff
@@ -4,103 +4,86 @@
...
+  const [isLoading, _setIsLoading] = React.useState(false);
+  const [response, _setResponse] = React.useState(null);
+  const [error, _setError] = React.useState(null);
+
+  const isMounted = () => _isMounted.current;
+  const isUnblocked = () => !_isBlocked.current;
+
+  const setIsLoading = _isLoading => doIf(isMounted, _setIsLoading, _isLoading);
+  const setResponse = _response => doIf(isMounted, _setResponse, _response);
+  const setError = _error => doIf(isMounted, _setError, _error);
```
joshxg, line 32: now that we're wrapping the state setters, their references are unstable and may cause unnecessary/hard to find re-renders - Maybe `doIf` becomes a factory for producing wrapped memoized callbacks? 🤣 `doIfMountedFactory` 🤣  


But hmm.. maybe we're holding onto the `doIf` now just because it's funny?

`const setIsLoading = _isLoading => doIf(isMounted, _setIsLoading, _isLoading);`

```
const setIsLoading = _isLoading => {
	if (isMounted) _setIsLoading(_isLoading)
};
```

src/hooks/useFetch.js
```diff
@@ -4,103 +4,86 @@
...
+    _isMounted.current = true;
     return () => {
-      isMounted.current = false;
+      _isMounted.current = false;
+      if (_controller.current) _controller.current.abort();
+    };
+  }, []);
+
+  const _fetch = async (path, init, opts) => {
+    const { responseHandler = res => res, errorHandler = err => err, timeout = 10000 } = opts;
```
joshxg, line 49: nit: I like having `opts = defaultOpts` here.

In `useFetch = (url, defaultOpts = DEFAULT_OPTIONS)` and having `DEFAULT_OPTIONS = { responseHandler = res => res, errorHandler = err => err, timeout = 10000 }` set outside the function?

joshxg: Also a bit clearer if these default functions are `async`?

src/hooks/useFetch.js
```diff
@@ -4,103 +4,86 @@
...
+        const res = await tryFetch();
+        await onResponse(res);
+      } catch (err) {
+        await onError(err);
+      } finally {
+        afterFetch();
+      }
+    };
+
+    doIf(isUnblocked, doFetch);
```
joshxg, line 85: This is way clearer now, nice! My personal taste (and makes no difference whatsoever), just for the different perspective would be to not use the functions `beforeFetch`, `tryFetch` and `AfterFetch`, `doFetch` and to just have the code be written sequentially. Maybe it's a bit ugly but it's just so easy to read. At the moment I think I read it more similarly to not using any `async/await`, and just using `then` calls

src/sync/lookupApiUtils.js
```diff
@@ -32,6 +37,18 @@ const PARAMETERS = {
...
+
+class BugsnagError extends Error {
+  constructor(message, data, ...args) {
+    super(message, ...args);
+    bugsnagClient.notify(this, report => {
+      report.errorMessage = message;
+      report.metadata = data;
+    });
+  }
+}
```
joshxg, line 50: Nice. Do you think `notify` shouldn't be in the constructor? What's the point of this being a class?

src/hooks/useFetch.js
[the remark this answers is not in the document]
joshxg: Cool. I guess I don't really know why they need to 'clear' the response or error. Could you just have in `beforeFetch` a clear of the error/response? anyway doesn't matter!

src/sync/lookupApiUtils.js
wlthomson: Just for consistency with other "native" errors: see "Custom Error Types" [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error).

src/hooks/useFetch.js
wlthomson: No worries, I'm not too bothered on `await` v. `then`, generally whichever is cleanest! I think this should be resolved now anyways with the simplication of the API to make the handlers both synchronous.

src/hooks/useFetch.js
wlthomson: Both nice and sync now 😄.

src/hooks/useFetch.js
wlthomson: I think the `useDoIf` and `useProtectedState` hooks should cover this (using `useCallback` as we talked about earlier). Have tested with unmounting etc. and is working (but always a risk of weird cases in the wild when we are jestless!).

src/hooks/useDoIf.js
```diff
@@ -0,0 +1,20 @@
...
+const doIf = (doThis, ifThis) => {
+  if (ifThis()) return doThis();
+  return null;
+};
+
+export const useDoIf = (doThis, ifThis) =>
+  React.useCallback(
+    (...rest) => {
+      doIf(() => doThis(...rest), ifThis);
+      return null;
```
joshxg, line 17: just `return doIf(..` I think?

src/sync/lookupApiUtils.js
joshxg: Ah yup. still not sure if the notify needs to be in the constructor rather than 

`const newBugsnag = new BugsnagError(error)`
`newBugsnag.notify()`

Two actions so making it explicit in two calls - personal preference

joshxg: I just realised it would be more like

`const newBugsnag = new BugsnagError(error)`
`const newBugsang.notify()`
`throw newBugsnag`

Basically, I am questioning the need for this to be an error type rather than a creating a `notify` wrapper:

`notifyWrapper(error)`
`throw error`

but again I think personal preference

src/sync/lookupApiUtils.js
```diff
@@ -120,8 +135,26 @@ const processInsuranceResponse = response =>
...
+    if (responseError) throw new BugsnagError(responseError, { url, headers });
+    if (!json.length) throw new Error(ERROR_CODES.EMPTY_RESPONSE);
+    return json;
+  }
+  switch (status) {
+    case 400:
+    default:
+      throw new Error(ERROR_CODES.CONNECTION_FAILURE);
+    case 401:
+      throw new BugsnagError(ERROR_CODES.INVALID_PASSWORD, { url, headers });
```
joshxg, line 151: Do you think putting this switch into the error handler passed into the hook could be a good use? Rather than using the response handler for both error logic and ok logic - I always find error handling annoying 🤣 (no change just interested!)

src/sync/lookupApiUtils.js
mark-prins: Hm. I'm not seeing the raising of an error in the constructor of "Custom Error Types" and so agree with Josh's interpretation soz.

If the error is notified in the constructor, then I think the naming should reflect that
`BugsnagError` => ctor returns an object
`BugsnagNotification` => ctor could create an object and raise a notification
perhaps.

wlthomson: Just succinctly, it's just polymorphism, i.e. an `Error` should behave like an error (so having a `notify` method on an error doesn't seem like a good idea). A wrapper seems a bit overkill, as the basic idea here is to be able to throw an error as you would normally, but also have it log to bugsnag (so if you wanted to have a specific bugsnag error, you could then inherit from that). I think in this case inheritance is better than composition (considering that we are talking about OOP). I would agree it's mainly personal preference, but I think it's pretty consistent with standard OOP practices. 

@mark-prins Sorry, not following! The constructor doesn't raise any errors, the object is an error itself (so when it's raised, it also logs upstream by reporting the error to bugsnag). It's polymorphism, pretty much - it walks like an error, it talks like an error, so it is an error,,, but it also happens to log to bugsnag!

mark-prins: perhaps it's my wording. by 'raise an error' in the ctor I mean the call to `bugsnagClient.notify` (i.e. logging upstream). 

when you say 'raise an error' - do you mean create the error object? by which - calling `notify` in the ctor is all good. I was thinking that the ctor should just construct the error object. The `notify()` should be done when the error is `throw`n - that's how I thought the process usually runs.

all of which sounds exactly like your response to @joshxg - so perhaps we're saying the same thing?

src/sync/lookupApiUtils.js
wlthomson: I thought about this a bit and did think about it, but since this is a switch on a non-error response, it doesn't really make sense to put it in an error handler (no errors have been thrown yet). 

On revisiting it, I think a possible idea would be to have a custom `HTTPError` and do something like:

```
const processResponse = response => {
  const { ok, status, ... } = response;
  if (!ok) throw HTTPError(status);
  ...
}

const processError = error => {
  if (error instanceOf HTTPError) {
    switch (error.status) {
       ...
    }
  }
}
```

src/sync/lookupApiUtils.js
wlthomson: @mark-prins I agree with that, and I think we are all basically saying the same thing, which is that doing "logic" on instantiation seems wrong, so the notification should be decoupled from the constructor. Ideally, the error should notify when it is thrown, but the only way to really do that is when it is instantiated (as far as I know!). If there was a way to fire the notify on `throw` (dynamically, i.e. not explicitly in the calling code), that would be ideal, but I don't think there is? 

Maybe helps to clarify that the main reason for the custom error is to avoid having to have a conditional on certain errors (before they are thrown or after they are caught). The goal is for the error to be thrown just like any other error, but also to notify in the process (without the calling code having to call any methods that only exist on that subclass).

Does that make sense?

src/hooks/useDoIf.js
wlthomson: hmm_2

wlthomson: a8715ca1b42b5d4dfe88f5bd02cc51d8706a8ebc.

src/sync/lookupApiUtils.js
mark-prins: yep. all good - I don't know of any way to extend `throw` so this is probably as good as it gets

Issue #3085, opened 2020-08-12
Stock transfer is received without trans lines


anildahalsussol:
## Describe the bug

If we create a stock transfer from one tablet to another. Another Tablet gets transaction without trans lines.

### To reproduce

1. Create a Customer invoice on a tablet.
2. Then directly Finalised it. (Note need to finalize directly)
3. Sync finalized Customer invoice.
4. Sync in another tablet and check the Supplier Invoice.

### Expected behaviour

SI with trans-lines needs to be synced to the tablet.

### Proposed Solution

As discussed if we can change the order of syncing transaction if it's directly finalized. Maybe the first trans_line then transact. 

### Version and device info

- App version: 5.1.1
- Tablet model: Lenovo

### Additional context

Problem faced in MAM.

bijaySussol:
@joshxg @wlthomson @anildahalsussol : 

Could this be related with the item not being visible to an another tablet store?
This can happens on desktop, sender can do a stock transfer to another store even if that transferring item is not visible to another store - and the receiver store end up with blank lines against supplier invoice.
Just thought to share - things might be different - as 2 tablet access the 2 different store chances of having item not visible is high on either of the store and desktop does not bother to check the visibility.

joshxg:
Fixed by #3088

PR #3092, opened 2020-08-14
#3084 fix requisition sync no lines

### LINKED ISSUE #3084: Requisition is received without requisition lines
## Describe the bug

If we create request requisition from one tablet to another. Another Tablet gets requisition without requisition lines.

### To reproduce

1. Create Supplier requisition in a tablet.
2. Then directly Finalised it. (Note need to finalize directly)
3. Sync the finalized Supplier req.
4. Sync in another tablet and check customer requisition.

### Expected behaviour

Response Requisition with requisition lines needs to be synced to the tablet.

### Proposed Solution

As discussed if we can change the order of syncing requisition if it's directly finalized. May be the first requisition_line then requisition. 

### Version and device info

- App version: 5.1.1
- Tablet model: Lenovo

wlthomson:
Fixes #3084.

## Change summary

Phew, took a while to track this down... looks like a combination of two bugs!

- Fixes duplicate sync out records due to incorrect duplicate matching logic.
- Updates sync out queue ordering to ensure child records are synced before parents.

## Testing

- [ ] Bug described in #3084 cannot be replicated.
- [ ] Bug described in #3085 cannot be replicated.

@anildahalsussol Replicated the above bugs and this fixes it for me. Can you check that it fixes the original bug before merging (or do you need an `rc` build)?

Hoping this might fix a few of the bugsnag errors too (e.g. any cleanup logic based on the expectation that only one sync out record exists per `recordId` will have been broken!).

### DIFF
--- a/src/sync/SyncQueue.js
+++ b/src/sync/SyncQueue.js
@@ -103,9 +103,9 @@ export class SyncQueue {
           };
 
           if (this.isValidSyncOutRecord(syncOutRecord)) {
-            const existingSyncOutRecord = this.database.get('SyncOut', recordId);
+            const existingSyncOutRecord = this.database.get('SyncOut', recordId, 'recordId');
             if (existingSyncOutRecord) {
-              this.database.save('SyncOut', { ...existingSyncOutRecord, ...syncOutRecord });
+              this.database.save('SyncOut', { id: existingSyncOutRecord.id, ...syncOutRecord });
             } else {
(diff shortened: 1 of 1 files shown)


anildahalsussol:
@wlthomson @joshxg : I will need apk. My emulator has a problem again so I cannot build.

anildahalsussol:
@wlthomson @joshxg : Working perfectly... tested

PR #3101, opened 2020-08-16
#3099 Fix deleted stocktake in finalise state

### LINKED ISSUE #3099: Removing stock take > App crashed
## Describe the bug

Removing the stock take app crashed

### To reproduce

1. Create new stock take
2. Choose any 3 lines
3. Hit **Create**
4. Put the **Actual Quantity**
5. Tap on **Manage stocktake**
6. Sync it
7. Go back to stock take list
8. from the stock take check **Remove** against that stock take
9. Select that row take you to stock take input window
10. Tap on **Manage stocktake**
11. Unselect one line against from the existing stock take
12. Sync it
13. Go back
14. Click on **Remove** from the alert box which says: **Are you sure you want to remove these items?**
15. Wait > See app crashed

### Expected behaviour

No App crashing

### Proposed Solution

### Version and device info

- App version: v5.1.1
- Tablet model: Samsung Tablet Galaxy Tab A (SM-T515)
- OS version: Android 9

### Additional context

See video:
https://www.dropbox.com/s/x2wzelbyld2vtcc/video_2020-08-14_18-13-44.mp4?dl=0

wlthomson:
Fixes #3099.

After a lot of dead ends (most caused by logging!), it looks like this is the culprit (@bijaySussol could you confirm?).

## Change summary

- Update back navigation to always reset finalise item.

## Testing

- [ ] Navigating back from `StocktakeManagePage` to `StocktakesPage` and deleting the previously viewed stocktake does not cause an error.

From looking at the logic, it looks like it's safe to `resetFinaliseItem()` on all back navigation... but if I'm missing something, may need to revert back to a conditional (with an added edge case to handle this error)?

### DIFF
--- a/src/navigation/actions.js
+++ b/src/navigation/actions.js
@@ -56,7 +56,6 @@ export const goBack = () => dispatch => {
 
       const cleanUp = () => {
         dispatch(PrescriptionActions.deletePrescription());
-        dispatch(FinaliseActions.resetFinaliseItem());
       };
 
       batch(() => {
--- a/src/navigation/actions.js
+++ b/src/navigation/actions.js
@@ -10,7 +10,7 @@ import { NavigationActions, StackActions } from '@react-navigation/core';
 import { UIDatabase } from '../database';
 import { createRecord } from '../database/utilities';
 import { navStrings } from '../localization';
-import { ROUTES, FINALISABLE_PAGES } from './constants';
+import { ROUTES } from './constants';
 import { RootNavigator } from './RootNavigator';
 import { PrescriptionActions } from '../actions/PrescriptionActions';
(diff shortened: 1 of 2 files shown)


wlthomson:
> actually, im a little confused - this state update already occurs, doesn't it?
> 
> trigger:
> https://github.com/openmsupply/mobile/blob/0661fc3adc77d86c7cd43416246dfddd299c0ae5/src/navigation/actions.js#L53
> 
> state update:
> https://github.com/openmsupply/mobile/blob/5e535444d2c954908374191b1fbda62a9224ff16/src/reducers/FinaliseReducer.js#L42-L44

`Navigaton/BACK`... think we have a typo! Constants!

PR #3110, opened 2020-08-20
#3109 Add invalid sync queue migration code

### LINKED ISSUE #3109: Add migration code to cleanup invalid sync out records

See fix #3092 for background.

Add migration code to remove invalid sync out records.

## Implementation

See above. 

N/A.

## Additional context

#3092.

wlthomson:
Fixes #3109.

## Change summary

Adds migration code to delete invalid sync out records.

Tested locally by injecting invalid sync out records before upgrading. 

@bijaySussol Would be good if you could test this in production. Subject to approval will do a `v5.1.2` build and notify you. To ensure migration code runs correctly, will have to do a proper `v5.1.2` release (no more rcs).

@joshxg Any comments re. above? This patch has got quite big already, so I don't think there's any issue with bumping any other bugs to another patch, i.e. `v5.1.3`?

## Testing

- [ ] When upgrading to `v5.1.2`, all invalid sync out records are removed from sync queue (to test in production, export realm database or use realm explorer).

N/A.

### DIFF
--- a/src/dataMigration.js
+++ b/src/dataMigration.js
@@ -252,7 +252,7 @@ const dataMigrations = [
           const { recordType, recordId } = syncRecord;
           const record = database.get(recordType, recordId);
           return !record;
-        }, []);
+        });
         database.delete('SyncOut', invalidSyncRecords);
       });
     },
--- a/src/dataMigration.js
+++ b/src/dataMigration.js
@@ -239,6 +239,24 @@ const dataMigrations = [
       });
     },
   },
+
+  // 5.1.2 fixes a sync out queue bug which was preventing invalid sync out records being
+  // overwritten. On upgrade, the following migration code filters the sync queue for invalid
+  // sync out records and deletes any which are found.
+  {
+    version: '5.1.2',
+    migrate: database => {
+      database.write(() => {
+        const syncRecords = database.objects('SyncOut');
+        const invalidSyncRecords = syncRecords.filter(syncRecord => {
+          const { recordType, recordId } = syncRecord;
+          const record = database.get(recordType, recordId);
+          return !record;
+        }, []);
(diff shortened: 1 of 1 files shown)


src/dataMigration.js
```diff
@@ -239,6 +239,24 @@ const dataMigrations = [
...
+
+  // 5.1.2 fixes a sync out queue bug which was preventing invalid sync out records being
+  // overwritten. On upgrade, the following migration code filters the sync queue for invalid
+  // sync out records and deletes any which are found.
+  {
+    version: '5.1.2',
+    migrate: database => {
+      database.write(() => {
+        const syncRecords = database.objects('SyncOut');
+        const invalidSyncRecords = syncRecords.filter(syncRecord => {
```
joshxg, line 251: do we need to `slice()` to turn this into an array?

joshxg (changes requested):
quick q about array v realm list - not sure on the behaviour?

src/dataMigration.js
wlthomson: I believe `Realm.Results` inherits/implements `array.prototype.filter` now (not sure how recently), and behaves similarly to `array.prototype.map` (i.e. implicitly coerces to `array`). I also wasn't sure so cross-referenced with other parts of our codebase, and we do have it a few places, e.g. (from `PeriodSchedule.js`):

```
return isEmergency ? this.periods.slice() : this.periods.filter(filterValidPeriods);
```

Given that the code runs OK (although only tested by me), I think it's OK! But no harm in adding an explicit `slice`?

joshxg: ah yup - I remember now. 

Should be OK, but also no harm in adding the slice

joshxg (approved):

PR #3137, opened 2020-09-21
#3131 Usage denominator - *B

### LINKED ISSUE #3131: Usage should use 365/12 not 30

To be consistent, usage calculations should use 365/12 rather than 30

Used in the following places:
- `dailyUsage`
- `Item.js`
- `Requisition.js`
- `RequisitionItem.js`
- `createRecord.js`

## Implementation

Create a constant: `NUMBER_OF_DAYS_IN_A_MONTH =  365/12` and use it!

N/A

## Additional context

We need better folder structure - have no idea where this constant should actually go!

joshxg:
Fixes #3131

## Change summary

- Adds small adjustments to AMC calculation

## Testing

- [ ] Usage calculation is calculated with a denominator of 365/12 rather than 30, now

N/A


Chris-Petty (dismissed):
@craigdrown Would this be better as a preference that comes through sync?

src/database/DataTypes/Item.js
```diff
@@ -148,7 +148,7 @@ export class Item extends Realm.Object {
    * @return {Number} this items monthly usage based on a 30 day month.
```
mark-prins, line 148: :nit - the comment could be removed

src/database/DataTypes/Requisition.js
```diff
@@ -193,7 +193,7 @@ export class Requisition extends Realm.Object {
    * @param  {number}  months
    */
   set monthsToSupply(months) {
-    this.daysToSupply = months * 30;
+    this.daysToSupply = months * NUMBER_OF_DAYS_IN_A_MONTH;
   }
```
mark-prins, line 198: There's also a getter at line 132 - this would also need updating, I think: 

```
 get monthsToSupply() {
    return this.daysToSupply / 30;
  }
```

mark-prins (changes requested):
One extra usage of `30` to change I think.
Apart from that - looks good to me!

src/database/DataTypes/Item.js
joshxg: e4a3649a992c583cc1c5bab8258273515d1eef74

src/database/DataTypes/Requisition.js
joshxg: There was also default value in the schema which I updated. Wonder if I should write migration code. I think I'm going to put that in the too hard basket...........!!

joshxg: 2dc540c67c700d680e1a745983ad2b98b10127d1

src/database/DataTypes/Requisition.js
```diff
@@ -130,11 +130,11 @@ export class Requisition extends Realm.Object {
    * @return  {number}
    */
   get monthsToSupply() {
-    return this.daysToSupply / 30;
+    return this.daysToSupply / NUMBER_OF_DAYS_IN_A_MONTH;
   }
 
   /**
-   * Get the sum of required quantites for all items associated with requisition.
+   * Get the sum of required quantities for all items associated with requisition.
```
mark-prins, line 137: ha! didn't even notice that one.

mark-prins (approved):
Looking good!

src/database/DataTypes/Requisition.js
joshxg: Me either evidently!! 🤣

PR #3138, opened 2020-09-21
#3135 Refactor program based supplier requisitions & stocktakes to use name_tag records *A2

joshxg:
### Branched from: #3136 

Fixes #3135 

## Change summary

Have done the minimal amount of changes for this refactor
- `ByProgramModal` now uses `name_tag`
- Checking for 'all programs' now uses `name_tag`

## Testing

- [ ] Able to create a program supplier requisition with functionality no different to before
- [ ] Able to create a program stocktake with functionality no different to before

### Related areas to think about

N/A

PR #3139, opened 2020-09-22
#3132 Customer Requisitions ByProgramModal *A3

### LINKED ISSUE #3132: Create customer requisition ByProgramModal component

Need to make manually created, program customer requisitions

Create a `ByProgramModal` modal component (similar/same) as the current `ByProgramModal`, but customer requisition specific

## Implementation

Create a `CustomerRequisitionByProgramModal` component, which is essentially very similar/the same to the current `ByProgramModal` component, just customer requisition specific

Try to continue using the generic solution. Can investigate first.

## Additional context

N/A

joshxg:
### Branched from: #3138 

Fixes #3132

## Change summary

**Implemented differently to the issue describes**

- Initially thought it would be far too complex to add customers over suppliers into `ByProgramModal`. Turned out to be quite quick and simple (If you were the one to write it originally. Sorry).
- This should *still* be refactored into: Customer requisition `ByProgramModal`, Supplier Requisition and Stocktake variants as the logic in *general* is convoluted and extremely hard to follow.
- Essentially, this allows for passing `transactionType="customerRequisition"` as a prop which will have the steps "Customer -> Program -> Order type -> period"`.
- Not hooked up - can create a branch for testing if needed, otherwise can be tested in the next 'hook it up' PR

## Testing

*Internal changes, shouldn't be tested*

- At this point I am not sure how much more extensible we need to make the component - surely there's not going to be customer invoices by program ... ?!?! So although I hate this code and want to refactor it desperately, I think it's only worth it once the requirements are completed proper?

### DIFF
--- a/src/database/DataTypes/Period.js
+++ b/src/database/DataTypes/Period.js
@@ -31,7 +31,7 @@ export class Period extends Realm.Object {
     return this.requisitions.length;
   }
 
-  requisitionsForOrderType(program, orderType) {
+  numberOfSupplierRequisitionsForOrderType(program, orderType) {
     return this.requisitions.filtered(
       'program.id = $0 && orderType = $1 && type == "request"',
       program.id,
@@ -39,7 +39,7 @@ export class Period extends Realm.Object {
     ).length;
   }
 
-  customerRequisitionsForOrderTypeAndName(program, orderType, name) {
+  numberOfCustomerRequisitionsForOrderType(program, orderType, name) {
     return this.requisitions.filtered(
       'program.id = $0 && orderType = $1 && otherStoreName = $2',
       program.id,
--- a/src/database/DataTypes/Period.js
+++ b/src/database/DataTypes/Period.js
@@ -39,6 +39,15 @@ export class Period extends Realm.Object {
     ).length;
   }
 
+  customerRequisitionsForOrderTypeAndName(program, orderType, name) {
+    return this.requisitions.filtered(
+      'program.id = $0 && orderType = $1 && otherStoreName = $2',
+      program.id,
+      orderType.name,
+      name
+    ).length;
+  }
+
   addRequisitionIfUnique(requisition) {
     if (this.requisitions.filtered('id == $0', requisition.id).length > 0) return;
     this.requisitions.push(requisition);
--- a/src/database/DataTypes/Period.js
+++ b/src/database/DataTypes/Period.js
@@ -33,7 +33,7 @@ export class Period extends Realm.Object {
 
   requisitionsForOrderType(program, orderType) {
     return this.requisitions.filtered(
-      'program.id = $0 && orderType = $1',
+      'program.id = $0 && orderType = $1 && type == "request"',
       program.id,
       orderType.name
     ).length;
--- a/src/database/DataTypes/PeriodSchedule.js
+++ b/src/database/DataTypes/PeriodSchedule.js
@@ -28,12 +28,15 @@ export class PeriodSchedule extends Realm.Object {
     const filterValidPeriods = period => {
       if (customer) {
         return (
-          period.customerRequisitionsForOrderTypeAndName(program, orderType, customer) <
+          period.numberOfCustomerRequisitionsForOrderType(program, orderType, customer) <
           orderType.maxOrdersPerPeriod
         );
       }
 
-      return period.requisitionsForOrderType(program, orderType) < orderType.maxOrdersPerPeriod;
+      return (
+        period.numberOfSupplierRequisitionsForOrderType(program, orderType) <
+        orderType.maxOrdersPerPeriod
+      );
     };
 
     return isEmergency ? this.periods.slice() : this.periods.filter(filterValidPeriods);
--- a/src/database/DataTypes/PeriodSchedule.js
+++ b/src/database/DataTypes/PeriodSchedule.js
@@ -20,12 +20,21 @@ export class PeriodSchedule extends Realm.Object {
    * If the ordertype is an emergency order type, return all periods.
    * @param {MasterList}  program    The related program
    * @param {object}      orderType  The maxOrdersPerPeriod of the orderType
+   * @param {Name}        customer   Optional customer param to filter by.
    */
-  getPeriodsForOrderType(program, orderType) {
+  getPeriodsForOrderType(program, orderType, customer) {
     const { isEmergency } = orderType || {};
 
-    const filterValidPeriods = period =>
-      period.requisitionsForOrderType(program, orderType) < orderType.maxOrdersPerPeriod;
+    const filterValidPeriods = period => {
+      if (customer) {
+        return (
+          period.customerRequisitionsForOrderTypeAndName(program, orderType, customer) <
+          orderType.maxOrdersPerPeriod
+        );
+      }
+
+      return period.requisitionsForOrderType(program, orderType) < orderType.maxOrdersPerPeriod;
+    };
 
     return isEmergency ? this.periods.slice() : this.periods.filter(filterValidPeriods);
   }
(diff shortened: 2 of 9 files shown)


src/database/DataTypes/Period.js
```diff
@@ -33,12 +33,21 @@ export class Period extends Realm.Object {
...
   }
 
+  customerRequisitionsForOrderTypeAndName(program, orderType, name) {
+    return this.requisitions.filtered(
+      'program.id = $0 && orderType = $1 && otherStoreName = $2',
+      program.id,
+      orderType.name,
+      name
+    ).length;
+  }
```
joshxg, line 49: Just finding requisitions which are specific to the customer. Additional filter otherwise, as we'll now have customer requisitions for the program and order type which shouldn't be counted toward the 'count' for supplier requisitions for the period/program/ordertype

src/database/DataTypes/PeriodSchedule.js
```diff
@@ -20,12 +20,21 @@ export class PeriodSchedule extends Realm.Object {
...
-      period.requisitionsForOrderType(program, orderType) < orderType.maxOrdersPerPeriod;
+    const filterValidPeriods = period => {
+      if (customer) {
+        return (
+          period.customerRequisitionsForOrderTypeAndName(program, orderType, customer) <
+          orderType.maxOrdersPerPeriod
+        );
+      }
+
+      return period.requisitionsForOrderType(program, orderType) < orderType.maxOrdersPerPeriod;
```
joshxg, line 36: This is finding the 'valid' period for an order type - those which are 'available'.. or have not had 'too many' requisitions - adding an additional customer 'filter', although I don't think this is the best implementation

src/reducers/ByProgramReducer.js
```diff
@@ -121,6 +131,17 @@ export const byProgramReducer = (state, action) => {
...
+      return {
+        ...state,
+        customer: value,
+        program: null,
+        orderType: null,
+        period: null,
+        name: '',
+        isModalOpen: false,
+        complete: false,
+      };
```
joshxg, line 144: Just a case in the program reducer which, when selecting a customer, sets everything else to null

src/utilities/byProgram.js
```diff
@@ -10,6 +10,30 @@ import { SETTINGS_KEYS } from '../settings';
...
+    .objects('NameTag')
+    .filtered('subquery(nameTagJoins, $joins, $joins.name.id == $0).@count > 0', id)
+    .map(({ description }) => description.toLowerCase());
+
+  return database
+    .objects('MasterListNameJoin')
+    .filtered('name.id = $0 && masterList.isProgram = $1', id, true)
+    .filter(({ masterList }) => masterList.getStoreTagObject(customersTags))
+    .map(({ masterList }) => masterList);
+};
```
joshxg, line 35: New utility method which just finds all the programs for a specific customer

src/utilities/getModalTitle.js
```diff
@@ -43,6 +49,10 @@ export const getModalTitle = modalKey => {
   switch (modalKey) {
     default:
       return '';
+    case MODAL_KEYS.PROGRAM_REQUISITION:
+      return `${navStrings.requisition} ${programStrings.details}`;
+    case MODAL_KEYS.PROGRAM_STOCKTAKE:
+      return `${navStrings.stocktake} ${programStrings.details}`;
```
joshxg, line 55: Noticed a bit of a bug, where titles weren't showing!

src/widgets/Step.js
```diff
@@ -152,7 +155,6 @@ const localStyles = StyleSheet.create({
   containerStyle: {
     flexDirection: 'row',
     width: '100%',
-    justifyContent: 'center',
```
joshxg, line 155: wasn't needed 🤷

src/widgets/modals/ByProgramModal.js
```diff
@@ -31,11 +36,15 @@ import { FlexColumn } from '../FlexColumn';
...
-const modalProps = ({ dispatch, program, orderType }) => ({
+const modalProps = ({ dispatch, program, orderType, customer }) => ({
   program: {
     queryStringSecondary: 'name CONTAINS[c] $0',
     onSelect: value => dispatch(selectProgram(value)),
   },
+  customer: {
+    secondaryFilterProperty: 'code',
+    onSelect: value => dispatch(selectCustomer(value)),
+  },
```
joshxg, line 47: this `modalProps` function passes the props for the `AutocompleteSelector` modal, for choosing out of the options for the current step. The primary filter property default is to filter by `'name'`, and the `onSelect` being passed will just trigger the reducer, passing back the `customer` record and storing it in state, as well as resetting all other steps.

For the actual data filling the modal for selecting a customer, it is below in a function `getCustomers`

src/widgets/modals/ByProgramModal.js
```diff
@@ -73,7 +82,9 @@ const modalProps = ({ dispatch, program, orderType }) => ({
       const { requisitions } = getLocalizedStrings();
       const { maxOrdersPerPeriod, isEmergency } = orderType;
 
-      const requisitionsInPeriod = item.requisitionsForOrderType(program, orderType);
+      const requisitionsInPeriod = !customer
+        ? item.requisitionsForOrderType(program, orderType)
+        : item.customerRequisitionsForOrderTypeAndName(program, orderType, customer);
```
joshxg, line 87: This is for the properties for the modal: the row displays stats: how many requisitions so far in the period for example. So, it's changed to calculate correctly the number of requisitions in the period!

src/widgets/modals/ByProgramModal.js
```diff
@@ -130,13 +142,19 @@ export const ByProgramModal = ({ settings, database, transactionType, onConfirm
     .map(({ description }) => description.toLowerCase());
 
   // Helper methods for fetching modal data for user selection
-  const getPrograms = () => getAllPrograms(settings, database);
+  const getPrograms = () => {
+    if (transactionType === 'customerRequisition') {
+      return getAllProgramsForCustomer(customer, database);
+    }
+    return getAllPrograms(settings, database);
```
joshxg, line 149: Finding the correct programs, given the customer (or no customer for supplier requisition)

src/widgets/modals/ByProgramModal.js
```diff
@@ -266,7 +296,7 @@ export const ByProgramModal = ({ settings, database, transactionType, onConfirm
 
   return (
     <FlexColumn flex={1} alignItems="center">
-      <ProgramToggleBar />
+      {transactionType !== 'customerRequisition' ? <ProgramToggleBar /> : null}
```
joshxg, line 299: Cheap hack hiding the toggle bar (Creating general requisitions not part of the spec)

src/database/DataTypes/Period.js
```diff
@@ -33,12 +33,21 @@ export class Period extends Realm.Object {
...
   requisitionsForOrderType(program, orderType) {
     return this.requisitions.filtered(
-      'program.id = $0 && orderType = $1',
+      'program.id = $0 && orderType = $1 && type == "request"',
       program.id,
       orderType.name
     ).length;
   }
 
+  customerRequisitionsForOrderTypeAndName(program, orderType, name) {
```
Chris-Petty, line 42: 2 nitpicks: 

- Might be clearer if you change `requisitionsForOrderType` to `supplierRequisitionsForOrderType`
- The method names I was expecting the actual realm results object to returned, not the length. But I can see that only length is used so 🤷‍♂️

Chris-Petty (approved):

src/database/DataTypes/Period.js
joshxg: Great point. I made them more descriptive, but now a bit long. 🤷 


1791f74

joshxg:
Just re-requesting to double check that last commit, please sir!

Chris-Petty (approved):
Double checked, does appear you renamed the method everywhere 😄

src/database/DataTypes/Period.js
Chris-Petty: Yea the length I was a little worried about which is why I had approved regardless haha

joshxg:
Thanks!

PR #3159, opened 2020-10-05
#3158 Missing variable in deps list for ByProgramModal

### LINKED ISSUE #3158: ByProgramModal fields not updating correctly
## Describe the bug

- When changing a the current field from one selection to another, the selection isn't updated correctly

### To reproduce

1. Go to `ByProgramModal`
2. Select a supplier/customer
3. Check programs
4. Select a new supplier/customer which has a different set of programs available
5. See the selection not updated

### Expected behaviour

The list should be updated

### Proposed Solution

Probably memozation issue

### Version and device info

- App version: N/A 
- Tablet model: N/A 
- OS version: N/A

### Additional context

N/A

joshxg:
Fixes #3158 

## Change summary

- Updated variable to refetch modal data once the data of the step has changed

## Testing

- [ ] Selection is updated (as described in issue)

Another indication a refactor needed?

### DIFF
--- a/src/widgets/Step.js
+++ b/src/widgets/Step.js
@@ -59,6 +59,7 @@ const getLocalisation = ({ stepKey }) => {
 export const Step = memo(props => {
   const [modalData, setModalData] = useState([]);
   const { data, field, type, stepKey, status, getModalData, onPress } = props;
+
   // Each stepKey used with this component has a placeholder/error string in getLocalisation.
   const { placeholder, errorString } = getLocalisation({ stepKey });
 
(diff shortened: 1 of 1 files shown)


Chris-Petty (approved):

src/widgets/Step.js
```diff
@@ -73,7 +74,7 @@ export const Step = memo(props => {
   useEffect(() => {
     if (status !== 'CURRENT' || !getModalData) return;
     setModalData(getModalData());
-  }, [status, internalStatus]);
+  }, [status, internalStatus, data]);
```
Chris-Petty, line 77: Did it take long to find this? One of the arguments I read against prematurely optimising with memoization (might have been from the man Dan?) was that it makes tracing bugs a bit more difficult, so is a compromise against the value of the optimisation

joshxg: This one, no - but I have had others that I've had issues with. To be honest, getting pretty good at spotting when it's a memoization problem just from the behaviour of the interface.. 🤣

PR #3166, opened 2020-10-21
#3165 fix outgoing sync null program id

### LINKED ISSUE #3165: Server throws runtime error on null requisition program id
## Describe the bug

When syncing to mSupply `v4.13` (possibly `>=v4.13`), a runtime error occurs when a requision is synced with `requisition.programId === null`.

### To reproduce

1. Create a non-program requisition.
2. Sync to desktop.
3. See error.*

*@anildahalsussol can you provide more details on this?

### Expected behaviour

`requisition.programId` should synced as blank string if `undefined` or `null`.

### Proposed Solution

`outgoingSyncUtils.js:135`: 

```
case 'Requisition': {
  return {
    ...
    programID: record?.program?.id ?? '',
    ...
  };
}
```

### Version and device info

- App version: `v5.1.2`
- Tablet model: N/A.
- OS version: N/A.

### Additional context

Raised by @anildahalsussol.

wlthomson:
Fixes #3165.

## Change summary

Server is throwing runtime errors on `requisition.programID === null`. 

## Testing

- [ ] No runtime errors on syncing non-program requisitions to desktop (@anildahalsussol).

Minimal regression surface for changes, not sure if similar changes should be applied to e.g. `periodID`?


wlthomson:
@anildahalsussol are you able to test with a local build or do you need an APK?

joshxg (commented):
As much as I am sure this will fix this issue (I haven't tested, though), why is this only occurring for `programID`? This pattern is used a lot throughout the `outgoingSyncUtils` file - Should we change them all? Is there something specific about `programID` which causes this issue?

I thought if `null` was synced it mapped to the default for that type - and if undefined was sent it was ignored? Maybe @Chris-Petty can help?

wlthomson:
> As much as I am sure this will fix this issue (I haven't tested, though), why is this only occurring for `programID`? This pattern is used a lot throughout the `outgoingSyncUtils` file - Should we change them all? Is there something specific about `programID` which causes this issue?

Currently lacking a bit of context. @anildahalsussol has let me know there is a runtime error occurring for `programID` due to it being `null`. I'm waiting on some confirmation from him, so hopefully he will be able to give more info!

anildahalsussol:
In Yangon datafile where requisition is created from tablet to desktop. They don't use the program while making the requisition. We are planning to upgrade them to 413 and found that when we create req from tablet to desktop and while opening requisition in desktop it's throwing run time error due to null value set in the programID. In code, we are just checking for blank not null. And while checking code I found it's been used in few more places so I thought if we can send blank from the tablet if no program then it will fix the problem in 413. 
Record:
Screenshot 2020-10-20 at 13 17 04

 Issue:
https://github.com/sussol/msupply/issues/7099

wlthomson:
@anildahalsussol Can you test that this branch fixes the issue?

I guess the main Q here is getting a consistent mobile -> desktop sync convention, i.e. do we want to default strings to `''` or to `null`? I think the answer to that will determine how we proceed!

joshxg:
Excuse my lack of 4D knowledge, but shouldn't the `Map Null values to blank values` feature on the field mean this field can never be `null`? I checked the latest structure and this is set, so I am confused as to how this is set as `null`? IMO this seems like a symptom of a different problem: Maybe the field does not have this set on the structure/version they are using?

wlthomson:
> Excuse my lack of 4D knowledge, but shouldn't the `Map Null values to blank values` feature on the field mean this field can never be `null`? I checked the latest structure and this is set, so I am confused as to how this is set as `null`? IMO this seems like a symptom of a different problem: Maybe the field does not have this set on the structure/version they are using?

I believe the problem is to do with serialisation/parsing, e.g. desktop is parsing `programId` as `"null"` (as opposed to `null`). I'm still not clear on when this bug was introduced, but it does seem like it's likely to be a desktop side cause? @anildahalsussol thoughts?

anildahalsussol:
@wlthomson @joshxg : Will update once I will test properly with 413 and mobile version 5.1.2.

anildahalsussol:
@wlthomson @joshxg : It seems while parsing data it's casting null as a string.
{"ID":"e0a6fb80136011ebadc93babdc8c0fe8","date_entered":"2020-10-21T05:47:17.183Z","user_ID":"AF75BC5A678C4B1FAB4B034E301F2D12","name_ID":"DC2A3E4642351046B8E33601575FC287","status":"fn","daysToSupply":"30","store_ID":"395DC80373086546BA1D464E1FFA8B10","serial_number":"320","requester_reference":"173","comment":null,"type":"request","programID":null,"periodID":null,"orderType":null,"custom_data":null}}]

joshxg:
Having the string `"null"` makes a bit more sense to me! I can see how that's possible - pretty big one I think. Means we maybe should do this change for all the fields, write update code for desktop? Ugh.

anildahalsussol:
@joshxg : i can do check for programID in 4d server it can be either blank or null. Just I need to check in a few places. If agreed then I will update the agreed solution in the desktop issue and close this one. Hmm and I think if we are querying for blank for few more fields like "periodID": null," order type": null,". Then it will also most probably throw run time error.

joshxg:
Is it causing really big problems *right now* for users @anildahalsussol ?

I think it is important we make a change in mobile, but probably for all the fields: do you agree @wlthomson ? 

But if there are a small amount of records on desktop that are causing issues, we could update them (But we should make sure they are not synced back to mobile) manually. I would also claim the server might want to be defensive against receiving `null` as it is valid JSON and and non-intuitive that sending `null` in JSON will be parsed to a string - question for the server expert I think! @adrianwb ? (TLDR: Mobile has been sending `null`, 4D server parses it to `"null"` - will fix on mobile, should the server be more defensive also?)

anildahalsussol:
@joshxg : So far not. But we are planning to upgrade MAM and if we upgrade then it can be a problem. As we need to upgrade MAM due to another issue fixed in 413. Anyway I think I have found a simple fix in 4d I will make a pr then you guys can add your review.

joshxg:
Great. We should get this fixed on mobile soon, but the data already on the server will need a hand. Thanks so much @anildahalsussol and sorry for holding up the PR @wlthomson !

wlthomson:
>  I would also claim the server might want to be defensive against receiving null as it is valid JSON and and non-intuitive that sending null in JSON will be parsed to a string

I agree, I think the main issue here is desktop side (`null` is being incorrectly parsed as `"null"`). There is still an issue re. string fields being assigned `null` (although JSON is untyped, both mobile and desktop expect a string type here), so I think there are two issues here:

1. Mobile syncing out `null` instead of `""` for string typed fields.
2. Desktop parsing `null` as a string instead of `null` (I imagine this is an issue with the JSON parsing logic server side).

I think any urgent fixes should probably be done desktop side, e.g. substitute `"null"` for `null` or `""` wherever the error is occuring. This would just be a bandaid fix though, next steps would be to fix the mobile-side issue and ensure `null` is being correctly parsed as `null` (i.e. revert any quick fixes that may have been introduced to map `"null"` to `""`).

wlthomson:
>  I would also claim the server might want to be defensive against receiving null as it is valid JSON and and non-intuitive that sending null in JSON will be parsed to a string

I agree, I think the main issue here is desktop side (`null` is being incorrectly parsed as `"null"`). There is still an issue re. string fields being assigned `null` (although JSON is untyped, both mobile and desktop expect a string type here), so I think there are two issues here:

1. Mobile syncing out `null` instead of `""` for string typed fields (the actual outgoing logic is setting it to falsey, but I believe the sync logic maps this to `null`?).
2. Desktop parsing `null` as a string instead of `null` (I imagine this is an issue with the JSON parsing logic server side).

I think any urgent fixes should probably be done desktop side, e.g. temporarily substitute `"null"` for `Null` or `""` where the error is occurring. This would just be a bandaid fix though, next steps would be to fix the mobile-side issue (note: once the proper mobile-side fix is done, should also ensure that `null` is being correctly parsed as `Null` on server-side, i.e. revert any quick fixes that may have been introduced to map `"null"` to `""`).

wlthomson:
> sorry for holding up the PR @wlthomson !

No worries, agreed on all points. 

Have updated the issue with the fields that need updating; interestingly some fields are already mapped to `""`, so I definitely agree it's a good idea to bring it all into consistency with a single PR.

wlthomson:
@joshxg Do we want to put this into the next release?

joshxg:
Into 6.0 defo, into a patch? I'm not sure, I don't think so. Although I am not sure what to do regarding the current data which is getting messed (And when this bug was introduced??) - Maybe we should be writing migration code to fix all the malformed data, if any?  + maybe some footrunner for desktop side? I dunno..!

wlthomson:
@joshxg 

I think migration code would be good (better safe than sorry). I *think* that we need to migrate on mobile and desktop side to ensure all munged data is fixed up. Ideally the desktop migration would occur before the mobile migration, but I don't think the order matters?

joshxg:
@wlthomson Technically we shouldn't touch the store data of a mobile store on desktop. But if we edit all the data on mobile it's a lot of syncing, which is why I was not sure if it was the correct thing to do.

Possibly you could update the records on both, and just have them not sync, but that's pretty open to mistakes, I'd say!

Sorry I'm not 100% sure on what the best path to take is at the moment

adrianwb:
To the best of my knowledge, we have never had `Null` fields in any mSupply tables - it has always used default values e.g. "" for string; 0 for numeric, `False` for boolean etc. => IMHO, mobile should also respect that. The server could technically check all incoming "null" values, but that's extra overhead for absolutely everything (and would potentially mess up conversion of non-string fields) and as has already been pointed out, "null" is actually valid JSON.

That makes it tricky for the server to run any update code - how does it know if any of the "null"'s are actually valid? I'm assuming that this is easier to do on mobile, so that would be an argument for doing the update on mobile and syncing changes back to the cloud?

However, it's probably reasonable enough to say that none of the "null"'s on the server are valid, although that would require a mega-query to cover all of the possible fields which could be affected! Technically though, updating mobile store data on the central server won't sync anything out so it's safe enough from that point of view, but obviously if the same record is then synced again from an un-patched mobile site then we're back to square one again! 

The only way that I can think of to get around that is to (also?) run update code for a store on the server when it gets the corresponding upgrade message from the store's mobile site. In that case, it would be OK to run update code on mobile but not sync the changes, as the upgrade message on the server would (in theory anyway) apply the exact same updates.

joshxg (changes requested):
Missed a couple - sorry, may aswell!

- L127:  name_ID: record.otherStoreName && record.otherStoreName.id,
- L215:  category_ID: record.category && record.category.id,

joshxg (dismissed):
Great

joshxg:
I just realised this is based on master - If we base on develop, I plan to release 6.0 ?

wlthomson:
> I just realised this is based on master - If we base on develop, I plan to release 6.0 ?

👍 .

joshxg (approved):

PR #3176, opened 2020-11-01
#3175 add temp disabling of vaccines

### LINKED ISSUE #3175: Disable vaccine module

We need to release 6.0. Vaccines isn't fully tested/finished.

Disable the preference which enables vaccine module!

## Implementation

N/A

N/A

## Additional context

N/A

joshxg:
Fixes #3175 

## Change summary

As per the issue!

## Testing

- [ ] Vaccine module can never be enabled

N/A


mark-prins (approved):
Looks like that will do the job!

Not convinced that it is necessary - if a user enables vaccines and then gets a bug it's their problem 🤷‍♂️

joshxg:
@mark-prins 

I agree - my thoughts were in general that if they got a bug that it would probably go through support and then that would be filtered through to me. Then it would be my problem. This seemed easier than dealing with the ticket.. 😆

PR #3177, opened 2020-11-01
#3174 Add xlargescreens support

### LINKED ISSUE #3174: Display size setting for tablet is controlled.
## Describe the bug

When enrolled the device onto MDM (MediaPad M5 Lite) for IC - mSupply app against MDM shows an alert message saying: **mSupply Mobile does not support the current display size setting xxxx** 

The display setting against the tablet is controlled.

### Proposed Solution

### Version and device info

- App version: 5.0.3
- Tablet model: MediaPad M5 Lite
- OS version: Android 9

### Additional context

Found this:
https://developer.android.com/work/managed-configurations

joshxg:
Fixes #3174 

## Change summary

Changes `supports-screens` android manifest option.

https://developer.android.com/guide/topics/manifest/supports-screens-element#compat-mode

Seems `requiresSmallestWidthDp` only allows for googly play store to apply filter. But google play store does not filter by it. 🤷 

## Testing

- [ ] @bijaySussol  ;-)

Stab in the dark.. no idea if it'll work!

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -8,7 +8,7 @@
   <uses-permission android:name="android.permission.BLUETOOTH_PRIVILEGED" />
   <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
   <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
-  <supports-screens android:smallScreens="false" android:normalScreens="false" android:largeScreens="false" android:requiresSmallestWidthDp="720" />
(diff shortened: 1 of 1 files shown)


Chris-Petty (approved):
[LGTM](https://developer.android.com/guide/topics/manifest/supports-screens-element)

Chris-Petty:
Even if this doesn't fix the problem, I think we should add it heh.

bijaySussol (approved):
@joshxg : 

As per i read the docs shared - I think, this should be it.

bijaySussol:
@joshxg : 

Can we create a app out this changes?
So, That i can push this through MDM and asked Ben to test this apk on that tablet, please

PR #3198, opened 2020-11-11
Add safe ID checking

joshxg:
Fixes #?

## Change summary

A program indicator is optional, so a indicator value may not have a related program indicator (??)

## Testing

- [ ] Viewing a response requisition which is created from a request requisition doesn't crash the app

Odd


mark-prins (approved):
should do the job :) LGTM

PR #3211, opened 2020-11-19
#3210 Excessive row rendering

### LINKED ISSUE #3210: Customer requisition rows excessive re-rendering
## Describe the bug

Turned `debug` on on `CustomerRequisitionPage` and there is some unneeded row renders

### To reproduce

Turned `debug` on on `CustomerRequisitionPage` and see some unneeded row renders

### Expected behaviour

Rows should only re-render when required

### Proposed Solution

N/A

### Version and device info

- App version: 6.0RC10 +
- Tablet model:
- OS version:

### Additional context

N/A

joshxg:
Fixes #3210

## Change summary

- Memoized columns and style object of data table row
- Fixed minor page info key issue

## Testing

- [ ] Editing a row doesn't cause 10's of re-renders

N/A

### DIFF
--- a/src/pages/CustomerRequisitionPage.js
+++ b/src/pages/CustomerRequisitionPage.js
@@ -463,6 +463,7 @@ CustomerRequisition.defaultProps = {
   usingIndicators: false,
   showIndicators: false,
   indicatorCodes: [],
+  indicatorColumns: [],
   currentIndicatorCode: '',
 };
 
@@ -504,5 +505,5 @@ CustomerRequisition.propTypes = {
   datePickerIsOpen: PropTypes.bool.isRequired,
   onDatePickerClosed: PropTypes.func.isRequired,
   isRemoteOrder: PropTypes.bool.isRequired,
-  indicatorColumns: PropTypes.object.isRequired,
+  indicatorColumns: PropTypes.array,
 };
--- a/src/pages/CustomerRequisitionPage.js
+++ b/src/pages/CustomerRequisitionPage.js
@@ -68,7 +68,6 @@ export const CustomerRequisition = ({
   keyExtractor,
   modalValue,
   searchTerm,
-  columns,
   getPageInfoColumns,
   onEditComment,
   onFilterData,
@@ -79,7 +78,6 @@ export const CustomerRequisition = ({
   onEditSuppliedQuantity,
   route,
   onToggleColumnSet,
-  columnSet,
   onEditOpeningStock,
   onEditNegativeAdjustments,
   onEditPositiveAdjustments,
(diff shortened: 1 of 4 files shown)


src/widgets/DataTable/DataTableRow.js
```diff
@@ -372,6 +375,7 @@ const DataTableRow = React.memo(
         rowData={rowData}
         rowState={rowState}
         rowIndex={rowIndex}
+        debug
```
wlthomson, line 378: Oops, or intentional?

src/widgets/DataTable/DataTableRow.js
```diff
@@ -87,11 +87,14 @@ const DataTableRow = React.memo(
...
-      ...(!isValidated && { borderColor: SUSSOL_ORANGE, borderWidth: 1, borderRadius: 5 }),
-    };
+    const rowStyle = useMemo(
+      () => ({
+        ...(rowIndex % 2 === 0 ? alternateRowStyle : basicRowStyle),
+        ...(isSelected && selectedRowStyle),
+        ...(!isValidated && { borderColor: SUSSOL_ORANGE, borderWidth: 1, borderRadius: 5 }),
+      }),
+      [rowIndex, isSelected, isValidated]
+    );
```
wlthomson, line 97: Could you squeeze a bit more performance here but separating out into individual `useMemo`s for different use cases, e.g. `rowIndex` for sorting, `isSelected` for selecting, etc.?

src/pages/CustomerRequisitionPage.js
```diff
@@ -91,9 +89,24 @@ export const CustomerRequisition = ({
...
   const { isFinalised, comment, program, isResponse, createdDate, entryDate } = pageObject;
 
+  const currentColumns = useMemo(() => {
+    const formEntryColumnSet = isRemoteOrder
+      ? 'editableCustomerRequisitionFormEntry'
+      : 'customerRequisitionFormEntry';
+    if (usingIndicators && showIndicators) {
+      return indicatorColumns;
+    }
+    return columnSet === 'a'
```
wlthomson, line 105: Any particular reasoning for this literal? Could perhaps benefit from a constant (or a comment, in the meantime)?

wlthomson (approved):
lgtm

src/widgets/DataTable/DataTableRow.js
joshxg: oops 857b7aa

src/widgets/DataTable/DataTableRow.js
joshxg: I think only for when the row index changes and the `% 2` result is the same? Don't think it's worth it

src/pages/CustomerRequisitionPage.js
joshxg: it's just the state that's set, not sure what comment I could write that would help

src/widgets/DataTable/DataTableRow.js
wlthomson: I mean't more that you could make it so that you only needed to worry about the selected row style when `isSelected` changes and the validated styles when `isValidated` changes etc. (rather than calculating them all when any of the variables changes). But it's such a small thing, agree it's not really worth it!

src/pages/CustomerRequisitionPage.js
wlthomson: Is there a reason the `columnSet` is `'a'` rather than a constant? Does the literal have an obvious meaning that I'm missing?

joshxg: No, the action/reducer switches between "a" and "b" (This is just taken from the `mapStateToProps`)

PR #3220, opened 2020-11-22
#1680 Create a stocktake batch when creating a StocktakeItem

### LINKED ISSUE #1680: Create a StocktakeBatch when creating a StocktakeItem
## Describe the bug
Related to #673 

App does not sync uncounted stocktake lines, then on initialisation they are gone burgers.

### To reproduce

1. Create a stocktake with items
Screen Shot 2019-12-13 at 11 23 33
2. Count a couple lines
Screen Shot 2019-12-13 at 11 24 04
3. Sync the stocktake (note there should be far more records pending)
Screen Shot 2019-12-13 at 11 24 15
4. Wow sync is done
Screen Shot 2019-12-13 at 11 24 33
5. Clear app data and resync the app
6. Witness that all lines except those that were counted have vanished.
Screen Shot 2019-12-13 at 11 26 05

### Expected behaviour

That the lines in the original stocktake are persisted on the server through sync

### Proposed Solution

I suspect this occurs if the site synced has no stock, so when adding `StocktakeItems` no `StocktakeBatches` are created. The server only keeps track of the latter, but we're only sending the former... I think. If this is correct, one solution would be to make at least 1 `StocktakeBatch` when creating a `StocktakeItem` if none already exist.

Doing it server side is fraught with issues due to sync.

### Version and device info

- App version: 3.2.0
- Tablet model: emulator
- OS version: android 9

joshxg:
Fixes #1680

## Change summary

- Create a basic `StocktakeBatch` when creating a `StocktakeItem`

## Testing

- [ ] Creating a Stocktake and don't count anything. Sync it to desktop: On desktop, you should see some lines

N/A

### DIFF
--- a/src/database/DataTypes/Stocktake.js
+++ b/src/database/DataTypes/Stocktake.js
@@ -93,9 +93,13 @@ export class Stocktake extends Realm.Object {
       const stocktakeItem = createRecord(database, 'StocktakeItem', this, item);
 
       // Add all item batches currently in stock to the stocktake item as stocktake batches.
-      item.batchesWithStock.forEach(itemBatch => {
-        createRecord(database, 'StocktakeBatch', stocktakeItem, itemBatch);
-      });
+      if (item.batchesWithStock.length > 0) {
+        item.batchesWithStock.forEach(itemBatch => {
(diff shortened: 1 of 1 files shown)


wlthomson (dismissed):
lgtm

PR #3226, opened 2020-11-22
#3225 Stop pruning remote customer requisitions

### LINKED ISSUE #3225: Stop pruning non remote order customer requisitions
## Describe the bug

Customer requisitions are pruned of all lines if the required quantity is 0. This will prune lines even if the user has added some details: i.e. ins/outs. 

### To reproduce

1. Go to a manually created customer requisition, in a line add DOS = some X and no required quantitiy.
2. Finalise

### Expected behaviour

Lines which have had some data entered shouldn't be pruned.

### Proposed Solution

Just stop pruning customer requisitions which are not remote orders. In the future, could add a `touched` field to the line to determine if it has been touched and prune all those that haven't.

### Version and device info

- App version: 6.0RC10+
- Tablet model: N/A
- OS version: N/A

### Additional context

N/A

joshxg:
Fixes #3225 

## Change summary

- Just add to the boolean to stop pruning requisition items which are from a remote order.

## Testing

- [ ] Finalising a remote customer requisition does not prune any lines
- [ ] Finalising a non-remote customer requisition does prune any item where the required quantity is 0.

N/A

### DIFF
--- a/src/database/DataTypes/Requisition.js
+++ b/src/database/DataTypes/Requisition.js
@@ -387,7 +387,7 @@ export class Requisition extends Realm.Object {
   pruneRedundantItems(database) {
     const itemsToPrune = [];
     this.items.forEach(requisitionItem => {
-      if (requisitionItem.requiredQuantity === 0) {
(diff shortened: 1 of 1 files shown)


wlthomson (approved):
lgtm_2

PR #3229, opened 2020-11-23
#3082 Dont enforce personal policy numbers

### LINKED ISSUE #3082: Personal Policy numbers are enforced as mandatory, but shouldn't be
## Describe the bug

inconsistent validation behaviour of Personal Policy Number is disabling the Save button

### To reproduce

1. Go to Dispensary-> Choose Patient -> CHoose Prescriber
2. Choose one or two item
3. Add a new insurance such that fill Family policy number first
4. Now enter value in personal policy number. erase it. See error Validation keep showing this disabling save button
5. Now go to family policy number append or type something in this field. Save button is enabled now.

https://recordit.co/1f70XRun4V

### Expected behaviour

Since Personal Policy Number is not mandatory field hence in any condition it should not disable the Save button.
Also validation message shows wrong "Must be a unique policy number between 0 and 50 characters" since this field is accepting alphanumeric characters.

### Proposed Solution

### Version and device info

- App version: v6.0.0 rc4
- Tablet model: lenovo
- OS version: android

joshxg:
Fixes #3082 

## Change summary

- removes the lowerbound character length check of personal policy numbers

## Testing

- [ ] Can create a an insurance policy without a personal policy number

N/A

### DIFF
--- a/src/utilities/formInputConfigs.js
+++ b/src/utilities/formInputConfigs.js
@@ -187,7 +187,7 @@ const FORM_INPUT_CONFIGS = seedObject => ({
     type: FORM_INPUT_TYPES.TEXT,
     initialValue: '',
     key: 'policyNumberPerson',
-    validator: input => input.length > 0 && input.length < 50,
+    validator: input => input.length < 50,
     isRequired: false,
     invalidMessage: formInputStrings.unique_policy,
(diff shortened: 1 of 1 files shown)


wlthomson (approved):
lgtm

PR #3230, opened 2020-11-23
#2753 Debounce toggling of data sets

### LINKED ISSUE #2753: ToggleButton should be debounced when toggling data sets
## The suggestion

In a few places where `ToggleBar` is used, it should debounce. For example: `CustomerInvoices`.

**Why?**
It can take a little bit of time when there are thousands of records to toggle between - about a second or so. It's easy to double click accidentally and is a bit annoying!

N/A

## Additional context

I think a new prop should be added to `ToggleBar` to debounce the call back

joshxg:
Fixes #2753

## Change summary

- Adds a `useDebounce` hook cause it's fun
- Adds debouncing of toggle set toggling

## Testing

- [ ] Toggling the current->past data set on customer invoices, supplier invoices, supplier requisitions, customer requisitions and stocktakes are debounced

N/A


src/hooks/useDebounce.js
```diff
@@ -0,0 +1,31 @@
...
+    return (...args) => {
+      const doItLater = () => {
+        timeout = null;
+        if (!immediate) func(...args);
+        return null;
+      };
+
+      // If the immediate flag is true and there is no timeout
+      // already, just execute.
+      const doItNow = immediate && !timeout;
```
wlthomson, line 18: Assuming `!timeout` means clearing `timeout` will be redundant, can you not do:

```suggestion
      if (immediate && !timeout) return func(...args);
```

and remove the `if (doItNow)` conditional at the end? This would possibly remove the need to handle double executions that could occur otherwise?

src/hooks/useDebounce.js
```diff
@@ -0,0 +1,31 @@
...
+
+export const useDebounce = (func, waitAmount, immediate) => {
+  const debounceIt = () => {
+    // Closure timeout ref for all subsequent calls
+    let timeout;
+
+    // Return a func wrapper with the timeout var ref closure
+    return (...args) => {
+      const doItLater = () => {
+        timeout = null;
```
wlthomson, line 11: If you use `clearTimeout` here does it do some weird stuff, since this is the function being called by the timeout?

src/hooks/useDebounce.js
```diff
@@ -0,0 +1,31 @@
...
+        if (!immediate) func(...args);
```
wlthomson, line 12: Why the `!immediate` condition? To prevent executing twice? If so, are there possible race conditions here?

src/hooks/useDebounce.js
```diff
@@ -0,0 +1,31 @@
...
+        return null;
```
wlthomson, line 13: Prob can just have no `return` here, maybe `undefined` makes more sense in this context than `null`?

wlthomson (dismissed):
lgtm_3

src/hooks/useDebounce.js
joshxg: clearing timeout is redundant if there isn't one already, yeah, but you still need to set one after you execute the func so it debounces any further invocations

Doing it this way you would never set a timer, though or am I misunderstanding? So what you suggested would be equivalent to `debounce = (func, wait, imm) => func`..?

src/hooks/useDebounce.js
joshxg: I dont see the race?

src/hooks/useDebounce.js
joshxg: yea

joshxg (commented):
I'll wait for clarification /confirmation/ comments on this one before merging

src/hooks/useDebounce.js
wlthomson: Wouldn't you still set the timer is `immediate` was falsey? Also, if `immediate` is truthy doesn't it stop `doItLater` from executing anyway?

src/hooks/useDebounce.js
wlthomson: I mean it's pretty unlikely, but the `immediate` conditional could be executed, then `immediate` set to `false`, then the function run (despite `immediate` being falsey at the time of execution).

src/hooks/useDebounce.js
joshxg: Ah now I see: 

Yes, you would set the timer if immediate was false and it would work fine. But,if it was true, then the timer would never start and you would never debounce.

With this implementation, when immediate is true, you always execute on the first invocation, and then start the timer which will block any further invocations from executing the func, but the func is not executed the second time.

src/hooks/useDebounce.js
joshxg: I'm not sure how the immediate is changed within the closure?

PR #3231, opened 2020-11-23
#2995 Add populating of sentPackSize

### LINKED ISSUE #2995: Populate sent_pack_size

A new field has been added to the `trans_line` to store the pack size of an item when supplied.
see [this issue](https://github.com/sussol/msupply/issues/6365)

- Populate the `sent_pack_size` when creating a supplier invoice.
- Add a field `sentPackSize` to the `TransactionBatch` class.
- Add to `incomingSyncUtils` setting of `sentPackSize === record.sent_pack_size` [Just for storing the value for incoming SI's sent from the server, so when we send it back it is preserved -- MAYBE THIS IS NOT NEEDED? If we send an SI without a `sent_pack_size` field, does it override the value on the server?]
- Add to `outgoingSyncUtils` for a `TransactionBatch`, `sent_pack_size === record.sentPackSize ?? record.packSize` for `Transaction` with `type === 'supplier_invoice'` [Sending back either the `sentPackSize` that has been preserved from an incoming supplier invoice, otherwise, it's a manually created SI for an external supplier and we just use the normal packSize (which is 1, because mobile only uses 1)

joshxg:
Fixes #2995

## Change summary

- Adds sent pack size to the schema

## Testing

- [ ] Create a CI for the mobile store with an item with the pack size > 1 (and doesnt have pack to one set to true) and see the sent pack size field populated to be equal to the pack size sent

N/A

### DIFF
--- a/src/sync/incomingSyncUtils.js
+++ b/src/sync/incomingSyncUtils.js
@@ -883,6 +883,7 @@ export const createOrUpdateRecord = (database, settings, recordType, record) => 
           'VaccineVialMonitorStatus',
           record.vaccine_vial_monitor_status_ID
         ),
+        sentPackSize: parseNumber(record.sent_pack_size ?? 0),
       };
       const transactionBatch = database.update(recordType, internalRecord);
       transaction.addBatchIfUnique(database, transactionBatch);
--- a/src/sync/incomingSyncUtils.js
+++ b/src/sync/incomingSyncUtils.js
@@ -883,7 +883,7 @@ export const createOrUpdateRecord = (database, settings, recordType, record) => 
           'VaccineVialMonitorStatus',
           record.vaccine_vial_monitor_status_ID
         ),
-        sentPackSize: parseNumber(record.sent_pack_size ?? 0),
+        sentPackSize: parseNumber(record.sent_pack_size) || packSize,
       };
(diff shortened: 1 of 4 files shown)


mark-prins:
I tested. It made me sad 😢 
`Response error message for remote sync id 6: Field number not found for trans_line:sentPackSize`

joshxg:
@mark-prins I'm sorry!! e9ad3b6936d531d440b3e8a2781a0eb8326f5cdf i'm a muppet

src/sync/incomingSyncUtils.js
```diff
@@ -883,6 +883,7 @@ export const createOrUpdateRecord = (database, settings, recordType, record) =>
           'VaccineVialMonitorStatus',
           record.vaccine_vial_monitor_status_ID
         ),
+        sentPackSize: parseNumber(record.sent_pack_size) || packSize,
```
wlthomson, line 886: Nit: `??`

wlthomson (approved):
ok

src/sync/incomingSyncUtils.js
joshxg: Different logic

mark-prins (approved):
Code looks fine. unable to get it working using the harry potter sample datafile. but it's me. not you..

PR #3234, opened 2020-11-24
#1657 Bold the currently highlighted row

### LINKED ISSUE #1657: Bold/ Highlight the current row

joshxg:
Fixes #1657

## Change summary

- Adds some state to data table row and a callback to manage focused rows

## Testing

- [ ] When editing a row in a data table page (`CustomerRequisitionPage`, `SupplierInvoicePage`, `CustomerInvoicePage`, `SupplierRequisitionPage`, `StocktakeEditPage`)

Implementation differs to what I originally thought. 

1. I thought it would be best to manage this in `dataState` in the redux store for each page, and have a cell `onFocus`, dispatch the action to store in state that the row was focused. However, this caused a redux update which isn't ideal and is a bit slow.
2. Also I realised there needs to be an `onBlur` callback, for when you, for example, focus the searchbar the row needs to unfocus - this can cause some serious jank, as you need to manage the flow where a single row has multiple text inputs, and when you push enter, the next column on the same row is focused. So if you just un-focus on blur, the highlighting would blink! ..

So: 1) Manage state in `datatablerow` is a bit quicker than managing in redux 2) debounce the focusing such that when `blur` event is triggered, it does not execute for a few milliseconds, and then the next call to `focus` cancels the `blur` call (if its the same row).

### DIFF
--- a/src/widgets/DataTable/DataTableRow.js
+++ b/src/widgets/DataTable/DataTableRow.js
@@ -9,7 +9,7 @@ import PropTypes from 'prop-types';
 
 import currency from '../../localization/currency';
 
-import { dataTableStyles, SUSSOL_ORANGE } from '../../globalStyles';
+import { ROW_BLUE, dataTableStyles, SUSSOL_ORANGE } from '../../globalStyles';
 
 import Row from './Row';
 import Cell from './Cell';
@@ -68,6 +68,7 @@ const DataTableRow = React.memo(
     rowIndex,
     getCellError,
     isValidated,
+    onFocus,
   }) => {
     const {
       cellText,
@@ -84,16 +85,16 @@ const DataTableRow = React.memo(
       iconCell,
     } = dataTableStyles;
 
-    const { isSelected = false } = rowState || {};
-    // If the row is selected, use selectedRow style, otherwise alternate row style on index.
+    const { isFocused = false, isSelected = false } = rowState || {};
 
     const rowStyle = useMemo(
       () => ({
         ...(rowIndex % 2 === 0 ? alternateRowStyle : basicRowStyle),
         ...(isSelected && selectedRowStyle),
         ...(!isValidated && { borderColor: SUSSOL_ORANGE, borderWidth: 1, borderRadius: 5 }),
+        ...(isFocused && { backgroundColor: ROW_BLUE }),
       }),
-      [rowIndex, isSelected, isValidated]
+      [rowIndex, isSelected, isValidated, isFocused]
     );
 
     // Callback for rendering a row of cells.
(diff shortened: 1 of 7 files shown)


mark-prins:
on the stocktake batch modal I'm still getting a janky update when moving from the expiry date edit to another column in the same row. Maybe just on emulator, dunno. Don't think that it's bad enough to need addressing, just being a tester and adding a grey cloud into an otherwise clear sky..

src/widgets/DataTable/DataTableRow.js
```diff
@@ -371,6 +376,7 @@ const DataTableRow = React.memo(
         rowData={rowData}
         rowState={rowState}
         rowIndex={rowIndex}
+        debug
```
mark-prins, line 379: needs removing?

mark-prins (approved):
I really like it - and nice work with the debounce, the UI looks good

src/widgets/DataTable/DataTableRow.js
joshxg: 261a9c9

joshxg:
Nice - Turns out this cell was actually calling its callback twice - causing two re-renders and the jank!

PR #3244, opened 2020-11-24
Add corrected order of operations

### LINKED ISSUE #3243: CustomerRequisition AMC has a calculation error
## Describe the bug

`return Math.max(outgoingStock / (numberOfDaysInPeriod ?? 0 - daysOutOfStock), 0);`

should be

`return Math.max(outgoingStock / ((numberOfDaysInPeriod ?? 0) - daysOutOfStock), 0);`

### Expected behaviour

Be right

### Proposed Solution

N/A

### Version and device info

- App version: 6.0+
- Tablet model: N/A
- OS version: N/A

### Additional context

N/A

joshxg:
Fixes #3243 

## Change summary

- some bedmas!

## Testing

- [ ] Customer requisition daily usage is correct

N/A

### DIFF
--- a/src/utilities/dailyUsage.js
+++ b/src/utilities/dailyUsage.js
@@ -108,5 +108,5 @@ export const programDailyUsage = (item, period) => {
  */
 export const customerRequisitionProgramDailyUsage = requisitionItem => {
   const { outgoingStock, daysOutOfStock, numberOfDaysInPeriod } = requisitionItem;
-  return Math.max(outgoingStock / (numberOfDaysInPeriod ?? 0 - daysOutOfStock), 0);
+  return Math.max(outgoingStock / ((numberOfDaysInPeriod ?? 0) - daysOutOfStock), 0);
 };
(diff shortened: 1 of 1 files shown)


src/utilities/dailyUsage.js
```diff
@@ -108,5 +108,5 @@ export const programDailyUsage = (item, period) => {
  */
 export const customerRequisitionProgramDailyUsage = requisitionItem => {
   const { outgoingStock, daysOutOfStock, numberOfDaysInPeriod } = requisitionItem;
-  return Math.max(outgoingStock / (numberOfDaysInPeriod ?? 0 - daysOutOfStock), 0);
+  return Math.max(outgoingStock / ((numberOfDaysInPeriod ?? 0) - daysOutOfStock), 0);
```
mark-prins, line 111: Maybe a check for the case where `numberOfDaysInPeriod` is 0 and `daysOutOfStock` is too - returning `0` in that case makes more sense than infinity..maybe?

mark-prins (dismissed):
LGTM! 
I've just added a comment about possible division by zero which is really outside the scope of the PR 😄

src/utilities/dailyUsage.js
joshxg: Ah of course, Doh! I added more slightly (too?/more?) defensive code .. 
feeb4a4

mark-prins (approved):
Looking good. I think it's a super unlikely scenario.. but just in case 😄

PR #3246, opened 2020-11-24
#3050 login screen should show store name

### LINKED ISSUE #3050: Login screen should show store name

Currently it shows the site name. If you have a messy or codified name for sites, this looks bad.

Show the store name instead, or additionally.

## Implementation

TBD

Being less picky

katherine-sussol:
Fixes #3050 

## Change summary

- Adds store name to the login screen under the site name and some labels to distinguish between the two
- Removed a now unused local style, unsure about styling but seems reasonably clear/usable(?)
- Don't know translations for these words so English only for now

## Testing

- [ ] Open login page, check that the site name and store name are correctly displayed

- Unsure - maybe check the rest of the login page still looks acceptable and it still looks ok on different sized tablets?

### DIFF
--- a/src/widgets/modals/LoginModal.js
+++ b/src/widgets/modals/LoginModal.js
@@ -152,7 +152,7 @@ export class LoginModal extends React.Component {
             </View>
             <View style={[globalStyles.verticalContainer, { flex: 1 }]}>
               <Text style={[globalStyles.authFormTextInputStyle]}>
-                {authStrings.store}:{' '}
+                {authStrings.store}:
                 {
                   UIDatabase.objects('Name').filtered(
                     'id == $0',
--- a/src/widgets/modals/LoginModal.js
+++ b/src/widgets/modals/LoginModal.js
@@ -145,22 +145,18 @@ export class LoginModal extends React.Component {
               style={globalStyles.authFormLogo}
               source={require('../../images/logo_large.png')}
             />
-            <View style={[globalStyles.verticalContainer, { flex: 1 }]}>
-              <Text style={[globalStyles.authFormTextInputStyle]}>
-                {authStrings.site}: {settings.get(SETTINGS_KEYS.SYNC_SITE_NAME)}
-              </Text>
-            </View>
-            <View style={[globalStyles.verticalContainer, { flex: 1 }]}>
-              <Text style={[globalStyles.authFormTextInputStyle]}>
-                {authStrings.store}:
-                {
-                  UIDatabase.objects('Name').filtered(
-                    'id == $0',
-                    settings.get(SETTINGS_KEYS.THIS_STORE_NAME_ID)
-                  )[0]?.name
-                }
-              </Text>
-            </View>
+            <Text style={[globalStyles.authFormTextInputStyle, { flex: 0 }]}>
+              {authStrings.site}: {settings.get(SETTINGS_KEYS.SYNC_SITE_NAME)}
+            </Text>
+            <Text style={[globalStyles.authFormTextInputStyle, { flex: 0, marginVertical: 10 }]}>
+              {authStrings.store}:
+              {
+                UIDatabase.objects('Name').filtered(
+                  'id == $0',
+                  settings.get(SETTINGS_KEYS.THIS_STORE_NAME_ID)
+                )[0]?.name
+              }
+            </Text>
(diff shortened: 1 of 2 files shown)


src/widgets/modals/LoginModal.js
```diff
@@ -144,9 +145,20 @@ export class LoginModal extends React.Component {
...
+              </Text>
+            </View>
+            <View style={[globalStyles.verticalContainer, { flex: 1 }]}>
+              <Text style={[globalStyles.authFormTextInputStyle]}>
+                {authStrings.store}:
+                {
+                  UIDatabase.objects('Name').filtered(
+                    'id == $0',
+                    settings.get(SETTINGS_KEYS.THIS_STORE_NAME_ID)
+                  )[0].name
```
joshxg, line 160: for safety: `[0]?.name`

src/widgets/modals/LoginModal.js
```diff
@@ -144,9 +145,20 @@ export class LoginModal extends React.Component {
               style={globalStyles.authFormLogo}
               source={require('../../images/logo_large.png')}
             />
-            <View style={globalStyles.horizontalContainer}>
-              <Text style={[globalStyles.authFormTextInputStyle, localStyles.syncSiteName]}>
-                {settings.get(SETTINGS_KEYS.SYNC_SITE_NAME)}
+            <View style={[globalStyles.verticalContainer, { flex: 1 }]}>
```
joshxg, line 148: On my emulator I felt there was a bit to much vertical space:

image


So I gave it a spin trying to reduce it :
image

I dunno.


Also removed the `store`, `site` because I thought they were a bit ugly, but I think that could be confusing so I am not sure if thats a good idea.



I just did it like this:

```
 <Image
   resizeMode="contain"
   style={globalStyles.authFormLogo}
   source={require('../../images/logo_large.png')}
 />
 <Text style={[globalStyles.authFormTextInputStyle, { flex: 0 }]}>
  {settings.get(SETTINGS_KEYS.SYNC_SITE_NAME)}
 </Text>
 <Text style={[globalStyles.authFormTextInputStyle, { flex: 0, marginVertical: 10 }]}>
 {
  UIDatabase.objects('Name').filtered(
  'id == $0',
  settings.get(SETTINGS_KEYS.THIS_STORE_NAME_ID)
  )[0].name
 }
</Text>
```

joshxg (changes requested):
I think there's to much space, but happy to be told i'm stupid, usually the stuff I make looks super ugly.

I think this component needs some serious refactoring so I'm not concerned, but normally I might suggest storing the value in a variable maybe even a function in `utilities` and maybe making new styles rather than adding the margins and stuff inline.. actually nah I'd probably just do it inline 🤣

src/widgets/modals/LoginModal.js
katherine-sussol: Oh, yep good point 9f788c58

src/widgets/modals/LoginModal.js
katherine-sussol: Hmm yeah, the vertical space looks much worse on your emulator than it did on mine! I'll have to set up another device for testing purposes I think. I dunno about the text, are real site names quite obviously distinct to users from store names? If so then yeah it's probably redundant and I'll just steal your implementation 🤭

Chris-Petty: I agree with @joshxg on the spacing. I haven't flexbox'd for literally years, but there should be a props on the container and items that determine how they space (e.g. spread out evenly in the container, smallest possible from the top, stretch to fit w/ even spacing...).

When ever I'm confused I end up playing with this website https://cvan.io/flexboxin5/. Is web flexbox but it's close enough to figure out what you want.

Chris-Petty: @joshxg Regarding the label on the values I agree that it isn't the prettiest but also think it's ambiguous/random seeming without, so would keep. Dunno if it'd improve much but could use icons maybe? Unfortunately there isn't a pattern I know that I can lean into here haha.

joshxg (approved):
lgtm-kid

Chris-Petty:
Sorry didn't pick this up with PR image. But site text has a space (👍 ) store text does not (😿 ) 
image

PR #3255, opened 2020-11-26
factor out retrieving store name to not be in the render code

### LINKED ISSUE #3254: Login modal needs a space after store label text
## Describe the bug

Login page the site label has a space, but the store label does not.
image

### Expected behaviour

Both should have spaces.

### Proposed Solution

Add a space.

### Version and device info

- App version:
- Tablet model:
- OS version:

### Additional context

@joshxg @katherine-sussol For tweaks like this we used to allow making PRs with prefix #xxxx, e.g. `#xxxx-add-space-to-login-store-label` kind of says all you need to know without making an issue like this one haha. Do we want to bring that back? Will it be misused (e.g. #xxxx-rewrote-sync)? Can we just handle that if it happens?

Chris-Petty:
Fixes #3254

## Change summary

Linter rules didn't like it being a one liner, but also didn't like trailing spaces! So couldn't get a space in without rearranging, but this is a bit more readable anyway 🙂 

## Testing

- [ ] Look at the login modal. It has a space between `Store: ` and `my store name`
image

### DIFF
--- a/src/widgets/modals/LoginModal.js
+++ b/src/widgets/modals/LoginModal.js
@@ -129,6 +129,10 @@ export class LoginModal extends React.Component {
   render() {
     const { isAuthenticated, settings } = this.props;
     const { authStatus, username, password, appVersion, isLanguageModalOpen } = this.state;
+    const storeName = UIDatabase.objects('Name').filtered(
+      'id == $0',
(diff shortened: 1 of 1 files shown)


joshxg (approved):
lgtm-elmo

PR #3266, opened 2020-11-29
#3258 react-native bump

### LINKED ISSUE #3258: Dependencies upgrade
## The suggestion

Need to upgrade dependencies. I suspect it might be a good idea to separate this into multiple issues, if there are real code changes required when upgrading react-native.

**Why?**
Cause we're very behind.

N/A

## Additional context
N/A

joshxg:
Fixes #3258 

## Change summary

Bumps RN

## Testing

- [ ] App still worky

Technically flipper works now, but it doesn't work well with realm, still - for me, anyway!

### DIFF
--- a/src/widgets/InfoBadge.js
+++ b/src/widgets/InfoBadge.js
@@ -46,6 +46,10 @@ const InfoBadgeComponent = ({
     []
   );
 
+  console.log('-------------------------------------------');
+  console.log('ref', ref);
+  console.log('-------------------------------------------');
+
   return (
     <View>
       {children}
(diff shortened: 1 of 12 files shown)


src/widgets/InfoBadge.js
```diff
@@ -54,7 +58,7 @@ const InfoBadgeComponent = ({
           <Badge ref={ref} value={unfinalisedCountText} onPress={show} />
           <Popover
             isVisible={visible}
-            fromView={ref.current}
+            from={ref}
```
joshxg, line 57: Actually this was from a couple prs ago when I pumped `popover-view` - they changed the API

src/widgets/KeyboardSpacing.js
```diff
@@ -10,6 +10,7 @@ export const KeyboardSpacing = React.memo(() => {
     toValue: height,
     duration: 100,
     easing: Easing.linear,
+    useNativeDriver: false,
```
joshxg, line 13: Now required to explicitly state true/false

joshxg (commented):
Just followed https://react-native-community.github.io/upgrade-helper/

mark-prins (approved):
LGTM!

PR #3275, opened 2020-12-03
#2724 Update lookup modal to use SimpleTable

### LINKED ISSUE #2724: Update lookup modal table

The lookup modal table should:

- Display column headers.
- Style rows consistently with data table (alternating colours).

Add column headers and row styling.

## Implementation

The lookup modal form uses `FlatList` to render results

- Styles can be added using a conditional in the `renderItem` callback.
- `FlatList` directly supports column headers with `listHeaderComponent`, `stickyHeaderIndices` props (latter isn't documented for `FlatList` but seems to be passed down to the internal virtualised list).

N/A.

## Additional context

@joshxg @wlthomson Feedback from dispensary demo (17/04/20).

katherine-sussol:
Fixes #2724 

## Change summary
- Swapped SearchForm modal components to use SimpleTable (which provides styling and headers)
- Modified SimpleTable to handle dates

## Testing

- [ ] Lookup patients (Dispensary -> Patients -> Lookup Patient -> 'Search') and check Table format looks acceptable
- [ ] Lookup prescribers (Dispensary -> Prescribers -> Lookup Prescriber -> 'Search') and check Table format looks acceptable
- [ ] Verify patient/prescriber selection works in the above two modals as per before 

- [ ] SimpleTable is also used by the ItemSelect component (used when picking prescription items) - check it still works 
- [ ] Kat has no idea what she's doing so plz tell if anything is messed up


src/widgets/SimpleTable.js
```diff
@@ -38,12 +39,17 @@ export const SimpleTable = React.memo(
         rowData => {
           const { id } = rowData;
 
-          return columns.map(({ key, width, alignText }, index) => {
+          return columns.map(({ key, width, alignText, type }, index) => {
             const disabledStyle = disabledRows?.[id] || isDisabled ? { color: GREY } : {};
             const textStyle = { ...cellText[alignText], ...disabledStyle };
+            const valueText =
+              type === 'date'
+                ? rowData[key]?.toDateString() ?? generalStrings.not_available
```
joshxg, line 47: Believe it's the same format but can we use this `formatDate` method used here: https://github.com/openmsupply/mobile/blob/311babb8092fb578a062364c1be18e950ec07c4b/src/widgets/DataTable/DataTableRow.js#L282

joshxg (approved):
amazing - minor minor change

lgtm-crocodileman1

joshxg:
I meant to push request changes oops

src/widgets/SimpleTable.js
katherine-sussol: Cool format looks slightly different but I think the formatDate version makes more sense - committed it with e9f640d9:
- toDateString() => Wed Jul 29 1998
- formatDate() => 29/7/1998

joshxg (dismissed):
Cool - Oh, I think it's for localization - we use an internal language choice rather than what would be smart and the device OS' selection, so the built-in date formatting won't localize the names "Wed" etc.

I think

joshxg (approved):
Oh, also changed the base from master -> develop.

Develop isn't our default branch - I'm not sure why.

katherine-sussol:
> Oh, also changed the base from master -> develop.
> 
> Develop isn't our default branch - I'm not sure why.

Oops - thanks!

PR #3292, opened 2020-12-10
Add remove rounding days to supply

### LINKED ISSUE #3290: Requisition suggested quantities giving wrong values
## Describe the bug

When creating an automatic order the requisition is populated with lines. The suggested quantity column seems to be botching the math.

Used automatic order. Note Max MOS is `2`. The suggested quantity should be `2*AMC-SoH`. It is something else.
Screen Shot 2020-12-10 at 5 33 02 PM

I change Max MOS to `1`. We get `AMC-SoH` as you'd expect.
Screen Shot 2020-12-10 at 5 35 01 PM

Changing back to `2` we get the correct calculation of `2*AMC-SoH`!
Screen Shot 2020-12-10 at 5 34 54 PM

### To reproduce

Literally the description

### Expected behaviour

The math for suggested quantity works correctly when I create an automatic order.

### Proposed Solution

### Version and device info

- App version: 6.0.0 master build on dev
- Tablet mode: Emulator
- OS version:

joshxg:
Fixes #3290

## Change summary

- I think this is mainly cause of the rounding of the days to supply when creating a requisition. The changing of floor and ceil is probably not needed, technically, cause it should be some multiple of 365/12, always?

## Testing

- [ ] When creating a plain supplier requisition, the max MOS is default to 1
- [ ] Suggested quantities are correct as per the issue

Nice find.

Need unit tests.

Putting this into a patch so based on master

### DIFF
--- a/src/database/DataTypes/Requisition.js
+++ b/src/database/DataTypes/Requisition.js
@@ -130,7 +130,7 @@ export class Requisition extends Realm.Object {
    * @return  {number}
    */
   get monthsToSupply() {
-    return Math.ceil(this.daysToSupply / NUMBER_OF_DAYS_IN_A_MONTH);
+    return Math.floor(this.daysToSupply / NUMBER_OF_DAYS_IN_A_MONTH);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


mark-prins (approved):
works for me!

PR #3293, opened 2020-12-10
#3291 Correct disabling of a row

### LINKED ISSUE #3291: Can select finalised (Past) records for deletion
## Describe the bug

You can go to the past filter and select rows for deletion. You can't delete them, but this is a regression from the old behaviour where they'd be greyed out.

can select finalised for delete

This is particularly bad in Supplier invoices. The "Current" list should grey out the button for invoices that came from other stores, but allow selection for manually created ones (because they are allowed to be deleted).

### To reproduce

1. Go to past CI/SI/Req/Stocktake...
2. Select rows
3. Press delete

### Expected behaviour

Should not be able to select rows that you cannot delete. The radio button should be greyed out.

### Proposed Solution

TBD something something state management I'm unfamiliar with heh.

### Version and device info

- App version: 6.0
- Tablet model: emulator
- OS version:

joshxg:
Fixes #3291

## Change summary

- Disables a row when its finalised
- Disables a row when its a remote order - I think??

## Testing

- [ ] Cannot delete remote reqs/transacts
- [ ] Cannot check finalised rows

I synced an SI in - it had no linked req, but did have a linked transct - I think this is right??

### DIFF
--- a/src/database/DataTypes/Transaction.js
+++ b/src/database/DataTypes/Transaction.js
@@ -257,6 +257,15 @@ export class Transaction extends Realm.Object {
   }
 
   /**
+   * Get if this transaction is linked to a transaction.
+   *
+   * @return  {boolean}
+   */
+  get isLinkedToTransaction() {
+    return !!this.linkedTransaction;
+  }
+
+  /**
    * Get if this transaction includes a given item.
    *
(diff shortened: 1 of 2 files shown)


mark-prins (approved):
looks good to me

PR #3297, opened 2020-12-16
#3285 Saving temperature logs

### LINKED ISSUE #3285: Saving temperature logs

The logic to save temperature logs isn't crazy, but it's not simple either. It's also duplicated with mSupply ColdChain [here](https://github.com/openmsupply/msupply-cold-chain/blob/develop/src/features/Bluetooth/Download/DownloadManager.js)

Utilize the code written within mSupply ColdChain in mSupply mobile in a way where it can be shared.

## Implementation

- Take `DownloadManager`
- Rename it `TemperatureLogManager` (?)
- Adjust it in a way where it can be shared between the projects: namely changing the use of the database and also how UUIDs are generated.

N/A

## Additional context

We need a solution which will allow us to use this code in both projects. The main obstacles here are:

- Database usage. The class uses a `databaseService` which is not compatible with Realm (Although, it is very close).
	- Could wrap realm
    - Could change the shared code to not use the database, instead passing in a callback for saving things.
- UUID Generation
    - same as above
- Object fields
- Coupling the implementation where both projects are required to have the same field names exactly is also not ideal.
    - Callback where values are passed (temperature, timestamp) => TemperatureLog rather than hard coding the creation in the shared code?

joshxg:
Fixes #3285 

## Change summary

- Adds `TemperatureLogManager` class with tests for slightly changed up logic

## Testing

Internal, no black box?

I'm not sure why Jest went nutty on the `yarn.lock`?!


src/bluetooth/TemperatureLogManager.js
```diff
@@ -0,0 +1,83 @@
...
+  temperature,
+  logInterval,
+});
+
+export class TemperatureLogManager {
+  constructor(dbService, utils) {
+    this.db = dbService;
+    this.utils = utils;
+
+    this.createLogFunc = createTemperatureLog;
```
joshxg, line 17: Schemas differ slightly - allowing to setter inject a new creator for temperature logs.

src/bluetooth/TemperatureLogManager.js
```diff
@@ -0,0 +1,83 @@
...
+  sensor,
+  timestamp: new Date(timestamp),
+  temperature,
+  logInterval,
+});
+
+export class TemperatureLogManager {
+  constructor(dbService, utils) {
+    this.db = dbService;
+    this.utils = utils;
```
joshxg, line 15: Could use `import {UIDatabase, generateUUID} from './database'` but makes testing a bit painful with the imports and stuff. 

I dunno, I like passing stuff whenever possible

src/bluetooth/TemperatureLogManager.js
```diff
@@ -0,0 +1,83 @@
...
+  get createLog() {
+    return this.createLogFunc;
+  }
+
+  // Calculates the number of sensor logs that should be saved from some given starting
+  // point. Where the starting point is the timestamp for the next log.
+  calculateNumberOfLogsToSave = (
+    nextPossibleLogTime = 0,
+    logInterval,
+    timeNow = moment().unix()
```
joshxg, line 33: Just having this whole class work from unix timestamps (in secs) as I find it pretty easy (Dates suck!) and seems to be a pretty easy format to convert into whatever format is required by the caller (Mobile actually uses `Date` objects`)

joshxg (commented):
The mocks were old mocks from a time when jest was apparently used, then removed!

I removed them as they were causing issues with testing and weren't used (there exists no other tests)

src/bluetooth/TemperatureLogManager.js
```diff
@@ -0,0 +1,83 @@
...
+    // has not been recorded yet.
+    return Math.floor(secondsBetween / logInterval) + 1;
+  };
+
+  createLogs = (logs, sensor, maxNumberToSave, mostRecentLogTime, timeNow) => {
+    const { logInterval } = sensor;
+    const sliceIndex = logs.length - maxNumberToSave;
+    const logsToSave = logs.slice(sliceIndex);
+
+    let initial;
```
katherine-sussol, line 55: Hmm, so if there are no previous logs is this kind of a calculated guesstimate based on the current time?

katherine-sussol (commented):
Thanks for sending the PR for inspection! I think I'm generally on board with the implementation - I just have a few questions (potentially just me missing some 🧩  pieces)...

src/bluetooth/TemperatureLogManager.js
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  get createLog() {
+    return this.createLogFunc;
+  }
+
+  // Calculates the number of sensor logs that should be saved from some given starting
+  // point. Where the starting point is the timestamp for the next log.
+  calculateNumberOfLogsToSave = (
+    nextPossibleLogTime = 0,
```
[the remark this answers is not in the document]
joshxg: I have learned that most of the time, things are personal opinions on 'niceness'. Surely this isn't one of those times.. (oops)

1c03d3a

src/bluetooth/TemperatureLogManager.js
```diff
@@ -0,0 +1,83 @@
...
+
+    // For example, if there are 1 log interval between the starting time and now,
+    // then the times are for example, 0955 and 1000 - so, we save both the 0955 log
+    // and the 1000 log. If there was less than one log interval between, then the
+    // times would be 0955 and 0957 - so we only save a single log, as the 1000 log
+    // has not been recorded yet.
+    return Math.floor(secondsBetween / logInterval) + 1;
+  };
+
+  createLogs = (logs, sensor, maxNumberToSave, mostRecentLogTime, timeNow) => {
```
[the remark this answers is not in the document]
joshxg: I am also curious c447d54

src/bluetooth/TemperatureLogManager.js
```diff
@@ -0,0 +1,83 @@
...
+
+    let initial;
+    if (!mostRecentLogTime) {
+      initial = moment.unix(timeNow);
+      initial.subtract((logsToSave.length - 1) * logInterval, 'seconds');
+    } else {
+      const now = moment();
+      now.subtract(mostRecentLogTime);
+
+      const numberOfLogIntervalsUntilNow = Math.floor(now.format('x') / (logInterval || 1)) + 1;
```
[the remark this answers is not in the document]
joshxg: Yup you're right as


fee925b

src/bluetooth/TemperatureLogManager.js
joshxg: Yeah. Think the goal was to try and make it simpler to follow by calculating a time to calculate dates from, rather than counting dates backward.... 🙄

joshxg:
OK 
5e19a99

Lucky you're here.

Turns out the test cases weren't really capturing the case where there already existed some logs and therefore not covering this branch at all 😰 

The diff isn't too helpful but to try and explain: The `initial` is actually meant to be just that - the `initial` timestamp for the logs to be integrated. Which means that it actually should be `one log interval` greater than any existing temperature logs OR in the case where there are none, the guesstimate of when it was taken. But this wasn't doing that.

src/bluetooth/TemperatureLogManager.js
```diff
@@ -0,0 +1,75 @@
...
+    return Math.floor(secondsBetween / logInterval) + 1;
+  };
+
+  createLogs = (logs, sensor, maxNumberToSave, mostRecentLogTime, timeNow = moment().unix()) => {
+    const { logInterval } = sensor;
+    const sliceIndex = logs.length - maxNumberToSave;
+    const logsToSave = logs.slice(sliceIndex);
+
+    let initial;
+    if (mostRecentLogTime == null) {
```
joshxg, line 56: unfortunate line to capture `undefined` or `null` but not other falsey values like `0`, `''`

katherine-sussol (approved):
Thanks for the explanation(s), makes sense to me now in conjunction with the changes! 👍

PR #3314, opened 2021-01-10
#3308-1 TextInputWithAffix component

### LINKED ISSUE #3308: New editing rows
*EPIC: #3161*

*Related Marvel Designs: https://marvelapp.com/prototype/55aea8f*
*Related Component Breakdown: https://app.diagrams.net/#G1wLg12JAQyrgB6D3F_X0O9LBZwJs8PcJM

In the new vaccine module designs there are different style of editing rows:

Examples:

image
image
image

This issue is to discuss the implementation of these editable areas to try and create a solution which easily enables different styles of rows to edit different fields.

Unsure 100% just yet.

Still thinking about the API for this. I think I would like to see some sort of `Row` style component which can be composed of different `RowEdit` components?

Something like:

```
<Row>
    <RowLabel icon="X" text="Y"/>
    <RowTemperatureEdit currentValue={} someCallbacks/>
    <RowDurationEdit currentValue={} someCallbacks/>
</Row>
```

## Implementation

Create components:

- [ ] `TextInputWithMask`
- [ ] `TextInputWithAffix`
- [ ] `BaseStepperButton`
- [ ] `BaseStepperInput`
- [ ] `StepperInputWithLabel`
- [ ] `DurationEditor`
- [ ] `TemperatureEditor`
- [ ] `TimeEditor`
- [ ] `DateEditor`
- [ ] `TextEditor`
- [ ] `EditorRow`

TBD

## Additional context

N/A

joshxg:
Fixes #3308 **Wrong branch name- oops**

## Change summary

*Note* I named a prefix or a suffix 'affix'. I have no idea if that is correct english. I also don't know if what I am calling a prefix or suffix is a prefix or suffix. It's what I came up with 🤷 

- Adds `TextInputWithAffix` Component
- Essentially this is to fill the role of having an input with, for example: `10 C` or `10 Mins`
- Creating an input with the affix within the editable input area would be a lot more work than you'd think (Or, at least, from what I can figure out)

## Testing

N/A - internal code only white box testing

N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,6 +89,7 @@
     "react-native-safe-area-context": "^0.7.3",
     "react-native-screens": "^2.2.0",
     "react-native-svg": "^11.0.1",
+    "react-native-text-input-mask": "^2.0.0",
     "react-native-ui-components": "^0.5.0",
     "react-native-vector-icons": "^7.1.0",
     "react-navigation": "^4.2.2",
--- a/src/widgets/TextInputs/TextInputWithAffix.js
+++ b/src/widgets/TextInputs/TextInputWithAffix.js
@@ -1,0 +1,37 @@
+import React from 'react';
+import { TextInput as RNTextInput, TextInputProps as RNTextInputProps } from 'react-native';
+import propTypes from 'prop-types';
+
+import { FlexRow } from '../FlexRow';
+
+export const TextInputWithAffix = React.forwardRef(
+  ({ TextInputImpl, SuffixComponent, PrefixComponent, flexRowProps, ...textInputProps }, ref) => (
+    <FlexRow {...flexRowProps}>
+      {PrefixComponent}
+      <TextInputImpl ref={ref} {...textInputProps} />
+      {SuffixComponent}
+    </FlexRow>
+  )
+);
+
+TextInputWithAffix.displayName = 'TextInputWithAffix';
+
+TextInputWithAffix.defaultProps = {
+  TextInputImpl: RNTextInput,
+  SuffixComponent: <></>,
+  PrefixComponent: <></>,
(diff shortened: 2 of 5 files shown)


katherine-sussol (approved):
👀  looks good to me.

Word usage makes sense too, I always get confused between when to use 'Suffix' or 'Postfix' for things after words but context seems clear enough here 🤷

Chris-Petty:
> _Note_ I named a prefix or a suffix 'affix'. I have no idea if that is correct english. I also don't know if what I am calling a prefix or suffix is a prefix or suffix. It's what I came up with 🤷
> 

I wonder about just calling it `TextInput`, where it's just our own with default text style and what not set. Having a prefix/suffix would each just be props extending the RN component. Risk is mild overhead where it's not needed (e.g. possibly noticeable somewhere busy, like in a table, though). Maybe it'll cause too much confusion with the RN base component. If it's useful, I see it like overloading a method. Thoughts? Probably in the bucket of "We need to clean up widgets at some stage" stuff heh.

katherine-sussol:
I just noticed there's already a `TextInput` component under widgets, is this technically a parent, child or sibling of that one? 🤔   Maybe the prefix/suffixes should be added there? Or this component could use the custom `TextInput` instead of the RN one? Too many options 🙀

Chris-Petty:
> I just noticed there's already a `TextInput` component under widgets, is this technically a parent, child or sibling of that one? 🤔 Maybe the prefix/suffixes should be added there? Or this component could use the custom `TextInput` instead of the RN one? Too many options 🙀

🤣  what a mess. Should we start an issue where we just note down all the various muddy bits of UI, so we have a reference of what to clean up down the road?

PR #3315, opened 2021-01-10
#3308-2 Add react-native-text-input-mask

### LINKED ISSUE #3308: New editing rows

joshxg:
Fixes #3308

## Change summary

- Adds the package `react-native-text-input-mask`
- I tried a bunch of packages out and this was the best. It stops at the native component level a lot of jank. For example, if you were to use a `TextInput` as a controlled component and passed it `value` prop - you can actually type in the `TextInput` and NOT update the `value` and it will enter into the component. This is hard to explain but the `TextInput` basically has it's own internal state and if it's not in sync it causes odd behaviour, like if I had a mask which only allowed two digits and that was enforced through JS `onChangeText` prop, then I could type an `a` and it would be rendered for .5second, and then reverted back. This package however does the masking at the native level and this jank didn't seem to happen. Get it?

## Testing

Internal

Not creating the masks to use in the future yet. Will do as used.

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,6 +89,7 @@
     "react-native-safe-area-context": "^0.7.3",
     "react-native-screens": "^2.2.0",
     "react-native-svg": "^11.0.1",
+    "react-native-text-input-mask": "^2.0.0",
     "react-native-ui-components": "^0.5.0",
     "react-native-vector-icons": "^7.1.0",
     "react-navigation": "^4.2.2",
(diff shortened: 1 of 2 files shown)


katherine-sussol (approved):
Looks good to me 👀

PR #3318, opened 2021-01-10
#3286 creating temperature breaches

### LINKED ISSUE #3286: Creating temperature breaches

The logic to create temperature breaches isn't crazy, but it's not simple either. It's also duplicated with mSupply ColdChain [here](https://github.com/openmsupply/msupply-cold-chain/blob/develop/src/features/Breach/ConsecutiveBreach/ConsecutiveBreachManager.js)

Two options:

- Utilize the code written within mSupply ColdChain in mSupply mobile in a way where it can be shared.
- Rewrite the code in a way which can be utilized in both projects. The current solution isn't great!

## Implementation

- Take `ConsecutiveBreachManager`
- Adjust it in a way where it can be shared between the projects: Possibly using a completely different and simpler algo.

N/A

## Additional context

We need a solution which will allow us to use this code in both projects. The main obstacles here are:

- Database usage. The class uses a `databaseService` which is not compatible with Realm (Although, it is very close).
	- Could wrap realm
    - Could change the shared code to not use the database, instead passing in a callback for saving things.
- UUID Generation
    - same as above
- Object fields
- Coupling the implementation where both projects are required to have the same field names exactly is also not ideal.
    - Callback where values are passed (temperature, timestamp) => TemperatureLog rather than hard coding the creation in the shared code?

katherine-sussol:
Fixes #3286

## Change summary
- Mostly a copy of https://github.com/openmsupply/msupply-cold-chain/blob/develop/src/features/Breach/ConsecutiveBreach/ConsecutiveBreachManager.js 
- Extracted DB access methods to separate class and modified for mobile DB schema (may be a better way to structure these?) 
- Extracted createBreachRecord as it differs between CCE and Mobile
- Adjusted breach creation logic to transfer config to breach when creating, and read from breach (instead of config) when continuing
- Copy pasta-ed tests and adjusted to accommodate modified stuff

Next steps
- Cumulative breaches to be added separately?
  
## Testing

- [ ] Run the tests and visual inspection? 

- N/A yet but later there will be stuff that calls this


src/bluetooth/BreachManager.js
```diff
@@ -0,0 +1,189 @@
...
+
+    const updatedBreaches = this.db.upsertBreaches(breaches);
+
+    const mapped = temperatureLogs.map(({ id, temperature, timestamp }) => ({
+      id,
+      temperature,
+      timestamp,
+    }));
+
+    // TODO: Copied from CCE app - not sure if needed here
```
joshxg, line 168: I think remove it - I should fix whatever the stupid thing I did in CCE app heh

src/bluetooth/DataAccess.js
```diff
@@ -0,0 +1,51 @@
...
+    this.db = dbService;
+  }
+
+  getTemperatureLogsFrom = (sensorId, timeToCheckFrom) =>
+    this.db
+      .objects(VACCINE_ENTITIES.TEMPERATURE_LOG)
+      .filtered('sensor.id == $0 AND timestamp >= $1', sensorId, timeToCheckFrom)
+      .sorted('timestamp');
+
+  getMostRecentBreachLog = sensorId => {
```
joshxg, line 22: This should be the most recent log which was a part of a breach. 

- really it's used to determine when to find logs 'from'. I.e. if I download a log now, but to trigger a breach we need 1 hour of logs, we need to chec previous logs to determine if 'this is the one' which will create a breach. 
- Technically, we can look backwards and check only the duration of a breach - i.e. if a breach requires 1 hour worth of logs, we would technically only need to look back 1 hour.
-  It's just easier (imo!) to always look back to the most recent breached log (where we can guarantee that either this new log that has just been downloaded is a part of the same breach, or all of the logs since that breach could make up the new one) rather than looking up all the breach configs, determining which is the highest duration and looking back that amount. It's certainly not the most efficient, though!


So, I believe adding `AND temperatureBreach != $2, ..., null)` (I think, I always forget how to check for not exists... maybe `NOT null`..?)

src/bluetooth/DataAccess.js
```diff
@@ -0,0 +1,51 @@
...
+      .filtered('sensorId == $0', sensorId);
+    const [mostRecentBreach] = allBreaches.sorted('startTimestamp', true);
+
+    return mostRecentBreach;
+  };
+
+  getBreachConfigs = () => {
+    const configs = this.db
+      .objects(VACCINE_ENTITIES.TEMPERATURE_BREACH_CONFIGURATION)
+      .filtered('id == "HOT_BREACH" OR id == "COLD_BREACH"');
```
joshxg, line 43: We're gonna use `type` field now just to mix it up. The types will be `HOT_CUMULATIVE`, `HOT_CONSECUTIVE`, `COLD_CUMULATIVE` and `COLD_CONSECUTIVE`. This will need to return the `_CONSECUTIVE` ones

src/bluetooth/BreachManager.js
```diff
@@ -0,0 +1,189 @@
+import moment from 'moment';
+
+// Default breach creation method.
+const createBreachRecord = (
```
joshxg, line 4: Hmm there is the file `createRecord` ⚠️ 

Think we should probably use it (Can edit the func how you need)

src/bluetooth/DataAccess.js
```diff
@@ -0,0 +1,51 @@
...
+ * mSupply Mobile
+ * Sustainable Solutions (NZ) Ltd. 2021
+ */
+
+/**
+ * Helper methods to perform data access functionality for breach calculations
+ */
+import VACCINE_ENTITIES from '../utilities/modules/vaccines/constants';
+
+export class DataAccess {
```
joshxg, line 11: I wonder if we should have a name which is somewhat related to cold chain stuff `ColdChainDataAccess` 😕 ..? 🤷

joshxg (changes requested):
It's great - thanks so much!

Couple of minor changes!

src/bluetooth/DataAccess.js
katherine-sussol: Hm good point, DataAccess is a bit vague 😆 , is it more appropriate to prepend these things with `Vaccine` or `ColdChain`? I guess it's not always going to be vaccines? I created some constants below using `Vaccine` but am now reconsidering renaming those too 🤔 

I also kind of struggled with whether this should go in the Bluetooth folder or the `utilities/modules/vaccines` folder, lol.

src/bluetooth/BreachManager.js
katherine-sussol: Cool, have removed - I guess we will find out when we call it if Realm does weird stuff with it? c553c4ad

src/bluetooth/DataAccess.js
katherine-sussol: Oh right that makes sense, I see it in the updated schema too - updated with 4aaf415b

src/bluetooth/DataAccess.js
katherine-sussol: Whoops I just realised this method is just grabbing _all_ the temperature logs and returning the most recent 🤦 

Using NOT seemed to throw an invalid predicate error, but != appears to be at least a valid query (is there an easier way to manually populate the DB with data or is spamming create record methods the way to go?)

4aaf415b

src/bluetooth/BreachManager.js
katherine-sussol: 🙀  wow. Ok thanks I'll give this a go

src/bluetooth/DataAccess.js
joshxg: Yeah I'm not sure. I generally search files by name so I usually don't care so much for this type of thing.. I think either vaccines or coldchain is suitable in our context/names we use for stuff

src/bluetooth/DataAccess.js
joshxg: I think @Chris-Petty was working on a way to inject records easily ;-)

But really - we should probably make a test db, but in this epic is a zip `createData.zip` which is some 4D code to create records, which you could sync to your store.

But seems to be how you do it! https://github.com/sussol/msupply/issues/7271#issue-734959779

src/bluetooth/DataAccess.js
katherine-sussol: 👍  went with Vaccine for now ab82519d

src/bluetooth/BreachManager.js
Chris-Petty: On that note, `createRecord` is getting unwieldy (i.e. 1000+ lines). I'm wondering if it'd be better/plausible to use constructors in the realm classes and what that'd look like, how they'd be used?

[The realm docs](https://realm.io/docs/kotlin/latest/#limitations-models) would suggest it's possible but with minor(?) limitation:
>7. If a custom constructor is provided, a public no-arg constructor must also be present.

katherine-sussol: After looking at `createRecord` I'm not convinced this particular method is an ideal candidate for the file - as this is just building up an object, whereas the general pattern for `createRecord` seems to be to create + insert/update to DB in one go (which is also less clean?).

I briefly tried experimenting with adding a constructor to the TemperatureBreach Realm Object (cos it sounded like a really neat idea) but trying to do a DB update using the constructed object didn't seem to work, I think there's more magic going on behind the scenes I don't understand. Also the docs above seemed to be kotlin related, tried to multi-constructor in JS and got told off haha. Probably doing something wrong but maybe something still worth exploring?  Thought I'd stop digging before spending too long.

joshxg (approved):

src/bluetooth/BreachManager.js
joshxg: Yeah I've looked into realm constructors in the past as it would be pretty wicked for di and customizations - but I don't think it's worthwhile looking at them as we seem to be pretty set on sqlite if there were to be any big db moves. You're right about `createRecord`, what you've done is good

PR #3331, opened 2021-01-13
#3298 development ble manager

### LINKED ISSUE #3298: Development BleManager

Using `react-native-ble-plx` BLE manager throws a runtime if you try and use it on a device which doesn't support bluetooth. This is all emulators.

Port this class https://github.com/openmsupply/msupply-cold-chain/blob/develop/src/common/services/Bluetooth/DevBleManager.js over and use it when `__DEV__ === true` 

Kind of like https://github.com/openmsupply/msupply-cold-chain/blob/43f882c09732480d4b211536b98b2561de803070/src/ui/containers/DependencyContainer.js#L54

I dunno if you can determine if you're using an emulator or not, if you can do that, then it would probably be good to use that as the condition.

## Implementation

I think we will initialize an instance of this object in `mSupplyMobileAppContainer`, probably. So check the condition there.

N/A

## Additional context

N/A

katherine-sussol:
Fixes #3298 

## Change summary

- Added copy of https://github.com/openmsupply/msupply-cold-chain/blob/develop/src/common/services/Bluetooth/DevBleManager.js 
- Modified BleService into singleton (if causes problems later - investigate passing instance into thunk(?))
- Added [isEmulator()](https://github.com/react-native-device-info/react-native-device-info#isemulator) check and Initialise BleService Singleton in mSupplyMobileApp 
  - Done inside `usingVaccines` check, as it looks like we only request BT Permissions right now if this module is enabled

## Testing

- [ ] Dev Testing    
Make `componentDidMount` async, and add some methods into it on both emulator and real device and compare/observe results. e.g
```
    await this.initialiseBtService();
    await BleService().blinkWithRetries('E7:6F:FC:15:13:F2', 3);
    await BleService().getInfoWithRetries('E7:6F:FC:15:13:F2', 3);
```
Main issue/fix is to verify no RSOD on emulator when using BleService

Not sure... we don't really use the BleService anywhere yet

### DIFF
--- a/src/bluetooth/BleService.js
+++ b/src/bluetooth/BleService.js
@@ -24,7 +24,7 @@ const base64FromString = string => Buffer.from(string, 'utf-8').toString('base64
  *
  *
  */
-export class BleService {
+class BleService {
   constructor(manager = new BleManager()) {
     this.manager = manager;
   }
@@ -366,3 +366,14 @@ export class BleService {
     );
   };
 }
+
+let BleServiceInstance;
+
+export const getBleServiceInstance = manager => {
+  if (!BleServiceInstance) {
+    BleServiceInstance = new BleService(manager);
+  }
+  return BleServiceInstance;
+};
+
+export default getBleServiceInstance;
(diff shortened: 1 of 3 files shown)


joshxg (approved):
Works perfectly. I had a few thoughts about exporting the right instance from the get go using something like [top level await](https://v8.dev/features/top-level-await#:~:text=Top%2Dlevel%20await%20enables%20developers,they%20start%20evaluating%20their%20body. ) or some other hacky things, but I prefer your solution having it quite obvious and non-magical

PR #3337, opened 2021-01-15
#3336 Paper component

### LINKED ISSUE #3336: Paper background component

A paper-looking component.

image

Container component 

## Implementation

Just a simple component which renders children in a paper-like container. With a header also. 

N/A

## Additional context

Similar component used in dispensing

joshxg:
Fixes #3336 

## Change summary

- Adds two components `Paper` and `PaperSection`.
- Two components because there are some areas with a header and some without.
- Thought abstracting the `Paper` away from the component with the dedicated header gave more flexibility.
- Unsure about `PaperSection` naming.

## Testing

Internal

N/A

### DIFF
--- a/src/widgets/Paper.js
+++ b/src/widgets/Paper.js
@@ -1,0 +1,44 @@
+/* eslint-disable react/forbid-prop-types */
+import React from 'react';
+import PropTypes from 'prop-types';
+import { View } from 'react-native';
+
+import { WHITE } from '../globalStyles';
+
+export const Paper = ({ width, height, style, children }) => {
+  // If width and height are passed hard fix them in the container style,
+  // let the axis which isn't passed flex. Plain width or height properties
+  // sometimes act as suggestions when the container is flexing so use
+  // max/min properties.
+  const internalContainerStyle = { ...style };
+  if (width) {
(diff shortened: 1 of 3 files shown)


src/widgets/Paper.js
```diff
@@ -0,0 +1,44 @@
...
+import { View } from 'react-native';
+
+import { WHITE } from '../globalStyles';
+
+export const Paper = ({ width, height, style, children }) => {
+  // If width and height are passed hard fix them in the container style,
+  // let the axis which isn't passed flex. Plain width or height properties
+  // sometimes act as suggestions when the container is flexing so use
+  // max/min properties.
+  const internalContainerStyle = { ...style };
```
mark-prins, line 13: an alternative is possibly to disallow flex grow & shrink and to set the initial size:

`flex: '0 ${width}px'`

which retains flex, but then doesn't help with height so is perhaps not useful

mark-prins (approved):
Looks good and compiles fine!

I'm not seeing the value of having both Paper and PaperSection, but I expect that's just me.

I would have gone with just Paper and had the heading element display if a header text is passed in

PR #3357, opened 2021-01-18
#3351 bluetooth scan start/stop thunks

### LINKED ISSUE #3351: Bluetooth Thunks - Scan Start/Stop
Child of: #3319

Thunks required to start and stop a scan for BT sensors.

Thunks available for UI to utilise

## Implementation
N/A - Refer to Feature

N/A

## Additional context
References
- CCE Impl: https://github.com/openmsupply/msupply-cold-chain/tree/develop/src/features/Bluetooth/Scan
- Mobile prev Impl: https://github.com/openmsupply/mobile/blob/master/src/actions/TemperatureSyncActions.js

katherine-sussol:
Fixes #3351

## Change summary

- Have added actions to start/stop scanning and adds found sensors to the store (not sure if `stop` was needed but thinking if 'connect' is clicked on the UI or someone navigates away scanning should stop?)
- Used the toasts from `TemperatureSyncActions` when BT/Locations are not available, not sure if appropriate or not

## Testing

- [ ] Internal only 

N/A

### DIFF
--- a/src/actions/VaccineActions.js
+++ b/src/actions/VaccineActions.js
@@ -45,12 +45,12 @@ const scanForSensors = () => async dispatch => {
   BleService().scanForSensors(deviceCallback);
 };
 
-// eslint-disable-next-line no-unused-vars
 const stopSensorScan = () => async dispatch => {
+  dispatch(scanStop());
   BleService().stopScan();
-  console.log('scan stopped');
 };
 
 export const TemperatureSyncActions = {
   startSensorScan,
+  stopSensorScan,
 };
--- a/src/actions/VaccineActions.js
+++ b/src/actions/VaccineActions.js
@@ -5,7 +5,7 @@
 
 import { ToastAndroid } from 'react-native';
 import { PermissionSelectors } from '../selectors/permission';
-import selectScannedSensorAddresses from '../selectors/vaccine';
+import selectScannedSensors from '../selectors/vaccine';
 import { PermissionActions } from './PermissionActions';
 import BleService from '../bluetooth/BleService';
 import { syncStrings } from '../localization/index';
@@ -51,7 +51,7 @@ const scanForSensors = () => async (dispatch, getState) => {
     const { id: macAddress } = device;
 
     if (macAddress) {
-      const alreadyScanned = selectScannedSensorAddresses(getState());
+      const alreadyScanned = selectScannedSensors(getState());
       const alreadySaved = UIDatabase.get('Sensor', macAddress, 'macAddress');
 
       if (!alreadyScanned?.includes(macAddress) && !alreadySaved) {
--- a/src/actions/VaccineActions.js
+++ b/src/actions/VaccineActions.js
@@ -51,10 +51,10 @@ const scanForSensors = () => async (dispatch, getState) => {
     const { id: macAddress } = device;
 
     if (macAddress) {
-      const alreadyFound = selectScannedSensorAddresses(getState());
+      const alreadyScanned = selectScannedSensorAddresses(getState());
       const alreadySaved = UIDatabase.get('Sensor', macAddress, 'macAddress');
 
-      if (!alreadyFound.includes(macAddress) && !alreadySaved) {
+      if (!alreadyScanned?.includes(macAddress) && !alreadySaved) {
         dispatch(sensorFound(macAddress));
       }
     }
--- a/src/actions/VaccineActions.js
+++ b/src/actions/VaccineActions.js
@@ -48,14 +48,14 @@ const scanForSensors = () => async (dispatch, getState) => {
   dispatch(scanStart());
 
   const deviceCallback = device => {
-    const { id } = device;
+    const { id: macAddress } = device;
 
-    if (id) {
+    if (macAddress) {
       const alreadyFound = selectScannedSensorAddresses(getState());
-      const alreadySaved = UIDatabase.get('Sensor', id, 'macAddress');
+      const alreadySaved = UIDatabase.get('Sensor', macAddress, 'macAddress');
 
-      if (!alreadyFound.includes(id) && !alreadySaved) {
-        dispatch(sensorFound(id));
+      if (!alreadyFound.includes(macAddress) && !alreadySaved) {
+        dispatch(sensorFound(macAddress));
       }
     }
   };
--- a/src/actions/VaccineActions.js
+++ b/src/actions/VaccineActions.js
@@ -5,6 +5,7 @@
  */
 
 import { PermissionSelectors } from '../selectors/permission';
+import selectScannedSensors from '../selectors/vaccine';
 import { PermissionActions } from './PermissionActions';
 import BleService from '../bluetooth/BleService';
 
@@ -13,14 +14,14 @@ export const VACCINE_ACTIONS = {
   ERROR_LOCATION_DISABLED: 'Vaccine/errorLocationDisabled',
 
   SCAN_START: 'Vaccine/sensorScanStart',
-  SCAN_COMPLETE: 'Vaccine/sensorScanComplete',
   SCAN_ERROR: 'Vaccine/sensorScanError',
   SCAN_STOP: 'Vaccine/sensorScanStop',
+  SENSOR_FOUND: 'Vaccine/sensorFound',
 };
 
 const scanStart = () => ({ type: VACCINE_ACTIONS.SCAN_START });
-// eslint-disable-next-line no-unused-vars
 const scanStop = () => ({ type: VACCINE_ACTIONS.SCAN_STOP });
+const sensorFound = macAddress => ({ type: VACCINE_ACTIONS.SENSOR_FOUND, payload: { macAddress } });
 
 const startSensorScan = () => async (dispatch, getState) => {
   const bluetoothEnabled = PermissionSelectors.bluetooth(getState());
@@ -30,16 +31,22 @@ const startSensorScan = () => async (dispatch, getState) => {
   if (!locationPermission) await dispatch(PermissionActions.requestLocation());
 
   if (!(bluetoothEnabled && locationPermission)) return null;
-  await dispatch(scanForSensors());
-  // TODO
+  dispatch(scanForSensors());
+
+  // Do we need to do something here?
   return null;
 };
 
-const scanForSensors = () => async dispatch => {
+const scanForSensors = () => async (dispatch, getState) => {
   dispatch(scanStart());
 
   const deviceCallback = device => {
-    console.log(device);
+    const alreadyFound = selectScannedSensors(getState());
+
+    // TODO: Filter out already saved sensors?
+    if (!alreadyFound.includes(device?.id)) {
+      dispatch(sensorFound(device?.id));
+    }
   };
 
   BleService().scanForSensors(deviceCallback);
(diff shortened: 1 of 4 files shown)


src/reducers/VaccineReducer.js
```diff
@@ -0,0 +1,39 @@
...
+      return {
+        ...state,
+        isScanning: false,
+      };
+    }
+    case VACCINE_ACTIONS.SENSOR_FOUND: {
+      const { payload } = action;
+      const { macAddress } = payload;
+
+      const { scannedSensors } = state;
```
mark-prins, line 28: this may be ok, but I think it's sailing close to the wind - in that it is almost mutating an object in state.
better to use `[...scannedSensors, macAddress]` though I would go slightly further and only update if there's a new one added 😃

src/reducers/VaccineReducer.js
```diff
@@ -0,0 +1,39 @@
...
+      scannedSensors.push(macAddress);
```
mark-prins, line 29: perhaps `sensorAddresses`? though that feels a little clumsy to me. I'm just wondering at the equivalence of a mac address and a 'sensor' in the naming of the collection

src/actions/VaccineActions.js
```diff
@@ -0,0 +1,74 @@
...
+
+  BleService().scanForSensors(deviceCallback);
+};
+
+const stopSensorScan = () => async dispatch => {
+  dispatch(scanStop());
+  BleService().stopScan();
+};
+
+export const TemperatureSyncActions = {
```
mark-prins, line 71: I think should be `VaccineActions` to match the filename

src/actions/VaccineActions.js
```diff
@@ -0,0 +1,74 @@
...
+  dispatch(scanStart());
+
+  const deviceCallback = device => {
+    const { id } = device;
+
+    if (id) {
+      const alreadyFound = selectScannedSensors(getState());
+      const alreadySaved = UIDatabase.get('Sensor', id, 'macAddress');
+
+      if (!alreadyFound.includes(id) && !alreadySaved) {
```
mark-prins, line 57: aha! you've cleverly pre-empted a comment I made in the reducer!
Or perhaps I've reviewed in the wrong order. anyhoo.. the reducer is taking the `id` here and calling it `macAddress` - would you mind changing the param in reducer to be `id`?
or save the whole `device`? that would get around my other naming concern, about storing an address/id in a collection called `sensors`

src/actions/VaccineActions.js
```diff
@@ -0,0 +1,74 @@
...
+
+export const VACCINE_ACTIONS = {
+  SCAN_START: 'Vaccine/sensorScanStart',
+  SCAN_STOP: 'Vaccine/sensorScanStop',
+  SENSOR_FOUND: 'Vaccine/sensorFound',
+};
+
+const scanStart = () => ({ type: VACCINE_ACTIONS.SCAN_START });
+const scanStop = () => ({ type: VACCINE_ACTIONS.SCAN_STOP });
+const sensorFound = macAddress => ({ type: VACCINE_ACTIONS.SENSOR_FOUND, payload: { macAddress } });
```
mark-prins, line 22: I prefer `id` rather than `macAddress`

mark-prins (commented):
Looks great! thanks - I have some naming questions, apologies for reviewing from the bottom up, that may make things a little unclear (!)

src/actions/VaccineActions.js
```diff
@@ -0,0 +1,74 @@
...
+
+const startSensorScan = () => async (dispatch, getState) => {
+  const bluetoothEnabled = PermissionSelectors.bluetooth(getState());
+  const locationPermission = PermissionSelectors.location(getState());
+  // Ensure the correct permissions before initiating a new sync process.
+  if (!bluetoothEnabled) await dispatch(PermissionActions.requestBluetooth());
+  if (!locationPermission) await dispatch(PermissionActions.requestLocation());
+
+  if (!bluetoothEnabled) {
+    ToastAndroid.show(syncStrings.please_enable_bluetooth, ToastAndroid.LONG);
```
mark-prins, line 32: both of these strings include `We tried syncing temperatures, ` which isn't really relevant here.  Perhaps use `bluetooth_disabled` instead in this case.

src/actions/VaccineActions.js
katherine-sussol: I think so too! e9b62e87

src/actions/VaccineActions.js
katherine-sussol: Hmm, I didn't want to get it confused with the DB field Sensor.id (which is different from Sensor.macAddress) but then this isn't really anything to do with the DB at this stage so can potentially do a rename.

mark-prins: what about stuffing the whole 'device' in rather than just the id? or macAddress? is that overkill?
I just ask because I needed the `id` to make the sensor blink, but I think we'd want the `macAddress` for display purposes ( it matches the label on the back of the sensor ).
Disclaimer.. all of this is specific to Blue Maestro, so we should probs think about generic objects too!

src/selectors/vaccine.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * mSupply Mobile
+ * Sustainable Solutions (NZ) Ltd. 2021
+ */
+
+export const selectScannedSensors = ({ state }) => {
```
mark-prins, line 6: Side note: I had trouble when using this - as I was passing in the whole state, and the reducer isn't called, so there isn't event a `vaccine` object in state. To get things working, I've added the vaccine reducer to the index.js and have used `vaccine` here rather than `state` 😃

src/reducers/VaccineReducer.js
katherine-sussol: Oops, I agree it's a little misleading as we are not really storing sensors 😆 . I tried `scannedSensorAddresses`? Too long? 8dbd7ef0

src/actions/VaccineActions.js
katherine-sussol: Oh thanks - forgot to double-check `syncStrings.json`! d5ab9069

src/reducers/VaccineReducer.js
joshxg: yeah `push` mutates the array which will mean there won't be any diff and the UI won't re-render. 


[ I suspect this is copied from `ColdChain` which uses https://github.com/immerjs/immer - which is a library which helps you update state by using a proxy of your state object which you update mutably, but causes immutable updates to the real state ]

src/reducers/VaccineReducer.js
```diff
@@ -0,0 +1,39 @@
...
+
+      return {
+        ...state,
+        scannedSensorAddresses,
+      };
+    }
+    default:
+      return state;
+  }
+};
```
joshxg, line 39: can add this to the redux store by adding it to `src/reducer/index.js` (If at any point there are realm objects added into the state, we will need to blacklist the state from being persisted in `src/store.js`

src/actions/VaccineActions.js
joshxg: `get` can also return `null` https://github.com/sussol/react-native-database/blob/81e18edcfe935d796412fa8ffcae6ba69c2a4e94/src/Database.js#L84-L87

and `null.includes()` throws - suggest `?.includes()`

src/actions/VaccineActions.js
joshxg: Annoying thing is the `id` in the `device` object returned by the scanner is actually a mac - it's a `device` object which is returned by the native module and not a `sensor` - so it is a consistent object returned (where the id is the mac, and the mac/id is used to send command via ble)

All that to say that it might be better to rename `id` below to `macAddress`?

joshxg (changes requested):
Looks great - couple of suggestions. Wasn't sure if this wasn't added to redux on purpose (if so, all good!)

src/actions/VaccineActions.js
mark-prins: me too! 😆  only noticed because I broke something else while testing..

src/reducers/VaccineReducer.js
mark-prins: nice tip re `immer` thanks @joshxg I'd noticed that earlier and hadn't made the connection

src/actions/VaccineActions.js
mark-prins: yup. thanks - you are spot on Mr G.

katherine-sussol: Sounds good, should be a lot clearer now 👍  7bf78972

src/actions/VaccineActions.js
katherine-sussol: Cool thanks for the repo link, that helps clarify things! In this instance I think `alreadyFound` is from the state which is empty array by default but agree better to be safe 👍. I also renamed to `alreadyScanned` because this made me reconsider the original naming hehe.  6b603bf7

src/reducers/VaccineReducer.js
katherine-sussol: Good to know, thanks @mark-prins for the suggestion and @joshxg for the link! 4ee6470a

src/reducers/VaccineReducer.js
katherine-sussol: Added with c239fc74. I did wonder about adding it if it wasn't being used yet. But it's probably more helpful than not to 😅

katherine-sussol:
Thanks both @mark-prins @joshxg for comments, I think I have covered them all but let me know if I missed anything 🙀

src/actions/VaccineActions.js
mark-prins: ah. nice - the renaming makes it clear! thanks.

src/selectors/vaccine.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * mSupply Mobile
+ * Sustainable Solutions (NZ) Ltd. 2021
+ */
+
+export const selectScannedSensorAddresses = ({ state }) => {
+  const { scannedSensorAddresses } = state;
+  return scannedSensorAddresses;
+};
```
mark-prins, line 9: I used this while testing, as I was using the conditional for `isScanning`. The `state` property won't exist on the top level state now that it has been added as `vaccine` :) 

```suggestion
export const selectScannedSensors = ({ vaccine }) => {
  const { scannedSensorAddresses = [] } = vaccine || {};
  return scannedSensorAddresses;
};

export const selectIsScanning = ({ vaccine }) => {
  const { isScanning = false } = vaccine || {};
  return isScanning;
};
```

mark-prins (approved):
One more tweak, in the selector, please
Apart from that.. all the changes look great. Thanks!

src/selectors/vaccine.js
katherine-sussol: Ahh oops sorry about that. Thanks for the tweak 😅

src/reducers/index.js
```diff
@@ -50,4 +51,5 @@ export default combineReducers({
   breach: BreachReducer,
   rowDetail: RowDetailReducer,
   permission: PermissionReducer,
+  vaccine: VaccineReducer,
```
joshxg, line 54: I kind of have a feeling we might want this named something more appropriate to bluetooth but if we do we can change it then heh

joshxg (approved):
amazing 🎉

PR #3363, opened 2021-01-21
#3348 Step one of new sensor flow

### LINKED ISSUE #3348: New Sensor Flow/Page

The new sensor flow and page

As per the marvel prototype

## Implementation

Quite a big page with a lot of steps. Going to break it down into the following tasks:

- [ ] `Stepper`
- [ ] Make `IconButton` have a `Rectangle` variant?
- [ ] Make step1 component
- [ ] Make step2 component
- [ ] Make step3 component
- [ ] Logic to send bluetooth commands + save a new sensor

N/A

## Additional context

Resources

- https://marvelapp.com/prototype/55aea8f/screen/75648024
image

- https://drive.google.com/file/d/1wLg12JAQyrgB6D3F_X0O9LBZwJs8PcJM/view?usp=sharing
image

- https://drive.google.com/file/d/15SsY4YCmhpSgiihU1Q7FpZMnqCUkZcIp/view?usp=sharing
image

- https://drive.google.com/file/d/15SsY4YCmhpSgiihU1Q7FpZMnqCUkZcIp/view?usp=sharing
image

joshxg:
Fixes #3348 

## Change summary

- Added a button to settings page to navigate to the new sensor page
- Added first tab of the new sensor flow:
	- Scans when you navigate to it, might need to add some other logic to handle the case when bluetooth is not working good
	- Blink a sensor, disables all others, shows a spinner
	- clicking connect navigates to the next tab
- Added `react-native-animatable` for easy animations
- Added `NewSensor` state- will have multiple `temperatureBreachConfig`s and other various sensor details as plain objects before saving at the end of the process
- Updated blink state

## Testing

Internal?

- Possibly should be the scanning state into the new sensor state?
- Not pixel perfect to the design. Getting there? Also factoring in time to change existing components and time to do minor minor things which make a difference but.. time?

newsensor1


src/Store.js
```diff
@@ -35,6 +35,8 @@ const persistConfig = {
     'rowDetail',
     'permission',
     'cashTransaction',
+    'vaccine',
+    'newSensor',
```
joshxg, line 39: Just added this to the blacklist, expecting realm objects

src/actions/VaccineActions.js
```diff
@@ -48,23 +52,27 @@ const withPermissions = async (dispatch, getState, func) => {
     return null;
   }
 
-  return func(dispatch, state);
+  return func(dispatch, getState);
```
joshxg, line 55: passing `getState` as just `state` gets stale in the scan call closure

src/bluetooth/BleService.js
```diff
@@ -373,6 +373,11 @@ export const getBleServiceInstance = manager => {
   if (!BleServiceInstance) {
     BleServiceInstance = new BleService(manager);
   }
+
+  if (manager) {
+    BleServiceInstance = new BleService(manager);
+  }
```
joshxg, line 379: added this to be able to update the service with a manager if needed.. think it shouldn't make much change?

src/bluetooth/DevBleManager.js
```diff
@@ -1,3 +1,16 @@
+const delay = ms => new Promise(resolve => setTimeout(resolve, ms));
+
+const randomMacs = [
+  'AA:BB:CC:DD:EE:FF',
+  '12:34:56:78:09:10',
+  'AB:CD:EF:GH:IJ:KL',
+  'TH:IS:IS:TO:TE:SA',
+  'RE:AL:MA:CO:MG:AA',
+  'HE:HE:WO:ND:ER:IF',
+  'AN:YO:NE:WI:LL:NO',
```
joshxg, line 10: Added this to have multiple options when scanning

src/bluetooth/DevBleManager.js
```diff
@@ -82,21 +95,21 @@ export class DevBleManager {
...
 
-    setInterval(
-      () =>
-        callback(_, {
-          id: 'AB:CD:EF:GH:IJ:KL',
-          manufacturerData: 'MwEBDAN0AFkBtwEMA3QAWQG3AMwCrAAAAAAA',
-        }),
-      1000
-    );
+    this.scannerInterval = setInterval(() => {
```
joshxg, line 98: Wasn't stopping the scan when deving

src/pages/NewSensor/NewSensorPage.js
```diff
@@ -12,13 +13,11 @@ const TabThree = () => (
...
+  { component: NewSensorStepOne, name: 'prescription', title: '' },
   { component: TabTwo, name: 'prescription', title: '' },
   { component: TabThree, name: 'prescription', title: '' },
 ];
 
 export const NewSensorPage = () => (
-  <>
-    <Wizard tabs={tabs} useNewStepper />
-  </>
+  <Wizard captureUncaughtGestures={false} tabs={tabs} useNewStepper />
```
joshxg, line 22: scrolling wasn't working well with this capturing gestures? I dunno. Weird one

src/widgets/Stepper.js
```diff
@@ -200,8 +199,6 @@ export const Stepper = ({ useNewStepper, ...stepperProps }) =>
 
 const localStyles = StyleSheet.create({
   oldContainer: {
-    borderColor: SHADOW_BORDER,
```
joshxg, line 203: updating styles - bit weird but dispensing looks the same

src/actions/VaccineActions.js
mark-prins: yeah, I wondered about that. thanks - over eager optimisation on my part that one

src/bluetooth/DevBleManager.js
mark-prins: 🤣

src/actions/NewSensorActions.js
```diff
@@ -0,0 +1,7 @@
+export const NEW_SENSOR_ACTIONS = {
+  SELECT: 'SensorActions/select',
+};
+
+const select = macAddress => ({ type: NEW_SENSOR_ACTIONS.SELECT, payload: { macAddress } });
+
+export const NewSensorActions = { select };
```
mark-prins, line 7: will `NewSensorActions` will become `SensorActions` at a later stage?
the const is `NEW_SENSOR_ACTIONS` though the text string is `SensorActions` which made me think that this file is for sensor actions. not for new sensors.

src/bluetooth/BleService.js
mark-prins: took me quick review to figure out how this worked with the intention of a singleton instance
to help a simpleton like me to use the singleton like this, a comment might help avoid confusion. maybe?
I had to check the usage, and saw that we only use without params, so is fine.
```
/**
 * Exports a singleton instance of the BleService
 * import the default and call without parameters to return the singleton
 * Call with a manager parameter to re-instantiate
 */

File: src/bluetooth/DevBleManager.js (hunk 4 above)
Comment (mark-prins) [reply src/bluetooth/DevBleManager.js:10]: Is it missing a 'T'?
`THIS IS TO TES[T] A REAL MAC OMG AA HE HE WONDER IF ANYONE WILL NOTICE THIS JOKE`

**Review by mark-prins — APPROVED** [review]
Awesome! 🎉 

looking really good. slightly troubled by the use of 'new' in the naming, as I'm worried that we'll have files called 'new' in the codebase forever. checking that they'll be renamed later, if that's appropriate, or that they are actually for new stuff, i.e. finding new sensors. 
couple of follow on questions
- if location permission is required, you have to enable, get told in the Toast that is blocked and come back to the sensor scanning screen. no biggie, but would be nicer if it scanned after enabling.

- when scanning it raised an error after finding three sensors ( but not mine 😢  )
```Possible Unhandled Promise Rejection
null is not an object (evaluating 'BleManager.sendCommand')
```

src/actions/NewSensorActions.js
joshxg: I thought this particular case would stay as `new` as it is the 'new' sensor flow/feature. Happier? Sadder?

joshxg:
I think the locations thing needs to be handled but i'm not quite sure the best way. Maybe if scanning isn't enabled, a button 'start scan' and some feedback about enabling OS features to enable the button?

src/bluetooth/BleService.js
joshxg: Possibly better to add a field `setManager` on the class?

src/bluetooth/DevBleManager.js
joshxg: woah totes crazy!

mark-prins:
> I think the locations thing needs to be handled...

That's a good idea. what about disabling the scanning button, and have an 'enable permissions' button or similar. have the UI re-render on permissions change so that the buttons enable

joshxg:
@mark-prins Cool. I might do that in a new PR cause I suspect it will be more work than you might expect as the permissions are at the OS level, might need to sort out some listeners for re-rendering etc

src/actions/NewSensorActions.js
mark-prins: sill vaguely confused.. but I think all good. you know what you're doing.
this is for working with new sensors, so the 'New' is appropriate and will stay. is that it?

mark-prins:
yah - I thought it might be fussy to actually get working. and it's not a big deal.. so I'd say create an issue and do it later.
it only affects users once, after all.

src/actions/NewSensorActions.js
joshxg: Yeah exactly (except for the part of me knowing what I am doing) my thinking - I figure it will be clearer when there aren't the .. "other new prefixes" confusing things...

src/actions/VaccineActions.js
```diff
@@ -48,23 +52,27 @@ const withPermissions = async (dispatch, getState, func) => {
...
 
   const deviceCallback = device => {
     const { id: macAddress } = device;
 
     if (macAddress) {
-      const alreadyScanned = selectScannedSensors(state);
+      const alreadyScanned = selectScannedSensors(getState());
       const alreadySaved = UIDatabase.get('Sensor', macAddress, 'macAddress');
 
+      console.log(alreadyScanned, macAddress);
```
katherine-sussol: Looks like this might be some test code?

katherine-sussol (commented):
Nice! 🤩  This is quite a satisfying page to load

src/actions/VaccineActions.js
joshxg: cd0adf6091f4f55a1838f563f51d0304e24eab25

src/bluetooth/BleService.js
joshxg: 53b8542a5869aabbe799e6012d4b90aee3c53314

joshxg:
Thanks everyone

PR #3366, opened 2021-01-21
#3354 update logging interval

### LINKED ISSUE #3354: Bluetooth Thunks - Update Logging Interval
Child of: #3319

Thunks required to send a command to update a sensor's logging interval. Note that updating the logging interval will clear the sensor's logs.

Thunks available for UI to utilise

## Implementation
N/A - Refer to Feature

N/A

## Additional context
References
- CCE Impl: https://github.com/openmsupply/msupply-cold-chain/tree/develop/src/features/Bluetooth
- Mobile Prev Imp: https://github.com/openmsupply/mobile/blob/master/src/actions/TemperatureSyncActions.js (resetLogFrequency - this just updates it to 300s)

mark-prins:
Fixes #3354 

Adds actions and reducers for setting the log frequency. I've added the option to call with only the mac address, which will reset the frequency to the default of 300. Don't know if that will be useful or not?

Also not sure if you want to continue with the term 'frequency' or use 'interval' which is what BlueMaestro are using

## Change summary

tested with a couple of sensors 
- E7:F6:88:07:03:DA which is working nicely and is now set to 200s
- FC:C0:EE:9D:80:A3 which is not playing ball and throws an error. Good for error testing 😉

### DIFF
--- a/src/actions/VaccineActions.js
+++ b/src/actions/VaccineActions.js
@@ -16,20 +16,20 @@ export const VACCINE_ACTIONS = {
   SCAN_START: 'Vaccine/sensorScanStart',
   SCAN_STOP: 'Vaccine/sensorScanStop',
   SENSOR_FOUND: 'Vaccine/sensorFound',
-  SET_LOG_FREQUENCY_ERROR: 'Vaccine/setLogFrequencyError',
-  SET_LOG_FREQUENCY_START: 'Vaccine/setLogFrequencyStart',
-  SET_LOG_FREQUENCY_SUCCESS: 'Vaccine/setLogFrequencySuccess',
+  SET_LOG_INTERVAL_ERROR: 'Vaccine/setLogIntervalError',
+  SET_LOG_INTERVAL_START: 'Vaccine/setLogIntervalStart',
+  SET_LOG_INTERVAL_SUCCESS: 'Vaccine/setLogIntervalSuccess',
 };
 
 const scanStart = () => ({ type: VACCINE_ACTIONS.SCAN_START });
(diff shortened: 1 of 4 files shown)


src/actions/VaccineActions.js
```diff
@@ -75,6 +84,35 @@ const scanForSensors = (dispatch, state) => {
   BleService().scanForSensors(deviceCallback);
 };
 
+const setLogFrequency = (macAddress, frequency) => async dispatch => {
+  let ok = false;
+  let error = '';
```
joshxg, line 89: I think this error is always going to be that the regex failed, unless it is overridden on L98 so maybe we can just have that as the default?

joshxg (approved):
amazing 👍

src/actions/VaccineActions.js
mark-prins: I've managed to get 
`[BleError: Device ? is already connected]`
and 
`[BleError: Device FC:C0:EE:9D:80:A4 was disconnected]`
returned as errors - so I think worth retaining.

joshxg: But at the moment `error` is only ever assigned `e` - in the case where regex fails, it isn't assigned anything is it?

mark-prins: Huh! you and your 'logic' 🙄 

thanks - 8dc3dd67

katherine-sussol (commented):
W.r.t naming I prefer 'interval' just because that's what's used in the UI and the temperature log manager, that being said it's also pretty obvious what `frequency` means 🤷 .

mark-prins:
Thanks @katherine-sussol! 😄 
9559a9d8

PR #3360, opened 2021-01-19
#3352 bluetooth thunks blink

mark-prins:
Fixes #3352 

## Change summary

Adds the blink action to the vaccines actions.

## Testing

To test I added a button to `MenuPage` which called the blink action with a specific mac address

- [ ] Call action with mac address

### Related areas to think about

If there are any general areas of the codebase your changes might have side affects on, mention them here

PR #3369, opened 2021-01-22
#3355 bluetooth thunk log download

### LINKED ISSUE #3355: Bluetooth Thunks - Download temperatures
Child of: #3319

Thunks required to send a command to download temperature logs from a specified sensor

- Will happens continuously in the background(?)
- Synchronous download per sensor (too difficult with BT to send multiple commands and know which logs belong to which sensor)
- No progress bar required

## Implementation
- Utilise TemperatureLogManager and BleService

## Additional context
References
- CCE Impl: https://github.com/openmsupply/msupply-cold-chain/tree/develop/src/features/Bluetooth/Download
- Mobile: https://github.com/openmsupply/mobile/blob/develop/src/actions/TemperatureSyncActions.js (DownloadLogs)

katherine-sussol:
Fixes #3355

## Change summary

- Adds logic to download and save sensor logs
- Will create a separate issue to create and save breaches

## Testing

- [ ] Stole @mark-prins button and created a button with startDownloadLogs action https://github.com/openmsupply/mobile/pull/3360/commits/37fb179829dc5dc454ec4d76fa048556c86ba6b7   

Fiddled with the TemperatureLogManager a bit due to some weird issues observed


src/bluetooth/TemperatureLogManager.js
```diff
@@ -74,3 +75,14 @@ export class TemperatureLogManager {
 
   saveLogs = async logsToSave => this.db.upsertTemperatureLog(logsToSave);
 }
+
+let TemperatureLogManagerInstance;
+
+export const getTemperatureLogManagerInstance = (dbService, utils) => {
```
katherine-sussol, line 81: I'm not sure about this - would it be better to expose a setter similar to what we've done for BleService?

src/bluetooth/TemperatureLogManager.js
```diff
@@ -65,7 +66,7 @@ export class TemperatureLogManager {
 
     return logsToSave.map(({ temperature }, i) => {
       const offset = logInterval * i;
-      const timestamp = Number(moment(initial).add(offset, 's').format('X'));
+      const timestamp = Number(moment(initial).add(offset, 's'));
```
katherine-sussol, line 69: Logs were being generated in the 1970s (pretty sure the sensor wasn't around then!).
Alternatively could multiply the offset by 1000 and then `format('X')`?

src/actions/VaccineActions.js
```diff
@@ -25,6 +32,73 @@ const blinkStop = () => ({ type: VACCINE_ACTIONS.BLINK_STOP });
...
+});
+
+const downloadAllLogs = () => async dispatch => {
+  dispatch(downloadLogsStart());
+
+  // Ensure there are some sensors which have been assigned a location before syncing.
+  const sensors = UIDatabase.objects('Sensor').filtered('location != null && isActive == true');
+
+  if (!sensors.length) {
+    // TODO: Should we do something if there are errors?
```
mark-prins, line 52: Could add a Toast notification for now - though I think not needed for log downloading. If there are no sensors then there's nothing to do.

src/bluetooth/TemperatureLogManager.js
mark-prins: Good like this, I think.

src/bluetooth/TemperatureLogManager.test.js
```diff
@@ -96,6 +96,7 @@ describe('DownloadManager: createLogs', () => {
     // a log for anytime past 600, so we assume any excess logs are passed records which have been
     // saved previously and just save the two 'most recent'.
     const sensor = { id: 'a', logInterval: 300 };
+    const MILLISECONDS_PER_SECOND = 1000;
```
mark-prins, line 99: or `import { MILLISECONDS_PER_SECOND } from '../database/utilities/constants';` at the top of the file, with an `export` in the constants.js?

mark-prins (approved):
Looks Good! 🎉

src/bluetooth/TemperatureLogManager.js
joshxg: Hm yeah, CCA using just seconds. Good to use ms though

joshxg (changes requested):
Nice!

Just the `await`ing in a for each which should change (annoying one). Th eother suggestion regarding the store updating can wait till another issue if you prefer, but I am basically thinking about state for the `3m since last sync`.. it can technically be calculated from the timestamp of the last temperature log but knowing when the last download failed would be helpful, rather than trying to determine the last sync time v log interval v last download attempt (somehow)

image

src/actions/VaccineActions.js
joshxg: I thought about this more/again and I think it'd be better to not do my suggestion... 🤣

src/bluetooth/TemperatureLogManager.test.js
katherine-sussol: Hm, yeah I think I went over the top with trying to not include dependencies, constants should be pretty safe haha. Also found `src/utilities/constant` which has a MILLISECONDS constant so I used that one 61d41166

src/actions/VaccineActions.js
```diff
@@ -25,6 +32,73 @@ const blinkStop = () => ({ type: VACCINE_ACTIONS.BLINK_STOP });
...
+    dispatch(downloadLogsError());
+    return null;
+  }
+
+  sensors.forEach(async sensor => {
+    await dispatch(downloadLogsFromSensor(sensor));
+  });
```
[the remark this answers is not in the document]
katherine-sussol: Oops, good catch, thanks. I've swapped to a plain for loop with 94dedfd8

src/actions/VaccineActions.js
katherine-sussol: I was thinking about this a bit too and think we may need to go back to the temperature log table anyway to fetch the last sync timestamp per sensor? There will likely be a period of time when a user first launches the app and won't know the last sync time until a successful one is done (I think? If my understanding of the way people use the app is correct - unless people just never close the app and tend to keep it running? Then this doesn't apply).

Maybe something like on start, fetch last temp logs and calculate/guess timestamps based on the record, then after each successful log download, update the timestamp/sync state? 

I also wanted to add some 🍞 🔥 to the UI  but I couldn't think of a way that would be helpful rather than annoying haha, I agree a failure indicator somewhere might be nice. Could that be an idea for when to turn a sensor orange/red on the sensor page? https://marvelapp.com/prototype/55aea8f/screen/75648034. Or turn it into a sync fail icon?? 🤷

joshxg: You're right about determining the last sync date. Also good idea regarding the circle: The last sync failing would be easier as state (which we can persist between app starts, if needed, kind of like a database!) I'm not sure exactly where the state should live, though. Maybe a new issue? Can determine exactly what state should be stored regarding the downloading of temperatures?

katherine-sussol: Cool yeah, I think best for a new issue, have created #3380 for it. That's good to know about persisting between app starts!

joshxg (approved):

PR #3388, opened 2021-01-26
#3375 passively downloading temperatures

### LINKED ISSUE #3375: Passively downloading temperatures

Need to use the new bluetooth logic to passively download temperatures

- A way to schedule a task to download temperatures in the 'background'

## Implementation

TBD

TBD

## Additional context

N/A

katherine-sussol:
Fixes #3375

## Change summary

- Swap over the existing temperature sync scheduler to use the newly created temperature download action
- Got rid of macAddress param on `downloadAllLogs()` method as it wasn't needed
- Add check to not start download process if one is already in progress

## Testing

- [ ] Added some `console.log` statements into the `downloadAllLogs()` function to check it was being called at the scheduled interval and set `SYNC_INTERVAL` to something short (in `mSupplyMobileApp`) for testing.

- There is a UI component that looks like it was tied into the previous scheduler which also allows manual sync. Not sure about whether to keep it, get rid of it, or rework it. I think it's currently a little bit vague about what exactly is being synced, but probably a useful indicator to keep?
  Screen Shot 2021-01-26 at 5 38 44 PM

### DIFF
--- a/src/actions/VaccineActions.js
+++ b/src/actions/VaccineActions.js
@@ -6,7 +6,7 @@
 import moment from 'moment';
 import { ToastAndroid } from 'react-native';
 import { PermissionSelectors } from '../selectors/permission';
-import { selectScannedSensors, selectIsSyncingTemps } from '../selectors/vaccine';
+import { selectScannedSensors, selectIsDownloadingLogs } from '../selectors/vaccine';
 import { PermissionActions } from './PermissionActions';
 import BleService from '../bluetooth/BleService';
 import TemperatureLogManager from '../bluetooth/TemperatureLogManager';
@@ -166,8 +166,8 @@ const scanForSensors = (dispatch, getState) => {
 const startDownloadAllLogs = () => async (dispatch, getState) => {
   // Ensure there isn't already a download in progress before starting a new one
   const state = getState();
-  const isDownloadingTemps = selectIsSyncingTemps(state);
-  if (isDownloadingTemps) return null;
+  const isDownloadingLogs = selectIsDownloadingLogs(state);
+  if (isDownloadingLogs) return null;
 
   await withPermissions(dispatch, getState, downloadAllLogs());
   return null;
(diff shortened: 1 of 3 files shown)


src/actions/VaccineActions.js
```diff
@@ -164,8 +163,13 @@ const scanForSensors = (dispatch, getState) => {
   BleService().scanForSensors(deviceCallback);
 };
 
-const startDownloadAllLogs = macAddress => async (dispatch, getState) => {
-  await withPermissions(dispatch, getState, downloadAllLogs(macAddress));
+const startDownloadAllLogs = () => async (dispatch, getState) => {
+  // Ensure there isn't already a download in progress before starting a new one
+  const state = getState();
+  const isDownloadingTemps = selectIsSyncingTemps(state);
```
mark-prins, line 169: I was confused for a moment because the method is `DownloadAllLogs` and the condition is `DownloadingTemps` - it would help me if they were both the same.. either logs or temps

mark-prins (approved):
working nicely! 
I'd prefer the names to change - and Josh may say something about using `getState`..? apart from that, all good 😄

joshxg (approved):
👍 

I gots no problem with `getState` for this 🤣 Only when it's possible to not use it or if it's state that is set in a specific workflow such that one action is fully dependent on a specific order of previous actions.. maybe that makes sense 🤷

src/actions/VaccineActions.js
katherine-sussol: Thanks - yeah that was a bit confusing, just keeping everyone on their toes 😆 . 
Have stuck with `DownloadAllLogs`, also renamed the selector to avoid confusion between Syncing and Downloading 3dc74637

joshxg:
> There is a UI component that looks like it was tied into the previous scheduler which also allows manual sync. Not sure about whether to keep it, get rid of it, or rework it. I think it's currently a little bit vague about what exactly is being synced, but probably a useful indicator to keep?

Sorry I missed this comment.

I've had almost everyone has tell me individually they don't like it. Although I agree it's useful to have, I'm not sure if it's the right place. It makes that area of the UI a bit too busy. Possibly reworking the original sync icon thing to also show some detail on bluetooth downloading/syncing would be better?

Anyway, gonna make an issue to remove it and the associated logic.

PR #3403, opened 2021-01-28
#3401 with permission should recheck permissions

### LINKED ISSUE #3401: Permission checker withPermissions doesn't recheck after enabling

https://github.com/openmsupply/mobile/blob/199067d4f9b2371ea89ab49e8386ffbc772eb8ff/src/actions/VaccineActions.js#L142-L163

After `withPermission` tries to enable location/bluetooth, there is no re-check to determine if it was enabled - it will just always reject the action 

Have the actions checking the permissions/feature returning a bool flag whether it wi

## Implementation

As above

N/A

## Additional context

Might be nice to also return true for bluetooth when on an emulator

joshxg:
Fixes #3401 

## Change summary

- updates `withPermission` to re-eval if permissions were granted
- Also hard enable vaccines for development!!!!

## Testing

internal?

n/a

### DIFF
--- a/src/reducers/ModulesReducer.js
+++ b/src/reducers/ModulesReducer.js
@@ -49,7 +49,7 @@ const initialState = () => {
     usingPayments,
     usingDashboard,
     usingDispensary,
-    usingVaccines,
+    usingVaccines: true,
     usingCashRegister,
     usingModules,
     usingInsurance,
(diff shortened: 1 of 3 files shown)


src/reducers/ModulesReducer.js
```diff
@@ -49,7 +49,7 @@ const initialState = () => {
     usingPayments,
     usingDashboard,
     usingDispensary,
-    usingVaccines,
+    usingVaccines: true,
```
katherine-sussol, line 52: Hehe, is there a way to check if dev here? Either way don't want to hold up PR/Dev 😅

katherine-sussol (approved):
Nice - works well 👍

joshxg:
So I don't cause a conflict with https://github.com/openmsupply/mobile/pull/3411 I'll do the __DEV thing in a different branch and merge this one

PR #3415, opened 2021-01-29
#3385 Improved New/Edit Sensor flow performance #1

### LINKED ISSUE #3385: Improve new sensor flow performance

The performance in the new sensor flow is sub optimal and switching tabs could be improved.

Not sure exactly, but it feels slow and not good. Would need a bit of investigation.

## Implementation

TBD

N/A

## Additional context

N/A

joshxg:
Fixes #3385 

## Change summary

- Added optimistic state with direct manipulation to update the long press components quicker.
- Added calling to navigate directly on the navigator ref when navigating from step one->two. This is quite a bit faster. 
- Throttled the bluetooth scan. Just means less redux diffs.

## Testing

- [ ] The new sensor flow feels less janky and slow.

- There are UI issues which will be fixed in later PRs. Nothing to do with this PR, although it's hard to tell.
- Should change the Wizard component to use the navigator ref rather than the redux integration which in general is significantly slower.


src/actions/Bluetooth/ScanActions.js
```diff
@@ -21,17 +22,21 @@ const sensorFound = macAddress => ({ type: SCAN_ACTIONS.SENSOR_FOUND, payload: {
...
+        if (!alreadyScanned?.includes(macAddress) && !alreadySaved) {
+          dispatch(sensorFound(macAddress));
+        }
       }
-    }
-  };
+    },
+    500,
+    { leading: true }
+  );
```
joshxg, line 39: Just throttling this to limit the number of dispatches

src/hooks/useOptimisticUpdating.js
```diff
@@ -0,0 +1,54 @@
...
+  };
+
+  const newOnChange = newValue => {
+    wrappedOnChange(withOptimisticState(newValue));
+  };
+
+  const newValue = postChangeHook(optimisticValue.current);
+
+  return [ref, newValue, newOnChange];
+};
```
joshxg, line 54: Whole thing basically returns a ref which you apply to a text input, then wraps the onChange for the textinput and directly updates the dom element and debounces the real callback.

reasonably common use case to optimistically update state and think this is a pretty OK solution. This is the main change of the PR, improving the stepper performance

src/pages/NewSensor/ScanRow.js
```diff
@@ -41,32 +42,39 @@ RectangleButton.propTypes = {
...
-          Icon={<ChevronRightIcon color={DARKER_GREY} />}
-        />
-      </FlexRow>
-    </WithFixedDimensions>
-  </Animatable.View>
-);
+          <IconButton
+            onPress={() => {
+              // Navigating to the next tab whose name is '1'
+              navigation.navigate('1');
```
joshxg, line 64: Other main change is using the navigation ref to navigate here as it's quite a bit quicker. Would like to change the whole `Wizard` component to just use this style rather than the redux-integration that it has

src/actions/Bluetooth/ScanActions.js
```diff
@@ -21,17 +22,21 @@ const sensorFound = macAddress => ({ type: SCAN_ACTIONS.SENSOR_FOUND, payload: {
...
-      if (!alreadyScanned?.includes(macAddress) && !alreadySaved) {
-        dispatch(sensorFound(macAddress));
+        if (!alreadyScanned?.includes(macAddress) && !alreadySaved) {
+          dispatch(sensorFound(macAddress));
+        }
       }
-    }
-  };
+    },
+    500,
```
mark-prins, line 37: :nit would be nice to have as a CONST

src/hooks/useOptimisticUpdating.js
mark-prins: not sure if it is because of this.. the stepper is behaving very strangely now. default of 5mins. Click `+` and it goes to 6 mins. click `-` and it goes to 1 mins. Go up again and it is 3537 mins and then keeps going up in about 3000 increments

mark-prins: other than that, the wrapper looks nifty. nice idea.

src/pages/NewSensor/NewSensorStepTwo.js
```diff
@@ -27,30 +28,32 @@ export const NewSensorStepTwoComponent = ({
...
-        updateDuration={updateDuration}
-        updateTemperature={updateTemperature}
-      />
-      <BreachConfigRow
-        type="COLD_CUMULATIVE"
-        {...coldCumulativeConfig}
-        updateDuration={updateDuration}
-        updateTemperature={updateTemperature}
-      />
+      <AfterInteractions placeholder={null}>
```
mark-prins, line 31: the stepper for minutes on this page is funky - working on page 3 though

mark-prins (commented):
Working very nicely!
I have an issue with the time stepper which I'm looking into.
Also, the array assigned to the `style` prop on `EditorRow` threw a warning. I changed to this: 
```style={{ ...localStyles.containerStyle, ...containerStyle }}```
instead to get it working - does the array work ok for you?

mark-prins (commented):
The `BreachConfigRow` is dividing the displayed value by `SECONDS.ONE_MINUTE` which is making things a little crazy. the safest option would be to do everything as minutes, except the call which updates the sensor

src/pages/NewSensor/NewSensorStepTwo.js
joshxg: Yeah, sorry! 
a537571

Done this here as a quick fix as I am actually changing the implementation of this state in a couple PRs down the line..

src/actions/Bluetooth/ScanActions.js
joshxg: 4e86931

joshxg:
The styles prop type thing: 

- There were quick a few prop type/style mix ups from this pr https://github.com/openmsupply/mobile/pull/3409 
- think you're referring to this prop type mistake fixed here: 255804f ?
- Styles can technically be an object or an array of style objects

mark-prins (approved):
Much nicer!
Another small bug - if you long-press the `-` button and get the value down to the minimum, then press `+` it takes a very long time to get a response. Feels like there is a backload of debounced reduce value actions to get through before an increase action can happen. Would be nice to discard invalid actions

joshxg:
Will sort that bug out in a new one - nice find!

mark-prins:
> The styles prop type thing:
> 
> * There were quick a few prop type/style mix ups from this pr #3409
> * think you're referring to this prop type mistake fixed here: [255804f](https://github.com/openmsupply/mobile/commit/255804f1244767775e608e8a8241bf0c60c6bd0f) ?
> * Styles can technically be an object or an array of style objects

yeah, that's it. thanks - I thought that styles could take arrays of objects.. didn't check the propTypes 🙄

Chris-Petty:
> style={{ ...localStyles.containerStyle, ...containerStyle }}

@mark-prins I probably introduced this, I found that restarting the app seemed resolve till moving branch, so figured it was RN being weird. I suspect in certain cases RN doesn't evaluate it as it normally would (i.e. flatten the array for StyleSheets) if any of the array elements are not objects (e.g. `null`, or `false`). But not all the time 🤷

PR #3411, opened 2021-01-29
#3402 extract blink actions/reducers

katherine-sussol:
Fixes #3402

## Change summary

- Nothing too fancy, moved the blink stuff out of the vaccine actions/reducer 
- Moved withPermissions into PermissionsActions 
- Manual merge of permission changes from PR #3403 (trying to pre-emptively avoid the conflict 😕  maybe there is a better way to do this)

## Testing

Steps to reproduce or otherwise test the changes of this PR:

- [ ] Check sensor blinks still work on emulator and device? (add a new sensor flow) 

### Related areas to think about
N/A

PR #3417, opened 2021-01-29
#3416 new vaccine page to be real #2

### LINKED ISSUE #3416: Replace old VaccinePage with NewVaccinePage

Make the component `NewVaccinePage` the real `VaccinePage`

As above

## Implementation

Copy and paste

Renaming files. Not as good.

## Additional context

N/A

joshxg:
Fixes #3416 

## Change summary

- `NewVaccineModulePage` copy pasted into `VaccinePage`
- Added `PropType` fixes
- Added use of `AfterInteractions` to make navigation more fluid.
- Removed all the buttons in `SettingsPage` since we can navigate from `VaccinePage` now.

## Testing

- [ ] Can navigate to `VaccinePage`

N/A


src/pages/VaccinePage.js
```diff
@@ -1,125 +1,196 @@
...
-
 import React from 'react';
-import { connect } from 'react-redux';
-import { StyleSheet, Text, FlatList } from 'react-native';
+import { FlatList, StyleSheet } from 'react-native';
+import moment from 'moment';
 import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import * as Animatable from 'react-native-animatable';
+import {
```
joshxg, line 8: This is just a copy paste of `NewVaccineModulePage` and then I deleted `VaccinePage` - there are a couple things to fix in this component where we have made issues, but I have no fixed anything in here. OK maybe a prop type or two, I think

src/pages/SettingsPage.js
```diff
@@ -46,15 +45,7 @@ const exportData = async () => {
...
-const Settings = ({
-  toRealmExplorer,
-  currentUserPasswordHash,
-  requestStorageWritePermission,
-  toFridgeDetail,
-  toNewVaccineModulePage,
-  toNewSensorPage,
-  toEditSensorPage,
-}) => {
+const Settings = ({ toRealmExplorer, currentUserPasswordHash, requestStorageWritePermission }) => {
```
joshxg, line 48: Removed the settings buttons to navigate around as they're not needed anymore

src/pages/VaccinePage.js
```diff
@@ -1,125 +1,196 @@
...
+});
+
+const stateToProps = state => {
+  const { fridge } = state;
+  const { fridges } = fridge;
+  return { fridges };
+};
+
+const dispatchToProps = dispatch => ({
+  blinkSensor: macAddress => dispatch(VaccineActions.startSensorBlink(macAddress)),
```
katherine-sussol, line 190: Oh. Think I missed updating this with https://github.com/openmsupply/mobile/pull/3411 to `BlinkActions.startSensorBlink`. I suppose can be slotted in with #3428 though.

src/pages/VaccinePage.js
```diff
@@ -1,125 +1,196 @@
...
-  maxLine: PropTypes.array.isRequired,
-  breaches: PropTypes.oneOfType([PropTypes.object, PropTypes.array]),
-  maxDomain: PropTypes.number.isRequired,
-  minDomain: PropTypes.number.isRequired,
-  fridges: PropTypes.oneOfType([PropTypes.array, PropTypes.object]).isRequired,
-  onSelectFridge: PropTypes.func.isRequired,
-  onOpenBreachModal: PropTypes.func.isRequired,
-  onViewAdminPage: PropTypes.func.isRequired,
-  isAdmin: PropTypes.bool.isRequired,
+  fridges: PropTypes.object,
```
katherine-sussol, line 160: Hm when the page is first loaded for me (before setting up anything), I'm getting a complaint fridges is being supplied as an array to the VaccinePage. I think `FridgeReducer` has an array default for the initial state that should maybe be converted to an object? (`locationsWithASensor`)

katherine-sussol (changes requested):
👍  nice!

Might also be a good time in `NewSensorStepThree` to swap over 
`dispatch(goToSettings())` to `dispatch(goToVaccines())` now (line 111)?

That and one other prop concern (my suggestion might be off but the warning is there), otherwise looks good!

src/pages/VaccinePage.js
katherine-sussol: Or maybe it is supposed to allow arrays. The name fridges does suggest an array but now I think I confused myself.

joshxg: It is being passed a realm collection which has the shape:

```
{ "0": someFridge, "1": someOtherFridge } 
```

which acts as an array but is technically an object!

katherine-sussol: Right, thanks that makes a lot more sense hah. Must just be the initialState combined with my setup causing the warning then!

src/pages/VaccinePage.js
joshxg: b466ddc

src/pages/VaccinePage.js
joshxg: d18f503

joshxg:
Regarding the navigation action: it's done in I think 2 PR's down the line heh

katherine-sussol (approved):
🎉

PR #3437, opened 2021-02-01
#3426 Create a matching Location #5

### LINKED ISSUE #3426: When creating a new sensor, also create a matching Location

- Sensors and Locations are one-to-one
- We need a sensor to link to a location
- We don't want a separate interface for creating locations.

- To limit the complexity in mobile, it was decided that the process of creating a sensor will also create a matching Location with the sensor name being the location description and the sensor code being the location code.

## Implementation

in the `createRecord` case?

Creating a full UI to create locations

## Additional context

- Doesn't account for selecting 'old' locations or 'switching' the location of a sensor.

joshxg:
Fixes #3426 

## Change summary

- Adds `LocationReducer`
- Creates a Location when moving through the new sensor flow, which is saved after connecting and sending commands a long with temperature breach configs and the sensor iteself.

## Testing

- [ ] After moving through the new sensor flow, there is a new sensor, a new location and 4x temperature breach configs created.

- Need to figure out what to set as the location description, how to edit location, etc?

### DIFF
--- a/src/actions/Entities/LocationActions.js
+++ b/src/actions/Entities/LocationActions.js
@@ -1,0 +1,41 @@
+import { generateUUID } from 'react-native-database';
+import { selectNewLocationId } from '../../selectors/Entities/location';
+
+export const LOCATION_ACTIONS = {
+  CREATE: 'LOCATION/create',
+  UPDATE: 'LOCATION/update',
+  SAVE_NEW: 'LOCATION/saveNew',
+};
+
+const createDefaultLocation = () => ({
+  id: generateUUID(),
+  description: '',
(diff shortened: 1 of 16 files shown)


src/actions/Entities/LocationActions.js
```diff
@@ -0,0 +1,41 @@
+import { generateUUID } from 'react-native-database';
+import { selectNewLocationId } from '../../selectors/Entities/location';
```
joshxg, line 2: This all very similar to `SensorActions`

mark-prins (approved):
LGTM 👍

PR #3441, opened 2021-02-02
#3440 duration stepper unusable

### LINKED ISSUE #3440: New sensors: duration stepper is unusable
## Describe the bug

When creating a new sensor, the duration field shows `999 minutes` and cannot be changed

### To reproduce

1. Go to vaccines and click on `Add Sensor`
2. Select a sensor. On Step 2, try to change the value of duration

### Expected behaviour

Default should be minimum (1) and you should be able to change up to the maximum (99)

mark-prins:
Fixes #3440 

## Change summary

Changed the duration stepper and storage to use minutes. The display and user interaction is in minutes, it seems clumsy to work around this and store in seconds.

Have updated the usage to reflect what I think it should be.. though I'm not entirely sure about that

## Testing

- [ ] Add a sensor, adjust duration.
- [ ] Default should show `20` for a new sensor
- [ ] minimum value of `1` is enforced

The usage which, I think, was comparing milliseconds against (n seconds / 1000). That would mean a configured value of 20 minutes would be stored as 1200 ( i.e. in seconds ) and then used as ```logsDuration < 1200 / 1000``` ??


src/bluetooth/BreachManager.js
```diff
@@ -68,7 +69,7 @@ export class BreachManager {
     const { timestamp: endTimestamp } = logs[logs.length - 1];
     const { timestamp: startTimestamp } = logs[0];
     const logsDuration = endTimestamp - startTimestamp;
-    if (logsDuration < duration / 1000) return false;
+    if (logsDuration < duration * MILLISECONDS.ONE_MINUTE) return false;
```
mark-prins, line 72: This is the line I need confirmation on.

joshxg (changes requested):
You're right. There's plenty of weird inconsistencies I've been noticing :(

Unfortunately there's a bunch of desktop work which assumes duration is in seconds. The UI for TBConfigs, as well as on `TemperatureBreach` the `duration` is 'copied' over so in that form also.

image

Additionally the `BleService` method `updateLogInterval`

```
  updateLogInterval = async (macAddress, logInterval) => {
    await this.connectAndDiscoverServices(macAddress);
    return this.writeWithSingleResponse(
      macAddress,
      `${BLUETOOTH.COMMANDS.UPDATE_LOG_INTERVAL}${logInterval}`,
      stringFromBase64
    );
  };
``` 

The BlueMaestro sensors use the command `*lintXXX` (I believe it is 60-6535 and the original inspiration duration is in seconds), will need to be changed.

I'm not opposed to changing duration to seconds (will also require minor changes to CCA), if we want to do this work.

joshxg:
Think I didn't emphasise enough that I really want to make it minutes - also not sure if Dhanya will blow a gasket if I do a desktop PR

joshxg:
OH - I am getting completely confused between log intervals and breach config durations!!!!!

mark-prins:
Thanks @joshxg - that validates the logic in the `BreachManager`. I've also commented to clarify.
I'll change to milliseconds and try to get the UI to work, I don't see that we have time to make big changes right now.. but we do need the UI to be usable.

This also affects the logInterval then? as that was changed to use minutes to get it working in another PR

joshxg:
That's where the mess becomes very obvious! Log interval should be in seconds. So yeah, saved persistently as seconds but displayed as minutes

mark-prins:
Ready to review again. Now storing duration as milliseconds, while displaying as minutes.
Tests updated and code commented.

joshxg (approved):
Nice !

I merged develop in and also made this commit (cause I had it sitting staged to get the sensor edit page to work) - 57ddd28 - Seems to work perfectly! Thanks so much

joshxg:
🤞 hope that commit was OK 🤣

PR #3445, opened 2021-02-02
#3432 FridgeHeader on FridgeDetailPage A2

### LINKED ISSUE #3432: Add a header to the FridgeDetail paper area

The current `FridgeDetail` page needs a header on the `Paper` area.

Currently:

image

Should be something like:

image

## Implementation

Unsure. Would be nice to have a component for the 3 different places a similar component is used. The above two and also in the `SensorEditPage`:

image

I think theres a bit of difference and only the `DetailPage` and `EditPage` will be able to share the component, where there would be a conditional to show or hide the `Cog`.

I think it would possibly be worthwhile using this same component in the `VaccinePage` also, and just also use the battery level and last sync time in the header (Since it is one to one sensor to location it's OK)

N/A

## Additional context

N/A

joshxg:
Fixes #3432 

## Change summary

- Added some state for the last download status and the last download time. With these being able to be persisted as well as there soon being a `bluetooth/download` slice, I figured this could be nicely encapsulated there. Getting the last download time from the logs means you know when the last successful download was, but not the last attempt (Maybe you should only be displaying the last 'attempt' anyway), so have a status and time means we can display in the UI the last download attempt time, as well as show whether it was OK or not.
- Added a `FridgeHeader` component. This is mainly ripped out of `SensorEditPage`. Added its use in `SensorEditPage` and `FridgeDetailPage`.
- Added `LastSensorDownload` (looking for a new name if you can think of one!) component: can be given a mac address and placed anywhere to give up to date info on when the last download was etc. There ended up being a decent amount of code to handle updating itself and hooking into redux which shouldn't need to be repeated.
- Added `BlinkSensorButton` (looking for a new name if you can think of one!) component: Again can be given a mac and placed anywhere and it will handle sending the command+state updates+disabling itself when another sensor is blinking and spinning when itself is blinking.
- Used `BlinkSensorButton` on `VaccinePage`

## Testing

*Fridge Detail Page*
- [ ] Has a header component on top of the chart and the details: sensor name, battery level and last sync time are correct
- [ ] Last sync time flashes when that sensor is downloading
- [ ] Last sync icon changes when the last sync fails to be a fail icon
- [ ] Last sync time updates continuously (i.e. will say 'few seconds ago' then 'a minute ago', 'two minutes ago' etc without touching the ui at all
- [ ] when blinking a sensor, all other sensor blink buttons are disabled (can see on vaccine page)

*Sensor edit page*
- [ ] Has a header component on top of the chart and the details: sensor name, battery level and last sync time are correct
- [ ] Last sync time flashes when that sensor is downloading
- [ ] Last sync icon changes when the last sync fails to be a fail icon
- [ ] Last sync time updates continuously (i.e. will say 'few seconds ago' then 'a minute ago', 'two minutes ago' etc without touching the ui at all
- [ ] when blinking a sensor, all other sensor blink buttons are disabled (can see on vaccine page)

- I think this `FridgeHeader` should be used on the `VaccinePage` also and remove the last sync/battery level from the flat list item. Can add the `Circle` and just use that everywhere.


src/hooks/useIntervalReRender.js
```diff
@@ -0,0 +1,18 @@
...
+ */
+export const useIntervalReRender = interval => {
+  const [, setState] = useState({});
+  const intervalRef = useRef();
+
+  useEffect(() => {
+    intervalRef.current = setInterval(() => setState({}), interval);
+    return () => clearInterval(intervalRef.current);
+  }, [interval]);
+};
```
joshxg, line 18: This hook is will re-render the "last download 1 min ago" component every so often which will cause a recalc the delta of the last sync time and the time now - without this, theres no re-render when time changes and if you sit on a page for 10 mins it'll just say "a few seconds ago"

src/pages/SensorEditPage.js
```diff
@@ -2,8 +2,7 @@
 /* eslint-disable react/forbid-prop-types */
 /* eslint-disable react/jsx-wrap-multilines */
 import React from 'react';
-import { View, Text, StyleSheet } from 'react-native';
-import moment from 'moment';
+import { StyleSheet, ToastAndroid } from 'react-native';
```
joshxg, line 5: Main change in this component is extracting `FridgeHeader` into its own file and using it in here.

src/pages/VaccinePage.js
```diff
@@ -14,25 +14,24 @@ import {
   DownloadIcon,
   TemperatureIcon,
   IconButton,
-  LightbulbIcon,
-  WifiIcon,
   CogIcon,
   Circle,
   FlexRow,
   AlarmClockIcon,
+  WifiIcon,
```
joshxg, line 21: Just added the use of `BlinkSensorButton` (Name is trademarked on this one, sorry) in this component - don't have to handle disabling/spinner etc again

src/reducers/VaccineReducer.js
```diff
@@ -1,14 +1,78 @@
 import { VACCINE_ACTIONS } from '../actions/VaccineActions';
 
+const LAST_DOWNLOAD_STATUS = {
+  OK: 'OK',
+  ERROR: 'ERROR',
+};
```
joshxg, line 6: OK - I am keen for ideas of what this really should be.. 🤷 

At the moment just want to show the `WiFiIcon` or `NoWifiIcon` so two states is fine

- could make it `LAST_DOWNLOAD_OK` with a boolean.... but reasonably certain we might want to put a reason at some point - "Couldn't find sensor", "Couldn't connect with sensor". 
- So maybe it should be LAST_DOWNLOAD_ERROR and be `null` if it's OK? 
- Then I thought having "OK" was nicer than `null` to represent `OK` - ok?

joshxg (commented):
To test the state changes of the `Wifi` icon/text, in `FridgeHeader`, I added into `dispatchToProps`: 

```
const syncTemperatures = () => dispatch(VaccineActions.startDownloadAllLogs());
```

And passed it to the `onPress` of the download icon button.

I also added into `downloadLogsFromSensor` in `VaccineActions`:

```
if (macAddress === 'AB:CD:EF:GH:IJ:KL') throw new Error()
```

To test the error state.

katherine-sussol (approved):
Looks good!

Names sound pretty reasonable to me 🤷

joshxg:
Thanks @katherine-sussol ❤️ 

I actually branched this from https://github.com/openmsupply/mobile/pull/3445 so will wait for that to get a review before merging this.

I also merged into develop (and then back down into this) the changes you and Mark made to the `VaccineActions/Reducer` etc so hopefully it's still all good 🤣

src/reducers/VaccineReducer.js
mark-prins: The current enum meets requirements, so I think all good to have as is.
We can extend when new scenarios come up - to your point, Kat, we could have `OK` and `everything else` - so we show the ok icon if the status is ok and an error state otherwise. 
What sort of error conditions could there be? ( I see little value in second guessing possible states.. all I've seen is 'sensor disconnected' / 'connection error' sort of things. )

Also, why is it a wifi icon? these are bluetooth sensors! ( this has already confused me )

src/selectors/vaccine.js
```diff
@@ -21,3 +21,27 @@ export const selectIsSyncingTemps = ({ vaccine }) => {
...
+
+export const selectLastDownloadTime = ({ vaccine }, macAddress) => {
+  const { lastDownloadTime } = vaccine;
+  return lastDownloadTime[macAddress] ? new Date(lastDownloadTime[macAddress]) : null;
+};
+
+export const selectLastDownloadFailed = ({ vaccine }, macAddress) => {
+  const { lastDownloadStatus } = vaccine;
+  const status = lastDownloadStatus[macAddress];
+  const lastDownloadFailed = status === 'ERROR';
```
mark-prins, line 38: or `status !== 'OK'` to allow for other error states?

mark-prins (approved):
Not that it needs a second approval. Just clicking Approve because it looks marvellous

mark-prins:
I am getting `Failed prop type: Invalid prop 'lastDownloadTime' of value 'Thu Feb 04 2021 10:26:56 GMT+1200 (+12)' supplied to 'LastSensorDownloadComponent', expected one of [null,null].` with this branch. I think merge and when all the branches are merged I'll recheck.

joshxg:
Proptype fix: 56901c5

src/selectors/vaccine.js
joshxg: Nice! 
1f417e7

PR #3447, opened 2021-02-03
#3394 CSV Exporting

### LINKED ISSUE #3394: Exported CSV report

Need to create a report of temperature logs/breaches/sensor info to be emailed.

Rushed very quickly don't judge me: https://github.com/openmsupply/msupply-cold-chain/blob/45e0433d5b5bffa5f24f6ed6946928f6ebdd016d/src/services/device/DeviceService.js#L120-L227

example report (probably doesnt have to be exactly the same):
[Example of CCE Export.xlsx](https://github.com/openmsupply/mobile/files/5879098/Example.of.CCE.Export.xlsx)

## Implementation

Not sure. Think it should just be a very specific and direct massive function to write the CSV in a non-smart way as there are no other reports generated at the moment.

I am pretty sure we will have to make reports for other things in the future, where a more generic way of doing this would be better, but wait till then, I guess.

N/A

## Additional context

N/a

joshxg:
Fixes #3394 

## Change summary

- Adds packages required to create and email CSV reports
- Adds a method to create a very specific CSV report
- Adds buttons in SettingsPage to test it (Will remove after UI in?)

TODO: Lots of localisation required. I wanted to get some opinions on the report/text before doin em all.

This is what I think is the minimum required data for a report. Be good to get opinions on "if we can live with this for now" - if it's way to bad, I can improve!

## Testing

- [ ] Creates a report where the values are correct!
- [ ] Prompts to email the report with the report being an attachment.

- This isn't a great way to do things. Could be far more extensible but should be decent/ok for creating csv reports for now?

[03-02-2021-1445.csv.zip](https://github.com/openmsupply/mobile/files/5916023/03-02-2021-1445.csv.zip)


src/utilities/vaccineReport.js
```diff
@@ -0,0 +1,238 @@
...
+  const directory = '/Download/mSupplyMobile_data';
+  const now = moment().format('DD-MM-YYYY-HHmm');
+  const file = `/${now}.csv`;
+  const path = `${RNFS.ExternalStorageDirectoryPath}${directory}${file}`;
+
+  try {
+    await RNFS.mkdir(`${RNFS.ExternalStorageDirectoryPath}${directory}`);
+    await RNFS.writeFile(path, content, 'utf8');
+    // eslint-disable-next-line no-empty
+  } catch (e) {}
```
mark-prins, line 222: maybe Toast on error? And don't try to email if the file cannot be written.

mark-prins: ```EACCES (Permission denied), open '/storage/emulated/0/Download/mSupplyMobile_data/04-02-2021-1115.csv'```

mark-prins (approved):
I think that's all very cool. I agree - once people see this, they will want emailed reports for all sorts of things, and at that point we'll need to generalise the work done here and think about making it useful elsewhere.

I had an issue in that it doesn't ask for permission to the file system - just silently bombs if no permission. Exporting realm data asks, and once granted then I can use the csv export ok.

Could you prompt for file access, and also not send an email if the csv fails?

src/utilities/vaccineReport.js
joshxg: Ah! Of course, totally forgot about permissions 🙄 Another one that I forgot to implement was a check for if there was any temperature logs - getting an empty report is a bit average!

I just spent a bit thinking of the best way to do this and checking out the `UIDatabase` export data thingy. I think it might be best to do the checks before this is called, like export data. Keeps the toast out of these functions which have nothing to do with UI stuff and also allows for easier use of checking the permission in the redux store and dispatching an action if the permission is updated etc.

PR #3450, opened 2021-02-03
#3431 Remove DataTablePageView from capturing gestures to enable scrolling

### LINKED ISSUE #3431: VaccinePage can't scroll
## Describe the bug

- When there is enough content to require the need for scrolling on the `VaccinePage`, you can't scroll.

### To reproduce

1. Go to `VaccinePage` with like 8+ sensors
3. Scroll down to '....' !!

### Expected behaviour

Should be able to scroll.

### Proposed Solution

Not sure exactly. Something to do with the gesture being captured by something. Not sure I've seen the behaviour before.

### Version and device info

- App version: 6.1 Development
- Tablet model: N/A
- OS version: N/A

### Additional context

Can scroll by holding down press on a different clickable - i.e. the Cog icon and then scrolling.

joshxg:
Fixes #3431 

## Change summary

- `DataTablePageView` captures the gesture event and handles it, the flat list doesn't get a chance to handle it.

## Testing

- [ ] Have heaps of fridges, you can scroll

N/A

### DIFF
--- a/src/widgets/DataTablePageView.js
+++ b/src/widgets/DataTablePageView.js
@@ -16,8 +16,11 @@ const dismiss = () => Keyboard.dismiss();
 /**
  * Simple template container for a standard data table page.
  * Handles placement of content and wraps the page with a
- * Touchable, dismissing the keyboard when an event propogates
+ * Touchable, dismissing the keyboard when an event propagates
  * to this level.
+ *
+ * WARNING: If this component captures events they are not propagated further.
+ *          Be careful with components which scroll.
  */
 export const DataTablePageView = React.memo(({ children, captureUncaughtGestures, style }) => {
   // Use a Fragment over TouchableWithoutFeedback so no gesture events are caught.
--- a/src/pages/VaccinePage.js
+++ b/src/pages/VaccinePage.js
@@ -139,7 +139,10 @@ FridgeDisplay.propTypes = {
 };
 
 export const VaccinePageComponent = ({ fridges, toNewSensorPage, ...dispatchers }) => (
-  <DataTablePageView style={{ paddingHorizontal: 20, paddingVertical: 30 }}>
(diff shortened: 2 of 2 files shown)


src/widgets/DataTablePageView.js
```diff
@@ -16,8 +16,11 @@ const dismiss = () => Keyboard.dismiss();
 /**
  * Simple template container for a standard data table page.
  * Handles placement of content and wraps the page with a
- * Touchable, dismissing the keyboard when an event propogates
+ * Touchable, dismissing the keyboard when an event propagates
```
katherine-sussol, line 19: 👍

katherine-sussol (approved):
This is working nicely after the change 🥳 
I'm not sure I would have been able to find this - the comment helps 👍 . 

So just checking I understand it correctly - `captureUncaughtGestures` is mainly to stop the keyboard popping up and being annoying?

joshxg:
It's mainly for being able to 'click somewhere' to make the keyboard 'disappear' - so when you focus a text input, the default behaviour is for the keyboard to pop up (good) and when you lose focus, it drops down, but you need something to take the focus (a button or something). In some cases, you want to edit the list of quantities and then just be viewing the list again, so you wanna be able to click somewhere (just like, anywhere kinda thing) which will drop the keyboard down

katherine-sussol:
> It's mainly for being able to 'click somewhere' to make the keyboard 'disappear' - so when you focus a text input, the default behaviour is for the keyboard to pop up (good) and when you lose focus, it drops down, but you need something to take the focus (a button or something). In some cases, you want to edit the list of quantities and then just be viewing the list again, so you wanna be able to click somewhere (just like, anywhere kinda thing) which will drop the keyboard down

Ah gotcha, that makes sense, thanks 😸

PR #3451, opened 2021-02-03
#3433 Remove old vaccine UI components B1

### LINKED ISSUE #3433: Remove old Vaccine UI components
## The suggestion

There are a few old Vaccine UI components which should be removed.

- `TemperatureSyncState`
- `TemperatureSync`
- `VaccineAdminPage`

+ All imports and any referencing code i.e. navigation actions.

**Why?**

Not used.

N/A

## Additional context

N/A

joshxg:
Fixes #3433 

## Change summary

- Removes (all?) the old unused vaccine components

## Testing

- [ ] App still works

N/A


src/widgets/HeaderRight.js
```diff
@@ -5,30 +5,11 @@
...
-import { TemperatureSyncState } from './TemperatureSyncState';
-
-const mapStateToProps = state => {
-  const usingVaccines = selectUsingVaccines(state);
 
-  return { usingVaccines };
-};
-
-const HeaderRightComponent = ({ usingVaccines }) => (
+export const HeaderRight = () => (
```
mark-prins, line 11: just curious - the sensor sync bits from the header are removed and not returning?
There's no manual sync option now?

mark-prins (approved):
I didn't realise that the sync bits from the header were going.. This seems to do the job if they are!

src/widgets/HeaderRight.js
joshxg: Yeah. Majority one that it didn't look good in the header and also that manual downloading of temperatures wasn't required 🤷

katherine-sussol (approved):
Hehehe @mark-prins beat me to the punch (🦥  <-- Me) but here's a double approval anyway 🥳 . App still works and nice to de-clutter the UI. 

I guess the manual download button in the fridge header kind of does the same thing, just at the per fridge level anyway?

joshxg:
Ah interesting!

The download button is to download/export the temperature logs! like from #3447

katherine-sussol:
> Ah interesting!
> 
> The download button is to download/export the temperature logs! like from #3447

Ohhhh, thanks for clarifying. So there's download from sensor -> tablet (background) and then download from the tablet -> file export (manual)... that makes sense I think, not sure why I thought it'd do a download from the sensor, that would be more of a sync icon I guess.

joshxg:
Yup defo confusing though hmm

PR #3453, opened 2021-02-03
#3392 Removed usage of location type

### LINKED ISSUE #3392: Remove use of restricted location type
## The suggestion

Original vaccine had references to `defaultRestrictedLocationType` which is the location type which must related to a location that the item is assigned to. Mobile currently will use only fridge location types and this is not needed. We should remove traces of this.

**Why?**
Simplifies the code.

N/A

## Additional context

N/A

joshxg:
Fixes #3392 

## Change summary

- Removed a lot of uses of location type. 
- Removed a few other bits of code which aren't used anymore.
- Various code is still in which is removed in #3450 and #3451

## Testing

- [ ] App still works?

Think that should be all the old code gone out of our LIVES

### DIFF
--- a/src/database/DataTypes/Item.js
+++ b/src/database/DataTypes/Item.js
@@ -238,22 +238,6 @@ export class Item extends Realm.Object {
   }
 
   /**
-   * Get this items restricted LocationType - the location type for which Location records must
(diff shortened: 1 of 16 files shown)


mark-prins (approved):
lgtm!

PR #3457, opened 2021-02-04
#3406 extract bt download actions

### LINKED ISSUE #3406: Extract and update bluetooth sensor 'Download' Actions/Reducers
Child of #3380 

## The suggestion

- Currently all bluetooth actions/reducers live in the same place (`VaccineActions`/`VaccineReducers`).
- The suggestion is to extract out specific concerns into their own slices and use `combineReducers` to compose the Bluetooth state.
- This specific issues applies to the following actions:
  - `VACCINE_ACTIONS.DOWNLOAD_LOGS_START`
  - `VACCINE_ACTIONS.DOWNLOAD_LOGS_ERROR`
  - `VACCINE_ACTIONS.DOWNLOAD_LOGS_COMPLETE`

This will also involve adding some new state to track the download timestamps/status for each sensor (store in lookup table keyed by sensor mac address). These could also be potentially separated out into their own actions/reducers too.

**Why?**
Clearer separation of concerns, easier state management. Ability to track at a more granular level what is happening with the downloads.

Leaving as is. But it's getting fairly large now.

## Additional context
N/A

katherine-sussol:
Fixes #3406 

## Change summary

- Last of the BT actions/reducer restructuring: Moved Download Actions/Reducers into separate files
- Deleted VaccineReducer/VaccineActions/VaccineSelector
- Split selectors into function specific files 
- Snuck in a small change to use the MAX_BLUETOOTH_ATTEMPTS constants in a few places

## Testing

- [ ] Check sensor scanning, downloading and blinking still functions as per before

Soz for the weird differences in `VaccineActions` merge did something weird :/


joshxg (approved):
Nice :) 👍 🎉

PR #3458, opened 2021-02-04
#3393 NewModal and export data from UI A1

### LINKED ISSUE #3393: New Modal components

image

image

Not actually removing sensors right now.

This Modal

- A genericish component?
- `<InputModal/>` ? `<InputDialog/>` ? Not sure
- To be a child of `ModalContainer`?

## Implementation

Think its OK to just have a child for `ModalContainer` specific to entering a line or two and a OK button?

N/A

## Additional context

N/a

joshxg:
Fixes #3393 

## Change summary

- Adds a component: `PaperModalContainer` which is sort of just a square `Paper` as a modal.
- Adds a component: `PaperInputModal` which is intended to be a child of `PaperModalContainer` which is a button, icon and some text which puts children in the middle.
- Adds a component: `ExportTemperatureDataButton` which is intended to be able to slow in anywhere, pass it a sensor and bobs your aunty

## Testing

- [ ] Can export temperatures from `VaccinePage`
- [ ] Can export temperatures from `FridgeHeader` in `SensorEditPage`/`FridgeDetailPage`

N/A


src/utilities/vaccineReport.js
```diff
@@ -227,12 +228,17 @@ const writeReport = async content => {
...
-export const emailVaccineReport = async (sensor, user, comment, recipients) => {
+export const emailVaccineReport = async (macAddress, user, email, comment) => {
+  const sensor = UIDatabase.get('Sensor', macAddress, 'macAddress');
+
+  if (!sensor) throw new Error('Cannot find sensor');
+  if (sensor?.logs <= 0) throw new Error('No temperature logs');
+
   const content = await vaccineReport(sensor, user, comment);
   const path = await writeReport(content);
   const subject = getSubject(sensor);
```
katherine-sussol, line 239: I'm very confused, when `sensor.name` is null (e.g. save a sensor with no name) the subject is coming up empty 😕 . Surely` sensor.name ?? sensor.macAddress` should return the macAddress? What 🧙  is going on

src/widgets/PaperModal/PaperModalContainer.js
```diff
@@ -0,0 +1,58 @@
...
+          flex={1}
+          justifyContent="center"
+          alignItems="center"
+          style={{ backgroundColor: TRANSPARENT_GREY }}
+        >
+          <Paper paddingHorizontal={0}>
+            <View style={{ height: height / 2, width: width / 2 }}>
+              <TouchableWithoutFeedback
+                onPress={() => {
+                  /* This Touchable is capturing gestures such that only the transparent overlay
```
katherine-sussol, line 38: Hmm this is just a theory, but for me, when trying to download from a sensor low enough in the page (below the midpoint...ish?), touching the textbox is dismissing the modal as well.

katherine-sussol (approved):
Modal and report looks good! 

I'm going to be annoying and point out some small 🐛, - not showstoppers but maybe just usability heh (isn't it great how it's way easier to find problems with other people's code than actually writing it 😹 )

src/widgets/PaperModal/PaperModalContainer.js
joshxg: @katherine-sussol 

Interesting - I can't really recreate this at all but it's possible that I don't understand.

I made this GIF which was going to be a really good idea but you can't see me clicking, so it's kind of useless. But I went to the effort to make it, so you get it anyway!!!

hello1

This is me clicking anywhere on the white modal area (it doesn't close) from a modal opened from the bottom sensor...

🤷

katherine-sussol: Haha I like the GIF idea, thanks for sharing, shame about the clicking! That's weiiiird I'm gonna try cleaning up the DB and setting up some new sensors from scratch to make sure I'm not going crazy. It only happens when I focus a text input (e.g. Email or Comment).

joshxg: Oh - I thought you meant to white area.

If it's to do with the text input, it could be to do with your resolution - maybe the text input is in such a position where it gets fully covered with the keyboard and the keyboard has to close, i've seen that before - but not causing the whole modal to close also 😕

joshxg:
I fixed the naming bug: b5e5ee5

sensor name:

I think `sensor.name` must be `''` - this means `sensor.name ?? sensor.macAddress === ''` - I was about to do a PR to fix it but I found a couple other things so I think I'll just do a separate PR rather than create conflicts for my other PR, heh.

Issue #3422, opened 2021-01-31
Vaccine module page fridge circles; More colours?


Chris-Petty:
## Is your feature request related to a problem? Please describe.

Currently each fridge has a green circle rendered next to their name. The green doesn't actually mean anything, maybe we should do something with it?

## Describe the solution you'd like

Give the circle a purpose? Ideas:

- Orange circle if breached recently?
  - Red if currently breached?
- Change to different icons?
  - Over kill? The sensor info row of the fridge gives breach details already

## Implementation

TBD

## Describe alternatives you've considered

Do nothing. Ask a UI designer.

## Additional context

https://github.com/openmsupply/mobile/pull/3409#discussion_r566471691

joshxg:
I think green for good and (flashing?) red for DANGER - low battery/in breach.

katherine-sussol:
The designs have the circle turning orange (sorry...`DANGER_RED`) on some of the other pages if anything (battery/no sync/breach) is in a DANGER state, with the specific danger-causing thing also orange, I kinda like it as it makes it more obvious what's concerning? Although that maybe too much `DANGER_RED`...

joshxg:
I feel like @katherine-sussol is making fun of DANGER_RED and she better be DANGER_CAREFUL

zing

PR #3462, opened 2021-02-04
#3421 Use `FridgeHeader` in `VaccinePage` B1

### LINKED ISSUE #3421: VaccinePage UI should update when events occur.

Currently the `VaccinePage` uses a realm object and each of the UI elements extract their values from the realm models getters. This means certain UI elements won't update when a download occurs

- Current temperature
- Last download time
- Last Breach Time

Also need to update and check the current battery level
- Battery level

Additionally when a sensor is edited and saved, navigating back won't cause an update to the sensor/fridge name.

Likely need to use redux selectors in `mapStateToProps` which will cause an update after a download occurs.

## Implementation

- Would suggest passing a `fridgeID` to `FridgeDisplay` which is then connected to redux, and uses that ID to pick slices of state rather than connecting the `Page` to redux.

Just literally causing a re-render of the page when certain events occur

## Additional context

N/A

joshxg:
Fixes #3421 

## Change summary

- Using `FridgeHeader` in `VaccinePage` "though could it be renamed to something a little generic?"  - I couldn't think of a better name :(
- Slight adjustments required from not using a realm sensor object - passing IDs etc.
- Slight adjustment to `FridgeHeader` to add the CIRCLE - #3422 to make it do something

## Testing

Internal??

image

image

image


src/pages/SettingsPage.js
```diff
@@ -173,13 +173,13 @@ const Settings = ({
...
         name: 'sensor 3',
-        macAddress: 'sen2',
+        macAddress: 'sen3',
         batteryLevel: 30,
         location: fridge3,
       });
       createRecord(UIDatabase, 'Sensor', {
         name: 'sensor 4',
-        macAddress: 'sen3',
+        macAddress: 'sen4',
```
joshxg, line 182: bit offroad

src/pages/VaccinePage.js
```diff
@@ -10,128 +12,66 @@ import {
...
-      <IconButton Icon={<DownloadIcon color={BLACK} />} containerStyle={localStyles.iconButton} />
-      <BlinkSensorButton macAddress={sensor?.macAddress} />
-      <IconButton
-        Icon={<CogIcon color={BLACK} />}
-        onPress={() => toEditSensorPage(sensor)}
-        containerStyle={localStyles.iconButton}
-      />
-    </>
-  );
+const FridgeDisplay = ({ sensor, toFridgeDetail }) => {
```
joshxg, line 32: Main change to the PR to use `FridgeHeader` here. 

`FridgeHeader` wants to be passed a sensor. Could also change to be passed a location, but they're pretty much same same I dunno why it was made a sensor hmm 🤔  🤷

src/pages/VaccinePage.js
```diff
@@ -10,128 +12,66 @@ import {
...
+  const { id, macAddress, currentTemperature, mostRecentBreachTime, locationID } = sensor;
 
   return (
     <Paper
-      // macAddress passed to fridgeHeader is quite naff. The settings page for the "fridge" is
-      // actually for a sensor, assuming 1 sensor per fridge.
-      Header={header}
+      Header={<FridgeHeader showTitle showCog sensor={sensor} />}
```
joshxg, line 37: Having a look now and it would be possible to pass just the `sensorID` to `FridgeHeader` and `FridgeHeader` itself can pick up all the values it needs since it's already connected to redux, by just using the `ID`. This might be a better solution as passing `sensor` as a prop will cause some unneeded re-renders when the object is mutated rather than passing only the primitives which are de structured ...

src/pages/VaccinePage.js
```diff
@@ -168,6 +109,7 @@ const localStyles = StyleSheet.create({
     flex: 0,
     height: 60,
     paddingHorizontal: 10,
+    width: 120,
```
joshxg, line 112: Not sure if this is good - was going for a set width so the icons align..

src/widgets/FridgeHeader.js
```diff
@@ -2,24 +2,50 @@ import React from 'react';
...
+  editSensor,
+  showCog,
+  showTitle,
+}) => (
   <>
-    <FlexView justifyContent="center">
-      <Text style={localStyles.paperTitleText}>{name}</Text>
-    </FlexView>
+    <FlexRow flex={1} alignItems="center">
+      {showTitle && (
```
joshxg, line 35: was sort of trying to avoid this, but it looks a bit silly on the `SensorEditPage` where the `Sensor Name` editing input is directly under the title!

src/widgets/LastSensorDownload.js
```diff
@@ -72,6 +73,8 @@ const localStyles = StyleSheet.create({
   headerTextWithIcon: {
     flex: 0,
     paddingHorizontal: 8,
+    width: 150,
```
joshxg, line 76: Again not sure about this set width. Think it looks OK when the `last download time` is not `N/A`, but when it is it looks silly - but that won't be for long, usually, so..?

src/reducers/Entities/SensorReducer.js
```diff
@@ -40,6 +46,19 @@ export const SensorReducer = (state = initialState(), action) => {
...
+      const { sensor } = payload;
+      const { id, currentTemperature } = sensor;
+      const { byId } = state;
+      const oldSensor = byId[id];
+
+      const newSensor = { ...oldSensor, currentTemperature };
+      const newById = { ...byId, [id]: newSensor };
+
+      return { ...state, byId: newById };
+    }
```
joshxg, line 60: This is picking up after a download happens to update the `currentTemperature`. When we properly have functionality creating breaches, we can add another case here to pick that up and  display when the last breach was, properly.

mark-prins:
I'm getting some lack-of-padding with the time and arrow - could that be due to the fixed widths?

Screen Shot 2021-02-05 at 1 47 27 PM

src/reducers/Entities/SensorReducer.js
mark-prins: @katherine-sussol maybe when the breaches are calculated, we could do this too?

src/widgets/LastSensorDownload.js
mark-prins: is this to make the battery icons align vertically?
if so - are we allowed to change the order, and have this on the left of the battery? The text for the battery is `100%` - `1%` so there's very little width change

mark-prins (approved):
marvellous! looks great, and the vaccine page feels much quicker, though I dunno if that's just me?

src/pages/VaccinePage.js
joshxg: 🤣 


7f10cfd

src/pages/VaccinePage.js
joshxg: 48f2f1c

src/widgets/LastSensorDownload.js
joshxg: 75307e0

joshxg: great idea

PR #3467, opened 2021-02-05
#3464 display data on bar chart

### LINKED ISSUE #3464: Implement boxplot chart

If possible, would like to implement the chart which is mocked up here: https://marvelapp.com/prototype/55aea8f/screen/75648017

Screen Shot 2021-02-05 at 9 46 12 AM

There's a suggestion to try with [BoxPlot](https://formidable.com/open-source/victory/docs/victory-box-plot/)

The current chart looks good, this new version shows breaches nicely and has a little more detail. It's also pretty.
Could/should we have both available?

mark-prins:
Fixes #3464 

## Change summary

Have added three bars to display the hot-mid-cold colour bars.
To do so, I've added a selector to grab the upper and lower breach values - it was unclear what, from the mockup, should be used out of the consecutive and cumulative options. ideas?

This version replaces the current line chart - should we have both options available?

This is how it looks with the generated data:

Screen Shot 2021-02-05 at 1 13 31 PM

## Testing

- [ ] View fridge detail
- [ ] adjust the breach thresholds and check the chart again

### DIFF
--- a/src/database/DataTypes/Item.js
+++ b/src/database/DataTypes/Item.js
@@ -241,7 +241,7 @@ export class Item extends Realm.Object {
    * A vaccine item can have multiple doses per 'unit' or 'pack'. For example, you can have
    * a single vaccine which has two doses. Once a vaccine has been opened, the doses must be
    * given quickly. When recording outgoing stock, the doses and quantity (or number of packs)
-   * are recorded seperately. The amount of open vial wastage (that is, the amount of doses
+   * are recorded separately. The amount of open vial wastage (that is, the amount of doses
    * which are left in a vial and wasted) is the difference between the total number of 'packs'
    * that are outgoing, multiplied by the amount of doses in each pack and the total number of
    * doses actually given.
--- a/src/globalStyles/colors.js
+++ b/src/globalStyles/colors.js
@@ -23,3 +23,4 @@ export const WHITE = '#FFFFFF';
 export const COLD_BREACH_BLUE = '#70b4f0';
 export const DANGER_RED = '#FA6400';
 export const BLACK = '#000000';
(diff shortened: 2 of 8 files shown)


joshxg (approved):
Wow that is looking great! When I played around I was trying to make the red/blue appear as the whisker - this seems much nicer!

> should be used out of the consecutive and cumulative options. ideas?

Consecutive, I reckon

> This version replaces the current line chart - should we have both options available?

I.. guess..? I dunno either..

- Wondering if the hazard icons aren't needed? Think it looks pretty great without? Could make the red/blue parts pressable (I think you can pass in a custom component?) instead?  (new issue?)
- Also just occurred to me about what happens when the breach thresholds change. At the moment we're calculating the breach thresholds from the current configs, but what if we had breaches a month ago at 8 degrees, and then the threshold was changed to 12 - we keep the thresholds on the `TemperatureBreach` record for this reason so we historically know what the thresholds were, but I'm not sure what we should do about this on the chart - maybe this is a good reason to keep the line chart around - easier to handle this with breach icons on a line chart? 🤣 

Also couple other things we should make issues for:
- Fix the date/time ticks along the x axis to align with the bars. Should also change them for the line chart to make them smaller and more frequent- adding some time in when theres more than one tick for the same day
- Grid lines like on the marvel? Think they look pretty nice

mark-prins:
I've added the grid lines and changed to use consecutive thresholds and removed breach hazard notification icons from the bar chart.

then went offroad.. there's now an option to select the chart type (top right)


Screen Shot 2021-02-05 at 5 21 25 PM

Screen Shot 2021-02-05 at 5 21 32 PM

sorry?

joshxg:
So coool!!! ❤️

Issue #3386, opened 2021-01-26
FridgeDetail performance


joshxg:
## Is your feature request related to a problem? Please describe.

The performance of the `FridgeDetail` page is suboptimal and takes a bit of time before being able to navigate to it.

## Describe the solution you'd like

Improved performance

## Implementation

TBD

## Describe alternatives you've considered

N/A

## Additional context

Think the selectors aren't helping much

joshxg:
There's also a bit of an issue with not displaying nulls: https://formidable.com/open-source/victory/gallery/victory-line-with-null-data/

like the above, when there is no data - especially a bigger problem now with pausing

joshxg:
Hmm.. digging into this problem a little bit and there seems to be a few issues with when the page actually updates. I.e. when a new log is downloaded, the page isn't updated. 

There's a bit of a pain point with using date objects and realm objects, so I think changing the state shape to use primitives will be worthwhile and should make things a bit simpler 🤞

PR #3468, opened 2021-02-08
refactor BreachCard component

### LINKED ISSUE #3423: Remove the BreachCard component in Vaccine Module Page
## The suggestion

Remove the Vaccine Module Page `BreachCard` component. It's just a style wrapper around BreachCard. We don't love it's existence.

**Why?**

Tidier.

Do nothing. I suppose I could say it's low priority, but I am a frustratingly a stickler for tidy.

## Additional context

https://github.com/openmsupply/mobile/pull/3409#discussion_r566459793

mark-prins:
Fixes #3423 

The issue says to remove the component. I've refactored and removed from the `FridgeDetailPage` which may not be what was intended, but pleased me. The `FridgeDetailPage` is smaller and neater as a result

## Change summary

Created a new widget and update the `FridgeDetailPage`

## Testing

- [ ] Check that the fridge page looks ok


src/widgets/BreachCard.js
```diff
@@ -0,0 +1,78 @@
...
+      headerText={headerText}
+      style={localStyles.card}
+      contentContainerStyle={{ alignItems: 'center', paddingBottom: 10 }}
+    >
+      <Text style={textStyle}>{text}</Text>
+      {icon}
+    </Paper>
+  );
+};
+
```
joshxg, line 50: I like this a lot, thanks.

I was wondering if we should actually take it further: the [mapStateToProps](https://github.com/openmsupply/mobile/blob/55ebde82acaf95465f393eb9859ed449db053c10/src/pages/FridgeDetailPage.js#L226-L276) is a bit of an issue as part of #3386 I was going to do this also- but was thinking to `connect` each breach card to redux directly, which would spread the time taken for the map state to props function and not front load it.

joshxg (approved):
Nice :)

A suggestion if you want to go further 🤣

joshxg:
sorry I made conflicts cause I forgot to merge my PR earlier :(

src/widgets/BreachCard.js
mark-prins: Nice! I've had a go, with this: c812b886

Wasn't sure how else to keep the shared elements in a single component while only referencing a specific part of the state. I think that this is quicker to load..?

joshxg (commented):
I actually like this, but I think I was thinking something like:

```
const BreachCard = ({config}) => {
...
}


const stateToProps = (state, props) => {
	const {type} = props

	if (type === "HOT_CONSECUTIVE"){
		return { config: selectHotConsecutive(state), type: .. }
    }
}
```

mark-prins:
ah, nice idea. thanks - will tidy up.

mark-prins:
Thanks @joshxg : 6f2051b5 🤞

PR #3469, opened 2021-02-08
update location name when saving

### LINKED ISSUE #3427: "Sensor name" in UI should update Sensor.name and Location.description

- Currently the `VaccinePage` shows the name of the sensor and the name of the location:

image

This is per the designs

- When we create a sensor we need to create a matching location, also - as per: #3426 

So, these will technically, in most cases, be the same text.

TBD

## Implementation

-  "Sensor name" in UI updates Sensor.name and Location.description

N/A

## Additional context

N/A

mark-prins:
Fixes #3427 

## Change summary

Updating `location.description` from `sensor.name` when saving sensors
Bonus task - when saving a sensor, skip trying to disable the button and set the log interval if the sensor has an invalid mac address, otherwise you cannot test this change when using an actual tablet and the generated dummy data.

## Testing

- [ ] Save a sensor. Check realm explorer to validate that the description and name are the same

### DIFF
--- a/src/actions/Bluetooth/SensorUpdateActions.js
+++ b/src/actions/Bluetooth/SensorUpdateActions.js
@@ -92,7 +92,14 @@ const startSetLogInterval = ({ macAddress, interval = 300 }) => async (dispatch,
   return result;
 };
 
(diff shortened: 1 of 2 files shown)


joshxg (approved):
❤️

PR #3470, opened 2021-02-09
#3461 Specific bluetooth state persistance

### LINKED ISSUE #3461: Bluetooth state persistence can enter an invalid state
## Describe the bug

- There is some state which is persisted, which tracks things like what sensor is currently being downloaded from, updated etc. If this state is set when the state is persisted, this value is retained when opening the app again.

### To reproduce

1. Go to update/download from a sensor, close the app
2. Reopen the app
3. The state still has the previous values set.

### Expected behaviour

Only a subset of state should be persisted. For example, the last download time/status.

### Proposed Solution

Creating a new case for each of the slices of the state that should be persisted.

Something like:

```
import { REHYDRATE } from 'redux-persist';

...

case REHYDRATE: {
		// payload here is the state slice which had been persisted. Pick the values to rehydrate
		// from it.
		const { payload } = action;
		const {bluetooth} = payload;
	 	const { ... } = bluetooth

		// Return the newly rehydrated state
		return { ...initialState(), [pickedKey]: {pickedValue} }
```

### Version and device info

- App version: 6.0
- Tablet model: N/A
- OS version: N/A

### Additional context

Since it's in the area it might be a good idea to transition from a blacklist to a whitelist for redux persisting

joshxg:
Fixes #3461

## Change summary

- Made the redux persist use a white list rather than a black list - didn't seem to have an effect on previous state 🤞 
- Only saving a small subset of the bluetooth state

## Testing

internal

N/A


src/Store.js
```diff
@@ -15,28 +15,7 @@ const persistConfig = {
...
-    'modules',
-    'supplierCredit',
-    'fridge',
-    'breach',
-    'rowDetail',
-    'permission',
-    'cashTransaction',
-    'entities',
-  ],
+  whitelist: ['bluetooth', 'sync'],
```
mark-prins, line 18: hoho. that's much neater!

mark-prins (approved):
looks good to me!
Working on the emulator and tablet

PR #3471, opened 2021-02-09
#3374 dont download before log delay

### LINKED ISSUE #3374: Don't download logs before the logging delay

We can sent a logging delay on a sensor. This logging delay should prevent any logs being downloaded before it.

- Add `loggingDelay` date field to `Sensor`.
- Prevent logs being downloaded before it.
- Make sure when creating a sensor it is saved correctly

## Implementation

TBD

N/A

## Additional context

N/A

joshxg:
Fixes #3374 

## Change summary

- Adds logDelay to schema
- Adds programmedDate to schema - Offroad
- Adds setting the log delay 'better' in `SensorActions`/`SensorReducer`.
- prevents download logs for the sensor when the logging delay is greater than the time now.

## Testing

- [ ] Create a new sensor with a logging delay, logs aren't downloaded till after it
- [ ] Can see the logging delay in the UI

- Not sure if this will display nicely on all screens
image
- Probably need to allow for users to change the logging delay after its been set... we are adding a remove button also, so maybe just make them remove it .. i dunno

### DIFF
--- a/src/actions/Bluetooth/SensorDownloadActions.js
+++ b/src/actions/Bluetooth/SensorDownloadActions.js
@@ -75,38 +75,43 @@ const downloadLogsFromSensor = sensor => async dispatch => {
   dispatch(sensorDownloadStart(sensor));
 
   try {
-    const { macAddress, logInterval } = sensor;
+    const { macAddress, logInterval, logDelay } = sensor;
 
-    const downloadedLogsResult =
-      (await BleService().downloadLogsWithRetries(
-        macAddress,
-        VACCINE_CONSTANTS.MAX_BLUETOOTH_COMMAND_ATTEMPTS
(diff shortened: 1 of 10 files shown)


mark-prins (approved):
Beautiful!

One minor complaint - if you sit on the `VaccinePage` and wait for the delay to pass, this page doesn't refresh itself and continues to show the 'logging delayed until..' message even when the time has passed.

I think the option of removing and re-adding is fine if someone needs to tweak the delay - that fits with what I understand normal usage to be.

joshxg:
Ah nice, thanks.

I shifted the logic from a selector, which wouldn't get re-run until there was an action dispatched to the component, which is re-rendering every 10 seconds with the hook: 
9bdaea7

PR #3485, opened 2021-02-10
#3463 ability to pause temperature downloading

### LINKED ISSUE #3463: Ability to pause temperature downloading

> function must be programmed to include a ‘pause period’ 

This is in the new spec for 30 day fridge loggers. We currently don't have this functionality.

We need:
- UI to pause a sensor.
- Display in the UI when a sensor is paused.
- Add logic to not download logs when a sensor is paused.

### UI
- Just a button in `EditSensorPage`
- Not sure

### Display when paused
- Where ever it says "last download time", display a paused icon?
- Have some state to display when a sensor is paused.. somehow/somewhere? Not sure where exactly to put this.

### Logic
- We still have #3374 to go which will check the `logDelay` time on `sensor` before starting to download.
- `logDelay` to be able to be `null` when paused - enables us to determine if a sensor is paused or delayed.
- `logDelay` defaults to `new Date(0)` (1/1/1970) which will always be less than the current time.
- Conditions for downloading become:
	- If `logDelay` is < current time, download
	- else don't download

## Implementation

TBD

???

## Additional context

spec:
[PQS_E006_TR06.3.pdf](https://github.com/openmsupply/mobile/files/5928338/PQS_E006_TR06.3.pdf)

mark-prins:
Fixes #3463 

## Change summary

Ability to pause temperature log download for a specific sensor.

- A play/pause icon button on the sensor settings page which toggles the state
- Saved to a new field ( `isPaused` )
- Sensor header shows the disabled wifi icon if the sensor is paused, with the text `PAUSED` ( found that having a pause icon here was confusing, to me at least )
- A sensor that is paused does not download temperature logs
- Couple of offroad fixes ( e.g. saving a sensor wasn't working )

## Testing

- [ ] Edit a sensor. Pause and check that the header shows 'PAUSED' and the icon and text on the button change
- [ ] Edit a sensor. Pause and don't save. Expected result: shouldn't save the state
- [ ] Edit a sensor. Pause and save. Expected result: should save the state
- [ ] Pause a sensor and check that it no longer syncs

### DIFF
--- a/src/actions/Bluetooth/SensorDownloadActions.js
+++ b/src/actions/Bluetooth/SensorDownloadActions.js
@@ -79,11 +79,11 @@ const downloadLogsFromSensor = sensor => async dispatch => {
   dispatch(sensorDownloadStart(sensor));
 
   try {
-    const { macAddress, logInterval, logDelay } = sensor;
+    const { macAddress, logInterval, logDelay, isPaused } = sensor;
 
     const timeNow = moment();
 
-    if (timeNow.isAfter(moment(logDelay))) {
+    if (timeNow.isAfter(moment(logDelay)) && !isPaused) {
       const downloadedLogsResult =
         (await BleService().downloadLogsWithRetries(
           macAddress,
--- a/src/actions/Entities/SensorActions.js
+++ b/src/actions/Entities/SensorActions.js
@@ -38,9 +38,10 @@ const update = (id, field, value) => ({
 
 const create = macAddress => async dispatch => {
   const defaultSensor = {
+    isPaused: false,
     location: {},
-    logInterval: 300,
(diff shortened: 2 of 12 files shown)


joshxg (approved):
beautifullll!

I thought the `Enable` was a bit weird but I was trying to think of alternatives and couldn't. I googled `unpause` and apparently it's a real word - don't think it's better.

I also thought the placement of the button on the edit page seemed out of place, but I couldn't think of a better place 🤣

mark-prins:
yeah, button placement was a problem. I thought maybe it was kinda related to the log delay ( and there was a bit of space in that row 🤷‍♂️  ) So, not ideal but works for now..

PR #3486, opened 2021-02-11
show permissions & services status

### LINKED ISSUE #3376: Handling of Bluetooth permissions/services

Would be good to have a clearer and persistent indication when the device is in a state which disallows bluetooth usage.

For example, if location services are not enabled, the user might receive popups which state that temperatures can't be downloaded, but may not know why - an indication for example in the nav bar which states: "Location services disabled, click to enable" would be a big benefit

TBD

## Implementation

TBD

TBD

## Additional context

n/a

mark-prins:
Fixes #3376 

## Change summary

Adds a wee icon to the top right - showing an exclamation if there's a problem:

Screenshot_2021-02-11-15-13-30-727

..or a settings icon if everything is fine:

Screenshot_2021-02-11-15-11-30-258

Pressing the icon shows a modal giving the status of the permission and services required. If something is disabled, pressing that line will prompt to enable:

Screenshot_2021-02-11-15-13-40-049

Screenshot_2021-02-11-15-11-37-358

One extra item is whether we need to display location services - which we currently don't have in the permissions checks. Was looking at `react-native-connectivity-status` - don't know if we need to do this though? seems to work ok as it is?

Also - I've only shown this when vaccines are enabled, as the permission checks are tied to vaccine enabled also. Should this be changed?

## Testing

- [ ] Disable storage permission and start the app. confirm that the settings icon shows an exclamation sign
- [ ] Press the settings/exclamation icon and confirm the modal shows
- [ ] With storage permission disabled, press the settings/exclamation icon and confirm that the 'Storage' shows as disabled
- [ ] With storage permission disabled, press the settings/exclamation icon. In the modal, press 'Storage'. confirm that the settings prompt shows
- [ ] With the Storage settings prompt showing, press 'deny'. confirm that the 'Storage' permission still shows as disabled
- [ ] With the Storage settings prompt showing, press 'allow'. confirm that the 'Storage' permission shows as enabled
- [ ] In the settings modal, press 'Storage' when storage is enabled. Confirm that the storage prompt does not show
- [ ] With location permission disabled, press the settings/exclamation icon and confirm that the 'Location' shows as disabled
- [ ] With location permission disabled, press the settings/exclamation icon. In the modal, press 'Location'. confirm that the settings prompt shows
- [ ] With the Location settings prompt showing, press 'deny'. confirm that the 'Location' permission still shows as disabled
- [ ] With the Location settings prompt showing, press 'allow'. confirm that the 'Location' permission shows as enabled
- [ ] In the settings modal, press 'Location' when location is enabled. Confirm that the location prompt does not show

### DIFF
--- a/src/widgets/SettingsIcon.js
+++ b/src/widgets/SettingsIcon.js
@@ -1,0 +1,153 @@
+/**
+ * mSupply Mobile
+ * Sustainable Solutions (NZ) Ltd. 2021
+ */
+
+import React from 'react';
+import { connect } from 'react-redux';
+import { StyleSheet, Text, TouchableOpacity, View } from 'react-native';
+import PropTypes from 'prop-types';
+
+import { selectUsingVaccines } from '../selectors/modules';
+import { PermissionSelectors } from '../selectors/permission';
+import { PermissionActions } from '../actions/PermissionActions';
+
+import { useToggle } from '../hooks/index';
+import { CancelIcon, CogIcon, ConfirmIcon, DisabledIcon, HazardIcon } from './icons';
+import { PaperModalContainer } from './PaperModal/PaperModalContainer';
+import { IconButton } from './IconButton';
+import { FlexRow } from './FlexRow';
+
+import { DARK_GREY } from '../globalStyles';
+import { modalStrings } from '../localization';
+
+const PermissionRow = ({ comment, enabled, onPress, label }) => (
+  <TouchableOpacity onPress={enabled ? null : onPress}>
+    <FlexRow alignItems="flex-start" flex={0} style={localStyles.permissionRowContainer}>
+      {enabled ? <ConfirmIcon /> : <DisabledIcon />}
+      <FlexRow style={{ flexDirection: 'column', marginLeft: 10 }}>
+        <Text style={localStyles.permissionRowText}>{label}</Text>
+        <Text style={localStyles.permissionRowComment}>{comment}</Text>
+      </FlexRow>
+    </FlexRow>
+  </TouchableOpacity>
+);
+
+PermissionRow.defaultProps = {
+  comment: '',
+};
+
+PermissionRow.propTypes = {
+  comment: PropTypes.string,
+  enabled: PropTypes.bool.isRequired,
+  label: PropTypes.string.isRequired,
+  onPress: PropTypes.func.isRequired,
+};
+
+const SettingsIconComponent = ({
+  bluetooth,
+  location,
+  requestBluetooth,
+  requestLocation,
+  requestWriteStorage,
+  usingVaccines,
+  writeStorage,
+}) => {
+  const [settingsModalOpen, toggleSettingsModal] = useToggle();
+
+  if (!usingVaccines) {
+    return null;
+  }
+
+  const icon =
+    location && bluetooth && writeStorage ? (
+      <CogIcon color={DARK_GREY} />
+    ) : (
+      <HazardIcon color={DARK_GREY} size={20} />
+    );
+
+  return (
+    <>
+      <IconButton Icon={icon} containerStyle={{ marginLeft: 8 }} onPress={toggleSettingsModal} />
+
+      <PaperModalContainer isVisible={settingsModalOpen} onClose={toggleSettingsModal}>
+        <View style={localStyles.container}>
+          <FlexRow
+            alignItems="flex-start"
+            justifyContent="flex-end"
+            style={localStyles.cancelContainer}
+            flex={0}
+          >
+            <IconButton onPress={toggleSettingsModal} Icon={<CancelIcon />} />
+          </FlexRow>
+
+          <Text style={localStyles.heading}>{modalStrings.permissions}</Text>
+          <PermissionRow
+            onPress={requestWriteStorage}
+            label={modalStrings.storage_permission}
+            comment={modalStrings.storage_permission_comment}
+            enabled={writeStorage}
+          />
+          <PermissionRow
+            onPress={requestLocation}
+            label={modalStrings.location_permission}
+            comment={modalStrings.location_permission_comment}
+            enabled={location}
+          />
+          <Text style={localStyles.heading}>{modalStrings.services}</Text>
+          <PermissionRow
+            onPress={requestBluetooth}
+            label={modalStrings.bluetooth_permission}
+            comment={modalStrings.bluetooth_permission_comment}
+            enabled={bluetooth}
+          />
+        </View>
+      </PaperModalContainer>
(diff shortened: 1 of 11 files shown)


mark-prins:
On my tablet - disabling bluetooth doesn't make the bluetooth service show as disabled.
Thoughts as to how to test this?

katherine-sussol (approved):
This looks neat and works well on my end!

I couldn't stick Bluetooth disabled on the Lenovo I have either 😆 , even with ✈️  mode enabled the app seems to just be able to turn on BT. (Commenting out Line 30 in `PermissionsActions` does it but then kind of makes it not testable without changing the component haha - at least we know it's successfully force-enabling BT? 😹 ).

Something minor I also noticed was if prompted for location permissions during the regular vaccines stuff before opening the permissions settings, if accepted, it still shows up as disabled in the modal (until you click it).

src/widgets/SettingsIcon.js
```diff
@@ -0,0 +1,153 @@
...
+  const icon =
+    location && bluetooth && writeStorage ? (
+      <CogIcon color={DARK_GREY} />
+    ) : (
+      <HazardIcon color={DARK_GREY} size={20} />
+    );
+
+  return (
+    <>
+      <IconButton Icon={icon} containerStyle={{ marginLeft: 8 }} onPress={toggleSettingsModal} />
```
joshxg, line 71: I have fat fingers and had a litttle bit of an issue clicking this - not sure if it's worth the time to cater for this though 🤣

src/widgets/SettingsIcon.js
```diff
@@ -0,0 +1,153 @@
...
+SettingsIconComponent.propTypes = {
+  bluetooth: PropTypes.bool.isRequired,
+  location: PropTypes.bool.isRequired,
+  requestBluetooth: PropTypes.func.isRequired,
+  requestLocation: PropTypes.func.isRequired,
+  requestWriteStorage: PropTypes.func.isRequired,
+  usingVaccines: PropTypes.bool.isRequired,
+  writeStorage: PropTypes.bool.isRequired,
+};
+export const SettingsIcon = connect(stateToProps, dispatchToProps)(SettingsIconComponent);
```
joshxg, line 153: I like this component a lot 😄 

Maaaaybe `PermissionSettingsIcon` might be more accurate. Until we put other stuff in there. Also services. 🤷 

Was also thinking of having some sort of extensibility plan - when we need to also have things for barcode scanners n stuff, the amount of conditionals might blow up, - constructing the 'list' of permissions with a builder pattern or something - buuuut, future problems..?

joshxg (approved):
This is great 😄 

The bug @katherine-sussol mentioned about enabling location permission and the modal still showing an old value is in https://github.com/openmsupply/mobile/blob/655d00910e1cee7cc0a1b2eaf3adce857c118e97/src/actions/PermissionActions.js#L55-L59

```
  const result = !locationPermission
    ? (await request(PERMISSIONS.ANDROID.ACCESS_COARSE_LOCATION)) === RESULTS.GRANTED
    : true;

  dispatch(setLocation(result));
```

Some stuff I've learnt:

### Location

Disabling `Location` when you bring down the 'top menu thing' where wifi and stuff is will just make the bluetooth not pick stuff up. Nothing will throw run time errors or anything like that. Can test this on a real device by turning it off, then clicking new sensor - can see you don't scan anything in the office. Then go back and restart the scanner with `Location` on, and you start picking them up. Really this will just be bad UX for people, I guess.

For enabling this programmatically:

https://github.com/nearit/react-native-connectivity-status has: 

```
Enable services
NOTE: Due to possible app rejection from Apple (caused by illegal usage of private URL Scheme "prefs:root" or "App-Prefs:root"), methods for enabling bluetooth and location services have been removed from this module.
```

The only package I could find which would allow programmatically enabling this was https://github.com/c19354837/react-native-system-setting - which if I remember correctly, opens up a devices `Settings` at the `Location` area, rather than straight enabling it.

This is all to say: I think we probably should use that package and have another row to enable it. I guess? Nice thing about native apps - bundle sizes don't matter as much? 🤣 

### Bluetooth

Bluetooth can be enabled programmatically on *most* devices - I've seen reports that you're not able to do this on some devices 🤷 

We're listening to the current bluetooth status changes and enable this on a listener, which is why it 'wont turn off' - here: https://github.com/openmsupply/mobile/blob/655d00910e1cee7cc0a1b2eaf3adce857c118e97/src/mSupplyMobileApp.js#L98

src/widgets/SettingsIcon.js
```diff
@@ -0,0 +1,153 @@
...
+
+  if (!usingVaccines) {
+    return null;
+  }
+
+  const icon =
+    location && bluetooth && writeStorage ? (
+      <CogIcon color={DARK_GREY} />
+    ) : (
+      <HazardIcon color={DARK_GREY} size={20} />
```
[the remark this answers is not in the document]
mark-prins: 😁  I sooo wanted to. I resisted because I didn't want to be too different from the current design.
Your prompting was needed, thanks!
it looks good, I think. I've used `SUSSOL_ORANGE` though, as that matches the `InfoBadge` and menu page icons.

Screen Shot 2021-02-12 at 9 48 36 AM

I've also demoted the cog from `DARK_GREY` to `GREY` as this is more in keeping with the rest of that header, and also the settings cog down the bottom.

Screen Shot 2021-02-12 at 9 55 21 AM

Another concern was whether the modal needed some introductory text, explaining what was happening and that you can press to enable. I think that it does, so have had a go..

Screen Shot 2021-02-12 at 10 04 00 AM

src/widgets/SettingsIcon.js
mark-prins: Yes. the name `SettingsIcon` is too generic. `PermissionSettingsIcon` works for me, I just worry that having `Services` in it? 

I'm also troubled at having a little cog icon at the top right, and having a settings button with a little cog icon down the bottom right. That's confusing isn't it?

mark-prins: And yes - re extensibility. absolutely. My preference is to look at that later 😉  I'll use the excuse of 'this issue is big enough already' although really the reason is 'that sounds like a lot of work and I'm lazy'

...though I would like to try the location services, will have a go at including the other package and testing that now

katherine-sussol: I thought the double cog icons weren't too confusing, the permissions one being next to the sync stuff was enough of an indication it was probably sync related in my mind 🤷 . Could use a wrench to be different if it's weird that there's two similar buttons doing different things? Of course I have to suggest a cat too, but I think that won't help here 😿 .

PR #3489, opened 2021-02-15
#3488 adjust cumulative breach units

### LINKED ISSUE #3488: Cumulative breach calculation using different units for Duration & LogInterval
## Describe the bug

Breach threshold duration is being stored in milliseconds, whereas LogInterval uses seconds.
The calculation to determine whether there is a cumulative breach or not should convert one of them to use the same unit when doing comparison (milliseconds?)

This results in the UI showing no cumulative breaches when there has actually been one 🙀 

### To reproduce

1. Setup test data (Settings -> Generate vaccine data)
2. Navigate to vaccines page and edit 'Sensor 2' (as it has lots of logs that should create hot breaches)
3. Set the 'Hot cumulative' threshold duration to 2 minutes and save:
Screen Shot 2021-02-15 at 5 13 41 PM
4. See no cumulative breaches 
Screen Shot 2021-02-15 at 5 15 56 PM

### Expected behaviour
Expected to see '16h' in the hot cumulative card (this figure is based on the test data that got generated on my machine - it may not be the same but there should be -some- breach).

### Proposed Solution
Convert logInterval to milliseconds before doing a comparison?

### Version and device info

- App version: develop
- Tablet model: Emulator
- OS version: Android 8.1.0

### Additional context
- Found while investigating for #3474 
- This also applies to cold breaches - use test 'sensor 3' for this one

katherine-sussol:
Fixes #3488 

## Change summary

- Adjusts cumulative breach calculation 

## Testing

- [ ] Follow steps in original issue https://github.com/openmsupply/mobile/issues/3488 - a hot cumulative breach should show up for test sensor 2 when the duration is set to 2 minutes, and a cold cumulative breach should show up for test sensor 3 when the duration is set to 2 minutes

- Might be a separate issue that the card doesn't update when a download has occurred in background, to investigate separately

### DIFF
--- a/src/selectors/fridge.js
+++ b/src/selectors/fridge.js
@@ -5,7 +5,8 @@
 import moment from 'moment';
 import { createSelector } from 'reselect';
 import { UIDatabase } from '../database';
-import { chunk } from '../utilities';
+import { chunk, MILLISECONDS } from '../utilities';
+
 import { CHART_CONSTANTS, VACCINE_CONSTANTS } from '../utilities/modules/vaccines/constants';
 
 export const selectSelectedFridgeID = ({ fridge }) => {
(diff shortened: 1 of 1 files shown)


mark-prins (approved):
LGTM!

PR #3499, opened 2021-02-15
#3495 Show temperature is n/a not 0 on vaccine module page

### LINKED ISSUE #3495: When adding a sensor for the first time, temperature should be N/A not 0

When a sensor has not had any temperature readings, we should display 'N/A' rather than 0.0C

When a sensor has not had any temperature readings, we should display 'N/A' rather than 0.0C

## Implementation

Not sure will have to check

N/A

## Additional context

N/A

joshxg:
Fixes #3495 

## Change summary

- Shows the temperature of a sensor with no logs is N/A instead of `0` by not assigning `0` as a default temperature, and returning the invalid pattern in `.temperature()`

## Testing

- [ ] Add a new sensor, the temperature should be N/A not `0.0`

N/A

### DIFF
--- a/src/database/DataTypes/Sensor.js
+++ b/src/database/DataTypes/Sensor.js
@@ -35,10 +35,6 @@ export class Sensor extends Realm.Object {
     return this.sendCommand('*logall');
   }
 
-  get breachConfigIDs() {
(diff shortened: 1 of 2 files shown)


mark-prins (approved):
yeah, much better. 👍

PR #3502, opened 2021-02-15
#3501 HazardIcon to use the full dataset

### LINKED ISSUE #3501: Breach hazard icons on the line chart aren't opening the breach modal
## Describe the bug

The breach icons on the line chart aren't pressable

### To reproduce

1. Go to `FridgeDetailPage` with a sensor which has breached
2. Click on the breach icon

### Expected behaviour

Should open the `BreachDisplay` modal

### Proposed Solution

TBD

### Version and device info

- App version: 7.0 dev
- Tablet model:
- OS version:

### Additional context

Not sure why

joshxg:
Fixes #3501 

## Change summary

- Seems the `dataComponent` was previously being passed the `datum` component which was a real breach object, and it is now only an object `{x: .., y: ..}`, so, it doesn't have an `id` to pass back to find the breach. So instead, pass the full data set to the icon component as it is passed the `idx` of which data element it is... 
- Also simplified `BreachDisplay` and removed the weird logic which was trying to limit the number of data points and displaying a spinner with `AfterInteractions`
- Also made the hazard icon more clickable by chucking a bigger rect around it, sort of. 

## Testing

- [ ] Clicking a hazard icon on the `FridgeDetailPage` opens a breach modal

There's plenty wrong with the `breachmodal` itself but.. for another issue, I think

image

### DIFF
--- a/src/widgets/HazardPoint.js
+++ b/src/widgets/HazardPoint.js
@@ -7,9 +7,9 @@
 
 import React from 'react';
 import PropTypes from 'prop-types';
-import Svg, { Path, G } from 'react-native-svg';
+import Svg, { Rect, Path, G } from 'react-native-svg';
 
-import { SUSSOL_ORANGE } from '../globalStyles/colors';
+import { SUSSOL_ORANGE, TRANSPARENT } from '../globalStyles/colors';
 
 const hazardPath = `M569.517 440.013C587.975 472.007 564.806 512 527.94 512H48.054c-36.937 0-59.999-40.055-41.577-71.987L246.423
  23.985c18.467-32.009 64.72-31.951 83.154 0l239.94 416.028zM288 354c-25.405 0-46 20.595-46 46s20.595 46 46 46 46-20.595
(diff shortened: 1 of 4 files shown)


src/widgets/HazardPoint.js
```diff
@@ -20,14 +20,22 @@ const hazardPath = `M569.517 440.013C587.975 472.007 564.806 512 527.94 512H48.0
...
   const { xOffset, yOffset, scale, fill } = hazardPointStyles;
 
-  const onPressWrapper = React.useCallback(() => onPress?.(datum?.id), []);
+  const onPressWrapper = React.useCallback(() => onPress?.(dataSet[index]?.id), []);
 
   return (
     <Svg>
-      <G onPressIn={onPressWrapper} delayPressIn={0}>
+      <G>
+        <Rect
```
mark-prins, line 31: Could you swap the order of the `Rect` and the `Path` please?
On my emulator I have to click just beside the hazard icon to get the modal to show, pressing on the hazard icon doesn't do anything. Changing the order here puts the rectangle on top and enables the press on the icon.

mark-prins (approved):
That fixes it! Thanks.
I have a minor request on ordering of SVG items, not essential.

src/widgets/HazardPoint.js
joshxg: Oh nice! 3b2301e

PR #3503, opened 2021-02-15
#3493 Default of 100 for battery level

### LINKED ISSUE #3493: When adding a sensor, it is already in danger

When adding a sensor for the first time, it is automatically in danger and has a flashing light.

I think this is because sensors by default have a battery level of 0.

Think a default battery level of 100 is better.

## Implementation

TBD

Not sure

## Additional context

Testing on a tablet so can't tell exactly

joshxg:
Fixes #3493 

## Change summary

- Adds a default of 100 for the battery level of a sensor. It's not ideal, ideally we'd scan straight away and pick up the real battery level, but that went in the too hard basket - in reality, it won't really be too long before a download occurs and the battery level is picked up correctly

## Testing

- [ ] When adding a sensor the sensor doesn't have a flashing red circle and has a battery level of 100

N/A

### DIFF
--- a/src/bluetooth/SensorManager.js
+++ b/src/bluetooth/SensorManager.js
@@ -40,7 +40,7 @@ class SensorManager {
       name,
       location,
       logInterval,
-      batteryLevel: 0,
+      batteryLevel: 100,
       isActive: true,
     });
   };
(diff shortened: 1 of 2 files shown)


katherine-sussol (approved):
Looks good! Definitely better than false sense of alarm 🔋

mark-prins (commented):
What about making `batteryLevel` optional in the DataType - giving it a default of `undefined` in `SensorManager` and putting an option in `formatBatteryLevel` to display `-` or blank or `?`
.. that way we aren't misleading with an incorrect number value. As you say, it will be updated pretty quickly with the correct value anyway.

( I'd also added a `batteryLevel && ` to the two selectors which are checking for low battery level . all seems to work ok )

joshxg:
I think you're right - much better solution

PR #3511, opened 2021-02-17
#3386 FridgeDetail page updates

### LINKED ISSUE #3386: FridgeDetail performance

joshxg:
Fixes #3386 

## Change summary

- OK well .. 'performance' is somewhat increased, I guess.
- Changed the `FridgeReducer` to `FridgeDetailReducer` so it was a little clearer it was for the `FridgeDetail` page specifically..?
- Changed the shape of `FridgeDetailReducer` to use primitives over objects. This is primarily so that we can use the `toDate` as the diff for updating when new logs are downloaded.
- As above, this is changed slightly such that the `toDate` will change, if you're currently viewing 'today' after a download of logs (the `mapStateToProps` func will run, causing the `toDate` to be recalculated, it will then be a different unix timestamp and cause a re-render).
- Changed back the sync interval to `10` minutes. I didn't mean to change it.
- Changed some selectors to 1) account for using unix timestamps and 2) calculate most of the chart data in a single selector, in a single loop, using database queries. This now accounts for `null` when there are no temperatures in a period - also now should be quite a bit quicker when there are 1000's of logs over a longer period (realm operations far quicker than array/math operations like `math.max` on an array of 100's/1000's of logs)

## Testing

- [ ] FridgeDetailPage operates as it did before - so good test case

I hope this is better ..


mark-prins (approved):
The changes looked ok to me!
Page itself is still a little sluggish, nothing too bad though. It loads much more nicely now - with components showing after the page loads. Overall, the page is looking really good.

PR #3519, opened 2021-02-17
#3517 is paused state is saved

### LINKED ISSUE #3517: Removed sensors keep their is paused state
## Describe the bug

After removing a sensor, if it is paused, when re-added, it is re-added in a paused state.

### To reproduce

1. Go to edit a sensor and pause it - save it.
2. remove the sensor
3. Go to add the same sensor

### Expected behaviour

Newly added sensors should not be paused

### Proposed Solution

N/A

### Version and device info

- App version: 7.0.0-RC3
- Tablet model:
- OS version:

### Additional context

N/A

joshxg:
Fixes #3517 

## Change summary

- When removing/replacing, explicitly set state correctly.

## Testing

- [ ] When removing a sensor which is paused and readding it, it is not paused.
- [ ] When adding a sensor which was previously removed, it is added to the list of sensors 

N/A

### DIFF
--- a/src/actions/Entities/SensorActions.js
+++ b/src/actions/Entities/SensorActions.js
@@ -127,6 +127,7 @@ const removeSensor = sensorId => dispatch => {
     UIDatabase.update('Sensor', {
       id: sensor.id,
       isActive: false,
+      isPaused: false,
     });
   });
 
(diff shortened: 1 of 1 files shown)


mark-prins (approved):
Working nicely 👍

PR #3538, opened 2021-02-17
#3525 default battery level null

### LINKED ISSUE #3525: Default batteryLevel should be null and display n/a
## Describe the bug

The default battery level should be `null` and and display `n/a`

### To reproduce

1. Add a new snsor

### Expected behaviour

As above

### Proposed Solution

One issue is that 4d can't store `null` in the db (or at least, more fields map `null` to a default value

### Version and device info

- App version: 7.0
- Tablet model:
- OS version:

### Additional context

N/A

mark-prins:
Fixes #3525

## Change summary

`batteryLevel` of a sensor is now optional with a default of undefined.
Displays `N/A` if the level is undefined. If a sensor is removed and re-added, the level is reset to `undefined` ( we don't know how long the sensor has been removed )
Have defaulted to `100` for outgoing sync, if the value is `undefined` as I thought desktop would be happier that way.

## Testing

- [ ] Add a new sensor, observe the battery level. This should be `N/A` until a log download occurs
- [ ] The battery level should update and display correctly within a minute of the delayed log download time
- [ ] Remove and re-add a sensor. The battery level should display as `N/A` until a sync occurs


src/database/schema.js
```diff
@@ -214,7 +214,7 @@ export const schema = {
     VaccineVialMonitorStatus,
     VaccineVialMonitorStatusLog,
   ],
-  schemaVersion: 19,
+  schemaVersion: 20,
```
joshxg, line 217: Master is currently `18` so I think it's best not to bump this past `19` for this release?

src/selectors/fridge.js
```diff
@@ -251,7 +251,7 @@ export const selectSelectedFridgeIsInColdBreach = state => {
 export const selectSelectedFridgeSensorIsLowBattery = state => {
   const fridge = selectSelectedFridge(state);
   const { batteryLevel } = fridge;
-  return batteryLevel <= VACCINE_CONSTANTS.LOW_BATTERY_PERCENTAGE;
+  return batteryLevel && batteryLevel <= VACCINE_CONSTANTS.LOW_BATTERY_PERCENTAGE;
```
joshxg, line 254: Had to check to make sure but if this is `0` then this returns `0` (or  `isLowBattery(0)` == `false`).

image

joshxg (changes requested):
Works well! Couple of minors - assuming the diff gets fixed up when I merge stuff

PR #3554, opened 2021-02-18
#3553 Unable to save a sensor

### LINKED ISSUE #3553: Navigating to the SensorEdit page from the FridgeDetail page has incorrect state
## Describe the bug

When navigating to the `SensorEditPage` from the `FridgeDetailPage` the `editingId` in the `LocationReducer` is not set correctly.

image

### To reproduce

1. Go to `SensorEditPage` from `FridgeDetailPage`
2. Click on Save

### Expected behaviour

Should be able to save nicely

### Proposed Solution

Add a `Sensor` getting `locationID`

### Version and device info

- App version: 7.0.0-RC4
- Tablet model:
- OS version:

### Additional context

Nice that it doesn't crash the app, at least!

joshxg:
Fixes #3553 

## Change summary

- Adds `locationID` to the sensor model which is the field which is on the plain object, which is what the sensor is when navigating from the `VaccinePage`.

## Testing

- [ ] Editing a sensor from the `FridgeDetailPage` can be saved.

N/A

### DIFF
--- a/src/database/DataTypes/Sensor.js
+++ b/src/database/DataTypes/Sensor.js
@@ -84,6 +84,10 @@ export class Sensor extends Realm.Object {
 
     return isOngoing && isHot;
   }
+
+  get locationID() {
+    return this.location?.id;
+  }
 }
 
 Sensor.schema = {


mark-prins (approved):
It works!

Issue #3564, opened 2021-02-18
Prescription confirmation laggy


mark-prins:
## Describe the bug

The final page of the dispensing process is very sluggish, and on investigation is re-rendering constantly

### To reproduce

Steps to reproduce the behaviour:

1. Go to Dispensing
2. Get to the final page

### Expected behaviour

Page should respond quickly to interactions

PR #3566, opened 2021-02-19
check for inactive on policies

### LINKED ISSUE #3565: Prescription: Inactive insurance policy applies discount
## Describe the bug

Selecting an inactive insurance policy will still apply discount to the payment.

### To reproduce

1. Create a prescription and add an item which has a price
2. On the confirmation screen, create or select an inactive insurance policy
4. Discount is applied

### Expected behaviour

Inactive policies should not discount the prescription

mark-prins:
Fixes #3565

## Change summary

Fixes the immediate problem. The confirmation screen could probably do with a more careful review, will leave that for #3564 

## Testing

- [ ] Create a prescription, add an item which has a cost, select an inactive policy. No discount should be applied.

### DIFF
--- a/src/selectors/insurance.js
+++ b/src/selectors/insurance.js
@@ -15,8 +15,8 @@ export const selectInsurancePolicyDiscountRate = ({ insurancePolicy }) => {
 };
 
 export const selectInsuranceDiscountRate = ({ insurance }) => {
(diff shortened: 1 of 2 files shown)


joshxg (approved):
Great. Yeah that whole confirmation screen is a bit of a nightmare!

PR #3574, opened 2021-02-20
#3569  Breaches only create, never continue

### LINKED ISSUE #3569: Temperature breaches are being created rather than continuing
## Describe the bug

When in breach, a temperature breach is created every download, rather than adding the logs to the previous breach.

### To reproduce

1. Go to add a sensor and let it download for a while while in a temperature that is in breach
2. check the realm and see there are a lot of breaches.

### Expected behaviour

Breaches should continue rather than be created contiuously

### Proposed Solution

TBD

### Version and device info

- App version: 7.0.0-RC5
- Tablet model:
- OS version:

### Additional context

N/A

joshxg:
Fixes #3569 

## Change summary

- Fair few changes. Primarily this was to ensure breaches are being continued rather than only created. To do this required mainly date format changes, which revealed a few other issues which i'll comment in the PR

## Testing
Set up a new sensor
- [ ] When the sensor goes in to a breach, all the logs for that breach are aggregated into a single temperature breach record
- [ ] When the sensor goes out of the breach, the end timestamp is correctly applied to the breach.

Trying to share the logic between the apps needed more thinking through, this is a bit of a mess sorry :(

### DIFF
--- a/src/actions/BreachActions.js
+++ b/src/actions/BreachActions.js
@@ -2,7 +2,7 @@
  * mSupply Mobile
  * Sustainable Solutions (NZ) Ltd. 2020
  */
-
+import moment from 'moment';
 import BreachManager from '../bluetooth/BreachManager';
 import { UIDatabase } from '../database';
 
@@ -23,18 +23,24 @@ const createConsecutiveSuccess = (sensor, updatedBreaches, updatedLogs) => ({
 });
 
 const createConsecutiveBreaches = sensor => async dispatch => {
-  const { id } = sensor;
+  const { id: sensorID } = sensor;
 
   try {
-    const logs = await BreachManager().getLogsToCheck(id);
+    const logs = await BreachManager().getLogsToCheck(sensorID);
+    const mappedToPlainObjects = logs.map(({ id, temperature, timestamp }) => ({
+      id,
+      temperature,
+      timestamp: moment(timestamp).unix(),
+    }));
     const configs = sensor.breachConfigs;
-    const mostRecentBreach = await BreachManager().getMostRecentBreach(id);
+    const mostRecentBreach = await BreachManager().getMostRecentBreach(sensorID);
     const [breaches, temperatureLogs] = await BreachManager().createBreaches(
       sensor,
-      logs,
+      mappedToPlainObjects,
       configs,
-      mostRecentBreach
+      mostRecentBreach?.toJSON()
     );
+
     const [updatedBreaches, updatedLogs] = await BreachManager().updateBreaches(
       breaches,
       temperatureLogs
--- a/src/bluetooth/BreachManager.js
+++ b/src/bluetooth/BreachManager.js
@@ -25,8 +25,8 @@ const createBreachRecord = (
     thresholdMaxTemperature,
     thresholdDuration,
     type,
-    startTimestamp,
-    endTimestamp,
+    startTimestamp: new Date(startTimestamp * MILLISECONDS.ONE_SECOND),
+    endTimestamp: endTimestamp ? new Date(endTimestamp * MILLISECONDS.ONE_SECOND) : endTimestamp,
     location,
     acknowledged: false,
   };
@@ -50,7 +50,7 @@ class BreachManager {
 
   closeBreach = (temperatureBreach, time) => {
     // eslint-disable-next-line no-param-reassign
-    temperatureBreach.endTimestamp = time;
+    temperatureBreach.endTimestamp = new Date(time * MILLISECONDS.ONE_SECOND);
     return temperatureBreach;
   };
 
@@ -69,6 +69,7 @@ class BreachManager {
     const { minimumTemperature, maximumTemperature, duration } = config;
     const { timestamp: endTimestamp } = logs[logs.length - 1];
     const { timestamp: startTimestamp } = logs[0];
+
     // logsDuration will be in seconds as it is based on unixtime
     // we are storing duration as milliseconds, so will have to scale to compare
     const logsDuration = MILLISECONDS.ONE_SECOND * (endTimestamp - startTimestamp);
@@ -78,6 +79,7 @@ class BreachManager {
       const { temperature } = log;
       return temperature <= maximumTemperature && temperature >= minimumTemperature;
     });
+
     return temperaturesWithinBounds;
   };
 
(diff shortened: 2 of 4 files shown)


src/actions/BreachActions.js
```diff
@@ -23,18 +23,24 @@ const createConsecutiveSuccess = (sensor, updatedBreaches, updatedLogs) => ({
 });
 
 const createConsecutiveBreaches = sensor => async dispatch => {
-  const { id } = sensor;
+  const { id: sensorID } = sensor;
 
   try {
-    const logs = await BreachManager().getLogsToCheck(id);
+    const logs = await BreachManager().getLogsToCheck(sensorID);
+    const mappedToPlainObjects = logs.map(({ id, temperature, timestamp }) => ({
```
joshxg, line 30: When we assign a `breach` in the `BreachManager` to a log, we are actually assigning to a realm object which would throw an error because you can only edit realm objects within a realm transaction, so mapped this to a plain object.. this only started happening when I corrected the the logic to actually close off breaches

joshxg: Maybe should add a `toJSON` to the `TemperatureLog` class

src/actions/BreachActions.js
```diff
@@ -23,18 +23,24 @@ const createConsecutiveSuccess = (sensor, updatedBreaches, updatedLogs) => ({
...
     const configs = sensor.breachConfigs;
-    const mostRecentBreach = await BreachManager().getMostRecentBreach(id);
+    const mostRecentBreach = await BreachManager().getMostRecentBreach(sensorID);
     const [breaches, temperatureLogs] = await BreachManager().createBreaches(
       sensor,
-      logs,
+      mappedToPlainObjects,
       configs,
-      mostRecentBreach
+      mostRecentBreach?.toJSON()
```
joshxg, line 41: Same logic with the log above, just now when assigning an `endTimestamp` to the breach

src/bluetooth/BreachManager.js
```diff
@@ -25,8 +25,8 @@ const createBreachRecord = (
     thresholdMaxTemperature,
     thresholdDuration,
     type,
-    startTimestamp,
-    endTimestamp,
+    startTimestamp: new Date(startTimestamp * MILLISECONDS.ONE_SECOND),
+    endTimestamp: endTimestamp ? new Date(endTimestamp * MILLISECONDS.ONE_SECOND) : endTimestamp,
```
joshxg, line 29: when creating a breach, the logs from the bluetooth are a unix timestamp

src/bluetooth/BreachManager.js
```diff
@@ -170,17 +174,11 @@ class BreachManager {
   getTemperatureLogsFrom = async (sensorId, timeToCheckFrom) =>
     this.db.getTemperatureLogsFrom(sensorId, timeToCheckFrom);
 
-  getMostRecentBreachLog = async sensorId => {
-    this.db.getMostRecentBreachLog(sensorId);
-  };
+  getMostRecentBreachLog = async sensorId => this.db.getMostRecentBreachLog(sensorId);
```
joshxg, line 177: these methods weren't returning anything

src/bluetooth/VaccineDataAccess.js
```diff
@@ -48,12 +48,18 @@ export class VaccineDataAccess {
...
+  upsertBreaches = breaches => {
     this.db.write(() => {
       breaches.forEach(breach => {
-        this.db.update(VACCINE_ENTITIES.TEMPERATURE_BREACH, breach);
+        this.db.update(VACCINE_ENTITIES.TEMPERATURE_BREACH, {
+          ...breach,
+          id: breach.id,
+          startTimestamp: new Date(breach.startTimestamp),
+          endTimestamp: breach.endTimestamp ? new Date(breach.endTimestamp) : breach.endTimestamp,
+        });
```
joshxg, line 59: Ensuring dates are Date's when saving to realm

src/database/DataTypes/TemperatureBreach.js
```diff
@@ -82,6 +82,21 @@ export class TemperatureBreach extends Realm.Object {
...
+      endTimestamp: this.endTimestamp?.getTime(),
+      locationID: this.location.id,
+      thresholdMaxTemperature: this.thresholdMaxTemperature,
+      thresholdMinTemperature: this.thresholdMinTemperature,
+      thresholdDuration: this.thresholdDuration,
+      acknowledged: this.acknowledged,
+      type: this.type,
+      sensorID: this.sensor.id,
+    };
+  }
```
joshxg, line 99: Is this name OK or `toObject` is better?

joshxg (commented):
Hm this took a while to figure out. Not happy about all the mess 😢

src/bluetooth/VaccineDataAccess.js
katherine-sussol: So many different kinds of dates and objects 😕  eep. I think the default here is (just a little) counter-intuitive, more explicit to just set to `null` if no `endTimestamp`?

src/database/DataTypes/TemperatureBreach.js
katherine-sussol: I like `JSON`, less ambiguity between regular JS and Realm object 🤔

katherine-sussol (approved):
Looking good 👍 . 

I got a little trapped by the fact the breach configs are copied to the breach record (i.e. can't 'end' a breach by simply increasing the threshold) 😆  but was able to successfully end a hot consecutive breach by mocking a temperature to something lower than the original threshold.

joshxg:
Thanks @katherine-sussol 

I think it's maybe a good thing that breaches are not updated with new configs settings, but I'm not sure. I think it's probably a bit of an edge case, anyway and people most likely wont be updating the breach configs very frequently.. 🤷

mark-prins (approved):
Looks fine to me :) Seems to work.. though I trust @katherine-sussol's testing more than mine 😉

mark-prins:
I see the `JSON` comment now..
Maybe `toJSONObject` 🤷‍♂️

katherine-sussol:
> I see the `JSON` comment now..
> Maybe `toJSONObject` 🤷‍♂️

Or what about `toJSObject`? 😆  I think it's the notation part that denotes stringy-ness? Hah now there are too many options. Oh no 🏃‍♀️

joshxg:
I'm gonna merge to get these pr's in - I think `toObject` is probably the way to go, so I'll make a new PR for that with the other ones I've added in other PR's, heh.

Issue #3571, opened 2021-02-20
isInHotBreach/isInColdBreach state is not updated after breaches are created


joshxg:
## Describe the bug

After downloading then creating breaches, the state of a sensor bein gin a breach is not updated.

### To reproduce

1. Go to Vaccine page and wait until a sensor is in breach
2. see the state not udpate.

### Expected behaviour

When on the vaccine page or fridge detail pages, after a sensor enters a breach, the state should update.

### Proposed Solution

Hook into the create_consecutive_success hook

### Version and device info

- App version: 7.0.0-RC5
- Tablet model:
- OS version:

### Additional context

N/A

PR #3576, opened 2021-02-21
#3572 battery level not updated

### LINKED ISSUE #3572: Battery level state is not updated
## Describe the bug

When on vaccine page or some other page showing the battery level, after a download, the battery level is not updated

### To reproduce

1. Go to add a new sensor - battery level will be `N/A`
2. Wait for a download to happen.

### Expected behaviour

The battery level should be updated.

### Proposed Solution

TBD
### Version and device info

- App version: 7.0.0-RC5
- Tablet model:
- OS version:

### Additional context

N/A

joshxg:
Fixes #3572 #3571 

## Change summary

- Added `toJSON` to `Location`/`Sensor`/`TemperatureConfig` classes, which is used to 'refresh' the state from a realm object.

## Testing

- [ ] add a new sensor, wait for a download. The battery level should have updated without navigating away from the vaccine page

N/A

### DIFF
--- a/src/actions/Entities/SensorActions.js
+++ b/src/actions/Entities/SensorActions.js
@@ -88,6 +88,7 @@ const save = () => (dispatch, getState) => {
     updatedLocation = UIDatabase.update('Location', location);
     updatedSensor = UIDatabase.update('Sensor', {
       ...sensor,
+      programmedDate: new Date(sensor?.programmedDate ?? 0),
       logDelay: new Date(sensor?.logDelay ?? 0),
       location: updatedLocation,
     });
--- a/src/database/DataTypes/Location.js
+++ b/src/database/DataTypes/Location.js
@@ -95,6 +95,14 @@ export class Location extends Realm.Object {
   get batteryLevel() {
     return this.sensor?.batteryLevel ?? 0;
   }
+
+  toJSON() {
+    return {
+      id: this.id,
+      description: this.description,
+      code: this.code,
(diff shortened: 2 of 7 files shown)


mark-prins:
so I guess we'll run with `toJSON` then 😄 

Any tricks for reloading the app? I just get an error saying that `location.toJSON` is not a function

joshxg:
🤔 

Weird. Whenever I get something weird I close all my terminals and run `yarn clean-mac-cache` and it usually fixes things....

I am happy to change all of the members to `toObject` - I think it does avoid any confusion!

mark-prins:
Nah, `toJSON` is fine.

joshxg:
It could be that the `location` object is not actually a realm instance, and so it really doesn't have a `toJSON` field, also.. 🤔 

Which flow are you doing which causes that to occur? I thought I tested them all...

You can also test if something is a realm object by checking if it has a `objectSchema` field.. if it does, it's some realm object instance (and so, should have the `toJSON` field

mark-prins:
just adding a new sensor, nothing tricksy.
I think I see it.. in `LocationActions` it creates a default location when going from step one to step two.
The default location is a plain object

joshxg:
@mark-prins ugh sorry about that: 844e2d9

joshxg:
@mark-prins 
c558428

mark-prins (approved):
Tested and working 😄

PR #3577, opened 2021-02-21
#3562 Set a logdelay when resuming a sensor

### LINKED ISSUE #3562: Logs are integrated for times when the sensor was paused.
## Describe the bug

- I pause a sensor at 12.
- No logs are downloaded 👍 
- I unpaused at 12.15
- All logs are downloaded and integrated back to 12 👎 

### To reproduce

1. Pause a sensor for awhile
3. Unpaused and check logs after a download

### Expected behaviour

Logs should not be integrated for the times when the sensor was paused.

### Proposed Solution

Need:

https://github.com/openmsupply/mobile/issues/3560
https://github.com/openmsupply/mobile/issues/3561

After those, if we set the logging delay to the time now after unpausing, this should fix it.

### Version and device info

- App version: 7.0.0-RC4
- Tablet model: N/A
- OS version: N/A

### Additional context

N/A

joshxg:
Fixes #3562 

## Change summary

- Technically not quite correct - you could resume while there was a log delay and bring the logdelay forward. Could also set a logdelay when the sensor was never actually paused. I think you very rarely set a sensor up and these edge cases won't hoping practically so I didn't bother accounting for them....... is that bad?

## Testing

- [ ] After resuming a sensor after some X minutes, no logs are saved for the time the sensor was paused.

N/A

### DIFF
--- a/src/pages/SensorEditPage.js
+++ b/src/pages/SensorEditPage.js
@@ -280,7 +280,12 @@ const dispatchToProps = (dispatch, ownProps) => {
     const field = isHot ? 'minimumTemperature' : 'maximumTemperature';
     dispatch(TemperatureBreachConfigActions.update(id, field, value));
   };
-  const updateIsPaused = isPaused => dispatch(SensorActions.update(sensorID, 'isPaused', isPaused));
+  const updateIsPaused = isPaused => {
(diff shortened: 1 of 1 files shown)


mark-prins (approved):
very cunning!
Haven't tested as I'm having issues while trying to test another issue. looks good to me

PR #3578, opened 2021-02-21
#3557 Merge state ratehr than overwriting after a sync occurs

### LINKED ISSUE #3557: Creating a new Sensor while sync occurs can create invalid state
## Describe the bug

While you're creating a sensor, if sync occurs, then the state becomes invalid.

### To reproduce

1. Go to create a new sensor, go to the sensor thresholds page
3. wait for sync to occur

### Expected behaviour

Sync should not effect the current state.

### Proposed Solution

- This case was added to handle new records coming in from sync after initializing, primarily: https://github.com/openmsupply/mobile/blob/3cfacf047f7e99cd3b58ec17136520eed16d622e/src/reducers/Entities/TemperatureBreachConfigReducer.js#L38-L42

I believe this needs to account for the previous by id state and not wipe the newly added ids in `newIds`

### Version and device info

- App version: 7.0.0-rc4
- Tablet model: N/A
- OS version: N/A

### Additional context

This might be similar to https://github.com/openmsupply/mobile/issues/3552 - maybe?

joshxg:
Fixes #3557 

## Change summary

- Added merged state rather than overwriting in Location/Sensor/Config reducers

## Testing

- [ ] When creating a sensor, wait for a sync to occur (Manually triggering should work I think). The state should be exactly the same.

N/A

### DIFF
--- a/src/reducers/Entities/LocationReducer.js
+++ b/src/reducers/Entities/LocationReducer.js
@@ -22,8 +22,12 @@ export const LocationReducer = (state = initialState(), action) => {
 
   switch (type) {
     case SYNC_TRANSACTION_COMPLETE: {
+      const { byId } = state;
+
       const newById = getById();
-      return { ...state, byId: newById };
+      const mergedById = { ...byId, ...newById };
+
+      return { ...state, byId: mergedById };
     }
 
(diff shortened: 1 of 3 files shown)


mark-prins (approved):
LGTM

PR #3579, opened 2021-02-21
#3555 maxValue prop in BreachConfigRow

### LINKED ISSUE #3555: Breach configs mistakenly have a duration limit of 10 minutes
## Describe the bug

Breach configurations are using the log interval limit of 10 minutes and shouldn't be.

### To reproduce

1. Go to add a new sensor
2. Click to increase the duration of a temperature breach config to be greater than 10

### Expected behaviour

Should have a much greater limit

### Proposed Solution

Pass the correct limit as a prop

### Version and device info

- App version: 7.0.0-RC4
- Tablet model: N/A
- OS version: N/A

### Additional context

Each input should also state what the limit is, probably.

joshxg:
Fixes #3555 

## Change summary

- Added prop to max value

## Testing

- [ ] Range of values for the duration for temperature breach configs is 1-999

N/A

### DIFF
--- a/src/pages/NewSensor/BreachConfigRow.js
+++ b/src/pages/NewSensor/BreachConfigRow.js
@@ -38,7 +38,11 @@ export const BreachConfigRow = React.memo(
         Icon={<Icon color={color} size={20} />}
         label={TYPE_TO_LABEL[type]}
       >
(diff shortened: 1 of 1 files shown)


mark-prins (approved):
👍  😃

PR #3581, opened 2021-02-21
#3537 regexp crash

### LINKED ISSUE #3537: Regexp used in MultiSelectList can cause a crash
## Describe the bug

This line here: https://github.com/openmsupply/mobile/blob/471c1e79f1d2aec239401e9e45c6ee8eea2295c3/src/widgets/modalChildren/MultiSelectList.js#L55

Can cause a crash if invalid regexp is entered

### To reproduce

1. Go to supplier requisition page and open the modal to add from master list
2. type something like a\a

### Expected behaviour

No crash

### Proposed Solution

Remove the regexp usage

### Version and device info

- App version: From about 5.0+
- Tablet model:
- OS version:

### Additional context

N/A

joshxg:
Fixes #3537 

## Change summary

- Uses toLowerCase and includes instead of regexp

## Testing

- [ ] Multiselect list works!

N/A

### DIFF
--- a/src/widgets/modalChildren/MultiSelectList.js
+++ b/src/widgets/modalChildren/MultiSelectList.js
@@ -52,11 +52,12 @@ export const MultiSelectList = ({
   };
 
   const filterArrayData = () => {
-    const regexFilter = RegExp(queryText, 'i');
+    const containsFilterString = string => string?.toLowerCase().includes(queryText);
+
     return options.filter(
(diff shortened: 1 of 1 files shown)


mark-prins (approved):
was unable to recreate the issue without cheating ( the masterlists are realm objects, so not arrays and not filtered by regexp.. ) 

This works for me though!

PR #3582, opened 2021-02-21
#3533 Sort batches by expiry date

### LINKED ISSUE #3533: Current stock shows batches out of order
## Describe the bug

Rowena: `wiki says batches will be ordered by expiry.  But mine doesn’t seem to be like this?`

### To reproduce

1. Go to `Current Stock`
2. Click on a row

### Expected behaviour

Order batches by expiry

### Proposed Solution

TBD

### Version and device info

- App version: Probably from ages ago
- Tablet model:
- OS version:

### Additional context

N/A

joshxg:
Fixes #3533 

## Change summary

- Sorts by `expiryDate` cause Rowena was complaining

## Testing

- [ ] Go to an SI and add 4 batches, with 3 different expiries (One with none). Check the batches in Current Stock Page. Should be ordered, earliest expiry first. N/A first

N/A

### DIFF
--- a/src/widgets/bottomModals/ItemDetails.js
+++ b/src/widgets/bottomModals/ItemDetails.js
@@ -33,7 +33,7 @@ export const ItemDetailsComponent = ({ item }) => {
   const getRow = (title, info) => ({ info, title });
 
   const getBatchColumn = field =>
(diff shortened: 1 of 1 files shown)


mark-prins (approved):
👍

PR #3583, opened 2021-02-21
#3531 Janky row detail opening

### LINKED ISSUE #3531: The opening the RowDetail for open/closed vial wastage is a bit janky
## Describe the bug

jank

### To reproduce

1. Go to SupplierRequisition page
2. Click on a vaccine item row

### Expected behaviour

The row detail should only try and open after the keyboard has closed

### Proposed Solution

Hook into `keyboardDidClose` event (or something) and only open after that has triggered

### Version and device info

- App version: 7.0.0
- Tablet model:
- OS version:

### Additional context

another issue being a row should have some sort of indication it's clickable 🤷

joshxg:
Fixes #3531 

## Change summary

- Only try and open the `RowDetail` component when the keyboard is closed.

## Testing

- [ ] Open the row detail from the Current Stock by clicking a row. Close it by opening the keyboard through the search bar - it's not crazy janky
- [ ] Open the row detail from clicking on a row when the keyboard is already open - it's not crazy janky

N/A

### DIFF
--- a/src/hooks/useKeyboardIsOpen.js
+++ b/src/hooks/useKeyboardIsOpen.js
@@ -1,0 +1,20 @@
+import { useState, useEffect } from 'react';
+import { Keyboard } from 'react-native';
+
+const KEYBOARD_DID_SHOW = 'keyboardDidShow';
+const KEYBOARD_DID_HIDE = 'keyboardDidHide';
+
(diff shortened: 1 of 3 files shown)


mark-prins (approved):
Very nice 😄 
Like the use of the hooks too!

Issue #3207, opened 2020-11-19
Supplier Requisition delete modal string is wrong


joshxg:
## Describe the bug

On trying to delete a supplier requisition: 

https://github.com/openmsupply/mobile/blob/c83576c1ecbbf6c92ff04713493b7adb8e026e5a/src/pages/SupplierRequisitionsPage.js#L173

### To reproduce

1. Go to `SupplierRequisitions` Page
2. Click on a requisition and try to delete it

### Expected behaviour

Should use `remove_these_requisitions` string

### Proposed Solution

as above

PR #3586, opened 2021-02-21
#3585 Correct the conditional on deleting remote orders

### LINKED ISSUE #3585: Cannot delete records created on a mobile, but can delete ones sent from the server
## Describe the bug

`Transaction`s/`Requisition`s which are sent from the server shouldn't be able to be deleted. At the moment they are. Vice versa for remote orders.

### To reproduce

1. Go to Supplier requisitions and create a new one
2. Go back and try delete it

### Expected behaviour

`Transaction`s/`Requisition`s which are sent from the server shouldn't be able to be deleted. At the moment they are. Vice versa for remote orders.

### Proposed Solution

Think there is some incorrect condition checking `isRemoteOrder`

### Version and device info

- App version: 6.0.0 + I think
- Tablet model:
- OS version:

### Additional context

N/A

joshxg:
Fixes #3585  #3207

## Change summary

- Corrects the conditional on deleting remote orders

## Testing

- [ ] Can delete requisitions/transactions which were created on a mobile store
- [ ] Cannot delete requisitions/transactions which were received via sync
- [ ] String is correct in the modal when deleting a supplier requisition

N/A

### DIFF
--- a/src/database/DataTypes/Transaction.js
+++ b/src/database/DataTypes/Transaction.js
@@ -593,6 +593,10 @@ export class Transaction extends Realm.Object {
   get hasVaccine() {
     return this.items.some(({ isVaccine }) => isVaccine);
   }
+
+  get isRemoteOrder() {
+    return !this.linkedRequisition;
+  }
 }
 
 Transaction.schema = {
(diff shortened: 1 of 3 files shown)


mark-prins (approved):
that's much better! 😃

PR #3588, opened 2021-02-22
#3529 should force sensor name and code

### LINKED ISSUE #3529: Should force a sensor name or code
## Describe the bug

When a user does not enter a sensor name or code, a location is created without a name or code. This means you can have a selection of locations like:

image

### To reproduce

1. Go to add a new sensor and don't give it a name or code
2. Go to stocktakes and try to assign an item to a location

### Expected behaviour

- Should probably force a code and description. On desktop you aren't able to create a location without a code or description (requires both) so we should probably mimic it.

### Proposed Solution

Have validation on `SensorEditPage` and `NewSensorStepThree` to ensure a name and code have been entered. Possibly we can just duplicate the name over to the code if a code isn't entered.

### Version and device info

- App version: 7.0.0-rc3
- Tablet model:
- OS version:

### Additional context

N/A

mark-prins:
Fixes #3529 

## Change summary

Adds validation to the new sensor and sensor edit pages.
Defaults the code of the new location to be the name of the sensor, if no code is specified.. open to better suggestions on that one.

Side note: was amused to find that you can use emojis as the name..

Screenshot_2021-02-22-16-37-51-910

## Testing

- [ ] Edit a sensor, verify that you are unable to save until a name is specified
- [ ] Edit a sensor, clear the `Code` field. Verify that the name is copied to the code field after saving
- [ ] Create a sensor and verify that a name is required
- [ ] Create a sensor with a blank code and verify that the name is used for the code

### DIFF
--- a/src/actions/Entities/SensorActions.js
+++ b/src/actions/Entities/SensorActions.js
@@ -83,6 +83,7 @@ const save = () => (dispatch, getState) => {
   const updatedConfigs = [];
 
   location.description = sensor.name;
+  location.code = location.code || sensor.name;
 
   UIDatabase.write(() => {
     updatedLocation = UIDatabase.update('Location', location);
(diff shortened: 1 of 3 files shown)


joshxg (approved):
Nice reusing of stuff! 😄

PR #3590, opened 2021-02-22
#3527 no vvm status records handling

### LINKED ISSUE #3527: No vvm status records not handled correctly
## Describe the bug

image

When having no vvm status records, the list is empty

### To reproduce

1. Go to `SI` and add a `vaccine` item
2. Click on vvm status column (with no vvm status records)

### Expected behaviour

- Not sure. Maybe the list should say "No VVM status' contact admin" but also 

### Proposed Solution

### Version and device info

- App version:7.0
- Tablet model:
- OS version:

### Additional context

N/A

mark-prins:
Fixes #3527 

## Change summary

Added placeholder when no vvm statuses are available. In supplier invoices and stocktake batches.
Slight detour: also handled the 'isActive == false' state, as that helped me test.
Diversion: noticed that the requisition lines were not handling the placeholder either
Full offroad: added the `isActive` check for reasons too

## Testing

Note: when testing, you have to have a supplier invoice with a vaccine in it; and the vaccines enabled; AND the payment module disabled; otherwise the payment columns show in supplier invoices, and not the vaccine columns 🙄 

- [ ] Check for no VVM status in supplier invoice and stocktake batch
- [ ] Check for no active VVM status in blah blah blah
- [ ] Check for no reasons in a program-based supplier requisition
- [ ] and make sure that inactive ones are ok for these too

### DIFF
--- a/src/localization/generalStrings.json
+++ b/src/localization/generalStrings.json
@@ -71,7 +71,8 @@
     "from": "from",
     "no_locations_for_batch": "This item is restricted to locations of type {0} and none are available. Contact your administrator.",
     "no_locations": "You don't have any locations to assign! Contact your administrator.",
-    "no_options": "No options available. Please contact your administrator!"
+    "no_options": "No options available. Please contact your administrator!",
+    "no_vvm_status": "No vaccine vial monitor statuses available. Please contact your administrator!"
   },
   "fr": {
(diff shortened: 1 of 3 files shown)


joshxg (approved):
Nice works well. Maybe we should make it prioritise the vaccine columns over the payment ones for a vaccine.. we really need a better column solution 🤷

PR #3591, opened 2021-02-22
disable blink when downloading

### LINKED ISSUE #3587: Blinking sensor while download is in progress hangs blinking state
## Describe the bug
https://user-images.githubusercontent.com/65875762/108653690-ffc3d600-752b-11eb-8a1b-637ccfd6a3e7.mp4

### To reproduce

1. Have a sensor setup in the Vaccine Module with downloads happening in background
2. Wait for a download to start (text by WiFi icon will blink when in progress)
3. While the download is happening, or very shortly after (1 or 2 seconds) try to blink the sensor
4. Blink fails because the BT connection can't be established
5. If the download wasn't finished, it gets interrupted with `Operation Cancelled` or `Device is not connected` error
6. Blink icon gets infinitely replaced with progress spinner and the app will forever think that sensor is blinking until it dies

### Expected behaviour

- Probably for the blinking state to reset

### Proposed Solution

- Probably need to catch blink failure and reset blinking state?

### Version and device info

- App version: v7.0.0-rc5
- Tablet model: Lenovo something
- OS version: 8.1.0

mark-prins:
Fixes #3587 

## Change summary

Disabled the blink button when a sensor is currently downloading. 
Resolves the problem, and why would anyone want to blink while it was actively downloading?

The blink button disabling logic wasn't clear - so I changed it. There's no reason to include `isBlinking` as this is already handled in the implementation - and I couldn't see why `!isBlinking` == disabled. Surely it should be disabled if it is already blinking?

## Testing

- [ ] Wait fo a sensor to download and try to blink

### DIFF
--- a/src/widgets/SensorHeader/BlinkSensorButton.js
+++ b/src/widgets/SensorHeader/BlinkSensorButton.js
@@ -7,6 +7,7 @@ import { IconButton } from '../IconButton';
 import { LightbulbIcon } from '../icons';
 import { SensorBlinkActions } from '../../actions/Bluetooth/SensorBlinkActions';
 import { selectSendingBlinkTo } from '../../selectors/Bluetooth/sensorBlink';
+import { selectIsDownloading } from '../../selectors/Bluetooth/sensorDownload';
 
 export const BlinkSensorButtonComponent = ({ isBlinking, blink, isBlinkDisabled }) =>
   isBlinking ? (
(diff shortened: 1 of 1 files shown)


joshxg (approved):
🎉 works nicely, thanks!

katherine-sussol:
Nice - avoids the problem of errant users/🐱 s  trying to break the app 😆

PR #3593, opened 2021-02-22
#3546 ble errors when updating sensor

### LINKED ISSUE #3546: BleErrors when updating sensor (when download in progress?)
## Describe the bug
When trying to edit a sensor (even non-Bluetooth settings, e.g. Breach Settings only) it will throw BleErrors if a download is currently happening in the background.
e.g.
- Operation was cancelled
- Device ? is already connected

Attached a screen recording - 
https://user-images.githubusercontent.com/65875762/108297509-3e8e1f00-7200-11eb-954a-870a02a2ebc1.mp4

### To reproduce

1. Go to Vaccines -> Click the cog on a sensor
2. Change some breach threshold temperature(s)
3. Click 'Save'
4. See error(s)

### Expected behaviour

1. Expected sensor to update with no errors (or at least for errors to be fairly rare)

### Proposed Solution

- Maybe one improvement could be to not send BT commands if there haven't been any updates that require them? e.g. if breach config change only
- Above doesn't solve the underlying issue though 🤷  - not sure how to fix that one yet. 

### Version and device info

- App version: 7.0.0 RC3
- Tablet model: Lenovo something (not sure of model name)
- OS version: Android 8.1

### Additional context
- From observations, seems like these errors are coming up/more frequent when downloads are happening/starting in the background. If there are lots of sensors a download sequence could take a while.
- Might need more testing to pin down the specifics, it's being really bizarre 🤔   - will leave issue here to add info as more details come in I guess?

katherine-sussol:
Fixes #3546 

## Change summary

- Update sensor save to not send BT command if the logInterval wasn't updated
- If a download is currently in progress, block the sensor from updating and display message to user to try again later

## Testing

- [ ] Have sensor(s) setup to download in the background
- [ ] Go to edit sensor (or create new sensor)
- [ ] If `logInterval` hasn't been modified/newly created BT command should not be sent out (used `console.log` in Action to verify) when clicking 'save'
- [ ] If there is a `logInterval` change but there is a download happening in the background, a message gets displayed to the user `A download is in progress, please try again soon`

https://user-images.githubusercontent.com/65875762/108782270-28051080-75d0-11eb-8360-831172d8e390.mp4

- Affects create a new sensor flow too
- Probably if an update is started and then a download starts in the background, BleErrors will pop up still. But shouldn't happen too often? And as per https://github.com/openmsupply/mobile/issues/3546#issuecomment-783060847 agree download failing once in a while isn't critical.


mark-prins:
@katherine-sussol very nice! I often have problems saving to random sensors in the office.

I wondered about taking out the `startSensorDisableButton` call - and only calling that when setting up a new sensor. Is there any point in waiting for that call, every time we save sensor details? The calls are pretty slow - fine for normal usage maybe, but when testing and calling this a lot, it's annoying.

How about creating a new action in `SensorUpdateActions` which is used for new sensors.. that would always call `startSetLogInterval` and `startSensorDisableButton`. Something like `updateNewSensor` ??

src/actions/Bluetooth/SensorUpdateActions.js
```diff
@@ -92,12 +95,24 @@ const startSetLogInterval = ({ macAddress, logInterval = 300 }) => async (dispat
   return result;
 };
 
-const updateSensor = sensor => async dispatch => {
+const updateSensor = sensor => async (dispatch, getState) => {
+  const state = getState();
```
mark-prins, line 99: :this is a really minor thing.. I'd move the `getState()` to below the `logInterval` check - as it might not be needed. That would give a very small optimisation..? Also, the state is only used once, so no need to use a const..

```if (selectIsSyncingTemps(getState())) {```

**Review by mark-prins — APPROVED** [review]
I like it a lot! 
there's a couple of minor suggestions offered, nothing essential.

**Comment by katherine-sussol** [comment]
> I wondered about taking out the `startSensorDisableButton` call - and only calling that when setting up a new sensor. Is there any point in waiting for that call, every time we save sensor details? The calls are pretty slow - fine for normal usage maybe, but when testing and calling this a lot, it's annoying.

Sounds like a good idea, it's been pretty slow for me too and I think decreasing the number of commands going out reduces the chances of conflicts between different calls happening too 👍 

> How about creating a new action in `SensorUpdateActions` which is used for new sensors.. that would always call `startSetLogInterval` and `startSensorDisableButton`. Something like `updateNewSensor` ??

Oh yeah, that would work I think. Was thinking could also just remove `startSensorDisableButton` from `updateSensor` and dispatch it separately in `NewSensorStepThree`?

File: src/actions/Bluetooth/SensorUpdateActions.js (hunk 1 above)
Comment (katherine-sussol) [reply src/actions/Bluetooth/SensorUpdateActions.js]: Nice, much tidier thanks 96a45bd

**Comment by katherine-sussol** [comment]
Have added an off-road to remove the disable sensor button BT command when doing updates. I thought it was small and kind of related so snuck it into this PR, requesting a re-review just to get a second pair of 👀  to confirm is still all ok after changes 😆

File: src/actions/Bluetooth/SensorUpdateActions.js [hunk 2]
```diff
@@ -92,19 +95,36 @@ const startSetLogInterval = ({ macAddress, logInterval = 300 }) => async (dispat
...
     return;
   }
 
+  const oldLogInterval = UIDatabase.get('Sensor', macAddress, 'macAddress')?.logInterval;
+  if (oldLogInterval && sensor.logInterval === oldLogInterval) {
+    // No updates required if no change
+    return;
+  }
+
+  if (selectIsSyncingTemps(getState())) {
```
mark-prins, line 117: Sorry! Only just thought of this.. should it be `selectIsDownloading(getState(), macAddress)`?
Does it matter if another sensor is downloading?

katherine-sussol: I've had really mixed results when another sensor is downloading (I tried the above)... on one hand it's really annoying to be blocked when another sensor is downloading, but on the other I've been seeing sensor disconnect problems if another one is downloading... maybe that is something worth digging into further 🕳️ ... hmm

mark-prins: Good enough for me! thanks.

mark-prins:
Awesome! all looking good thanks 😄

joshxg (approved):
awesome!

PR #3603, opened 2021-02-24
#233 trim white space off site and user names

### LINKED ISSUE #233: OS adding trailing space to username causing problems
## Describe the bug

Basically if the user lets auto complete do its thing, it will often put a space at the end of their user name.  The user can't see the space, so they don't realise it has happened.  When they try to log in, mSupply says that their credentials are wrong, but the user can't see anything wrong unless they know what they are looking for.

### To reproduce

1. Initialisation page
2. Enter username, take an auto complete suggestion
3. Enter password
4. Initialise
5. Good chance it gives an error message regarding wrong credentials.

Also, the user login modal. Same thing.

### Expected behaviour

Don't inconspicuously lull user into putting spaces in they don't want and aren't valid.

### Proposed Solution

Adjust the keyboard settings programatically to not leave trailing spaces to begin with:
- There is a [`TextInput` prop autoCompleteType](https://reactnative.dev/docs/textinput#autocompletetype) (that I don't think existed years ago when this issue was first created. The technology is finally here!)
  - Password fields should be 'password'. Username fields 'username'
  - URLs 'none'?
  - Note, this only works for Android, but we don't really support iOS at all so that's fine. 
- Also turn maybe `autoCapitalize` to 'none' if we haven't already.
- Totally feature creep, but could set server URL `defaultValue` to be "https://". Might be annoying for testers/non SSL usage though.

ALT: Removing trailing spaces:
- Just on site name and user name fields.
- Passwords should be exempt cause they shouldn't have auto complete. Spaces should be valid characters.

ALT: Remove ALL spaces:
- Not good, restrictive policy that does not take into account existing sites/users that may have intentionally added them.
- Doubly not good for passwords, as you can't unhash passwords to correct them retrospectively.

### Version and device info

- App version: all
- Tablet model: most
- OS version: most

### Additional context

Super annoying cause of many support tickets. Would be neat if we had a REST api for resetting passwords (via a link sent to associated email sort of thing)

Chris-Petty:
Fixes #233?

## Change summary

- Added `autoCompleteType` to textInputs. Didn't actually fix the issue unfortunately. Presumably what should be done for android at least.
- Added trimming of strings when submitting server url, site name and user name.

## Testing

**Initialisation/FirstUsePage**
- [ ] When submitting (pressing next or done on virtual keyboard) URL any leading and trailing white space is removed.
- [ ] When submitting site name any leading and trailing white space is removed.
- [ ] When submitting password any leading and trailing white space is **kept**.
- [ ] When changing focus (press/click elsewhere in UI, e.g. another field) from URL any leading and trailing white space is removed.
- [ ] When changing focus from site name any leading and trailing white space is removed.
- [ ] When changing focus from password any leading and trailing white space is **kept**.

**Login Page**
- [ ] When submitting user name any leading and trailing white space is removed.
- [ ] When submitting password any leading and trailing white space is **kept**.
- [ ] When changing focus from user name any leading and trailing white space is removed.
- [ ] When changing focus from password any leading and trailing white space is **kept**.

This is contentious haha. There are conceivably existing users that have a spaces leading or trailing their user name.

Creating patients or prescribers, we could/should(?) trim their names too. Though I think we enter "first" and "last" names separately, should still trim them?

### DIFF
--- a/src/pages/FirstUsePage.js
+++ b/src/pages/FirstUsePage.js
@@ -83,7 +83,19 @@ export class FirstUsePageComponent extends React.Component {
     }
   }
 
-  onChangeText = text =>
+  onChangeServerUrl = text =>
+    this.setState({
+      serverURL: text,
+      status: 'uninitialised',
+    });
+
+  onChangeSiteName = text =>
+    this.setState({
+      syncSiteName: text,
+      status: 'uninitialised',
+    });
+
+  onChangePassword = text =>
     this.setState({
       syncSitePassword: text,
       status: 'uninitialised',
@@ -123,8 +135,11 @@ export class FirstUsePageComponent extends React.Component {
               selectTextOnFocus
               autoCapitalize="none"
               autoCorrect={false}
-              onChangeText={text => this.setState({ serverURL: text, status: 'uninitialised' })}
+              onChangeText={this.onChangeServerUrl}
               onSubmitEditing={() => {
+                // Trim URLS. Any leading/trailing spaces lead to invalid URLs.
+                this.setState({ serverURL: serverURL.trim() });
+
                 if (this.siteNameInputRef) this.siteNameInputRef.focus();
               }}
             />
@@ -135,6 +150,7 @@ export class FirstUsePageComponent extends React.Component {
                 this.siteNameInputRef = reference;
               }}
               style={globalStyles.authFormTextInputStyle}
+              autoCompleteType="username"
               placeholderTextColor={SUSSOL_ORANGE}
               underlineColorAndroid={SUSSOL_ORANGE}
               placeholder="Sync Site Name"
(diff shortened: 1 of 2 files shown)


mark-prins:
@Chris-Petty should this be targeted for `master`?

src/pages/FirstUsePage.js
```diff
@@ -83,7 +83,19 @@ export class FirstUsePageComponent extends React.Component {
     }
   }
 
-  onChangeText = text =>
+  onChangeServerUrl = text =>
+    this.setState({
+      serverURL: text,
+      status: 'uninitialised',
```
mark-prins, line 89: :nit the string has been used three times, would be nice to have as a CONST

mark-prins (approved):
Looks good to me! 
Haven't tested as I'm running `develop` and it is a pain switching around.

src/pages/FirstUsePage.js
Chris-Petty: I'm gonna quickly do this!

Chris-Petty:
@mark-prins being able to release as a patch for 6.0.x would be mint for PNG, who are frequently having woes regarding this. The annoyance being as old as mobile itself 😅 .

I bumpted with commits for using CONST

Chris-Petty:
Huh do commits automatically dismiss reviews? I don't recall dismissing, whoops

mark-prins (approved):
Very nice! thanks

mark-prins:
> Huh do commits automatically dismiss reviews? I don't recall dismissing, whoops

yeah, I've had this too. not consistently though, so maybe it's a repo setting.

Ah. yes - it is. and it is set for `master` but not for `develop`

Screen Shot 2021-03-02 at 8 07 18 AM

wlthomson:
@ArpitaSussol Sorry to be a bit annoying, but could you test this for `v6` and `v7`?

ArpitaSussol:
> @ArpitaSussol Sorry to be a bit annoying, but could you test this for `v6` and `v7`?

i am into it :)

PR #3629, opened 2021-02-26
#3626 Update temperatureBreachConfig thresholds

### LINKED ISSUE #3626: Sensor Thresholds: HOT Cumulative showing lower value than COLD Cumulative
## Describe the bug

In Sensor Thresholds setting, HOT Cumulative showing lower value than COLD Cumulative

### To reproduce

1. Go to Vaccine-> Add Sensor -> Go to Step 2 of Sensor Threshold
2. Click on "-" icon for Hot cumulative in such a way that its value be lower than Cold Cumulative.
3. See error. It allows in case of Cumulative where as in case of Consecutive Hot/Cold it doesn't allow
photo_2021-02-25 12 31 49

### Expected behaviour

Temperatures are bound by the hot/cold in each group. HOT Cumulative cannot be lower than COLD Cumulative and vice versa. Also "-"  sign against values shown as dot "." when number's length increases. It would be good to enlarge space for Value field.
photo_2021-02-25 12 31 12

### Proposed Solution

### Version and device info

- App version: v7.0.0-rc7
- Tablet model: Lenovo
- OS version: 8.1

katherine-sussol:
Fixes #3626 

## Change summary

- Super tiny tweak to min/max thresholds so hot > cold when creating and editing breach configs for a sensor

## Testing

- [ ] Try to add new sensor
  - [ ] Try to set hot cumulative <= cold cumulative (shouldn't let you)
  - [ ] Try to set cold cumulative >= hot cumulative (shouldn't let you)
- [ ] Try to edit existing sensor
  - [ ] Try to set hot cumulative <= cold cumulative (shouldn't let you)
  - [ ] Try to set cold cumulative >= hot cumulative (shouldn't let you)

Creating new sensor was allowing hot/cold to be equal, adjusted to align with edit flow (min diff of 0.1)

### DIFF
--- a/src/selectors/Entities/temperatureBreachConfig.js
+++ b/src/selectors/Entities/temperatureBreachConfig.js
@@ -33,11 +33,11 @@ export const selectEditingConfigThresholds = state => {
     COLD_CUMULATIVE: coldCumulativeConfig = {},
   } = editingConfigs;
 
-  const coldConsecutiveThreshold = (hotConsecutiveConfig.minimumTemperature ?? 0) - 0.1;
-  const hotConsecutiveThreshold = (coldConsecutiveConfig.maximumTemperature ?? 0) + 0.1;
+  const coldConsecutiveThreshold = hotConsecutiveConfig.minimumTemperature ?? 0 - 0.1;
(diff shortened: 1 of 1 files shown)


joshxg (approved):
nice :)

PR #3636, opened 2021-03-01
#3617 vaccine dispensing redux shape

### LINKED ISSUE #3617: Add new vaccine dispensing flow redux shape

We will likely (?) use redux for the new vaccine dispensing flow to be consistent with the rest of the app.

We will need to design and add the shape for this redux state.

TBD

## Implementation

TBD

Not using redux???

## Additional context

Some things to consider:
- We want to have a step process for creating the prescription. The current dispensing flow creates a realm `Transaction` object, and if navigating away without finishing, deletes it. I think this isn't great and should be avoided, if possible by using an in-memory "transaction" object which is saved at the end of the flow.
	- Pros: We avoid adding more class instances and non-serialisable state into redux.
	- Pros: We don't have to hook into navigation events to delete things.
	- Pros: It's easier to validate things when you don't save them persistently straight away.
	- Pros: Less work when moving to SQLite
	- Cons: Can't utilise any of the realm class-method based logic
- This also goes for a `name`/`name_note` records. Ideally we don't update the name *straight* away, nor create new `name_note` records which we have to then track and delete.
- The general pattern for this is entity based reducers where you have something like:
```
{
byId: {
  [id]: entity,
  [id2]: entity2,
  ...,
 }
ids: [id, id2] // For sorting, if ever needed
editingId: id // this is the one we're editing
}
```
- However, I can imagine there being a LOT of `name_note` records, and there can even be a lot of `name` records (all of Tonga?!). So maybe we need to do a slightly different shape?

```
{
editingEntity: entity
}
```

or

```
// This is for when you're editing many records at once, i.e. name_note records
{
editingEntitiesById: {
    [id]: entity
    [id2]: entity2
	}
editingEntitiesIds: [id, id2]
}
```
- We may also want to think about a shape for `name_note` where we track the notes which are actually touched/edited as we only want to create new `name_note` records which have been changed, rather than for everything.

mark-prins:
Fixes #3617 

## Change summary

Add actions, reducers, selectors for `name`, `nameNote` and `vaccinePrescription`
Mostly just a guess at how we'll actually use these, and I think there'll need to be some tidy up of it. Is this the sort of thing you were thinking though?

## Testing

- [ ] Do something


src/database/DataTypes/Transaction.js
```diff
@@ -597,6 +597,36 @@ export class Transaction extends Realm.Object {
...
+    return {
+      id: this.id,
+      serialNumber: this.serialNumber,
+      otherParty: this.otherParty,
+      comment: this.comment,
+      entryDate: this.entryDate?.getTime(),
+      type: this.type,
+      status: this.status,
+      confirmDate: this.confirmDate?.getTime(),
+      // enteredBy: { type: 'User', optional: true },
```
mark-prins, line 611: was just checking that it is ok to remove all of these fields and have them undefined?
I think so.. wasn't sure.

src/database/DataTypes/Transaction.js
```diff
@@ -597,6 +597,36 @@ export class Transaction extends Realm.Object {
   get isRemoteOrder() {
     return !this.linkedRequisition;
   }
+
+  toJSON() {
+    return {
+      id: this.id,
+      serialNumber: this.serialNumber,
+      otherParty: this.otherParty,
```
joshxg, line 605: I've been in general keeping id's of other objects rather than objects. In this case, `otherParty` is a live realm object which shouldn't matter as we don't edit it, but would require editing in a transaction which would be weird when you think it's a plain JS object.

My argument for keeping IDs is
1) immutably updating nested objects is a pain
2) Keeping potentially multiple different versions of an object in memory can cause some issues with objects going stale so easier to just use FK id's

joshxg (approved):
Yeah pretty much what I was going for- just the boiler plate to get us rolling so theres less to get in the way with the 'real' issues, heh.

Thanks!

src/navigation/constants.js
```diff
@@ -56,6 +56,10 @@ export const ROUTES = {
   SENSOR_EDIT: 'sensorEdit',
   SENSOR_NEW: 'sensorNew',
   FRIDGE_DETAIL: 'fridgeDetail',
+
+  NAME_EDIT: 'nameEdit',
+  NAME_NOTE_EDIT: 'nameNoteEdit',
+  VACCINE_PRESCRIPTION: 'vaccinePrescription',
```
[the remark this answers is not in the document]
mark-prins: oh, yes. thanks - I did that first and decided we didn't need it. without going back and reviewing.

src/database/DataTypes/Transaction.js
mark-prins: thanks. and oops - I missed that one. Have removed, I don't think we'll need `otherParty` right now, and can add it back if needed.

PR #3642, opened 2021-03-01
#3640 Select widget

### LINKED ISSUE #3640: JSON Schema select widget

Need a select widget for selecting options in a JSON schema form

A `SelectWidget` component which is similar to `FormDropdown`

## Implementation

Take `FormDropdown` and adapt it to fit the new `JSONForm`

N/A

## Additional context

N/A

joshxg:
Fixes #3640 

## Change summary

- Adds a select widget
- Just really tested with some basic enums and it seems to work well.
- Went with the built in react-native picker even though it's deprecated because that's just what we have been using. Didn't use the `dropdown` component because it has a funny `header` thing implemented which was being annoying with implementing 'default/placeholder' values, with having the 'header' being selectable, which I didn't want to have to regression test.
- The `react-native-picker` package available to replace the built-in has exactly the same API, just didn't want to add native packages which can make building the app be funny
- I actually like the `mode="dialog"` a little better than "dropdown"?

## Testing

Internal?

image

### DIFF
--- a/src/widgets/JSONForm/JSONForm.js
+++ b/src/widgets/JSONForm/JSONForm.js
@@ -12,7 +12,7 @@ import { JSONFormWidget } from './widgets/index';
 import { JSONFormErrorList } from './JSONFormErrorList';
 import { PageButton } from '../PageButton';
 import { JSONFormContext } from './JSONFormContext';
-import { booleanSchema } from './testJSON';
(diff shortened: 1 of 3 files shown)


mark-prins (approved):
LGTM!
We could look at changing the picker, maybe not right now.

PR #3643, opened 2021-03-02
#3638 Date widget

### LINKED ISSUE #3638: JSON Schema date input

Need a date widget for picking dates in a JSON schema form

A `DateWidget` component which is similar to `FormDateInput`

## Implementation

Take `FormDateInput` and adapt it to fit the new `JSONForm`

N/A

## Additional context

N/A

joshxg:
Fixes #3638

## Change summary

- Adds .. a date widget!
- The format is for the JSON schema. I think we would probably use a different date format but I have gone for this for the simplicity for now. Can discuss and implement a way to use different formats later?
- This component being a weird text input thing is just copied from the current patient form. It's weird though.

## Testing

Internal

image


joshxg:
Not sure if this commit: 
3add008

is overkill.

Can revert easily - kinda nice to not have to rewrite the register and ref and stuff code in each component that wants to have focus controlled

src/widgets/JSONForm/widgets/DatePicker.js
```diff
@@ -0,0 +1,48 @@
+/* eslint-disable no-console */
```
mark-prins, line 1: :nit can remove

mark-prins (approved):
that all works nicely :)
will need to add some validation or parsing somewhere, to prevent people typing random stuff into the textinput

joshxg:
@mark-prins the validation is at the end (or we can live validate to have errors show up, but has performance concerns) and if its not valid, the submit callback won't be called

PR #3646, opened 2021-03-02
#3645 Field Template update

### LINKED ISSUE #3645: JSON Schema form Field template update

The JSON schema form Field template is a template for all of the different widgets 

Update the template so it looks a little nicer than what it current is

## Implementation

I think we should make the Title for `object` fields display better - maybe more of a title for a form, rather than when the template is used for a regular widget.

N/A

## Additional context

N/A

joshxg:
Fixes #3645 

## Change summary

- Added an update to the Field Template to handle objects and the root object differently to regular widgets.
- Removed the scrollview from focusController. If I put the form in a scrollview it does this automagically!
- Updated styles a bit cause it was annoying me
- Added tonga data to see whats up

Really I was mainly having a little play and checking things out, seeing what else needed to be done, putting the Tonga schema in to see how it was all coming together and I had some code that I thought was best to just commit.... Not sure if we should use this as the real styles

## Testing

Internal

formtexting

### DIFF
--- a/src/widgets/FormInputs/FormLabel.js
+++ b/src/widgets/FormInputs/FormLabel.js
@@ -14,14 +14,17 @@ import { SUSSOL_ORANGE, APP_FONT_FAMILY, APP_GENERAL_FONT_SIZE } from '../../glo
 import { formInputStrings } from '../../localization';
 
 export const FormLabel = ({ isRequired, value, textStyle, isRequiredStyle }) => (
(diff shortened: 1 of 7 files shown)


mark-prins (approved):
I like what you've done with the object fields - the paper and title sizing work really well!
Looks great, and it's good to have the test data in place to see how it will work

PR #3652, opened 2021-03-02
#3607 vaccine dispensing step one

### LINKED ISSUE #3607: VaccineDispensing step one

- We need the step one tab for the `VaccineDispensingPage` created here : https://github.com/openmsupply/mobile/issues/3606

- the `VaccineDispensingPageStepOne` component created

image

## Implementation

Need to change some details of the `SearchForm` implementation:
- Decouple from being used in the `DispensingPage` context.
- Change to lookup locally, and offer the user to look online afterward (??)

It may be better to create a new component rather than trying to utilise this one.

TBD

## Additional context

TBD

mark-prins:
Fixes #3607 

## Change summary

Added a page and tab for the first step, patient selection.
Only searches local patients, as that was the directive for the first release ( import _all_ patients first ) and doesn't handle a spinner / indication if the load takes some time. I'm testing with a small dataset atm.

Thought that we could extend with online search afterwards, once we have the basics working.

One issue is with the DOB search - once you've selected a DOB you cannot clear that field; would like to add a clear / reset option or allow blank for the date field. will raise an issue for that.

I've moved the 'create patient' button to a weird location - thinking that the 'NEXT' function is better handled by clicking on a patient, and moving the 'create patient' to the bottom saves screen space. Suggestions? 

## Testing

- [ ] Head to the Vaccine page, and click 'Dispensary'


src/actions/Entities/VaccinePrescriptionActions.js
```diff
@@ -49,11 +51,19 @@ const updateEditing = (value, field) => (dispatch, getState) => {
   dispatch(update(newVaccinePrescriptionId, field, value));
 };
 
+const cancel = () => dispatch => {
+  batch(() => {
+    dispatch(NavigationActions.back());
+    dispatch(reset());
+  });
+};
```
joshxg, line 59: At the moment this action isn't getting dispatched when the hardware back button is pressed or the extra cool back button on the top left of the nav bar.

Some options I can think of
1) Add dispatching of the reset action into `NavigationActions.back` - I did this for sensor work, I think it's the worst option.
2) Add a case in the reducer for the action `Navigation/BACK` - this is probably my preference
3) Leave it, it gets sorted when navigating to the page anyway, I think - this is probably the better option

src/reducers/Entities/NameReducer.js
```diff
@@ -35,6 +38,20 @@ export const NameReducer = (state = initialState(), action) => {
 
       return { ...state, editing: newName };
     }
+
```
joshxg, line 41: nit: The `.SAVE` action had me a little confused - it's more of a "select" or "start using this one" kind of action, where a "SAVE" for me implies persistently storing

joshxg (approved):
It's beautiful and works well, don't have much to add just some thoughts on the new patient/online situation. 

I played a little with a footer/empty component showing the buttons. Would show at the bottom fo the list of names where you've presumably scrolled to the bottom and need one of these options.. look online could throw a spinner up and then populate the list of the found options from the server, otherwise create can bring up the modal to create?

```
ListFooterComponent={
            <FlexRow
              flex={1}
              justifyContent="center"
              alignItems="center"
              style={{
                marginTop: 50,
              }}
            >
              <PageButton text="New Patient" />
              <PageButton text="Look Online" />
            </FlexRow>
        }
```

image


```
ListEmptyComponent={ ...
```

image

src/actions/Entities/VaccinePrescriptionActions.js
mark-prins: good point. I've added in 1fa783de
While I don't like the fact that it resets on every back navigation in the app.. it's better than not 🤷‍♂️

src/reducers/Entities/NameReducer.js
```diff
@@ -2,6 +2,9 @@ import { NAME_ACTIONS } from '../../actions/Entities';
 
 const initialState = () => ({
   editing: undefined,
+  isAscending: true,
+  searchParameters: {},
```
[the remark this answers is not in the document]
mark-prins: I had the same thought. so I've tested and adding in is fine. so I have. 8bf3f75d

src/reducers/Entities/NameReducer.js
mark-prins: me too - we have `save` elsewhere for this kind of action so I went for consistency.
I'll change though, thanks.

src/selectors/Entities/name.js
```diff
@@ -12,3 +15,46 @@ export const selectEditingName = state => {
...
+  const query = 'lastName BEGINSWITH[c] $0 AND firstName BEGINSWITH[c] $1';
+  const patients = UIDatabase.objects('Patient').filtered(query, lastName, firstName);
+
+  if (!dateOfBirth) {
+    return patients;
+  }
+
+  const dayAfterDOB = moment(dateOfBirth).add(1, 'days').toDate();
+  return patients.filtered('dateOfBirth >= $0 AND dateOfBirth < $1', dateOfBirth, dayAfterDOB);
+});
```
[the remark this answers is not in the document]
mark-prins: ok, thanks. the data I'm working with is sync'd from desktop so should be ok. but I like yours better.

PR #3654, opened 2021-03-03
#3627 fix sensor config reverting location updates

### LINKED ISSUE #3627: Vaccine->Sensor: Sensor code is showing previous value even if its left blank/empty
## Describe the bug

Sensor code is showing previous value even if its left blank/empty. 

### To reproduce

1. Go to Vaccine -> Sensor -> Add Sensor
2. Enter details Sensor name and Sensor code . Click save
3. Remove the sensor. Try adding same sensor with different sensor name and Sensor code (can be left blank as well)
4. See error. Sensor Code shows exisiting old value.

https://user-images.githubusercontent.com/58979436/109124783-6e46a500-7773-11eb-930b-7cc8604d64da.mp4

https://user-images.githubusercontent.com/58979436/109124899-8caca080-7773-11eb-8aa0-1d4364bb0542.mp4

### Expected behaviour

Sensor code should be updated and reflect the updated value

### Proposed Solution

### Version and device info

- App version: v7.0.0 RC7
- Tablet model: Lenovo
- OS version: 8.1

wlthomson:
Fixes #3627.

## Change summary

Took me a while to figure this one out! TLDR is that location edits are being overwritten by old location state persisted in the sensor configs. When the config locations are persisted to the database, the previous location updates are "reverted" back to the old state. This can be seen by examining `updatedLocation` after the configs are updated (as it is a live realm object, it reflects the most recent state of the database).

Note: what made this super confusing (for me) was the way concerns overlap between multiple live objects (e.g. as opposed to only updating config fields on the `TemperatureBreachConfiguration` object, linked location objects are also updated). I think some minor refactoring to only update certain fields rather than whole objects could make this code a bit nicer?

Note note: also just realised part of the reason for this is probably the fact that the code would be quite verbose now that realm objects can no longer be destructured!

## Testing

- [ ] Edit an existing sensor code. After saving and navigating back to the sensor edit page, the code has been updated.
- [ ] Delete an existing sensor. After reconnecting the sensor with a new code and navigating back to the sensor edit page, the code has been updated.

Good patterns for updating groups of live realm objects in thunks.

### DIFF
--- a/src/actions/Entities/SensorActions.js
+++ b/src/actions/Entities/SensorActions.js
@@ -177,7 +177,7 @@ const createNew = () => (dispatch, getState) => {
     if (existingLocation) {
       newLocation = UIDatabase.update('Location', {
         id: existingLocation.id,
-        description: sensor.name,
+        description: location.name,
         code: location.code,
       });
     } else {
--- a/src/actions/Entities/SensorActions.js
+++ b/src/actions/Entities/SensorActions.js
@@ -93,9 +93,11 @@ const save = () => (dispatch, getState) => {
       logDelay: new Date(sensor?.logDelay ?? 0),
       location: updatedLocation,
     });
-    configs.forEach(config =>
-      updatedConfigs.push(UIDatabase.update('TemperatureBreachConfiguration', config))
-    );
+
+    configs.forEach(config => {
+      config.location = updatedLocation;
+      updatedConfigs.push(UIDatabase.update('TemperatureBreachConfiguration', config));
+    });
 
     if (replacedSensor) {
       replacedSensor.location = null;
--- a/src/actions/Entities/SensorActions.js
+++ b/src/actions/Entities/SensorActions.js
@@ -178,6 +178,7 @@ const createNew = () => (dispatch, getState) => {
       newLocation = UIDatabase.update('Location', {
         id: existingLocation.id,
         description: sensor.name,
+        code: location.code,
       });
     } else {
       newLocation = createRecord(UIDatabase, 'Location', { ...location, description: sensor.name });
--- a/src/actions/Entities/SensorActions.js
+++ b/src/actions/Entities/SensorActions.js
@@ -95,8 +95,26 @@ const save = () => (dispatch, getState) => {
     });
 
     configs.forEach(config => {
-      config.location = updatedLocation;
-      updatedConfigs.push(UIDatabase.update('TemperatureBreachConfiguration', config));
+      const {
+        id,
+        minimumTemperature,
+        maximumTemperature,
+        duration,
+        description,
+        colour,
+        type,
+      } = config;
+      updatedConfigs.push(
+        UIDatabase.update('TemperatureBreachConfiguration', {
+          id,
+          minimumTemperature,
+          maximumTemperature,
+          duration,
+          description,
+          colour,
+          type,
+        })
+      );
     });
 
     if (replacedSensor) {
--- a/src/database/DataTypes/TemperatureBreachConfiguration.js
+++ b/src/database/DataTypes/TemperatureBreachConfiguration.js
@@ -5,7 +5,20 @@
 
 import Realm from 'realm';
 
-export class TemperatureBreachConfiguration extends Realm.Object {}
+export class TemperatureBreachConfiguration extends Realm.Object {
+  toJSON() {
+    return {
+      id: this.id,
+      minimumTemperature: this.minimumTemperature,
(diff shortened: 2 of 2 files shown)


joshxg (approved):
Ah weird.

I had done a `toJSON` override: https://github.com/openmsupply/mobile/commit/21f83dcd1b3bba9afddb9dce9a2a840de60a92be

But it's gone: https://github.com/openmsupply/mobile/blob/develop/src/database/DataTypes/TemperatureBreachConfiguration.js

🤷 

Means the default `toJSON` is being used, which is creating objects for all of the related entities, which is how the stale location is there.

I think it would be best to add back the `toJSON`, which will not have the `location` object and so won't override it (exactly what the `toJSON` override was trying to prevent - stale objects), but this works so 🤷

wlthomson:
> 
> 
> Ah weird.
> 
> I had done a `toJSON` override: [21f83dc](https://github.com/openmsupply/mobile/commit/21f83dcd1b3bba9afddb9dce9a2a840de60a92be)
> 
> But it's gone: https://github.com/openmsupply/mobile/blob/develop/src/database/DataTypes/TemperatureBreachConfiguration.js
> 
> 🤷
> 
> Means the default `toJSON` is being used, which is creating objects for all of the related entities, which is how the stale location is there.
> 
> I think it would be best to add back the `toJSON`, which will not have the `location` object and so won't override it (exactly what the `toJSON` override was trying to prevent - stale objects), but this works so 🤷

Ah yup. Don't know what happened to the `toJSON` 🤷 . So to clarify, there shouldn't be any live objects in the thunks? 

Nit: doesn't `toJSON` imply a serialised copy of the data (as opposed to `toObject`)?

katherine-sussol:
👋 Was just testing it already so thought I'd drop in my findings - the edit scenario is working nicely (where it previously wasn't) but the remove and re-add with different code is still reverting to the old code on my end. I'd offer a fix but I haven't dived in yet, will also double check in case I've missed something.

katherine-sussol (changes requested):
> 👋 Was just testing it already so thought I'd drop in my findings - the edit scenario is working nicely (where it previously wasn't) but the remove and re-add with different code is still reverting to the old code on my end. I'd offer a fix but I haven't dived in yet, will also double check in case I've missed something.

Ok I think I wasn't going crazy - looks like the functions are different for the create flow (`createNew()`) and the edit flow (`save()`) which explains it. I think the below bit is missing a `location.code` update.

https://github.com/openmsupply/mobile/blob/279304918b2dc95bbba70dccc9ca9b037d73ffd0/src/actions/Entities/SensorActions.js#L178

wlthomson:
> 
> 
> > 👋 Was just testing it already so thought I'd drop in my findings - the edit scenario is working nicely (where it previously wasn't) but the remove and re-add with different code is still reverting to the old code on my end. I'd offer a fix but I haven't dived in yet, will also double check in case I've missed something.
> 
> Ok I think I wasn't going crazy - looks like the functions are different for the create flow (`createNew()`) and the edit flow (`save()`) which explains it. I think the below bit is missing a `location.code` update.
> 
> https://github.com/openmsupply/mobile/blob/279304918b2dc95bbba70dccc9ca9b037d73ffd0/src/actions/Entities/SensorActions.js#L178

Nice! Fixed 594e5325.

joshxg:
> So to clarify, there shouldn't be any live objects in the thunks?

In the new `entities` reducers, I've been trying to avoid them
- SQLite soon *TM*
- Having to hold changes in memory rather than persistently stored straight away
- immutable updates are 'easier'

> Nit: doesn't toJSON imply a serialised copy of the data (as opposed to toObject)?

Yes, I think the preference of everyone is `toObject` but hadn't got round to changing it (I just overrode the new realm in built which in hindsight is doing something different, kinda)

src/actions/Entities/SensorActions.js
```diff
@@ -177,7 +177,7 @@ const createNew = () => (dispatch, getState) => {
     if (existingLocation) {
       newLocation = UIDatabase.update('Location', {
         id: existingLocation.id,
-        description: sensor.name,
+        description: location.name,
```
wlthomson, line 180: Minor nit for readability (equivalent as `location.description = sensor.name` on `SensorActions:170`).

katherine-sussol (approved):
> 
> Nice! Fixed [594e532](https://github.com/openmsupply/mobile/commit/594e53255d1039a288715e26a85ed18130c3dab8).

👍

PR #3660, opened 2021-03-04
Add packages

### LINKED ISSUE #3657: Replace the default JSON Form error messages

JSONForm is using the default AJV error messages.

The errors to be better than, for instance, when an integer is "a", the error message will read "should be integer"

## Implementation

TBd

N/A

## Additional context

There's a few possibilities to investigate.

Want to keep in mind
- Ideally error messages can be customisable from the `uiSchema` or `jsonSchema` which will allow messages to be context driven when required
- Localisation is a factor
- As little work as possible

joshxg:
Fixes #3657 

## Change summary

- Adds the ability to add a keyword into our JSON schema `errorMessage` like in [`ajv-errors`](https://www.npmjs.com/package/ajv-errors)

## Testing

Internal

N/A


src/widgets/JSONForm/testJSON.js
```diff
@@ -254,13 +257,15 @@ export const testTongaSurvey = {
           title: 'First Name',
           description: 'The first name of the respondent',
           type: 'string',
-          maxLength: 50,
+          maxLength: 2,
+          errorMessage: 'Hey! that names way to long, buddy!',
```
joshxg, line 261: This is now the error message for this field. It's not specific to the error - which we can do if we want: https://www.npmjs.com/package/ajv-errors#messages-for-keywords

But I think for simplicity, just adding a single message might be OK?

src/widgets/JSONForm/JSONForm.js
```diff
@@ -113,51 +127,97 @@ class FocusController {
...
+          formRef.current?.onSubmit(e);
+        },
+      }),
+      []
+    );
+
+    return (
+      <JSONFormContext.Provider value={options}>
+        <Form
+          validate={(formData, errorHandlers) => {
```
joshxg, line 151: This is all totally strange stuff to be able to add `ajv-errors` in .. :( Hopefully the comments help and not hinder

src/widgets/JSONForm/JSONForm.js
```diff
@@ -14,6 +15,19 @@ import { PageButton } from '../PageButton';
...
+const ajvErrors = require('ajv-errors');
+
+const ajv = new Ajv({
+  errorDataPath: 'property',
+  allErrors: true,
+  multipleOfPrecision: 8,
+  schemaId: 'auto',
+  unknownFormats: 'ignore',
+  jsonPointers: true,
+});
```
joshxg, line 27: This is basically the same ajv instance which the json schema package uses

src/widgets/JSONForm/testJSON.js
mark-prins, line 261: :pedantic the 'to' should be 'too' in the too long and too short messages..

mark-prins: single message is good

mark-prins (approved):
woo! it works. and is good.
Excited because I realised that the tester was screwing something up, I thought it wasn't working 🙄  and it was just poor testing.

👍

PR #3667, opened 2021-03-04
#3625 vaccines axis ticks

### LINKED ISSUE #3625: Sensor: Temperature Breach Graph axis (X, Y) UI enhancement
## Describe the bug

Some enhancement is needed in Sensor's Graph. For instance, in Detail dialog window of Temperature breach Y axis only shows Date irrespective of time that seems unclear.

### To reproduce

1. Go to Vaccine -> Sensor Detail page
2. One should have atleast one temperature Breach in a day.
3. Click on temperature breach icon from the graph
4. See error. Y axis is only showing Date irrespective of time
photo_2021-02-25 11 36 17

### Expected behaviour

It would be better if Temperature breach graph shows time as well

### Proposed Solution

### Version and device info

- App version: v7.0.0-rc7
- Tablet model: Lenovo
- OS version: 8.1

wlthomson:
Fixes #3625.

## Change summary

Fixes vaccine plot x-axis tick formatting.

## Testing

- [ ] Temperature line chart tick labels display both time and date.
- [ ] Temperature line chart x-axis has no more than 15 ticks.
- [ ] Temperature bar chart tick labels display both time and date.
- [ ] Temperature bar chart x-axis has no more than 15 ticks.

N/A.

### DIFF
--- a/src/utilities/formatters.js
+++ b/src/utilities/formatters.js
@@ -61,7 +61,9 @@ export const formatTimeDifference = duration => {
   );
 };
 
-export const getTickFormatter = () => tick => moment(tick).format('h:mma[\n](D/M)');
+export const timestampTickFormatter = tick => moment(tick).format('h:mma[\n](D/M)');
+
+export const temperatureTickFormatter = tick => `${tick} \u2103`;
 
 export const formatDate = date => (date ? moment(date).fromNow() : generalStrings.not_available);
 
--- a/src/utilities/formatters.js
+++ b/src/utilities/formatters.js
@@ -61,21 +61,7 @@ export const formatTimeDifference = duration => {
   );
 };
 
-export const getTickFormatter = () => {
-  let currentDay = null;
-  return tick => {
-    let formatted = '';
-    const nextCurrentDay = moment(tick).day();
-    if (currentDay !== nextCurrentDay) {
-      currentDay = moment(tick).day();
-      formatted = moment(tick).format('h:mma[\n](D/M)');
-    } else {
-      formatted = moment(tick).format('h:mma');
-    }
-
-    return formatted;
-  };
-};
+export const getTickFormatter = () => tick => moment(tick).format('h:mma[\n](D/M)');
 
 export const formatDate = date => (date ? moment(date).fromNow() : generalStrings.not_available);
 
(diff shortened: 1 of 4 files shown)


src/utilities/formatters.js
```diff
@@ -61,21 +61,7 @@ export const formatTimeDifference = duration => {
...
-      currentDay = moment(tick).day();
-      formatted = moment(tick).format('h:mma[\n](D/M)');
-    } else {
-      formatted = moment(tick).format('h:mma');
-    }
-
-    return formatted;
-  };
-};
+export const getTickFormatter = () => tick => moment(tick).format('h:mma[\n](D/M)');
```
joshxg, line 64: The screenshot is for the [`BreachChart`](https://github.com/openmsupply/mobile/blob/46b159479ad7a0e52a14658107e26ad0125c7a09/src/widgets/BreachChart.js#L60) which doesn't use this function (so doesn't fix the issue with that chart. Maybe we can just use this function in that component?

Personally I think it looks good if you don't display redundant dates, especially if you're viewing a small range

If we do keep it like this, I don't think there's any point to have it return a function given there's no closure

joshxg (approved):

joshxg:
I approved but as noted it doesn't actually fix the reported issue (which is to do with the breach chart) oops

src/utilities/formatters.js
wlthomson: Possibly, but given dates are only redundant when you've been logging for < 1 day, is it worth it for the potential resultant confusion?

joshxg: The only difference this change makes is that the date now shows when you're viewing logs on the same day. The original function would show a date if subsequent ticks were on different days and only when subsequent (am I using this right) ticks were on the same day would remove the day.

wlthomson:
@joshxg @katherine-sussol Updated to use same tick formatting for both vaccine charts.

src/utilities/formatters.js
wlthomson: > The only difference this change makes is that the date now shows on all logs.

That was the idea 😃. I think the conditional date rendering assumes that the user will understand that a time without a date implies the date is the current day.  As this data is quite mission critical, I would think including potentially redundant information is better than omitting it and assuming users will understand the implication.

TLDR; I might not have understood the argument for omitting the date for the current day, but AFAICS the UI looks fine either way, so I can't see a strong argument for adding potential ambiguity here?

But, if a user wanted it that way, I can fix to revert it back!

PR #3674, opened 2021-03-05
#3666 Filter for active VVM Status

### LINKED ISSUE #3666: SI: VVM status shows inconsistency in case of inactive and default values
## Describe the bug

VVM status by default shows first option as value. If that value becomes inactive still values keep showing.

### To reproduce

1. Go to Desktop. Go to Item ->Vaccine -> Show vaccine vial monitor status
2. Create 3-4 options as active. Sync
Screen Shot 2021-03-04 at 1 54 44 PM

3. Go to Mobile app -> Supplier Invoice -> Create an SI -> add vaccinated item
4. VVM status and Location is shown 
5. Location is assigned N/A where as VVM column is filled with First VVM option.
6. Now go to Desktop. Go to Vaccine -> Show Vaccine vial monitor status -> Inactive the first option
Screen Shot 2021-03-04 at 1 35 46 PM

7. Sync. Go to mobile app -> check the unfinalized SI created above
8. See Error. VVM column is still filled with inactive value
9. Create another new SI in mobile. Add vaccine item. 
10. See Error. in this Item list as well inactive VVM value shown

https://user-images.githubusercontent.com/58979436/109931207-a8bebd80-7cf0-11eb-99f6-47275255821d.mp4

### Expected behaviour

for new SI, default value should be the vvm status which has the lowest VaccineVialStatus.level is applied

I am not sure what should be the intended behavior if VVM's value is already assigned and then turn into inactive for unfinalised SI. 

### Proposed Solution

### Version and device info

- App version: v7.0.0 RC8
- Tablet model: Samsung
- OS version: 8.1

### Additional context

arises while testing #2913

katherine-sussol:
Fixes #3666 (partially)

## Change summary

- When auto assigning VVM status to items, filter out inactive ones.

## Testing

- [ ] Ensure there are some vaccine items available to the store for testing
- [ ] In Desktop, set the lowest VVM status to inactive (Item -> Vaccines -> Show Vaccine Vial Monitor Status)
- [ ] Try to do a stocktake or create a new supplier invoice and add a vaccine item line
- [ ] The VVM Status should default to the one with the lowest level that is active

- Historic records will not be updated (e.g. unfinalised SI or Stocktakes), a separate issue will be created for this

### DIFF
--- a/src/database/utilities/createRecord.js
+++ b/src/database/utilities/createRecord.js
@@ -649,6 +649,7 @@ const createItemBatch = (database, item, batchString, supplier) => {
     // Find a VVM Status with the lowest level to auto assign to new item batches.
     const vaccineVialMonitorStatus = database
       .objects('VaccineVialMonitorStatus')
+      .filtered('isActive == true')
       .sorted('level')[0];
 
     if (itemBatch.shouldApplyVvmStatus(vaccineVialMonitorStatus)) {


wlthomson (approved):
Replicated bug on `develop` and confirmed fixed here!

image

Issue #3671, opened 2021-03-04
Refactor NameNote redux state


mark-prins:
## Is your feature request related to a problem? Please describe.

In changing to store a JSON version of the entire survey, the `byId` & `creating` & `editing` are not used / required.
This can be refactored

## Describe the solution you'd like

Remove the `...ById` versions in state.
We should only require something like this
```
  CREATE:
    action to create a default - using a seed object if there is a previous version
    reducer to save state: { creating: newNameNote }
  SAVE:
    state: { creating: nameNote }
  RESET:
    reducer to save state { creating: undefined }
```

## Implementation

**A description of how the solution would best be implemented in the code base**

## Describe alternatives you've considered

**A clear and concise description of any alternative solutions or features you've considered.**

## Additional context

**Add any other context or screenshots about the feature request here.**

PR #3680, opened 2021-03-05
#3663 Step two validation

### LINKED ISSUE #3663: Validation in patient edit ( step two )

- Enable validation of the survey form
- Fix the validation of the patient edit form, which doesn't realise that it has valid data when loading an existing `[Name]` record
- Bug: go to step two and enter data which makes the patient edit form invalid. Go back to step one and the search form is now invalid

Would prefer to have the submit enabled if all required fields are filled in and disabled if there are some on either form to be filled in.

## Additional context

🤔

joshxg:
Fixes #3663 

## Change summary

- Seeds the JSONForm with the most recent name note of the patient or an empty object
- Pre-validates the seed form, enabling/disabling the next action of the wizard
- 

## Testing

Internal

- #3671 to clean up the name note reducer/state


joshxg:
Thinking about this more, I think this isn't right. Should be validate the `name_note.data` field against the schema being used in the page, rather than what it was originally created against.

So, going to make a change to have the `name_note.isValid` field take a schema parameter and pass in the schema being used in the UI.

I am not sure at that point whether we should link the `formSchema` to anything.. I think going back to no linkage anywhere right now, but maybe linking it to the `patient_event` would be wise.. at the moment we've got the form schema with a type `PatientSurvey`, which 'for some reason' links to a `patient_event` with a code `PCD` and we create `name_note`s linked to that `patient_event` which match the `PatientSurvey` `formSchema` and it doesn't feel very sturdy?

src/pages/VaccineDispensingPage.js
```diff
@@ -21,4 +21,4 @@ const tabs = [
   { component: PrescriptionConfirmation, name: 'prescription', title: dispensingStrings.finalise },
 ];
 
-export const VaccineDispensingPage = () => <Wizard tabs={tabs} />;
+export const VaccineDispensingPage = () => <Wizard captureUncaughtGestures={false} tabs={tabs} />;
```
joshxg, line 24: makes scroll betterer

src/reducers/Entities/NameNoteReducer.js
```diff
@@ -11,13 +11,30 @@ export const NameNoteReducer = (state = initialState(), action) => {
...
+      const { data, errors } = payload;
+      const { creatingId } = state;
+
+      const isValid = !(errors?.length > 0);
+
+      const newNameNote = { ...state.creatingById[creatingId], data, isValid };
+      const newById = { ...state.creatingById, [creatingId]: newNameNote };
+
+      return { ...state, creatingById: newById };
+    }
```
joshxg, line 25: Think this state is gonna get cleaned up soon, so just went with the flow - I think the `isValid` can probably be at the root of the state

src/widgets/FormInputs/FormDateInput.js
```diff
@@ -106,9 +106,12 @@ export const FormDateInput = React.forwardRef(
...
       const { timestamp } = nativeEvent;
-      if (!timestamp) return;
-      const newDate = moment(new Date(timestamp)).format('DD/MM/YYYY');
-      onUpdate(newDate, true);
+      if (!timestamp) {
+        setInputState(state => ({ ...state, datePickerOpen: false }));
+      } else {
+        const newDate = moment(new Date(timestamp)).format('DD/MM/YYYY');
+        onUpdate(newDate, true);
+      }
```
joshxg, line 114: fixing a bug where if there was no timestamp selected (canceled) the date picker would stay open

src/widgets/JSONForm/JSONForm.js
```diff
@@ -149,10 +151,13 @@ export const JSONForm = React.forwardRef(
     return (
       <JSONFormContext.Provider value={options}>
         <Form
-          validate={(formData, errorHandlers) => {
+          liveValidate
+          onChange={onChange}
```
joshxg, line 155: So this is live validating now and keeping the `formData` as state in the `NameNote` reducer and constantly feeds it back into the component. I thought to update the `formData` in state mutably so that it did not cause re-renders, but it wasn't too much of a performance hit so figured not doing weird stuff for performance wasn't needed.

src/widgets/Tabs/PatientEdit.js
```diff
@@ -79,7 +82,15 @@ const PatientEditComponent = ({
         <View style={localStyles.verticalSeparator} />
         <View style={localStyles.formContainer}>
           {surveySchema && (
-            <JSONForm ref={formRef} onSubmit={onSubmitSurvey} surveySchema={surveySchema}>
+            <JSONForm
+              onChange={data => {
+                updateForm(data.formData, data.errors);
+              }}
```
joshxg, line 88: passing the formdata and any errors (because live validation is enabled) up to the store to enable/disable the next button. Figured that's easier than checking when clicking next and warning, even though I kinda prefer on submit validation

mark-prins:
> I am not sure at that point whether we should link the `formSchema` to anything.. I think going back to no linkage anywhere right now, but maybe linking it to the `patient_event` would be wise.. at the moment we've got the form schema with a type `PatientSurvey`, which 'for some reason' links to a `patient_event` with a code `PCD` and we create `name_note`s linked to that `patient_event` which match the `PatientSurvey` `formSchema` and it doesn't feel very sturdy?

On reading your comment, I wondered about having a 'patient event code' field on the `form_schema` to tighten up the linking. Maybe have a more formal link between the tables - though I'm not too worried about having a general 'patient clinical data' event which is used to attach a random note. worth further discussion this one.

src/widgets/Tabs/PatientEdit.js
```diff
@@ -106,23 +117,32 @@ const PatientEditComponent = ({
...
 
-  return { onCancelPrescription, onSubmitSurvey, updatePatientDetails };
+  return { onCancelPrescription, onSubmitSurvey, updatePatientDetails, updateForm };
 };
 
 const mapStateToProps = state => {
   const currentPatient = selectEditingName(state);
   const completedForm = selectCompletedForm(state);
-  const canSaveForm = selectCanSaveForm(state);
+  const canSaveForm = selectCanSaveForm(state) && selectNameNoteIsValid(state);
```
mark-prins, line 133: there's still an issue with the validation and seeding in the `Name` form - first name, last name and DOB are required; even when populated with the existing data, the validation isn't firing, so the button isn't enabled.

mark-prins (approved):
That's way better - nice work!
The form is still re-rendering a lot, perhaps we look at that later?
There's a couple of bugs noted in #3663 not addressed - what do you think about spinning out a new issue for those?

joshxg:
- Merged everything in, needed some changes to make it worky
- ff7ef9d Did this one to fix the patient search not being valid when switching between step1/step2
- if there are other validation issues (I currently have been all good on the validation after that last commit) will need to make a new issue

PR #3685, opened 2021-03-06
#3671 Refactor of NameNote state

### LINKED ISSUE #3671: Refactor NameNote redux state

joshxg:
Fixes #3671

## Change summary

- Bit of a refactor of `NameNote` state/reducer/actions/selectors
- Bit of a tidy of `Name` and `VaccinePrescription` state/reducer/actions/selectors
- Made the form for editing a patient not auto focus - thoughts are that the name of a patient, and really, any of their details are unlikely to change. Arguably could have the JSONForm auto focus, that can be controlled by the `uiSchema`

## Testing

- [ ] `VaccinePrescription` flow is working good as

N/A

### DIFF
--- a/src/actions/Entities/NameActions.js
+++ b/src/actions/Entities/NameActions.js
@@ -1,5 +1,6 @@
 import { generateUUID } from 'react-native-database';
-import { selectEditingNameId, selectEditingName } from '../../selectors/Entities/name';
+import { batch } from 'react-redux';
+import { selectEditingName } from '../../selectors/Entities/name';
 import { createRecord, UIDatabase } from '../../database';
 
 export const NAME_ACTIONS = {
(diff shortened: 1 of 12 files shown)


joshxg:
You had `save` in the issue, but since then we've changed to live validation, so using `update` instead. Hopefully this is what you were thinking also!

mark-prins (approved):
So much nicer!
Tested and working just fine - some nice touches with the keyboard and auto focus too.. looking really good

Issue #3613, opened 2021-02-25
Update the patient edit modal to allow for editing a patients vaccine dispensing survey


joshxg:
## Is your feature request related to a problem? Please describe.

We need to allow users to create/update a patients vaccine dispensing survey when editing a patients record.

## Describe the solution you'd like

- Currently the patient editing modal is simply a `ModalContainer` with a child `FormControl`.
- We need to add, somehow (maybe two columns) an additional set of inputs for a patients vaccine dispensing survey, which will be rendered through the package/component added with #3611 

## Implementation

Maybe best, at this point, to create a new component `PatientEditModal` to put the `FormControl` and newly added components to create the additional form.

## Describe alternatives you've considered

N/A

## Additional context

N/A

joshxg:
Waiting on https://github.com/openmsupply/mobile/issues/3619#issuecomment-785658508

PR #3687, opened 2021-03-07
#3613 JSONForm in PatientEditModal

### LINKED ISSUE #3616: Add a createRecord case to create a vaccine prescription

- We need to create a 'special' vaccine prescription.
- The vaccine prescription will be created in-memory and have all details filled in memory. This is in contrast to regular operations, where we create a `realm` object and then update that `realm` object.
- So, we need a `createRecord` case, which will take an in-memory vaccine prescription and save a new CI 

- A new case in `createRecord` which accepts a vaccine-prescription object and saves it persistently

## Implementation

see `createRecord.js`

- We need to make the prescription in memory, which will have most of the details in the object already. So, this case might actually just end up being `database.update("Transaction", prescription)` without actually adding any value.
- The only real reason for suggesting this is to be consistent with the rest of the app.

## Additional context

N/A

joshxg:
Fixes #3616 

## Change summary

- Adds a `PatientEditModal` component to encapsulate some of the code a little
- Adds `JSONForm` into `PatientEditModal` in a somewhat crude way
- Adds creating new name notes from the `PatientEditModal`

## Testing

- [ ]  When editing a patient, can fill the survey with some details and have a name note created
- [ ] When editing a patient, can only save the form when all values, including those of the survey are valid
- [ ] No changes to behaviour when no survey

image

Few issues to come out of this
- Tidy up UI of this modal
- Create some sort of `touched` state which gets set after an `update` action, and only save/create a new name note if the form has been touched (Could check if the actual diff is different also, but a simple `touched` is gonna catch the majority of cases and is a bit simpler. Although could just do a `JSON.stringify` compare of the new name note data and the old name note data to see if anything has changed to check also.
- I don't really like the implementation in general, but I think we are gonna do more work in this area with AEFI's so kinda just wanna have it working and spend time on the area when we're doing AEFI's, otherwise we might do something we have to change when we do AEFI's anyway. Also AEFI's.

### DIFF
--- a/src/actions/Entities/NameNoteActions.js
+++ b/src/actions/Entities/NameNoteActions.js
@@ -71,9 +71,12 @@ const updateForm = (data, errors) => ({
   payload: { data, errors },
 });
 
-const saveEditing = () => (dispatch, getState) => {
+const saveEditing = optionalNameID => (dispatch, getState) => {
   const { nameID, patientEventID, ...nameNote } = selectCreatingNameNote(getState()) ?? {};
-  UIDatabase.write(() => createRecord(UIDatabase, 'NameNote', nameNote, patientEventID, nameID));
+
+  UIDatabase.write(() =>
+    createRecord(UIDatabase, 'NameNote', nameNote, patientEventID, nameID || optionalNameID)
+  );
   dispatch(reset());
 };
 
--- a/src/database/DataTypes/Name.js
+++ b/src/database/DataTypes/Name.js
@@ -139,7 +139,7 @@ export class Name extends Realm.Object {
       dateOfBirth: this.dateOfBirth?.getTime(),
       phoneNumber: this.phoneNumber,
       country: this.country,
-      billingAddress: this.billingAddress.toJSON(),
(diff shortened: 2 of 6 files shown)


mark-prins:
There's a separate issue with the validation when the schema is undefined ( #3689 )

Working around that - with no schema the layout is still expecting a survey form: can we stretch across the page like it used to?

Screen Shot 2021-03-08 at 10 34 52 AM

mark-prins (commented):
Changes look good! 
The note didn't save for me - am testing now :(
Also the layout when there is no schema, could you have a look at that please?

mark-prins:
ok, yeah. tester configuration error. saving nicely.

mark-prins (approved):
Code is good 👍 
would be nice to tweak the layout if the schema is undefined

joshxg:
Updated styles, but also realise I never tested making a new patient! Working on a fix for that now....!

joshxg:
OK a bit hacky but should be handling new names now: 
cb709ac

PR #3696, opened 2021-03-08
#3681 Add OK and Next functionality

### LINKED ISSUE #3681: Vaccine dispensing: "ok & next" button

Add an 'ok and next' button to the end of the process

This is to allow a user to finalise the current prescription and immediately create another

## Implementation

The current design has `Cancel` and `Confirm` buttons.
Add a third button which is called `Confirm & Next` to be right most, on the assumption that this is the most useful button. Cancel on the left, Confirm in the middle.

There is an additional click required without using this button. Not doing anything is only a minor difference
It is a slightly more confusing UI - is the saving of a button press worth it?
Should the 'Confirm' be 'Ok' on both buttons?

## Additional context

Wording and button position up for debate

joshxg:
Fixes #3681 

## Change summary

- OK And Next feature

## Testing

- [ ] Can OK and next

N/A

### DIFF
--- a/src/widgets/Tabs/VaccineSelect.js
+++ b/src/widgets/Tabs/VaccineSelect.js
@@ -219,8 +219,6 @@ const mapStateToProps = state => {
   const [selectedVaccine] = selectedVaccines;
   const vaccinator = selectSelectedVaccinator(state);
 
-  const vaccinator = selectSelectedVaccinator(state);
-
   return {
     vaccinator,
     hasRefused,
--- a/src/widgets/Tabs/VaccineSelect.js
+++ b/src/widgets/Tabs/VaccineSelect.js
@@ -34,7 +34,12 @@ import {
 import { getColumns } from '../../pages/dataTableUtilities';
 import { useLoadingIndicator } from '../../hooks/useLoadingIndicator';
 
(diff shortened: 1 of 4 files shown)


src/widgets/Tabs/VaccineSelect.js
```diff
@@ -162,6 +173,12 @@ const VaccineSelectComponent = ({
           isDisabled={selectedBatches.length === 0 && !hasRefused}
           onPress={confirmPrescription}
         />
+        <PageButton
+          text={generalStrings.ok_and_next}
+          style={{ marginLeft: 5 }}
+          isDisabled={selectedBatches.length === 0}
```
mark-prins, line 179: I would think that it should be enabled if the patient has refused too

```suggestion
          isDisabled={selectedBatches.length === 0 && !hasRefused}
```

mark-prins (approved):
It's a thing of beauty 👍 

I wondered about `Confirm & Next` to make the button more in keeping with `Confirm`.. but it is too long. `Ok & Next` is what desktop does, so it is consistent with that

src/widgets/Tabs/VaccineSelect.js
joshxg: Nice, thanks!

PR #3716, opened 2021-03-10
#3714 Default and recommended item/batches

### LINKED ISSUE #3714: Auto select the most recently used vaccine

When navigating to step 3 on vaccine dispensing, we should pre-select the most recently used vaccine

- pre-select the most recently used vaccine

## Implementation

When first navigating or enabling, make the selected vaccine in the most recently confirmed transaction which had a vaccine in it (just like, the first one, I guess??)

- Could select the vaccine which has the highest monthly usage? Either is better than nothing?

## Additional context

N/A

joshxg:
Fixes #3714 

## Change summary

- Adds preselected item/batch for vaccine select

## Testing

- [ ] When dispensing a vaccine, the most recently used item and the recommended batch are selected

N/A

### DIFF
--- a/src/actions/Entities/VaccinePrescriptionActions.js
+++ b/src/actions/Entities/VaccinePrescriptionActions.js
@@ -73,7 +73,7 @@ const create = () => ({
     prescription: createDefaultVaccinePrescription(),
     vaccinator: getDefaultVaccinator(),
     selectedVaccines: [getDefaultVaccine()],
-    selectedBatches: [getRecommendedBatch(getDefaultVaccine())],
+    selectedBatches: [getRecommendedBatch()],
   },
 });
 
--- a/src/actions/Entities/VaccinePrescriptionActions.js
+++ b/src/actions/Entities/VaccinePrescriptionActions.js
@@ -44,9 +44,37 @@ const getDefaultVaccinator = () => {
   return null;
 };
 
+const getDefaultVaccine = () => {
+  const [mostRecentTrans] = UIDatabase.objects('Transaction')
+    .filtered("type == 'customer_invoice' && (status == 'finalised' || status == 'confirmed')")
+    .sorted('confirmDate', true);
+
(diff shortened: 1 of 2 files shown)


src/selectors/Entities/vaccinePrescription.js
```diff
@@ -50,7 +50,7 @@ export const selectSortedVaccines = createSelector([selectVaccines], vaccines =>
 });
 
 export const selectSelectedRows = createSelector([selectSelectedVaccines], vaccines =>
-  vaccines.reduce((acc, vaccine) => ({ ...acc, [vaccine.id]: true }), {})
+  vaccines.reduce((acc, vaccine) => ({ ...acc, [vaccine?.id]: true }), {})
```
mark-prins, line 53: 🤔  - I'm not sure why there would be null entries in the array. will take your word for it

src/actions/Entities/VaccinePrescriptionActions.js
```diff
@@ -55,7 +83,7 @@ const reset = () => ({
 
 const selectVaccine = vaccine => ({
   type: VACCINE_PRESCRIPTION_ACTIONS.SELECT_VACCINE,
-  payload: { vaccine },
+  payload: { vaccine, batch: getRecommendedBatch() },
```
mark-prins, line 86: the call to `getRecommendedBatch` is using the default vaccine, but should be using the one which is passed in here: `vaccine`

mark-prins (changes requested):
Nice! really like the idea - good improvement to the workflow.
Just needs the default batch selection to use the selected vaccine

src/actions/Entities/VaccinePrescriptionActions.js
joshxg: 13b1a42 ha, those last minute untested changes aye

mark-prins (approved):

PR #3717, opened 2021-03-10
filter patient search form on change

### LINKED ISSUE #3711: Remove search button from patient selection form

With the on-screen keyboard, the search button is hidden and there's only one field showing.
Given that we are only searching local users, switching to a debounced live filter should be responsive enough and would remove the need for the search button.

Remove the search button
Search onChange of any of the fields on the form, with a suitable debounce

mark-prins:
Fixes #3711 

## Change summary

Removes the search button from the patient form again.. adds in a debounced onChange filter instead

## Testing

- [ ] Search for patients in step #1
- [ ] Lookup patients in normal dispensing to check nothing else is broken

The debounce on the date field is making it s-u-p-e-r-s-l-o-w when typing. Still works though, and that's outside of these changes, so won't address it here.

### DIFF
--- a/src/utilities/constants.js
+++ b/src/utilities/constants.js
@@ -15,3 +15,7 @@ export const MILLISECONDS = {
 export const SECONDS = {
   ONE_MINUTE: 60,
 };
+
+export const DATE_FORMAT = {
+  DD_MM_YYYY: 'DD/MM/YYYY',
+};
--- a/src/widgets/FormControl.js
+++ b/src/widgets/FormControl.js
@@ -276,7 +276,7 @@ const mergeProps = (stateProps, dispatchProps, ownProps) => {
   } = ownProps;
   const onInitialiseForm = () => initialiseForm(inputConfig);
   const onUpdateForm = (key, value) =>
-    onUpdate && !isDisabled ? onUpdate(key, value) : updateForm(key, value);
+    !isDisabled && (onUpdate ? onUpdate(key, value) : updateForm(key, value));
   const onSaveForm = () =>
     confirmOnSave && !isConfirmFormOpen ? showConfirmForm() : onSave(completedForm);
   const onCancelForm = () =>
--- a/src/widgets/FormControl.js
+++ b/src/widgets/FormControl.js
@@ -48,6 +48,7 @@ import { useDebounce } from '../hooks/useDebounce';
  * @param {bool}  isDisabled      Indicator whether this Form is disabled.
  * @param {Array} inputConfig     Configuration array of input config objects.
  *                                See { getFormInputConfig } from src/utilities/formInputConfigs.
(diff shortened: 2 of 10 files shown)


src/utilities/constants.js
```diff
@@ -15,3 +15,7 @@ export const MILLISECONDS = {
 export const SECONDS = {
   ONE_MINUTE: 60,
 };
+
+export const DATE_FORMAT = {
```
mark-prins, line 19: Completely not necessary.. it was troubling me slightly that we had a string value all over the place for this one.

joshxg: I like it :)

src/widgets/Tabs/PatientSelect.js
```diff
@@ -59,10 +61,16 @@ const PatientSelectComponent = ({
   onFilterData,
   patients,
   selectPatient,
+  updateForm,
 }) => {
   const columns = React.useMemo(() => getColumns(MODALS.PATIENT_LOOKUP), []);
   const { pageTopViewContainer } = globalStyles;
   const keyboardIsOpen = useKeyboardIsOpen();
+  const debouncedFilter = useDebounce(onFilterData, 500);
```
joshxg, line 69: I tried this with `250` and for me it works fine and has a little less lag. Totally dependent on the typist though and if it triggers to much it does get a bit janky (50 is too low!!). 500 is probably good

src/widgets/FormControl.js
```diff
@@ -271,9 +272,11 @@ const mergeProps = (stateProps, dispatchProps, ownProps) => {
     cancelButtonText,
     onCancel,
     canSave = true, // if not specified by calling component, set default
+    onUpdate,
   } = ownProps;
   const onInitialiseForm = () => initialiseForm(inputConfig);
-  const onUpdateForm = (key, value) => !isDisabled && updateForm(key, value);
+  const onUpdateForm = (key, value) =>
```
joshxg, line 278: I would've thought it doesn't matter what this fn is if the form is disabled, but it does have me curious if there was a reason it was not calling `updateForm` when it was disabled.

Currently if there is an `onUpdate` passed and it is disabled then `updateForm` is set- probably nothing 🤔

joshxg (approved):
Nice - working excellently. I was wanting some little indication that things had registered and it was about to do the search, like some sort of spinner, but it's probably just because I know what's going on, where a user likely doesn't.

src/widgets/FormControl.js
mark-prins: oopsie. 6634c014

src/widgets/Tabs/PatientSelect.js
mark-prins: Compromised.. went with `300`  though the `250` was fine for me too 🤷‍♂️

PR #3718, opened 2021-03-10
#3705 download state improvements

### LINKED ISSUE #3705: Sensor: updating logging interval and sensor code field frequently shows alert warning and doesn't allowed to update the Sensor's value
## Describe the bug

 updating logging interval and sensor code field frequently shows alert warning and doesn't allowed to update the Sensor's value. Its not showing consistent behavior. Sometimes error alert is shown sometimes its not. Not sure whats causing this behavior. I tried removing the Sensor followed by restarting the app and adding sensor again but after few updates it start showing same error alert message.

### To reproduce

1. Go to added sensor list
2. Click on cog icon
3. Change the Sensor code and then change the logging interval. Click on Save
4. It might saved/updated. If updating successful try repeating step 3 again
4. See error. It shows error " A download is in progress, please try again soon"

https://user-images.githubusercontent.com/58979436/110440044-53afec80-80e0-11eb-83c0-fb980557b4c8.mp4

### Expected behaviour

Its should update the values without any alert error warning

### Proposed Solution

### Version and device info

- App version: V7.0.0 rc9
- Tablet model: Lenovo
- OS version: 8.1

### Additional context

Low priority for now as its not blocker.

katherine-sussol:
Fixes #3705 (or at least improves it)

## Change summary

Have found that if the app is left running long enough the app schedules lots and lots of `startDownloadAll` actions (one job is created each time the main `mSupplyMobileApp` component updates) which eventually gets the app to the point that it's always 'downloading' and will be very difficult to update a sensor's logging interval 😆 . 

https://github.com/openmsupply/mobile/blob/4a79db0e40092612a8eeaacc5d66e084884d99ac/src/mSupplyMobileApp.js#L114

This PR just adds a small update to add some state to track if the passive download has been scheduled at least once.

## Testing

- [ ] Re-test #3705 and verify that the sensor log interval can be updated more consistently (it will still fail if there is a download happening in the background but should happen less after this change)
- [ ] Verify downloads are still happening every ~1min in the background (check wifi icon in UI for user testing or console logging for dev testing)

If there are lots of sensors (and especially if one is out of range) it may still be a little difficult to update the sensor interval, but this is more of an edge case...

### DIFF
--- a/src/actions/Bluetooth/SensorDownloadActions.js
+++ b/src/actions/Bluetooth/SensorDownloadActions.js
@@ -19,11 +19,13 @@ export const DOWNLOAD_ACTIONS = {
   DOWNLOAD_LOGS_START: 'Bluetooth/downloadLogsStart',
   DOWNLOAD_LOGS_ERROR: 'Bluetooth/downloadLogsError',
   DOWNLOAD_LOGS_COMPLETE: 'Bluetooth/downloadLogsComplete',
+  PASSIVE_DOWNLOAD_START: 'Bluetooth/passiveDownloadStart',
   SENSOR_DOWNLOAD_START: 'Bluetooth/sensorDownloadStart',
   SENSOR_DOWNLOAD_SUCCESS: 'Bluetooth/sensorDownloadSuccess',
   SENSOR_DOWNLOAD_ERROR: 'Bluetooth/sensorDownloadError',
 };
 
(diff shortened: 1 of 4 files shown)


katherine-sussol (commented):
Just realised `downloadAll` is still used to update the battery level, thinking this should still be done even if the sensor is 'paused' so undid my original change 🤔

joshxg (approved):
👍  nice :)

Had a quick look atthe other subscription in the cdu method also and it's a single subscription not an array of subscribers so that one should be all good! `BluetoothStatus.addListener()`

PR #3721, opened 2021-03-10
#3719 Add reason column for vaccines

### LINKED ISSUE #3719: Allow users to choose an open vial wastage reason for vaccines in a stock take

We need to allow for a user to select an open vial wastage reason in a stocktake.

For a vaccine item in a stoctake, allow a user to select `negativeInventoryAdjustment` OR `openVialWastage` reasons for a vaccine item with a negative difference.

## Implementation

Unfortunately, there's no room. 

I think maybe we need to remove the breach column..?

N/A

## Additional context

N/A

joshxg:
Fixes #3719 

## Change summary

- Adds a reason column to vaccines, and allows for applying an `openVialWastage`
- Quite a hacky solution. Not very happy about it, but 

## Testing

- [ ] Can apply a reason to a vaccine in a stocktake

image


src/widgets/modals/StocktakeBatchModal.js
```diff
@@ -89,14 +96,26 @@ export const StocktakeBatchModalComponent = ({
...
-    difference > 0
-      ? UIDatabase.objects('PositiveAdjustmentReason')
-      : UIDatabase.objects('NegativeAdjustmentReason');
+  let reasonsSelection = null;
+  if (usingReasons && !isVaccine) {
+    reasonsSelection =
+      difference > 0
+        ? UIDatabase.objects('PositiveAdjustmentReason')
+        : UIDatabase.objects('NegativeAdjustmentReason');
+  } else if (usingOpenVialWastageReasons && isVaccine) {
```
mark-prins, line 109: It is possible that `usingReasons` is false and it is a vaccine and there aren't any positive adjustment reasons, only the wastage reasons.

Though it really doesn't matter as the `REASON` column only shows if there are positive and negative reasons available. Really just wondering if the query could be simplified at all.

src/widgets/modals/StocktakeBatchModal.js
```diff
@@ -89,14 +96,26 @@ export const StocktakeBatchModalComponent = ({
...
+    reasonsSelection =
+      difference > 0
+        ? UIDatabase.objects('PositiveAdjustmentReason')
+        : UIDatabase.objects('Options').filtered(
+            '(type == $0 || type == $1) && isActive == true',
+            'negativeInventoryAdjustment',
+            'openVialWastage'
+          );
+  }
```
mark-prins, line 118:
```suggestion
  const reasonsSelection = (() => {
    if (difference > 0) {
      return UIDatabase.objects('PositiveAdjustmentReason');
    }

    if (usingReasons && !isVaccine) {
      return UIDatabase.objects('NegativeAdjustmentReason');
    }

    if (usingOpenVialWastageReasons && isVaccine) {
      return UIDatabase.objects('Options').filtered(
        '(type == $0 || type == $1) && isActive == true',
        'negativeInventoryAdjustment',
        'openVialWastage'
      );
    }

    return [];
  })();
```

mark-prins (approved):
that looks reasonable 😉 

all good - as you say, could perhaps be neater in the UI but works fine.

src/widgets/modals/StocktakeBatchModal.js
joshxg: oooo nice fancy pants

src/widgets/modals/StocktakeBatchModal.js
joshxg: Yeah, it's not great and I don't know what to do about it.. :(

PR #3679, opened 2021-03-05
#3364 disable location for non-vaccine items

wlthomson:
Fixes #3364,

## Change summary

- Disables location cell for non-vaccine items.
- Fixes `CURRENT_LOCATION` column key typo.

## Testing

Steps to reproduce or otherwise test the changes of this PR:

- [ ] Create a supplier invoice with a vaccine and non-vaccine item. The vaccine item location is orange and editable.
- [ ] Create a supplier invoice with a vaccine and non-vaccine item. The non-vaccine item location is grey and non-editable.

### Related areas to think about

N/A.

PR #3302, opened 2020-12-20
#3299 Update Schema

joshxg:
Fixes #3299 

## Change summary

- Updates schema as per the issue
- Updates sync with updating schema

## Testing

*Currently not testable as desktop needs an updated schema.. :)*

- [ ] Can sync with desktop

### Related areas to think about

N/A

PR #3724, opened 2021-03-10
#3679 remove duplicate temperature breach methods

### LINKED ISSUE #3697: Remove duplicated TemperatureBreach methods
## The suggestion

Remove unused `TemperatureBreach` methods:

- `willContinueBreach(temperatureLog)`
- `get colour()`

**Why?**

Unused and non-functional (will throw errors if called).

N/A.

## Additional context

N/A.

wlthomson:
Fixes #3697.

## Change summary

Just tidying up by removing some old methods. 

## Testing

N/A.

N/A.

### DIFF
--- a/src/database/DataTypes/TemperatureBreach.js
+++ b/src/database/DataTypes/TemperatureBreach.js
@@ -7,10 +7,6 @@ import Realm from 'realm';
 import moment from 'moment';
 
 export class TemperatureBreach extends Realm.Object {
-  get colour() {
-    return this.temperatureBreachConfiguration?.colour ?? null;
-  }
-
   get maximumTemperature() {
     return this.temperatureLogs.max('temperature');
   }
@@ -74,13 +70,6 @@ export class TemperatureBreach extends Realm.Object {
     );
 
     return incomingStock.sum('numberOfPacks') ?? 0 - outgoingStock.sum('numberOfPacks') ?? 0;
-  }
-
-  willContinueBreach(temperatureLog) {
-    const { minimumTemperature, maximumTemperature } = this.temperatureBreachConfiguration;
-    const { temperature } = temperatureLog;
-
-    return temperature >= minimumTemperature && temperature <= maximumTemperature;
   }
 
   toJSON() {
--- a/src/widgets/BreachChart.js
+++ b/src/widgets/BreachChart.js
@@ -19,17 +19,14 @@ import { timestampTickFormatter, temperatureTickFormatter } from '../utilities/f
 export const BreachChart = ({ lineData, x, y, xTickFormat, yTickFormat, breach }) => {
   const [width, height, setDimensions] = useLayoutDimensions();
 
-  const { minimumTemperature, maximumTemperature, colour } = breach;
+  const { minimumTemperature, maximumTemperature } = breach;
 
-  const lineStyle = React.useMemo(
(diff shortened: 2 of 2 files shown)


src/database/DataTypes/TemperatureBreach.js
```diff
@@ -7,10 +7,6 @@ import Realm from 'realm';
 import moment from 'moment';
 
 export class TemperatureBreach extends Realm.Object {
-  get colour() {
```
katherine-sussol, line 10: Just double checking if this line is a potential reference?  https://github.com/openmsupply/mobile/blob/c21dedf5f31bd667ab284dff76608adfc8241b2f/src/widgets/BreachChart.js#L22

`colour` doesn't ring a bell to me though and looks like it's usually `null`...unless configured in desktop somewhere 🤔 ?

src/database/DataTypes/TemperatureBreach.js
```diff
@@ -76,13 +72,6 @@ export class TemperatureBreach extends Realm.Object {
     return incomingStock.sum('numberOfPacks') ?? 0 - outgoingStock.sum('numberOfPacks') ?? 0;
   }
 
-  willContinueBreach(temperatureLog) {
```
katherine-sussol, line 79: 👍

src/database/DataTypes/TemperatureBreach.js
mark-prins: it is indeed.
@wlthomson would you care to comment? Would be nice to wrap up this PR

wlthomson: Well `this.temperatureBreachConfiguration?.colour ?? null` is always going to be `undefined` (as `temperatureBreachConfiguration` is not a field. I might be missing something, but not sure how desktop config would make a difference? Are you suggesting just replace the method with `return null`? In that case, is there much advantage of doing that over just leaving it without the method and returning `undefined`?

katherine-sussol: I see, that's a fair point, didn't see that `this.temperatureBreachConfiguration` wasn't even a field on `TemperatureBreach`. I was concerned there was a use case where a colour could be set up in Desktop and there was an expectation of colour changing in `BreachChart`, but now I see that it wouldn't work regardless of this PR haha. All good then 😆 .

katherine-sussol (approved):
👍

src/database/DataTypes/TemperatureBreach.js
wlthomson: I'm not sure, it looks like the changes were introduced here: https://github.com/openmsupply/mobile/pull/3302/files.

@joshxg any thoughts?

joshxg: Yeah there was intention to use colour on the config to colour the breach chart. Been scrapped.

The config relation on a breach was also changed to explicitly saved the values instead for ensuring history was kept.

Can be removed. It is also referenced elsewhere (breach chart) which could be removed

wlthomson:
Have removed the `colour` logic from `BreachChart`, might be good to get a quick re-review just in case (sorry!).

katherine-sussol:
Hmm ok, I think I mentally skipped this PR for ages, oops - latest commit looks good, should probs just merge lol.

PR #3743, opened 2021-03-11
#3730 Add new pref to enable editing of all patients

### LINKED ISSUE #3730: Handle the pref to enable editing of all patients

There is a new pref on desktop: `can_edit_patients_from_any_store`

When this preference is enabled, allow editing all patients.

## Implementation

TBD

TBD

joshxg:
Fixes #3730 

## Change summary

- Adds the pref to enable editing of all patients 🙃 

## Testing

- [ ] Looking up a patient from the lookup modal while having this pref


src/database/DataTypes/Name.js
```diff
@@ -4,6 +4,8 @@
  */
 
 import Realm from 'realm';
+import { UIDatabase } from '../index';
+import { PREFERENCE_KEYS } from '../utilities/preferenceConstants';
```
joshxg, line 8: This isn't ideal- adding UIDatabase import is a circular reference .. 

Alternatives:

- Put it into the `DataTableRow` (which is calling the getter on an instance of this class). Not great as it means this getter is actually not correct
- Put it into the component which calls `DataTableRow` and pass as some sort of `option` prop... not great as there is already plenty of props and the condition for enabling/disabling a row is already really crazy in `DataTableRow`, but the advantage is that this menans there's one db call

mark-prins: would it be crazy to lookup preferences when the app starts and store them globally?
Update the state on incoming sync?

mark-prins (approved):
I wonder about storing prefs in state, rather than looking up in realm each time. maybe later though?

joshxg:
🤔 

Maybe a good alternative. Also thought of saving editability against the name records themselves... (and not syncing back)

I think it would be good to do some testing with a lot of name records anyway, so we can see if it makes a difference there. Realm is really quick!

PR #3744, opened 2021-03-11
#3723 remove live validation

### LINKED ISSUE #3723: remove live validation of survey form

I've caused problems in the survey form by adding `island` and `village`
Removing the live validation should resolve the problem

- Disable live validation
- Enable `Next` button
- validate survey form on submit
- prevent navigation to the next page if validation fails
- display warnings after validation

Revert to a very long list of villages, with `...(island)` in the name to disambiguate the duplicates.
Remove the required attribute for village and island

## Additional context

mark-prins:
Fixes #3723 

## Change summary

Does what it says - removes the live validation. This fixes up the strangeness that `island` and `village` were having.
But.. the validation with `ajv` doesn't actually really work, and so you don't have to complete the required fields.
I've spent some time trying to get it to behave and have given up. Basically - the `onError` in `JSONForm` is firing correctly and populating with errors ( though often this doesn't on the first submit )
However, the validation run in `NameNoteActions` where there is a `required: ["island"]` and `patientClinicalData` is `{}` does not cause the validation to fail. 

Thought that I'd propose the PR regardless - the current live validation isn't working with the drop downs either.
Alternative? reformat the island and village lists perhaps.

## Testing

- [ ] Test validation on the patient edit form


joshxg:
What I did

- Added the following to the schema to make the properties required. Not ideal, but if we put a default rather than a placeholder for everything (meaning we can't ever select an invalid value) then I think we don't need the `patientDemographicData`
```
{
  "type": "object",
  "title": "Extra information",
  "required": ["patientDemographicData"],
  "properties": {
    "patientDemographicData": {
      "title": "Patient Demographic Data",
      "type": "object",
      "minProperties": 2,
```
- Made `Select` trigger an `onChange` if it's value is not in the available options. The components design only updates the onChange and doesn't check to change any dependencies or anything like that, so the old value is 'stale' in a sense. I think it's OK, generally, but it is a nice quality of life to have this swapped.
- Changed to validate on every `onChange`. Obviously not ideal but not to bad (also debounced).
- Changed custom error messages to only show one error, and to prefer the custom error message set in the json schema under `errorMessage`. This is because I was getting spammed with lots of messages and it was ugly as.


Not perfect. Still a bit to go - hopefully we have some time with the ADR work to improve on this.

PR #3751, opened 2021-03-12
#3750 UI tweaks

### LINKED ISSUE #3750: UI tweaks

Some feedback

1. more gap between field label and the field above (e.g. on patient edit window)
2. Cancel and save buttons should be full width down the bottom of same window
3. date picker icon a bit low res?

mark-prins:
Fixes #3750 

## Change summary

- Adds a little padding between heading and text
- makes the buttons span the page in patient edit

The date icon is an svg - so no change needed to make it sharper. unsure about this one.

Side effects of this PR: fixed the patient edit form which wasn't loading the survey and was creating patients rather than editing.

## Testing

- [ ] Edit a patient ( check the big buttons and functionality )
- [ ] look at the spacing on the survey form


src/widgets/modalChildren/PatientEditModal.js
```diff
@@ -89,22 +133,44 @@ PatientEditModalComponent.propTypes = {
   onUpdateForm: PropTypes.func.isRequired,
 };
 
+const mergeProps = (stateProps, dispatchProps, ownProps) => {
+  const { completedForm, nameNote, nameNoteIsValid, surveyForm } = stateProps;
```
joshxg, line 137: There's no `nameNote` is `stateProps`. Maybe best to just spread em (have been confused in the past when returning things from a `stateToProps` to not receive it in the component because a `mergeProps` wasn't returning it.

```
const {completedForm} = stateProps;
...

return {
	...ownProps,
	...stateProps,
}
```

src/widgets/modalChildren/PatientEditModal.js
```diff
@@ -23,11 +26,9 @@ export const PatientEditModalComponent = ({
...
   nameNoteIsValid,
 }) => {
-  const onSaveWithForm = completedForm => {
-    const id = generateUUID();
-
-    onSave({ id, ...completedForm });
-    onSaveSurvey(id);
+  const onSaveWithForm = () => {
+    onSaveForm();
+    onSaveSurvey();
```
joshxg, line 31: With this, creating a new name is not creating a name note:

I added this because when creating a new patient, we don't have the ID of the name yet (we create the name from the `completedForm` of the `FormControl` and we create the name note from the `formData` of the `JSONForm`.. but we need an `id` for the patient before creating the `name_note`, so I added an optional `id` to these for this case.

We have no `id` to pass into `createSurveyNameNote` when opening, so we don't have the `id` of the `name` when we call `onSaveSurvey`.

Think we need to add a prop `isCreating`, and if we're creating, do the same as before- generate a UUID and pass that into the actions, otherwise, don't pass it in

joshxg (changes requested):
Ohhh, Those buttons!

src/widgets/modalChildren/PatientEditModal.js
mark-prins: good point - I've updated, so when we extract patient from `ownProps`, if there isn't one then generate a new ID and use that throughout. No need for `isCreating` unless you think it would be better to be explicit?

5f45636b

joshxg (approved):
Yup nice

PR #3766, opened 2021-03-15
#3758 reset log interval on log download

### LINKED ISSUE #3758: Sensor log interval is not reset after downloading
## Describe the bug

After downloading logs from a sensor, the log interval should be reset.

This causes all logs to be reset.

if the logs are not reset, then the sensor will likely reach the max amount of logs (`~20k`) and then ever time we download, we'll download and parse 20k logs each download which is not ideal.

### To reproduce

1. Download some logs
2. Check the number of logs via blue maestro for example

### Expected behaviour

Should send a log interval set command after downloading logs.

### Proposed Solution

### Version and device info

- App version:
- Tablet model:
- OS version:

wlthomson:
Fixes #3758.

## Change summary

Resets sensor logs after saving them locally to device.

## Testing

- [ ] On sync, device log count is reset to `0` (view using Tempo Plus 2 app).

N/A.

### DIFF
--- a/src/actions/Bluetooth/SensorDownloadActions.js
+++ b/src/actions/Bluetooth/SensorDownloadActions.js
@@ -130,7 +130,11 @@ const downloadLogsFromSensor = sensor => async dispatch => {
 
             await TemperatureLogManager().saveLogs(temperatureLogs);
 
-            await BleService().updateLogInterval(macAddress, logInterval);
+            await BleService().updateLogIntervalWithRetries(
+              macAddress,
+              logInterval,
+              VACCINE_CONSTANTS.MAX_BLUETOOTH_COMMAND_ATTEMPTS
+            );
 
             await BleService().clearLogs(macAddress);
 
--- a/src/actions/Bluetooth/SensorDownloadActions.js
+++ b/src/actions/Bluetooth/SensorDownloadActions.js
@@ -129,6 +129,9 @@ const downloadLogsFromSensor = sensor => async dispatch => {
             );
 
             await TemperatureLogManager().saveLogs(temperatureLogs);
+
+            await BleService().updateLogInterval(macAddress, logInterval);
+
             await dispatch(BreachActions.createConsecutiveBreaches(sensor));
           } catch (e) {
             dispatch(sensorDownloadError(sensor, DOWNLOADING_ERROR_CODES.E_CANT_CONNECT));
--- a/src/actions/Bluetooth/SensorDownloadActions.js
+++ b/src/actions/Bluetooth/SensorDownloadActions.js
@@ -130,6 +130,8 @@ const downloadLogsFromSensor = sensor => async dispatch => {
 
             await TemperatureLogManager().saveLogs(temperatureLogs);
 
+            await BleService().updateLogInterval(macAddress, logInterval);
+
             await BleService().clearLogs(macAddress);
 
             await dispatch(BreachActions.createConsecutiveBreaches(sensor));
--- a/src/actions/Bluetooth/SensorDownloadActions.js
+++ b/src/actions/Bluetooth/SensorDownloadActions.js
@@ -130,7 +130,7 @@ const downloadLogsFromSensor = sensor => async dispatch => {
 
             await TemperatureLogManager().saveLogs(temperatureLogs);
 
-            await BleService().updateLogInterval(macAddress, logInterval);
+            await BleService().clearLogs(macAddress);
 
             await dispatch(BreachActions.createConsecutiveBreaches(sensor));
           } catch (e) {
--- a/src/bluetooth/BleService.js
+++ b/src/bluetooth/BleService.js
@@ -194,6 +194,23 @@ class BleService {
     return monitor;
   };
 
+  /** Facade for clearing logs.
+   *
+   * Connects with a sensor and clears all temperature logs.
+   *
+   * Returns a promise which resolves to string 'ok'.
+   *
+   * @param {String} macAddress
+   */
+  clearLogs = async macAddress => {
+    await this.connectAndDiscoverServices(macAddress);
+    return this.writeWithSingleResponse(
+      macAddress,
+      `${BLUETOOTH.COMMANDS.CLEAR}`,
+      stringFromBase64
+    );
+  };
+
   /**
    * Facade for downloading logs.
    *
--- a/src/bluetooth/constants.js
+++ b/src/bluetooth/constants.js
@@ -15,6 +15,7 @@ export const BLUETOOTH = {
   },
   COMMANDS: {
     BLINK: '*blink',
+    CLEAR: '*clr',
     DOWNLOAD: '*logall',
     INFO: '*info',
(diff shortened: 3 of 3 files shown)


wlthomson:
Update:

Did some testing with @katherine-sussol. Changing the log interval doesn't seem to be resetting the logs (tested on the Tempo Plus 2 app). Have tested the `*clr` command which seems to work more consistently. Will add a bluetooth constant + method to `BleManager`.

joshxg:
the docs here: https://drive.google.com/drive/u/2/folders/1AlMQEGAhVio2q8ouPYK4IKlTI8ZQhDu1

image

and also anecdotally, since I have been workin with the sensors, they have always deleted all logs and when changing the log interval (one is recorded straight after changing, which might be what you are seeing?)

katherine-sussol:
It was being super weird and inconsistent, with some of the blue sensors (and I think one red one) - updating the log interval (using `*lintX`) was resetting the logs, but testing on several of the red ones, the logs were just not clearing. 
E.g. my one (`FD9AC5C9`) had over 5000 logs and no matter what the log interval was set to after, kept all of its logs.

Is it possible there are firmware updates or something that we need to do on the sensors?

joshxg:
Weird. The red ones definitely have a bug in the firmware where sending an *info can sometimes reset the log interval to 1 hour 🤣 but I haven't experienced what you're talking about - there could definitely be a bug in the sensor.

I am unsure if we should code around their bugs 🤔 we already are going to recommend blue ones

wlthomson:
Putting aside the question of inconsistencies with `*lint`, is there a reason for erasing via changing the interval rather than using the clear command? Since a command exists for it, it seems like that would be intended usage (rather than clearing via a side-effect of changing the interval)? I might be missing something, but don't follow why doing it this way would mean we are coding around bugs. Isn't this way just using the command designed for this job?

joshxg:
Ensuring the log interval is set correctly is important. Having the log interval updated when it's safe to ensure its set correctly is good to do. Sending a bluetooth command is not free, more commands is more chances to fail.

Two birds one stone?

wlthomson:
> Ensuring the log interval is set correctly is important. Having the log interval updated when it's safe to ensure its set correctly is good to do.

Hm... but that seems contradictory to not wanting to code around bugs in the sensors? As I understand it, we have no way of handling if the device log interval deviates from that stored on the tablet? As we're not reading the log interval on the sensor before resetting it, I'm not sure how resetting it after every download would help? We'd have no way of ensuring the correctness of the data, anyway? Maybe missing something, but it seems like this would be more hiding or obfuscating the appearance of a problem rather than solving it?

> Sending a bluetooth command is not free, more commands is more chances to fail.

I don't understand why there are more commands? Either way you are sending one bluetooth command? Or are you suggesting resetting the log interval and clearing the logs? As above, I'm missing parts of the reasoning as to why we want to reset the log intervals after each download, but assuming it is something we need, I would still argue that explicitly using two commands means fewer chances of failure compared to relying on the side-effects of a single command which doesn't seem to behave consistently across devices.

> Two birds one stone?

@DhanyaHerath 

image

joshxg:
> Hm... but that seems contradictory to not wanting to code around bugs in the sensors?
 
I mentioned the bug in the firmware as precedent (is this right??) to firmware bugs, not as the primary reason we should always reset the log interval after downloading, think this has caused confusion, since this primary reason I put in the issue is for clearing the logs which your solution would solve better and now i've introduced this new thing that the log interval actually should be reset, not only because it resets the logs- sorry!

>  As we're not reading the log interval on the sensor before resetting it, I'm not sure how resetting it after every download would help?

It is hiding a problem- the primary problem not being bugs, but other sources changing the log interval

> We'd have no way of ensuring the correctness of the data, anyway?

You're right- I think it's an acceptable error level. We have a max log interval of 10 minutes. If the log interval is changed within that time, there might be a log that is ~9mins off, but it will realign after that one log.

> Either way you are sending one bluetooth command? Or are you suggesting resetting the log interval and clearing the logs? 

I am saying it is important we constantly realign the log interval with our persistent storage- the easiest way to do this is to send a command to reset the log interval after downloading.

> I'm missing parts of the reasoning as to why we want to reset the log intervals after each download, but assuming it is something we need

Yeah, sorry. It's because we need to sync up data stored in two places, where one of those places can be changed from outside sources.

There are a lot of other methods to do this (polling etc) but the easiest and quickest way is to just reset it after downloading, when it is safe to do so. I'm not against doing a more robust solution if you can think of one in the future

> I would still argue that explicitly using two commands

We can agree to disagree 👍 since this isn't exclusive of changing the log interval, it's fine by me 😁

wlthomson:
To be safe, going to go with explicitly setting with two commands. After more testing, I'm having red and blue devices not consistently resetting logs when the log interval is updated (maybe a firmware issue, I'm not sure). The log interval always updates, but the logs are only sometimes reset. The logs tend to reset when the interval is changed (if it's kept the same, they rarely seem to update for blue or red sensors)... this would still require two commands, so isn't really a solution. This is all in contrast to the `*clr` command which is always resetting the logs.

Although sending two commands is more expensive than one, I think it's better to err on the side of consistency. Arguably we don't need the logs to always clear, as we only need to keep them within a reasonable size. However, it seems like some devices never (or at least rarely enough that neither me nor @katherine-sussol have observed it) wipe logs when the interval is changed, so it's probably not a good idea to rely on the logs eventually wiping or wiping with enough frequency to be acceptable. 

TLDR; changing the log interval always updates the interval but doesn't reliably wipe the logs, clearing the logs always reliably wipes the logs. There is a cost to sending two commands, but it's an OK trade-off versus the potential cost of relying on one command with an inconsistent side-effect.

Note: all the above can be replicated using the Tempo Plus 2 app.

joshxg:
Weird as, but LGTM

p.s. these small inconsistencies with blue maestro (or.. firmware bugs) is what got us started looking at other sensors - i.e. the laird sensors

wlthomson:
@katherine-sussol Ready for re-review 😄 .

src/actions/Bluetooth/SensorDownloadActions.js
```diff
@@ -129,6 +129,11 @@ const downloadLogsFromSensor = sensor => async dispatch => {
             );
 
             await TemperatureLogManager().saveLogs(temperatureLogs);
+
+            await BleService().updateLogInterval(macAddress, logInterval);
```
katherine-sussol, line 133: Possibly worth including some retries here, if it's quite an important thing to have set? Less important for the log clearing probably since we don't really care if that fails now and again.

katherine-sussol (approved):
LGTM. A bit unfortunate to need so many BT commands going out on every download (think it's now `*info`, `*logall`, `*lintX` and `*clr` 🙀 ) but based on testing agree it's the most reliable.

src/actions/Bluetooth/SensorDownloadActions.js
wlthomson: 2989b51.

PR #3769, opened 2021-03-15
#3759 temperature log calculation inaccuracy

### LINKED ISSUE #3759: TemperatureLogs are not being saved by counting backward
## Describe the bug

`TemperatureLog` records are being saved counted from the most recent log

If, for example, the battery runs out of battery for a while, then when the sensor comes on, we would start saving logs from when the battery ran out of battery.

We should be saving logs by counting *back* from the time now, so when the first download occurs after a sensors battery is replaced, we have a 'gap' in the logs from when the sensor was off.

This will be the same for when a sensor is paused.

### To reproduce

1. Go to a sensor and pause it
2. wait for awhile
3. start downloading logs again.
4. View the logs (email yourself a csv, it's easier)
5. See error

### Expected behaviour

Logs when downloaded should be counted backward.

This test case isn't good: https://github.com/openmsupply/mobile/blob/e5897abcb85870e70cf43645a097011ae48169bd/src/bluetooth/TemperatureLogManager.test.js#L161-L185

This is only true because `timeNow = 600`, `logInterval=300` and `mostRecentLog=0`. The log saved should be `300`, because that is `300` less than `timeNow`. But it is correct and `300` because it is also `300` more than the most recent log.

### Proposed Solution

Change `createLogs` to count back from the `timeNow` like it is already doing when there is no `mostRecentLogTime`, except that the log should be aligned with the most recent log. I think something like:

The test case `Creates the correct number of logs` should be:

```
...
const maxNumberToSave = 1;
    const mostRecentLogTime = 1;
    const timeNow = 1200;

    const logs = [{ temperature: 10 }, { temperature: 11 }, { temperature: 12 }];
    const shouldBe = [
      {
        id: '1',
        temperature: 12,
        location: undefined,
        timestamp: new Date(901 * MILLISECONDS.ONE_SECOND),
        sensor,
        logInterval: 300,
      },
    ];
```

I think change to something like:

```
let initial;
    if (mostRecentLogTime == null) {
      initial = moment.unix(timeNow).subtract((logsToSave.length - 1) * logInterval, 'seconds');
    } else {
      const numberOfLogIntervalsUntilNow =
        Math.floor((timeNow - mostRecentLogTime) / logInterval) + 1;

      initial = moment.unix(
        mostRecentLogTime +
          (numberOfLogIntervalsUntilNow * logInterval - maxNumberToSave * logInterval)
      );
    }
```

### Version and device info

- App version:
- Tablet model:
- OS version:

katherine-sussol:
Fixes #3759

## Change summary

- Implemented proposed solution in the original issue (tested and works as expected)

## Testing

- [ ] Follow steps in #3759 and verify that when there is a gap in sensor downloads (either from pause button or sensor just dying) the gap is accurately reflected in the logs 

N/A

### DIFF
--- a/src/bluetooth/TemperatureLogManager.js
+++ b/src/bluetooth/TemperatureLogManager.js
@@ -64,6 +64,8 @@ class TemperatureLogManager {
       const numberOfLogIntervalsUntilNow =
         Math.floor((timeNow - mostRecentLogTime) / logInterval) + 1;
 
+      // This 'lookback' (as opposed to only counting forward) is necessary to account for
+      // potential gaps in logs (e.g. due to battery running out or sensor being paused)
       initial = moment.unix(
         mostRecentLogTime +
           (numberOfLogIntervalsUntilNow * logInterval - maxNumberToSave * logInterval)
--- a/src/bluetooth/TemperatureLogManager.js
+++ b/src/bluetooth/TemperatureLogManager.js
@@ -61,8 +61,6 @@ class TemperatureLogManager {
     if (mostRecentLogTime == null) {
       initial = moment.unix(timeNow).subtract((logsToSave.length - 1) * logInterval, 'seconds');
     } else {
-      initial = moment.unix(mostRecentLogTime).add(logInterval, 's');
-
       const numberOfLogIntervalsUntilNow =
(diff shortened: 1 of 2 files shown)


wlthomson (approved):
LGTM (subject to discussion in issue)!

PR #3790, opened 2021-03-17
#3783 Lookup patients online when vaccine dispensing

### LINKED ISSUE #3783: Vaccine dispensing step one should be able to look on the line for patients

Want to allow users to look on the line for patients from vaccine dispensing step1

- Add a footer to the flat list with a button `look on the line` which sends a request to the patient endpoint api for patients

## Implementation

There are some issues with the patient api endpoint:

- https://github.com/sussol/msupply/issues/8241
- https://github.com/sussol/msupply/issues/8267
- https://github.com/sussol/msupply/issues/8266

However, once resolved, we have code which does the fetch with params, so just reuse that, with additional pagination.

Thought maybe we should make the list of patients default to "enter search terms" instead of defaulting to a list of patients. It's very unlikely that you're going to scroll through without having searched, and it just adds rendering time for no real reason (same with the dispensing page, tbh!)

N/A

## Additional context

N/A

joshxg:
Fixes #3783 

## Change summary

- Adds online lookup to vaccine dispensing
- Lots of side effects !

## Testing

- [ ] Can vaccine dispense to an existing patient from local storage
- [ ] can vaccine dispense to a patient on the line
- [ ] can vaccine dispense to a new patient created
- [ ] Can create a new patient in normal dispensing
- [ ] Can create a patient in normal dispensing

- Normal dispensing edit/creation
- Vaccine dispensing edit/creation
- Errors in communicating with the server

### DIFF
--- a/src/actions/Entities/NameActions.js
+++ b/src/actions/Entities/NameActions.js
@@ -11,9 +11,10 @@ export const NAME_ACTIONS = {
   FILTER: 'NAME/filter',
   SORT: 'NAME/sort',
   SAVE: 'NAME/save',
+  FETCH_SUCCESS: 'NAME/fetchSuccess',
 };
 
-const createDefaultName = (type = 'patient', id) => ({
+export const createDefaultName = (type = 'patient', id) => ({
   id: id ?? generateUUID(),
   code: '',
   isCustomer: false,
@@ -25,9 +26,9 @@ const createDefaultName = (type = 'patient', id) => ({
   type,
 });
 
-const create = (type, id) => ({
+const create = name => ({
   type: NAME_ACTIONS.CREATE,
-  payload: { name: createDefaultName(type, id) },
+  payload: { name },
 });
 
 const reset = () => ({
@@ -42,13 +43,12 @@ const update = (id, field, value) => ({
 const updatePatient = detailsEntered => (dispatch, getState) => {
   const currentPatient = selectEditingName(getState());
   const { id } = currentPatient || {};
-  const currentKeys = Object.keys(currentPatient);
 
   if (!id) return;
 
   batch(() => {
     Object.entries(detailsEntered).forEach(([key, value]) => {
-      if (currentKeys.includes(key) && currentPatient[key] !== value) {
+      if (currentPatient[key] !== value) {
         dispatch(update(id, key, value));
       }
     });
--- a/src/actions/Entities/NameNoteActions.js
+++ b/src/actions/Entities/NameNoteActions.js
@@ -1,5 +1,6 @@
 import Ajv from 'ajv';
 import { generateUUID } from 'react-native-database';
+import moment from '../../../node_modules/moment/moment';
 import { createRecord, UIDatabase } from '../../database/index';
 import { selectCreatingNameNote } from '../../selectors/Entities/nameNote';
 import { selectSurveySchemas } from '../../selectors/formSchema';
@@ -43,12 +44,29 @@ const createDefaultNameNote = (nameID = '') => {
   };
 };
 
-const createSurveyNameNote = nameID => (dispatch, getState) => {
-  const name = UIDatabase.get('Name', nameID);
+const getMostRecentPCD = patient => {
+  const [pcdEvent] = UIDatabase.objects('PCDEvents');
+  if (!pcdEvent) return null;
 
+  const { id: pcdEventID } = pcdEvent;
+  const { nameNotes = [] } = patient ?? {};
+  if (!nameNotes.length) return null;
+
+  const filtered = nameNotes.filter(({ patientEventID }) => patientEventID === pcdEventID);
+  if (!filtered.length) return null;
+
+  const sorted = filtered.sort(
+    ({ entryDate: entryDateA }, { entryDate: entryDateB }) =>
+      moment(entryDateB).valueOf() - moment(entryDateA).valueOf()
+  );
+  const [mostRecentPCD] = sorted;
+  return mostRecentPCD;
+};
+
+const createSurveyNameNote = patient => (dispatch, getState) => {
   // Create the seed PCD, which is either their most recently filled survey or
   // and empty object.
-  const seedPCD = name?.mostRecentPCD ?? createDefaultNameNote(nameID);
+  const seedPCD = getMostRecentPCD(patient) ?? createDefaultNameNote(patient.id);
 
   // Get the survey schema as we need an initial validation to determine if
   // the seed has any fields which are required to be filled.
--- a/src/actions/Entities/NameNoteActions.js
+++ b/src/actions/Entities/NameNoteActions.js
@@ -1,6 +1,6 @@
 import Ajv from 'ajv';
 import { generateUUID } from 'react-native-database';
-import moment from '../../../node_modules/moment/moment';
+import moment from 'moment';
 import { createRecord, UIDatabase } from '../../database/index';
 import { selectCreatingNameNote } from '../../selectors/Entities/nameNote';
 import { selectSurveySchemas } from '../../selectors/formSchema';
--- a/src/actions/PatientActions.js
+++ b/src/actions/PatientActions.js
@@ -18,7 +18,7 @@ export const PATIENT_ACTIONS = {
 };
 
 const closeModal = () => ({ type: PATIENT_ACTIONS.COMPLETE });
-const createPatient = () => ({ type: PATIENT_ACTIONS.PATIENT_CREATION });
+const createPatient = patient => ({ type: PATIENT_ACTIONS.PATIENT_CREATION, payload: { patient } });
 
 const editPatient = patient => ({
   type: PATIENT_ACTIONS.PATIENT_EDIT,
--- a/src/database/DataTypes/Name.js
+++ b/src/database/DataTypes/Name.js
@@ -161,6 +161,7 @@ export class Name extends Realm.Object {
       thisStoresPatient: this.thisStoresPatient,
       nationality: this.nationality?.toJSON(),
       ethnicity: this.ethnicity?.toJSON(),
+      nameNotes: this.nameNotes?.map(nameNote => nameNote.toObject()),
     };
   }
 }
--- a/src/database/DataTypes/NameNote.js
+++ b/src/database/DataTypes/NameNote.js
@@ -15,6 +15,10 @@ export class NameNote extends Realm.Object {
     this._data = JSON.stringify(newValue);
   }
 
+  get patientEventID() {
+    return this.patientEvent?.id;
+  }
+
   toObject() {
     return {
       id: this.id,
--- a/src/hooks/useLocalAndRemotePatients.js
+++ b/src/hooks/useLocalAndRemotePatients.js
@@ -123,22 +123,20 @@ export const useLocalAndRemotePatients = (initialValue = []) => {
     if (dateOfBirth) {
       const dob = moment(dateOfBirth, DATE_FORMAT.DD_MM_YYYY, null, true);
       if (!dob.isValid()) {
-        return patients.sorted('lastName');
+        return patients;
       }
 
       const dayOfDOB = dob.startOf('day').toDate();
       const dayAfterDOB = dob.endOf('day').toDate();
 
-      patients = patients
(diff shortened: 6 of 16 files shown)


src/actions/Entities/NameActions.js
```diff
@@ -11,9 +11,10 @@ export const NAME_ACTIONS = {
   FILTER: 'NAME/filter',
   SORT: 'NAME/sort',
   SAVE: 'NAME/save',
+  FETCH_SUCCESS: 'NAME/fetchSuccess',
 };
 
-const createDefaultName = (type = 'patient', id) => ({
+export const createDefaultName = (type = 'patient', id) => ({
```
joshxg, line 17: rather than passing an id in and creating one here, started to create a default one in components and pass the patient.

This is because I would be passing a patient when selecting one from having been lookedup.

Just occurs to me I mightv'e been able to just quickly save the patient when selecting and made a lot less changes 🤔

src/actions/Entities/NameActions.js
```diff
@@ -42,13 +43,12 @@ const update = (id, field, value) => ({
...
   const currentPatient = selectEditingName(getState());
   const { id } = currentPatient || {};
-  const currentKeys = Object.keys(currentPatient);
 
   if (!id) return;
 
   batch(() => {
     Object.entries(detailsEntered).forEach(([key, value]) => {
-      if (currentKeys.includes(key) && currentPatient[key] !== value) {
+      if (currentPatient[key] !== value) {
```
joshxg, line 51: wasn't sure what this was really trying to do - but it stopped me from updating when for example I started with a 'default name' record and went to edit values like gender which weren't in the default name yet

src/actions/Entities/NameNoteActions.js
```diff
@@ -43,12 +44,29 @@ const createDefaultNameNote = (nameID = '') => {
...
+  const filtered = nameNotes.filter(({ patientEventID }) => patientEventID === pcdEventID);
+  if (!filtered.length) return null;
+
+  const sorted = filtered.sort(
+    ({ entryDate: entryDateA }, { entryDate: entryDateB }) =>
+      moment(entryDateB).valueOf() - moment(entryDateA).valueOf()
+  );
+  const [mostRecentPCD] = sorted;
+  return mostRecentPCD;
+};
```
joshxg, line 64: This is changed such that it can look up the most recent name note for an existing patient or one from the lookup API.

This will need this mSupply PR https://github.com/sussol/msupply/pull/8280 to work- I haven't tested it with it yet

src/actions/PatientActions.js
```diff
@@ -18,7 +18,7 @@ export const PATIENT_ACTIONS = {
 };
 
 const closeModal = () => ({ type: PATIENT_ACTIONS.COMPLETE });
-const createPatient = () => ({ type: PATIENT_ACTIONS.PATIENT_CREATION });
+const createPatient = patient => ({ type: PATIENT_ACTIONS.PATIENT_CREATION, payload: { patient } });
```
joshxg, line 21: This is for creating a patient in normal dispensing (because I guess I thought it was a good idea to have name actions and patient actions 🤔 )

This is setting a 'default name' as the one being edited now rather than sorting it out when saving, so that there is a stable ID to use for the name note.

src/database/DataTypes/Name.js
```diff
@@ -161,6 +161,7 @@ export class Name extends Realm.Object {
       thisStoresPatient: this.thisStoresPatient,
       nationality: this.nationality?.toJSON(),
       ethnicity: this.ethnicity?.toJSON(),
+      nameNotes: this.nameNotes?.map(nameNote => nameNote.toObject()),
```
joshxg, line 164: This is so the names from realm convert to an object which more closely resembles the ones sent from the patient lookup api

src/database/DataTypes/NameNote.js
```diff
@@ -15,6 +15,10 @@ export class NameNote extends Realm.Object {
     this._data = JSON.stringify(newValue);
   }
 
+  get patientEventID() {
+    return this.patientEvent?.id;
+  }
```
joshxg, line 20: Same as above- so the name note is more similar to those which will be sent from the api

src/hooks/useLocalAndRemotePatients.js
```diff
@@ -0,0 +1,153 @@
+import { useEffect, useReducer } from 'react';
```
joshxg, line 1: Aaaand this is where it gets a bit hairy.

I started off using thunks and redux to do the API calls.

It was getting pretty messy, knowing when data was coming from APIs or from realm, to know when to filter. It was also hard to track loading/error states through redux.

I thought to use the `useFetch` hook which handles most of this already.

What I ended up with was this hook which is an attempt to create an interface where data from two different sources are kind of merged to form a single source of state.

There will also need to be pagination logic added to this.

It's a bit complex but I think it is less so then mixing hooks and different redux stuff going on

src/pages/DispensingPage.js
```diff
@@ -349,10 +350,16 @@ const mapDispatchToProps = dispatch => ({
 
   editPatient: patient =>
     batch(() => {
-      dispatch(NameNoteActions.createSurveyNameNote(patient));
+      dispatch(NameNoteActions.createSurveyNameNote(UIDatabase.get('Name', patient)));
```
joshxg: `patient` was actually `patientID`- and `createSurveyNameNote` was changed to accept a patient, not the ID

src/pages/DispensingPage.js
```diff
@@ -349,10 +350,16 @@ const mapDispatchToProps = dispatch => ({
...
       dispatch(PatientActions.editPatient(UIDatabase.get('Name', patient)));
     }),
-  createPatient: () => dispatch(PatientActions.createPatient()),
+  createPatient: () =>
+    batch(() => {
+      const patient = createDefaultName('patient', generateUUID());
+      dispatch(PatientActions.createPatient(patient));
+      dispatch(NameNoteActions.createSurveyNameNote(patient));
```
joshxg, line 361: Creating these now with a stable patient ref

src/reducers/Entities/NameReducer.js
```diff
@@ -42,7 +42,11 @@ export const NameReducer = (state = initialState(), action) => {
       const { payload } = action;
       const { name } = payload;
 
-      return { ...state, editing: name.toJSON() };
+      // A name can be either a realm object, or a plain object if looked up through the
+      // patient API.
+      const asPlainObject = name.toJSON ? name.toJSON() : name;
+
+      return { ...state, editing: asPlainObject };
```
joshxg, line 49: this can now either be from the api or from realm

src/widgets/Tabs/PatientSelect.js
```diff
@@ -54,22 +61,78 @@ import { Paper } from '../Paper';
...
+
   const columns = React.useMemo(() => getColumns(MODALS.PATIENT_LOOKUP), []);
   const { pageTopViewContainer } = globalStyles;
   const keyboardIsOpen = useKeyboardIsOpen();
-  const debouncedFilter = useDebounce(onFilterData, 300);
+
   const handleUpdate = (key, value) => {
     updateForm(key, value);
-    debouncedFilter(key, value);
+    filter({ ...completedForm, [key]: value });
```
joshxg, line 126: need to do this here as completedForm is not correct until the above `updateForm` has been dispatched

src/widgets/modalChildren/PatientEditModal.js
```diff
@@ -22,62 +21,53 @@ export const PatientEditModalComponent = ({
...
-            textStyle={styles.saveButtonTextStyle}
-            text={generalStrings.save}
-          />
-          <PageButton
-            onPress={onCancel}
-            style={styles.cancelButton}
-            textStyle={styles.cancelButtonTextStyle}
-            text={modalStrings.cancel}
-          />
+}) => (
```
joshxg, line 26: don't need the `onOpen` now, because it is creating the required records in `DispensingPage` callbacks

src/widgets/modalChildren/PatientEditModal.js
```diff
@@ -123,26 +113,24 @@ PatientEditModalComponent.propTypes = {
...
   const { onOpen, onSave, onSaveSurvey, ...otherDispatchProps } = dispatchProps;
-  const { patient, surveySchema } = ownProps;
-  const { id = generateUUID() } = patient ?? {};
+  const { surveySchema } = ownProps;
+
   const onSaveForm = () => {
-    onSave({ id, ...completedForm });
-    if (surveySchema) onSaveSurvey(id);
+    onSave(completedForm);
+    if (surveySchema) onSaveSurvey();
```
joshxg, line 126: don't need to use these ids now because `dispensingPage` callbacks are creating a default `patient` which has an `id`.

mark-prins:
@joshxg If I enter a DOB on the search form, then I get the response 'there was a problem communicating with the server'

mark-prins:
Found another problem - searching using DOB doesn't work: enter a DOB and I'm told 'Enter the patients details...'
Entering a first name and DOB will find no patients.

mark-prins (approved):
👍 
Well, that's awesome. Genius to use a reducer in a hook to handle it all.
There's a couple of issues with DOBs - can push on an merge this one and fix those in another issue if that's helpful though.

I'm slightly confused about the merging of results.. the reducer is setting rather than merging. The API lookup returns all patients, so there isn't a merge, just a list from the server. some of them exist locally and some don't 🤷‍♂️

src/localization/generalStrings.json
```diff
@@ -1,5 +1,8 @@
 {
   "gb": {
+    "enter_patient_details": "Enter the patients details to start searching for them!",
```
mark-prins, line 3: :nit 
```suggestion
    "enter_patient_details": "Enter the patient's details to start searching for them!",
```
The details belong to the patient, so there should be an apostrophe, I think.

src/database/DataTypes/Name.js
```diff
@@ -161,6 +161,7 @@ export class Name extends Realm.Object {
       thisStoresPatient: this.thisStoresPatient,
       nationality: this.nationality?.toJSON(),
```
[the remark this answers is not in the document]
joshxg: Oh it's on the realm class which each of our models extends.

I should probably rename all our overrides to `toObject` like I had planned but keep forgetting!

joshxg:
- Changed all the strings: 19f3749, 78908919dc47e6059185b6e16109b76b0d08fda6
- Added sorting: c5b73e96c47c4feb676fe329561ac1bf62126dba ... e412e59a3ceab6237b48b970de7b0772cfc0af14

- Re DOB- 67472940a195e7b08264471b982b3f02a234917b - I was so hung up on handling errors when the server wasn't there I didn't handle them correctly when it is there! I hadn't realised that when there are no results, the response from `useFetch` was `null` and there was also an error set. I don't particularly like using the error message as it's pretty error prone, but changing the behaviour for not throwing an error when the result set is empty is a bit messy, so just went with this...  Also, when the DOB was invalid an action wasn't properly dispatched. Hopefully adding that all together it has more consistent behaviour now!
- Re 'merge' : wrong word, I think. I am trying to describe the 'funnelling' of two different 'things' going on- local state and actions with remote state/data and the async actions being 'funnelled' into a single 'state', where just reading 'loading' of this state will tell you if the local db is searching or the async request is being sent and you don't have to read from two different arrays for patients, dependent on whether the last request was through the api or local storage. I think i'm making it seem more confusing than it is...

joshxg:
Going to merge this now.

I did notice a couple other things which I will make new issues for:
- Disable search online when in the middle of another search
- The debouncing in `FormControl` is annoying
- Update comments to remove 'merge'?
-

mark-prins:
thanks - I get it now :) I had in mind that you'd have a single array of objects, some from the api and some from realm; which would be awful. On re-reading the comments, they're fine. No need to change those.
Have to confess that I didn't even notice the button not disabling while searching - that's usually a thing that bugs me. The search was quick ( all running locally 😉 ) and everything was so smooth - didn't even think to click the button twice.

joshxg:
Ha, yeah- I think mine was just running a bit slower because I was accidentally running on mobile data 🤣

PR #3792, opened 2021-03-17
#3727 Remove setting of state of unmounted components

### LINKED ISSUE #3727: AfterInteractions is trying to update an unmounted component
## Describe the bug

After having added that after interactions attempts to update based on focus rather than just on mounting, the component now tries to update state when poping a component off the stack (causing it to unmount)

### To reproduce

1. Go to the end of vaccinations and issue a vaccine

### Expected behaviour

No state update on unmounted component

### Proposed Solution

Use an `mounted` ref which can be used to conditionally update state, accept you aren't able to conditionally update state so I am not sure 🤔 

### Version and device info

- App version: 8.0 devleopment
- Tablet model:
- OS version:

### Additional context

N/A

joshxg:
Fixes #3727 

## Change summary

- Fixes `AfterInteractions` to not set state on an unmounted component
- Fixes `LoginModal` to not `setState` on an unmounted component by removing `onLogin` which I have no idea what it was doing 🤔 
- Fixes `useKeyboardIsOpen` to not set state on unmounted components

## Testing

Internal

N/a

### DIFF
--- a/src/hooks/index.js
+++ b/src/hooks/index.js
@@ -21,3 +21,6 @@ export { useLoadingIndicator } from './useLoadingIndicator';
 export { useToggle } from './useToggle';
 export { useIntervalReRender } from './useIntervalReRender';
 export { useConditionalAnimationRef } from './useConditionalAnimationRef';
+export { useIsMounted } from './useIsMounted';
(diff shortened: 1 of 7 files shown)


mark-prins (approved):
very nice 😁

PR #3793, opened 2021-03-17
#3767 Remove isSyncing prop from root container

### LINKED ISSUE #3767: mSupplyMobileAppContainer re-renders whenever sync occurs
## Describe the bug

- mSupplyMobileAppContainer re-renders whenever sync occurs

### To reproduce

Internal bug

### Proposed Solution

Check and see why it's re-rendering

### Version and device info

- App version: 1.0+
- Tablet model:
- OS version:

### Additional context

N/A

joshxg:
Fixes #3767 

## Change summary

- The last prop to remove to stop the root re-rendering constantly during sync woo (this has been like 2 years of removing props from this component 🤣 
- It's use to stop the synchronise func being executed - but it can't be executed if not initialised and the button is disabled if already syncing so it's redundant

## Testing

Internal

N?A

### DIFF
--- a/src/mSupplyMobileApp.js
+++ b/src/mSupplyMobileApp.js
@@ -26,7 +26,6 @@ import { SyncAuthenticator, UserAuthenticator } from './authentication';
 
 import { LoadingIndicatorContext } from './context/LoadingIndicatorContext';
 import { selectTitle } from './selectors/supplierCredit';
-import { selectIsSyncing } from './selectors/sync';
 import { selectCurrentUser } from './selectors/user';
 import { selectUsingVaccines } from './selectors/modules';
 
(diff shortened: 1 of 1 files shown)


mark-prins (approved):
👍 

It is possible for the scheduler to call the synchroniser twice now though isn't it? - if someone sets the sync interval to something very low - I wouldn't expect sync to take longer than 10 minutes, so currently it isn't a problem.

joshxg:
Yeah you're right and also right about it not being a problem at the moment.

(also the server has a semaphore per site for a sync so will throw an exception if it did take mroe than 10 minutes).

Will try to think of something- at the moment thinking of lifting the whole synchroniser out 🤔

PR #3820, opened 2021-03-20
#3819 Remove doses from batch records

### LINKED ISSUE #3819: Remove doses field
## The suggestion

We have changed the way we calculate open vial wastage to only use inventory adjustments which have an `openVialWastage` reason.

This was the primary reason for having a `doses` field. 

**Why?**

Now, we only allow `doses` to equal `quantity*number of doses per vial`, so we don't need to store the number of doses.

Leave it

## Additional context

- By removing this field, it is much better to calculate things. For example, if you change an item to be a vaccine and to have doses, then we need to figure out what to do with historic stuff. For example, do item batches now have a number of doses? So should we go and update all the item batches? Transaction batches? etc

joshxg:
Fixes #3819 

## Change summary

- Removes doses field

## Testing

- [ ] Issuing stock by doses in customer invoices correct decreases the item batch quantity
- [ ] stock changes in stocktakes correctly decreases/increases the item batch quantity
- [ ] Issuing stock in vaccine dispensing correctly decreases the item batch quantity

N/a


mark-prins:
If I try to create a CI, I'm told `This invoice has items with an invalid amount of doses!`
Comes back to the transaction `hasValidDoses > TransactionItem.hasValidDoses > TransactionBatch.hasValidDoses (removed)`

mark-prins:
I'm getting some strangeness with the stocktake quantity entry, which may not be related. Pressing backspace makes the number jump around.

https://user-images.githubusercontent.com/9192912/111936800-4f95bd00-8b2b-11eb-8701-6399dfad95aa.mov

joshxg:
Removed the remaining doses stuff and added a toFixed around the `StocktakeItem` - seemed to be just funny double precision stuff

mark-prins (approved):
Cool - all working nicely thanks

PR #3825, opened 2021-03-21
#3710 Create an NSJ on the server when using a remote patient

### LINKED ISSUE #3710: When using a patient from the patient API also send a request to create a name store join

- When we look up a patient, there is no visibility created. This means you will not receive any updates to the record, receive new but related data or receive the patient when initialising (arguably good)

- Create a `name_store_join` for a patient when looked up from the patient api

## Implementation

- Create a `name_store_join` for a patient when looked up from the patient api

Not doing this.

## Additional context

I think this is a good idea. I think if we want to limit init cost, we could potentially delete the nsj's on the server which aren't related to a patient whos home store is the mobile store?

joshxg:
Fixes #3710 

## Change summary

- Note: Requires the mSupply branch https://github.com/sussol/msupply/pull/8286

## Testing

- [ ] When selecting a remote patient to vaccine dispense to, a name store join is created on the server (if you have the internet)
- [ ] When selecting a remote patient to normal dispense to, a name store join is created on the server (if you have the internet)

### DIFF
--- a/src/actions/Entities/NameActions.js
+++ b/src/actions/Entities/NameActions.js
@@ -1,7 +1,10 @@
 import { generateUUID } from 'react-native-database';
 import { batch } from 'react-redux';
+import { ToastAndroid } from 'react-native';
 import { selectEditingName } from '../../selectors/Entities/name';
 import { createRecord, UIDatabase } from '../../database';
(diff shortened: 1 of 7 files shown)


mark-prins (approved):
Looks good! that works for me.

PR #3826, opened 2021-03-21
#3823 Pagination of patients in vaccine dispensing

### LINKED ISSUE #3823: Handle pagination in patient lookups

The patient api now handles pagination, we should implement this in mobile

In the patient lookup API and vaccination pages, implement search pagination

## Implementation

Pretty standard pagination requests- probably sending an additional request on scroll?

Not doing anything

## Additional context

N/A

joshxg:
Fixes #3823 

## Change summary

- This is pretty unnecessary, I think. Practically, the user is surely not going to be scroll through 100's of patients hoping to find the one, they would surely add more search details.. but here it is!
- It's not perfect- technically if you unmount in the middle of the fetch and there's no timeout for the request. This functionality is built in to `useFetch` but not really useable. I don't think we really have time to be making it perfect..
- I didn't add it to normal dispensing. Kinda don't want to 🤣 You get 50 patients.. surely you can just refine your search and it'll be OK for Tonga at least..

## Testing

- [ ] Have heaps of patients and do a lookup- you get 50 patients and then when you scroll down it does another lookup and gets you another 50

N/A


mark-prins:
Just noticed that earlier comments were pending..

Testing this morning, changed the limit to `5` so I didn't have to create lots of patients.
It returned 5 patients, said that it was searching and then..

```
[Wed Mar 24 2021 11:34:57.630]  WARN     Possible Unhandled Promise Rejection (id: 1):
Error: No records found
processResponse@http://localhost:8081/index.bundle?platform=android&dev=true&minify=false:173118:40
processPatientResponse@http://localhost:8081/index.bundle?platform=android&dev=true&minify=false:173136:33
getMorePatients$
tryCatch@http://localhost:8081/index.bundle?platform=android&dev=true&minify=false:25082:23
```

src/widgets/Tabs/PatientSelect.js
```diff
@@ -79,6 +79,18 @@ const Header = ({ onSearchOnline, onNewPatient }) => (
   </FlexRow>
 );
 
+const GettingMore = ({ gettingMore }) =>
+  gettingMore ? (
+    <FlexRow flex={1} justifyContent="center" alignItems="space-between">
```
mark-prins, line 84:
```suggestion
    <FlexRow flex={1} justifyContent="center" alignItems="space-between" style={localStyles.more}>
```

src/hooks/useLocalAndRemotePatients.js
```diff
@@ -99,15 +128,44 @@ export const useLocalAndRemotePatients = (initialValue = []) => {
...
+      `${getServerURL()}${getPatientRequestUrl(paramsWithLimits)}`,
+      {
+        headers: { authorization: getAuthorizationHeader() },
+      }
+    );
+
+    dispatch({
+      type: 'add_more_patients',
+      payload: {
+        data: processPatientResponse({ ...getMoreResponse, json: await getMoreResponse.json() }),
```
mark-prins, line 161: `processPatientResponse` throws an error if there are no results.. which happens when we hit the end of the patient list. 
I think it should dispatch an action that keeps the current list of patients and sets `gettingMore: false`

joshxg:
Added styling here: 
3f0640f

Messing around with the state a bit more here: 
a1ae6b3

Still not perfect..  I really prefer to use https://react-query.tanstack.com/ for fetching stuff like this as it sure does have so many cases you need to account for but I guess with it just being a single area to do a request its OK.. it's already so confusing, totally going to add it if theres anything else needing to be added 🤣

src/hooks/useLocalAndRemotePatients.js
```diff
@@ -62,13 +97,14 @@ const reducer = (state, action) => {
  * having to track multiple.
  */
 export const useLocalAndRemotePatients = (initialValue = []) => {
-  const [{ data, loading, searchedWithNoResults, error }, dispatch] = useReducer(
-    reducer,
-    initialValue,
-    initialState
-  );
+  const [
+    { data, loading, searchedWithNoResults, error, limit, offset, gettingMore },
```
mark-prins, line 101:
```suggestion
    { data, loading, searchedWithNoResults, error, limit, offset, gettingMore, noMore },
```

src/hooks/useLocalAndRemotePatients.js
```diff
@@ -62,13 +97,14 @@ const reducer = (state, action) => {
...
+    dispatch,
+  ] = useReducer(reducer, initialValue, initialState);
 
-  const { fetch, refresh, isLoading, response, error: fetchError } = useFetch(getServerURL());
+  const { fetch, refresh, isLoading, response, error: fetchError, noMore } = useFetch(
```
mark-prins, line 105:
```suggestion
  const { fetch, refresh, isLoading, response, error: fetchError } = useFetch(
```

mark-prins:
@joshxg cool - that's looking good

one more thing - the `noMore` should come from state, rather than fetch. the fetch is currently looping continuously

mark-prins (changes requested):
just the conflict and the change of `noMore`

joshxg:
woah! whoooops!
63b1d95

mark-prins (approved):
👍

PR #3830, opened 2021-03-22
#3824 Save NameNote records from remote patients

### LINKED ISSUE #3824: Save name notes from remote patients

name notes from remote patients aren't saved. They need to be

Not sure exactly, but there is an array of name notes sent from the server and we need to use them

## Implementation

TBD

TBD

## Additional context

- I imagine this is not going to be great

joshxg:
Fixes #3824 

## Change summary

- Saves name notes when theyre searched for and added in vaccine dispensing and normal dispensing

## Testing

- [ ] Search for a patient in vaccination dispensing which is NOT from the mobile store, but does have a previous clinical data - the clinical data is prepopulated on step 2
- [ ] Search for a patient in NORMAL dispensing lookup which is NOT from the mobile store, but does have a previous clinical data - the clinical data is prepopulated on step 2 of vaccination

This is Hard to test because you need to have a patient which has clinical data form that is not from the mobile store.

You can use the following code to create one with footrunner:

```
$nn:=ds.name_note.new()
$nn.ID:=UUID_generate
$nn.data:=data
$nn.name_ID:="SOME PATIENT ID YOU CAN FIND IN RECORD BROWSER"
$nn.patient_event_ID:="295C2B7E53264C1B98ADD57B720F3AF6"
$nn.save()
```

### DIFF
--- a/src/actions/Entities/NameNoteActions.js
+++ b/src/actions/Entities/NameNoteActions.js
@@ -109,7 +109,21 @@ const saveEditing = optionalNameID => (dispatch, getState) => {
 const createNotes = (nameNotes = []) => {
   UIDatabase.write(() => {
     nameNotes.forEach(nameNote => {
-      UIDatabase.update('NameNote', nameNote);
+      const { patientEventID, nameID } = nameNote;
+      const name = UIDatabase.get('Name', nameID);
+      const patientEvent = UIDatabase.get('PatientEvent', patientEventID);
+
+      if (name && patientEvent) {
+        const toSave = {
+          id: nameNote.id,
+          patientEvent,
+          name,
+          _data: JSON.stringify(nameNote?.data),
+          entryDate: new Date(nameNote?.entryDate),
+        };
+
+        UIDatabase.update('NameNote', toSave);
+      }
     });
   });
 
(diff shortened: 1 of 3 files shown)


src/actions/Entities/NameNoteActions.js
```diff
@@ -103,11 +106,22 @@ const saveEditing = optionalNameID => (dispatch, getState) => {
   dispatch(reset());
 };
 
+const createNotes = (nameNotes = []) => {
+  UIDatabase.write(() => {
+    nameNotes.forEach(nameNote => {
+      UIDatabase.update('NameNote', nameNote);
```
mark-prins, line 112: I had to do this for it to save notes:
```suggestion
      const { patientEventID, nameID } = nameNote;
      const name = UIDatabase.get('Name', nameID);
      const patientEvent = UIDatabase.get('PatientEvent', patientEventID);

      if (name && patientEvent) {
        const toSave = {
          id: nameNote.id,
          patientEvent,
          name,
          _data: JSON.stringify(nameNote?.data),
          entryDate: new Date(nameNote?.entryDate),
        };

        UIDatabase.update('NameNote', toSave);
      }
```

joshxg: Thanks!

mark-prins (approved):
🎉  LGTM!

PR #3836, opened 2021-03-22
#3834 Remove overriding of home store

### LINKED ISSUE #3834: Sending an updated name record from mobile updates the home store
## Describe the bug

Whenever we send a name record from a mobile store, the `supplying_store_ID` is set to the mobile store, overriding the real home store.

### To reproduce

1. Go to look up a patient whose home store is not the mobile store
2. Update the record with the store pref on
3. Sync it to the server and check the home store

### Expected behaviour

Shouldn't override home stores

### Proposed Solution

Change the line in outgoing sync utils which overrides the home store 

### Version and device info

- App version:
- Tablet model:
- OS version:

joshxg:
Fixes #3834 

## Change summary

- Don't override the home store!

## Testing

- [ ] Lookup a patient not from your store. Sync. The home store is not overriden to be the mobile store on desktop.

N/a

### DIFF
--- a/src/sync/outgoingSyncUtils.js
+++ b/src/sync/outgoingSyncUtils.js
@@ -89,7 +89,7 @@ const generateSyncData = (settings, recordType, record) => {
         date_of_birth: moment(record.dateOfBirth).format(),
         code: record.code,
         email: record.emailAddress,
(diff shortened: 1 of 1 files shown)


mark-prins (approved):
LGTM!

PR #3877, opened 2021-03-29
#3768 Remove afterinteractions from breach display

### LINKED ISSUE #3768: Clicking the breach hazard icon crashes the app [NOT in 7.0]
## Describe the bug

I added `useIsFocused` to `AfterInteractions` which needs to be nested within a navigation screen. The breach modal is not, so when it opens it crashes.

### To reproduce

1. Go to fridge detail page
2. Click on a breach hazard icon

### Expected behaviour

No 

### Proposed Solution

I think don't use `afterinteractions` in `breachmodal` 🤔 

### Version and device info

- App version: 8.0 dev
- Tablet model:
- OS version:

### Additional context

N/A

joshxg:
Fixes #3768 

## Change summary

- After interactions needs to be nested within the stack navigator, the breach modal is at the root and so isn't in a stack nav, so causes a crash. Not the best solution, but a solution..

## Testing

- [ ] Opening the breach modal does not crash the app

Also quick fix to the vaccine generated data

### DIFF
--- a/src/database/utilities/dataGenerators.js
+++ b/src/database/utilities/dataGenerators.js
@@ -1,3 +1,4 @@
+import moment from 'moment';
 import { ToastAndroid } from 'react-native';
 import BreachManager from '../../bluetooth/BreachManager';
 import { VaccineDataAccess } from '../../bluetooth/VaccineDataAccess';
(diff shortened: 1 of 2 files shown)


mark-prins (approved):
LGTM ( sorry! no breaches, so I haven't tested.. looks fine though )

PR #3879, opened 2021-03-30
#3865 Add different double handling as quantities behaviour

### LINKED ISSUE #3865: Trouble setting doses in a stocktake batch
## Describe the bug

When setting a value of `1` dose in a batch of a vaccine item which has a value for `doses` of `6` the entered value changes incorrectly. When entering `1` the value changes to `1.02`.. entering `20` changes to `19.98`

https://user-images.githubusercontent.com/9192912/112771881-794e6700-908a-11eb-9768-21077a1034a3.mov

Another issue is that when entering a value for `dose` in a batch, this value isn't shown in the stocktake; though it is saved correctly. The following sequence is for an item with a `doses` value of `10`:

stocktake

stocktake2

stocktake3

### To reproduce

1. Create a stocktake and add a vaccine item
2. Add a batch and set the doses value
3. Urgh.

### Expected behaviour

The value entered for `doses` should be the value shown

### Proposed Solution

🤷‍♂️

joshxg:
Fixes #3865 

## Change summary

- So.. just removed all the handling of doubles and only try and handle it at the interface ...??  Think this is better

## Testing

- [ ] Customer invoices doses column when entered, also adjusts the total quantity column to be doses*doses per vial
- [ ] stocktakes doses column when entered, also adjusts the total quantity column to be doses*doses per vial
- [ ] cannot see more than two decimal places for a quantity anywhere in the ui...

N/a

### DIFF
--- a/src/database/DataTypes/StocktakeBatch.js
+++ b/src/database/DataTypes/StocktakeBatch.js
@@ -212,7 +212,7 @@ export class StocktakeBatch extends Realm.Object {
   }
 
   setDoses(database, newValue) {
-    const newCountedTotalQuantity = Number(newValue / this.dosesPerVial).toFixed(2);
+    const newCountedTotalQuantity = Number(newValue / this.dosesPerVial);
     this.countedTotalQuantity = newCountedTotalQuantity;
     database.save('StocktakeBatch', this);
   }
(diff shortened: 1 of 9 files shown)


mark-prins (approved):
Nice! that removes the problem and it all seems to work for me.

mark-prins:
I did find another annoying triviality.. the value of the `doses` on the batch doesn't change the ACTUAL QUANTITY when you close the batch modal.

https://user-images.githubusercontent.com/9192912/112932279-1f2ccf00-917a-11eb-9a5a-0c3cb074e909.mov

joshxg:
Oh no, that's right. I totally forgot about that part! 🙄

joshxg:
will do another pr to fix the other bug

PR #3890, opened 2021-03-31
#3866 delete record missing types

### LINKED ISSUE #3866: Missing object types in `deleteRecord`
## Describe the bug

In the `deleteRecord` method we need to add some records into the delete record case here https://github.com/openmsupply/mobile/blob/d50346c2688f5e77a8013b090cdce6d45e68c0e6/src/database/utilities/deleteRecord.js#L26-L50 

### To reproduce

1. Go to desktop and delete a `FormSchema`
4. See error (it doesnt get deleted- you still have to fill the form in vaccination dispensing)

### Expected behaviour

- The `FormSchema` which is central data should be deleted. Technically we don't want to allow this, but we should still support it.. I guess?

### Proposed Solution

- There are a bunch missing here 🤷 Technically we should never be deleting system data and we should never receive a store data delete sync record.. so.. I dunno what to do. I guess it's consistent with the current impl to add all the tables we've added recently into this method. It's 

### Version and device info

- App version: 6.0+?
- Tablet model: 
- OS version:

### Additional context

@mark-prins whatcha reckon?

joshxg:
Fixes #3866 

## Change summary

- Means you can now delete records properly from the server (system data mainly)

## Testing

- [ ] sync with a `FormSchema`. Run the code `ds.form_schema.all().drop()` and then sync again. After syncing the second time, you do not have any form schemas (no extra patient info)

N/A

### DIFF
--- a/src/database/utilities/deleteRecord.js
+++ b/src/database/utilities/deleteRecord.js
@@ -15,40 +15,22 @@
 export const deleteRecord = (database, recordType, primaryKey, primaryKeyField = 'id') => {
   // 'delete' is a reserved word, |deleteRecord| is in the upper scope.
   const obliterate = () => {
-    const deleteResults = database
-      .objects(recordType)
-      .filtered(`${primaryKeyField} == $0`, primaryKey);
-    if (deleteResults && deleteResults.length > 0) {
-      database.delete(recordType, deleteResults[0]);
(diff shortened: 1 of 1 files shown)


mark-prins (approved):
woo! it worked :)

PR #3907, opened 2021-04-07
#3736 Adjust numberOfUsageDays calculation

### LINKED ISSUE #3736: Monthly usage is too high for newly added items
## Describe the bug

The number of usage days to divide usage is very small when just synced and just creating requisitions

### To reproduce

1. Sync a site
2. Create a CI with some quants
3. Create a SR, check monthly usage

### Expected behaviour

monthly usage should be calculated correctly

### Proposed Solution

https://github.com/openmsupply/mobile/blob/c21dedf5f31bd667ab284dff76608adfc8241b2f/src/utilities/dailyUsage.js#L51

Think this line should take the ceil? Or `math.max(.., 1)`

### Version and device info

- App version: 7.0 dev?
- Tablet model:
- OS version:

### Additional context

Not sure when this was introduced, think 6.0

katherine-sussol:
Fixes #3736

## Change summary

Adjust divisor to have min value of 1.

## Testing

- [ ] Replicate #3736 and don't see error (monthly usage calculation should be much lower when freshly synced)

- E.g. Before fix with usage of 5 over <= day, expected monthly usage was 6319.84 (or even higher depending on how soon the SR was created)
image
- E.g. After fix with usage of 5 over <= 1 day, expected monthly usage is 152.08
image

- Ran code formatter, it edited `programDailyUsage` but no functional changes

### DIFF
--- a/src/utilities/dailyUsage.js
+++ b/src/utilities/dailyUsage.js
@@ -48,7 +48,7 @@ export const dailyUsage = item => {
   const useLookbackDate = amcEnforceLookback || itemAddedDate.isBefore(lookbackDate);
   const startDate = useLookbackDate ? lookbackDate : itemAddedDate;
 
-  const numberOfUsageDays = moment.duration(dateNow.diff(startDate)).asDays();
+  const numberOfUsageDays = Math.max(moment.duration(dateNow.diff(startDate)).asDays(), 1);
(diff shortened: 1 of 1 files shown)


joshxg (approved):
Nice :)

PR #3949, opened 2021-04-16
#3757 fix breach boundary crossing detection

### LINKED ISSUE #3757: When temperatures of logs cross both breach boundaries, no breach is created
## Describe the bug

Crossing over breach boundaries causes incorrect behaviour.

The way breaches are calculated, logs are continually put into a stack until a breach is created. A breach is created when all the logs in the stack exceed a breach config threshold and the duration of all the logs exceed the config duration.

If there is a log which does not exceed the temperature threshold, the stack is reset.

This means if you have configs for less than `2` and greater than `8`, then if you have one log of `1` degree and then all the other logs are above `8`, then no breach is every created. For example and array of temperatures:

`[1,9,9,9,9,9,9,9,9]` - the stack is never reset, but no breaches are made, even though the sensor is in a hot breach.

### To reproduce

as above

### Expected behaviour

a breach should be created

### Proposed Solution

TBD

### Version and device info

- App version: 
- Tablet model:
- OS version:

### Additional context

Think this is very unlikely to occur. However, if it does, then no breaches would ever be created for the sensor which is quite bad.

katherine-sussol:
Fixes #3757

## Change summary

- Added new test for edge case of temperature logs crossing two breach boundaries (e.g. Too cold for one log, then too hot for 3 logs)
- Implemented logic to track which config the stack of 'potentially breach causing logs' is associated with, if a new log comes in associated with a different config to the stack, reset the stack 
- Makes the breach calculation logic even more confusing than before 😿 

## Testing

- [ ] Ran unit tests (could also probably be tested by setting up some mock data under the 'Generate vaccine data' button or playing around with the 'DevBleManager' to return fake logs breaching one config first, then breaching another)

- Previous breach calculation logic should work as before
- There's probably a nicer way of doing this 😆

### DIFF
--- a/src/bluetooth/BreachManager.js
+++ b/src/bluetooth/BreachManager.js
@@ -108,32 +108,52 @@ class BreachManager {
     return !(temperature >= thresholdMinTemperature && temperature <= thresholdMaxTemperature);
   };
 
-  couldBeInBreach = (log, configs) =>
-    configs.some(
+  getConfigInBreach = (log, configs) =>
+    configs.find(
       config =>
         log.temperature >= config.minimumTemperature && log.temperature <= config.maximumTemperature
     );
 
   createBreaches = (sensor, logs, configs, mostRecentBreach) => {
-    let potentialBreach = [];
+    const potentialBreach = { configInBreach: null, logs: [] };
     let currentBreach = mostRecentBreach;
 
     const breaches = mostRecentBreach && !mostRecentBreach.endTimestamp ? [mostRecentBreach] : [];
     const temperatureLogs = [];
 
     logs.forEach(log => {
-      const couldBeInBreach = this.couldBeInBreach(log, configs);
+      const configInBreach = this.getConfigInBreach(log, configs);
       const willCloseBreach = this.willCloseBreach(currentBreach, log);
       const willContinueBreach = this.willContinueBreach(currentBreach, log);
 
       if (willCloseBreach) {
         this.closeBreach(currentBreach, log.timestamp);
         currentBreach = null;
-        potentialBreach = [];
+        potentialBreach.logs = [];
+        potentialBreach.configInBreach = null;
       }
 
-      if (couldBeInBreach) potentialBreach.push(log);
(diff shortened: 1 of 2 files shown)


src/bluetooth/BreachManager.js
```diff
@@ -108,32 +108,52 @@ class BreachManager {
...
-        potentialBreach = [];
+        potentialBreach.logs = [];
+        potentialBreach.configInBreach = null;
       }
 
-      if (couldBeInBreach) potentialBreach.push(log);
-      else potentialBreach = [];
+      if (configInBreach) {
+        // Assign a config to the stack of logs
+        if (!potentialBreach.configInBreach && !logs) {
```
joshxg, line 138: Is `logs` always at least an empty array?

src/bluetooth/BreachManager.js
```diff
@@ -108,32 +108,52 @@ class BreachManager {
...
+          potentialBreach.configInBreach = configInBreach;
+        }
+
+        // If current log is in breach of the same config as the rest of the stack,
+        // push the current log onto the stack
+        if (potentialBreach.configInBreach === configInBreach) {
```
joshxg, line 144: I don't usualllyyy like to compare objects like this because a small change can cause this not to work.. maybe trying to compare ID's could be slightly safer

joshxg (approved):
nice.. jeez that function is loaded 🤣

src/bluetooth/BreachManager.js
katherine-sussol: Oops, yes it is. I think this block isn't actually needed as the calculations are still working with this never getting executed 🤦‍♀️ . Will remove

src/bluetooth/BreachManager.js
katherine-sussol: Oh yup - thanks! 8a472f44

src/bluetooth/BreachManager.js
katherine-sussol: ccddaedf

PR #3969, opened 2021-04-20
add reset option

### LINKED ISSUE #3968: Option to re-initialise mobile

**Possibly not required** feel free to argue this one.

I've had to reset tablets a lot recently, switching between environments to test production and dev servers.
We've also had an issue with tablets in Tonga, where the MDM is preventing admin users from accessing settings and clearing app data.

So.. there was a suggestion to allow clearing app data from within the app.

A protected (password? pin?) button in settings which resets the database and reloads the app.

mark-prins:
Fixes #3968 

## Change summary

Adds a button to settings: `Factory Reset` which prompts for the user's password and gives a small warning
If confirmed, the database is cleared and the app reloaded.

## Testing

- [ ] Settings: Factory Reset: cancel - should just return to the settings page
- [ ] Settings: Factory Reset: wrong password - toast showing that password is wrong and returns to the settings page
- [ ] Settings: Factory Reset: correct password - returns to the initial setup page

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,6 +97,7 @@
     "react-native-permissions": "^3.0.0",
     "react-native-popover-view": "^3.1.1",
     "react-native-reanimated": "^1.7.0",
+    "react-native-restart": "^0.0.22",
     "react-native-safe-area-context": "^0.7.3",
     "react-native-screens": "^2.2.0",
     "react-native-simple-radio-button": "^2.7.4",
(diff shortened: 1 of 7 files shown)


joshxg (approved):
Cool worked well for me.

Only concern is that any data that hasn't synced yet could be lost forever.. but 🤷

mark-prins:
Yes, good point about the sync. 
I was wondering about forcing an outgoing sync first - but maybe you are testing against the live server and you don't want to push dirty data back up to the server? Maybe you need to reset because something is wrong locally?
Also, not doing the sync first is easier. 
I'm now worried about the potential to lose data.

Issue #3922, opened 2021-04-13
8.1 dependency upgrades


joshxg:
## The suggestion

Been awhile since we did a round of dependency upgrades. We have some time now, with no clear deadline for 8.1 to get these done.

**Why?**
So we don't get to far behind.

## Describe alternatives you've considered

Doing nothing

## Additional context

N/A

katherine-sussol:
@joshxg Can pick this one up and give it a go (and hopefully don't break everything) - is there any general rule of thumb on what to upgrade?
Did a yarn audit and got `Severity: 132 Low | 1 Moderate | 25 High` - good idea to update anything showing up here?
Also looks like there's a newer react-native version released. Worth making separate issues for each dependency that needs changes?

joshxg:
@katherine-sussol A lot of the deps that come through with `audit` are peer deps from the packages we use.

This is a bit annoying. I think this was the last PR (maybe) https://github.com/openmsupply/mobile/pull/3264. I personally go through the `package.json` and upgrade each one individually checking release notes for potential breakages then commit each upgrade separately.. breaking it up into multiple PRs if there are a lot of upgrades needed (I think there will be).

Not sure if one PR per dep is needed, but could group them up say.. "linter/formatting" packages in a single PR kind of thing..

Anything that is to do with native code can be quite annoying so need to be cautious with those. Latest RN doesn't have too many new stuff so shouldn't break anything with a quick glance, and has also been out for a while and had a few patches so I think it is stable to upgrade to.

react-navigation I *think* is the main one where we are behind and also can cause a lot of issues to upgrade.

Hope that helps. I think it's important we get this done for 8.1. TLDR advice is: start with the 'easy' upgrades (where the code is pure javascript and only minor/patch version bumps) and move into the bigger ones where the package has some native code and/or has a major version bump (Where we need to evaluate if there are breaking changes that require us to change some code)

katherine-sussol:
Just realised last PR (#3970) broke regular dispensing ooops, don't know which dep broke it specifically yet, will try investigate:
 ![image](https://user-images.githubusercontent.com/65875762/115499673-edfc7600-a2c3-11eb-81ea-ec529eecd4dd.png)

joshxg:
@katherine-sussol looks like realm to me

katherine-sussol:
Done for now (seems there are a few 🐛  that crawled out from this Issue  - but being tracked as separate issues).

PR #3983, opened 2021-04-22
#3922 8.1 dependency upgrades rn

katherine-sussol:
Part 2 of #3922

## Change summary

- Currency.js (1.2.2 -> 2.0.3)
  - Removed formatWithSymbol param (ref: https://github.com/scurker/currency.js/blob/master/changelog.md)
- React (16.13.1 -> 17.0.1)
  - Didn't identify any obvious breakages (ref: https://reactjs.org/blog/2020/08/10/react-v17-rc.html#other-breaking-changes)
- React Native (0.63.4 -> 0.64) 
  - Applied these changes: https://react-native-community.github.io/upgrade-helper/?from=0.63.4&to=0.64.0
  - Fixed weird UI issues with dubious style changes
  - Before: image
  - After (in both cases I ignored the problem root cause and hacked the styles to make it look sort of ok): image
  - Didn't notice/identify anything else, ref: https://github.com/react-native-community/releases/blob/master/CHANGELOG.md
- react-native-vector-icons (7.1.0 -> 8.1.0)
  - No updates needed, seem mostly iOS related (ref: https://github.com/oblador/react-native-vector-icons/releases) 
- react-native-screens (2.2.0 -> 3.1.1)
  - Dropped support for old RN, no changes needed? (ref: https://github.com/software-mansion/react-native-screens/releases)
- @react-native-community/async-storage 1.8.1 -> @react-native-async-storage/async-storage 1.15.2
  - Package org moved
- Removed
  - react-native-reanimated
  - react-native-text-input-mask 

## Testing

- [ ] Build the app successfully (after cleaning everything)
- [ ] Run the app and hopefully it all works like before or better 🙏 

- Login password input looks bad, sounds like this issue: https://github.com/facebook/react-native/issues/30123
  - image
- I don't know, I feel like I broke more stuff


joshxg (approved):
👍 

Built dev+prod and seems to be OK

I'm sure there will be a couple more things come up but I'm sure we will find them as we continue to develop the next release and can fix them as they come up!

PR #4027, opened 2021-05-09
#4022 patient history lookup

### LINKED ISSUE #4022: Patient history lookup - for all patients

Add a button which will request patient history from the central server and display in a table.
This allows a user to review patient history for all patients, even for prescriptions which were not made on the current tablet.

There is already a patient history button in the dispensing section.
The team in Tonga would like to use the dispensing section, as they will verify using the registration tablets.

Add a button to the history window for 'Lookup history' which will:
- display a loading indicator
- fetch history from the server
- timeout if no connection is available
- display results if they are returned

## Implementation

TBC

## Additional context

The suggestion was made to show only vaccination history. Also to display from the final step of the vaccination process.
For Tonga, the latter request won't meed their need - also, in Tonga, only vaccinations are recorded in the system so by default only vaccinations history is shown.

Should we only lookup vaccination history?

mark-prins:
Fixes #4022 

## Change summary

Local and remote history in the dispensing and vaccine dispensing history modals.
Vaccine dispensing shows only vaccines, and sorts by date; dispensing history is sorted by name.

I've used the SimpleTable so the columns aren't sortable now - I thought that this would be ok for history, where the number of items is limited. Also haven't implemented offset and pagination with the api; thinking the number of prescriptions won't be too long.

The pref `can_view_all_patients_history` needs to be enabled for the store to be able to view online history.

## Testing

- Look at history through dispensing
  - [ ] Should show 'no history' if there isn't any; local history if there is
  - [ ] Able to look online for history; this shows 'no history' or history, if there is some
  - [ ] Entries are sorted by item name
- Look at history through vaccine dispensing
  - [ ] Should show 'no history' if there isn't any; local history if there is
  - [ ] Able to look online for history; this shows 'no history' or history, if there is some
  - [ ] Only vaccines are shown
  - [ ] Entries are sorted by date


katherine-sussol (approved):
Nice! Runs nicely with msupply #8643 🥳  (I hardcoded the pref to take a peek too). Would it be useful to have some kind of indication to pick out local vs remote (an icon? different colour?). Not sure how useful current end users would find it 😆 .

katherine-sussol:
Oh actually I probably spoke too soon, had a couple of issues after (but guess they can be looked at separately haha)

- Remote dispensing history was showing up for Jose Abad but not for Hermione Granger on my test data file (with both of them having some prescription history, might dig further and see what's going on) 😆 
- When there are local results but there's a problem fetching remote results, the whole modal displays text 'There was a problem communicating with the server', replacing the existing search results. Not a biggie (can always close and reopen) but maybe error is better as a toast or something?

PR #4049, opened 2021-05-13
#426 incorrect version number on install/update

### LINKED ISSUE #426: On install/update shows incorrect version number
Version Number: v1.1.1

Description: Version number not set correctly. Will show `v1.1.0` when updating to v1.1.1, or `v` on fresh install.

Reproducible: yes

Reproduction Steps: Uninstall app, install app, will show `v` for version on initialisation page

Comments: After first attempt to sync connect to server, the app version will update to correct value.

katherine-sussol:
Fixes #426 

## Change summary

- In `FirstUsePage` and `LoginModal` read the version from `package.json` rather than trying to access `AsyncStorage`, which doesn't get updated in time when an app is freshly installed to a device.
- Removed `getAppVersion()` from `settings/index.js` as it wasn't used anywhere else (contemplated moving the `package.json` reading to a separate helper under utilities but it's like 1 line so seems redundant?)
- Offroad, migrated a deprecated package `babel-eslint` to `@babel/eslint-parser` as I kept running into the same issue as this guy https://github.com/eslint/eslint/issues/13542 on `FirstLoginPage` when the pre-commit hook runs... as a sidenote, saw a bunch of linting errors running `eslint src` but left them to try not to explode the PR.

## Testing

- [ ] See https://github.com/openmsupply/mobile/issues/426 

Not sure...

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -4,7 +4,7 @@
   "globals": {
     "fetch": false
   },
-  "parser": "babel-eslint",
+  "parser": "@babel/eslint-parser",
   "plugins": ["prettier"],
   "rules": {
     "arrow-body-style": ["error", "as-needed"],
--- a/package.json
+++ b/package.json
@@ -115,8 +115,8 @@
   },
   "devDependencies": {
     "@babel/core": "^7.14.0",
+    "@babel/eslint-parser": "^7.14.2",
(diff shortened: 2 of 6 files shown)


mark-prins (approved):
😃  That works for me!

PR #4056, opened 2021-05-19
#3925 Update date parser

### LINKED ISSUE #3925: requisition created date is non-nullable on mobile but not desktop
## Describe the bug

- `[requisition]date_order_received` can be null on desktop.
- It cannot be on mobile
- this causes sync to halt during initialization

### To reproduce

1. Go to desktop and make a mobile stores requisition have `date_order_received = 00/00/00`
3. try init

### Expected behaviour

Should be able to init

### Proposed Solution

I think maybe the `parseDate` func should return `undefined` instead of `null` when receiving a date `00/00/00`

### Version and device info

- App version:
- Tablet model:
- OS version:

katherine-sussol:
Fixes #3925

## Change summary

Tested out the proposed solution for #3925 🤣 . Things seem to work as expected.

Record created after syncing requisition with `date_order_received` of 00/00/00:
image

## Testing

- [ ] Reproduce #3925 with PR and hopefully don't see error

Not sure

### DIFF
--- a/src/database/utilities/parsers.js
+++ b/src/database/utilities/parsers.js
@@ -24,7 +24,7 @@ export const parseDate = (ISODate, ISOTime) => {
     ISODate === '0000-00-00T00:00:00' ||
     ISODate === '0000-00-00T00:00:00Z'
   ) {
-    return null;
+    return undefined;
   }
   const date = new Date(ISODate);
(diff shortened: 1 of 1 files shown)


mark-prins (approved):
👍  That works for me!

Issue #4055, opened 2021-05-18
Sensor logs not downloaded


mark-prins:
## Describe the bug

From Georgia: "I just reconnected.. and there is no data"

![image](https://user-images.githubusercontent.com/9192912/118731022-9f270980-b88c-11eb-8b8e-647e2021e508.png)

Testing locally; the Tempo app shows data, but mobile is missing:

![graph](https://user-images.githubusercontent.com/9192912/118731149-deedf100-b88c-11eb-810a-ec1722380311.png)

### To reproduce

1. Add a sensor and log temperatures
2. Turn off tablet for some time. Turn on again and wait for sync

PR #4057, opened 2021-05-19
update logs to save & connectAndDiscover

### LINKED ISSUE #4045: Temperature log timestamps can be incorrect when some hunchy funchy happens
## Describe the bug

OK

- Hook up a sensor to a tablet
- Take the sensor away and use it elsewhere.
- Bring it back 4 days later with less then 4 days worth of logs
- App will try and start integrating from 4 days ago rather than counting back 4 logs (all the sensor has) from *now*

### To reproduce

```
it('Correctly calculates the timestamp of a sensor when there is a gap of logs', () => {
    // We are mocking the case where the sensor has been taken away and used elsewhere, for some reason.
    // This sensor is returning a few days later with only a few logs (they were deleted elsewhere).
    // In this case, we should be calculating timestamps backwards from now, rather than counting
    // forward from the saved timestamps.

    // Mocks
    const dbService = {};
    const utils = { uuid: () => '1' };
    const downloadManager = new TemperatureLogManager(dbService, utils);
    const sensor = { id: 'a', logInterval: 300 };

    // Important numbers!

    // We can save heaps of logs
    const maxNumberToSave = 100;
    // The most recent log we have saved is here.
    const mostRecentLogTime = 1000;
    // The time now though, is 100 log intervals into the future.
    const timeNow = mostRecentLogTime + 100 * 300;
    // Then, when we downloaded the logs, we only got two of them.
    const logs = [{ temperature: 10 }, { temperature: 11 }];

    // So when we save these two logs we downloaded, we should save them at
    // timeNow and timeNow - logInterval
    const shouldBe = [
      {
        id: '1',
        temperature: 10,
        timestamp: new Date((timeNow - sensor.logInterval) * 1000),
        sensor,
        logInterval: sensor.logInterval,
      },
      {
        id: '1',
        temperature: 11,
        timestamp: new Date(timeNow * 1000),
        sensor,
        logInterval: sensor.logInterval,
      },
    ];

    expect(
      downloadManager.createLogs(logs, sensor, maxNumberToSave, mostRecentLogTime, timeNow)
    ).toEqual(shouldBe);
  });
```

### Proposed Solution

I think if we just pass `Math.min(maxNumberToSave, shouldBe.length)` (but like, the right variables) then we will count back the number of logs a sensor actually has rather than what we 'think' it *should* have..
### Version and device info

- App version: 7.0+
- Tablet model:
- OS version:

### Additional context

Another option would be to make the download and integrate code less confusing 🙃

mark-prins:
Fixes #4045 #4055 

## Change summary

Disconnect and reconnect code from the cold chain app, which seems to resolve the problem with a ` BleError: Device MA:CA:DD:RE:SS was disconnected` error being thrown quietly, which was preventing logs from being downloaed.

Have included the fix ( from @joshxg :) )for #4045 too.

## Testing

- [ ] Download some logs; turn off the tablet for a while. turn on and hopefully, get some more logs.

### DIFF
--- a/src/bluetooth/BleService.js
+++ b/src/bluetooth/BleService.js
@@ -41,7 +41,8 @@ class BleService {
    * on the device.
    * @param {String} macAddress
    */
-  connectToDevice = async macAddress => this.manager.connectToDevice(macAddress);
+  connectToDevice = async (macAddress, options) =>
+    this.manager.connectToDevice(macAddress, options);
 
   /**
    * Connects to a device with the provided macAddress as well as discovering
--- a/src/bluetooth/BleService.js
+++ b/src/bluetooth/BleService.js
@@ -52,11 +52,16 @@ class BleService {
    * @param {String} macAddress
    */
   connectAndDiscoverServices = async macAddress => {
-    if (!(await this.manager.isDeviceConnected(macAddress))) {
-      await this.connectToDevice(macAddress);
+    // without the cancel & reconnect further commands
+    // were sometimes returning an error: "BleError: Device [mac address] was disconnected"
+    if (await this.manager.isDeviceConnected(macAddress)) {
+      await this.manager.cancelDeviceConnection(macAddress);
     }
 
-    return this.manager.discoverAllServicesAndCharacteristicsForDevice(macAddress);
+    const device = await this.connectToDevice(macAddress, { autoConnect: true });
+    await this.manager.discoverAllServicesAndCharacteristicsForDevice(macAddress);
+
+    return device;
   };
 
   /**
--- a/src/actions/Bluetooth/SensorDownloadActions.js
+++ b/src/actions/Bluetooth/SensorDownloadActions.js
@@ -124,7 +124,7 @@ const downloadLogsFromSensor = sensor => async dispatch => {
             const temperatureLogs = await TemperatureLogManager().createLogs(
               downloadedLogsResult,
               sensor,
-              numberOfLogsToSave,
+              Math.min(numberOfLogsToSave, downloadedLogsResult.length),
               nextTimestamp.unix()
             );
 


katherine-sussol (dismissed):
Ohh is that what that code was for in the cce app!
Feeding two 🐦  with one scone, nice one.

joshxg (dismissed):
Cool! 

Would it be better to put this into master so we can do a quick patch tomorrow?

src/bluetooth/BleService.js
```diff
@@ -52,11 +52,16 @@ class BleService {
...
-    if (!(await this.manager.isDeviceConnected(macAddress))) {
-      await this.connectToDevice(macAddress);
+    // without the cancel & reconnect further commands
+    // were sometimes returning an error: "BleError: Device [mac address] was disconnected"
+    if (await this.manager.isDeviceConnected(macAddress)) {
+      await this.manager.cancelDeviceConnection(macAddress);
     }
 
-    return this.manager.discoverAllServicesAndCharacteristicsForDevice(macAddress);
+    const device = await this.connectToDevice(macAddress, { autoConnect: true });
```
katherine-sussol, line 61: Hm, does the `autoConnect` arg get passed through to the BleManager if `connectToDevice` doesn't also pass  through? 🤔  https://github.com/openmsupply/mobile/blob/1f2149f1c6861883579072786747316b4fc9915e/src/bluetooth/BleService.js#L44

mark-prins: DOH! good spotting @katherine-sussol

mark-prins:
> Would it be better to put this into master so we can do a quick patch tomorrow?

Yes, probably. Was still rather hopeful that we could get the select-a-vaccinator-and-crash-the-app feature removed and do a full release, though this fix is probably good to release sooner rather than later. Would be nice to test it with them.

mark-prins:
@joshxg @katherine-sussol oops! I changed to master to see how bad that would be.. and it wasn't good, so I've switched back. but that had dismissed the reviews.. sorry. can you click again please? I'll make another PR for master 🙄

mark-prins:
actually.. I might have another look first:

gap-in-graph

joshxg:
oh noooo...

But that looks like it did download some logs this morning ?

mark-prins:
With further testing I can't repro - seems worth getting this out to the client and testing with them.

mark-prins:
@katherine-sussol or @joshxg would you mind re-reviewing? I'd like to merge this into develop before I forget

joshxg:
@mark-prins I've just merged master into develop which I think makes this redundant now?

mark-prins:
Quite right, thanks

PR #4077, opened 2021-06-02
#4068 Add barcode field to mobile schema

### LINKED ISSUE #4068: Add barcode field to realm schema

Child of #4066

The ability for mobile to store a patient's barcode

## Implementation

- Add barcode field to realm `Name` object
- Update sync utils to include barcode 

None really.

## Additional context
N/A

katherine-sussol:
Fixes #4068

## Change summary

- Add `[name]barcode` to realm schema. 
- Already exists in desktop with placeholder values (in the format `*[name]code*`)

## Testing

- [ ] Sync and check realm explorer to see if it's there?

Not sure

### DIFF
--- a/src/actions/PatientActions.js
+++ b/src/actions/PatientActions.js
@@ -41,6 +41,7 @@ const patientUpdate = patientDetails => async (dispatch, getState) => {
   const {
     id: currentPatientId,
     code: currentCode,
+    barcode: currentBarcode,
     name: currentName,
     firstName: currentFirstName,
     lastName: currentLastName,
@@ -69,6 +70,7 @@ const patientUpdate = patientDetails => async (dispatch, getState) => {
   const {
     id: patientId,
     code: patientCode,
+    barcode: patientBarcode,
     firstName: patientFirstName,
     lastName: patientLastName,
     dateOfBirth: patientDateOfBirth,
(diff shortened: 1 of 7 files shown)


mark-prins (changes requested):
The `schemaVersion` will need incrementing :)

mark-prins:
Does the field need to be populated when a patient is looked up?
Currently the field has `placeholderBarcode` in it.

katherine-sussol:
> The `schemaVersion` will need incrementing :)

🙀  thanks, I forgot about that!

katherine-sussol:
> Does the field need to be populated when a patient is looked up?
> Currently the field has `placeholderBarcode` in it.

Hm, probably 😿 , 757feee0 hopefully got everything needed for lookups, now I'm wondering if I've missed any other use cases 😅

mark-prins:
>Does the field need to be populated when a patient is looked up?

I think probably yes, but 🤦  the `/patient` api doesn't return it.
That will need to be updated if the barcode is to be looked up - I suspect that there is desktop work required anyway ( will need changes to be able to search by barcode )

mark-prins (approved):
Great!

katherine-sussol:
> > Does the field need to be populated when a patient is looked up?
> 
> I think probably yes, but 🤦 the `/patient` api doesn't return it.
> That will need to be updated if the barcode is to be looked up - I suspect that there is desktop work required anyway ( will need changes to be able to search by barcode )

Yeah, there is an issue for it here https://github.com/sussol/msupply/issues/8840, I just realised I'd modified my local mSupply to expose barcodes for testing which is why I was seeing them even though they aren't technically there. Jumped the 🔫   a bit but will re-test when it is done for reals

PR #4090, opened 2021-06-09
#4089: Hide QR Scan button behind DEV flag

### LINKED ISSUE #4089: Hide scan QR code button until it is releasable

Just so it doesn't show up when develop is merged to master

Hide behind DEV flag (or another appropriate setting)

## Implementation

Releasing not yet ready feature

katherine-sussol:
Fixes #4089

## Change summary

- Temporarily hide QR scan button until it is ready for release

## Testing

- [ ] Navigate to patient select screen and verify there is no 'Scan QR Code' button

Possibly could look into implementing an actual feature flag config if this type of thing comes up more frequently.

### DIFF
--- a/src/widgets/Tabs/PatientSelect.js
+++ b/src/widgets/Tabs/PatientSelect.js
@@ -1,3 +1,4 @@
+/* eslint-disable no-undef */
 /* eslint-disable react/forbid-prop-types */
 /**
  * mSupply Mobile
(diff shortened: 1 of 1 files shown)


mark-prins (approved):
👍  that works :)

I did wonder if we should be enabling via a preference; I'm expecting that not all clients will want or can support the QR scanning functionality ( currently we have 1 country only )

katherine-sussol:
> I did wonder if we should be enabling via a preference; I'm expecting that not all clients will want or can support the QR scanning functionality ( currently we have 1 country only )

Oh, yes probably a good idea. I guess (just because there's a bunch already there 😆 ) that would be best done as a store pref in desktop?

mark-prins:
@katherine-sussol yeah, store pref was what I was thinking. We should ask around to see if it is required or not - the button doesn't interrupt the process flow and can safely remain for everyone, I think. 🤷‍♂️

joshxg:
I vote for no pref. Only because I am lazy and adding a store pref is work

mark-prins:
Sound argument, thanks. I agree.

PR #4109, opened 2021-06-11
Reduce svg renders

### LINKED ISSUE #4053: Cold chain line graph doesn't have lines?
## Describe the bug

😿 
image

### To reproduce

1. Enable cold chain, set up a sensor and let it log for a while
2. View sensor details for the sensor and select line graph icon

### Expected behaviour

Lines? Something more like this (from `master`):
image

### Proposed Solution

### Version and device info

- App version: Develop 
- Tablet model: N/A
- OS version: Android 8.1

### Additional context
??

mark-prins:
Fixes #4053

If the `<Svg>` renders more than once, the lines don't show 🤷‍♂️ 

Have changed the line chart, so that it only renders the svg component if the dimensions are set. It was rendering twice due to the `onLayout` event.

A side issue is that this invalidates the functioning of `AfterInteractions` so I've added the loading indicator directly into the chart component.

### DIFF
--- a/src/pages/FridgeDetailPage.js
+++ b/src/pages/FridgeDetailPage.js
@@ -62,7 +62,7 @@ export const FridgeDetailPageComponent = ({
   onChangeToDate,
   onChangeFromDate,
 }) => {
-  const [chartType, setChartType] = useState('bar');
+  const [chartType, setChartType] = useState('line');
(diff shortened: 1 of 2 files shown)


joshxg (approved):
OK

Issue #4072, opened 2021-06-01
Remote patient searching via QR Code


katherine-sussol:
Child of #4066
Prerequisite: https://github.com/sussol/msupply/issues/8840

## Is your feature request related to a problem? Please describe.

As a user, I would like to be able to use a scanned QR code to search against online patient records (when a local search has returned no results), so that I don't need to manually enter search criteria into the app when dispensing a vaccine.

## Describe the solution you'd like
Given:
- The app has successfully retrieved a QR code, and 
- A local search has been performed with the QR code and no results were found
  - A search should be performed against the mSupply server patient API using the QR code
  - If there is a match, the user should be taken to Step 2 of the vaccine dispensing flow, with the relevant patient details prepopulated
  - If there is no match, the user should be told the patient couldn't be found, and be taken back to the vaccine dispensing page to either try again or enter details manually
 
## Implementation

TBD

## Describe alternatives you've considered

N/A

katherine-sussol:
Found a sample ID:
![Untitled copy](https://user-images.githubusercontent.com/65875762/121832084-52b0db80-cd1d-11eb-9230-426089c43cfd.png)
Encoded is the following: `VGID\00177678/8N2HYJL`

ArpitaSussol:
Tested in mSupply V8.2.0 - rc2

## Testing

Steps to reproduce or otherwise test the changes of this PR:

- [x] Have a patient that is not available locally but is available remotely, scan a QR code corresponding to that patient -> should be taken to step 2 of the patient vaccination page with remote patient details pre-populated

PR #4117, opened 2021-06-14
#4071 QR code patient lookup - local search implementation

### LINKED ISSUE #4071: Local patient searching via QR Code
Child of #4066
Prerequisite: https://github.com/openmsupply/mobile/issues/4068 

As a user, I would like to be able to use a scanned QR code to search against all patients stored on the device, so that I don't need to manually enter search criteria into the app when dispensing a vaccine.

- Given the app has successfully retrieved a QR code:
  - A search should be performed using the QR code against all patients stored locally in the app
  - If there is a match, the user should be taken to Step 2 of the vaccine dispensing flow, with the relevant patient details prepopulated
  - If there is no match, the user should be told to wait longer while an online search is done (?)

## Implementation

TBD

N/A

katherine-sussol:
Fixes #4071

## Change summary

- Adds some code to do some sort of local searching and select the patient if there is a match

## Testing

- [ ] Scan a QR code that contains a matching barcode for a patient. 

Some test codes (needs to exist in mobile for lookup to work):
`placeHolderBarcode`:
placeholderBarcode

`*asdasdasd*` (matches Jose Abad):
_asdasdasd_

### DIFF
--- a/src/widgets/Tabs/PatientSelect.js
+++ b/src/widgets/Tabs/PatientSelect.js
@@ -9,6 +9,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 import { ActivityIndicator, Keyboard, StyleSheet, Text, ToastAndroid, View } from 'react-native';
+import ReactNativeHapticFeedback from 'react-native-haptic-feedback';
 
 import { batch, connect } from 'react-redux';
 
@@ -138,8 +139,14 @@ const PatientSelectComponent = ({
   const withLoadingIndicator = useLoadingIndicator();
   const [isQrModalOpen, toggleQrModal] = useToggle();
 
+  const hapticFeedBackOptions = {
+    enableVibrateFallback: true,
+    ignoreAndroidSystemSettings: false,
+  };
+
   const onQrCodeRead = ({ data }) => {
     // TODO: Some validation/data sanitisation might be required here but don't know format yet...
+    ReactNativeHapticFeedback.trigger('notificationSuccess', hapticFeedBackOptions);
     toggleQrModal();
 
     const matchedLocalPatient = UIDatabase.objects('Name').filtered('barcode == $0', data)[0];
--- a/src/widgets/Tabs/PatientSelect.js
+++ b/src/widgets/Tabs/PatientSelect.js
@@ -139,9 +139,17 @@ const PatientSelectComponent = ({
   const [isQrModalOpen, toggleQrModal] = useToggle();
 
   const onQrCodeRead = ({ data }) => {
-    // TODO: Some validation might be good here but don't know format..
+    // TODO: Some validation/data sanitisation might be required here but don't know format yet...
     toggleQrModal();
-    ToastAndroid.show(`${data} scanned`, ToastAndroid.LONG);
+
+    const matchedLocalPatient = UIDatabase.objects('Name').filtered('barcode == $0', data)[0];
+
+    // Do local search
+    if (matchedLocalPatient) {
+      selectPatient(matchedLocalPatient);
+    } else {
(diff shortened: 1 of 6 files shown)


mark-prins:
When I scan the first barcode, it takes me to the patient creation screen. Scanning a different invalid code shows the toast as expected. I'm guessing that we have something funky with the placeholder text? So not really an issue, but amusing?

src/widgets/Tabs/PatientSelect.js
```diff
@@ -139,9 +139,17 @@ const PatientSelectComponent = ({
   const [isQrModalOpen, toggleQrModal] = useToggle();
 
   const onQrCodeRead = ({ data }) => {
-    // TODO: Some validation might be good here but don't know format..
+    // TODO: Some validation/data sanitisation might be required here but don't know format yet...
```
mark-prins, line 142: I'm not sure that it is necessary - it either matches the patient or doesn't 🤷‍♂️

src/widgets/Tabs/PatientSelect.js
```diff
@@ -139,9 +139,17 @@ const PatientSelectComponent = ({
...
     toggleQrModal();
-    ToastAndroid.show(`${data} scanned`, ToastAndroid.LONG);
+
+    const matchedLocalPatient = UIDatabase.objects('Name').filtered('barcode == $0', data)[0];
+
+    // Do local search
+    if (matchedLocalPatient) {
+      selectPatient(matchedLocalPatient);
+    } else {
+      ToastAndroid.show('No patient found via QR code', ToastAndroid.LONG);
```
mark-prins, line 151: The Toast is pretty subtle - I think probably ok, and certainly worth seeing what others think.

src/widgets/Tabs/PatientSelect.js
```diff
@@ -139,9 +139,17 @@ const PatientSelectComponent = ({
   const [isQrModalOpen, toggleQrModal] = useToggle();
 
   const onQrCodeRead = ({ data }) => {
-    // TODO: Some validation might be good here but don't know format..
+    // TODO: Some validation/data sanitisation might be required here but don't know format yet...
     toggleQrModal();
-    ToastAndroid.show(`${data} scanned`, ToastAndroid.LONG);
+
```
mark-prins, line 144: It would be awesome to have some haptic feedback if that's possible - it works for me with the covid tracer app anyway :)
Just something to let the user know that the scan happened and they can put the tablet down now

mark-prins (approved):
👍  It works 😄

src/widgets/Tabs/PatientSelect.js
katherine-sussol: Hm, mayyybe, was a bit unsure on this. I was thinking to provide some feedback if a user tried to scan unrelated/potentially huge QR codes instead of passing them through to our patient search API 😆 , but would probably do as part of #4072

src/widgets/Tabs/PatientSelect.js
katherine-sussol: Ooh this is a neat idea, will have a looksies 👀

katherine-sussol: Stole the package name from the covid tracer app and added some feedback on successful scan 🤣  ed3d4123

katherine-sussol:
> When I scan the first barcode, it takes me to the patient creation screen. Scanning a different invalid code shows the toast as expected. I'm guessing that we have something funky with the placeholder text? So not really an issue, but amusing?

Hmm yeah I think you are right, there are probably multiple name records that have `barcode: placeholderBarcode` without the Patient API change to expose barcodes to mobile 🤣

PR #4120, opened 2021-06-14
#4082 Data import/restore feature

### LINKED ISSUE #4082: Data import

When trying to reproduce user's issues, sometimes it's hard to know exactly what's wrong and how their data might be causing it.

It'd be useful to be able to take realm data exported from the app, and import it into the app on a different device (i.e. one in our offices).

@ujwalSussol says this would be really helpful for support trying to debug issues (one current support ticket trying to diagnose a sync issue), and for testers trying to reproduce things. I tend to agree!

A little more on said support issue, this would allow support to potentially recover un-synced data. There is a current support ticket where a client cannot sync, and has 100 invoices on the tablet. If the problem is due to a bad app state, clearing the tablet and resyncing would lose all those invoices. If support could get the data from the tablet, diagnose it, correct it, wipe the tablet and import, then mission accomplished! (And hopefully figure out what caused the bad state along the way.)

Support ticket : https://support.msupply.org.nz/cerb6/profiles/ticket/YQ-9957-RF

An import button somewhere on initialisation page (subtle cog in the corner?)

## Implementation

?

@andreievg once did something with loading a realm datafile into the app. Might have been at build time though.

Realm may have a way of doing this neatly with a realm datafile.

One thing to consider is that realm data from an older version of mobile would either A: need to go through migration or B: not work, they need to upgrade to a version that includes the import functionality (and therefore ran migrations needed).

🙅 

## Additional context

Related to #3277 ?

raviSussol:
Fixes #4082 

## Change summary
A data import/restore feature in the mobile app. With this a user (with a `view/edit preference` set in the desktop server for the mobile logged in user) can now import/restore the mobile data file into the current mSupply mobile app.

#### Setups

Setup a desktop mSupply server and use [these](https://we.tl/t-kWHLKcaDZe) apks to test below steps.

**To test data import**
- Install any version of apk from the above link and setup sync with the desktop server
- Login to the app and try modifying current datafile with adding some invoices or stocktakes.
- Goto the settings page and then click `Export data` - may ask to allow storage permission, go ahead give a permission.
- Now again try to modify data and then goto the settings page then click `Import data` - may ask to allow storage permission, go ahead give a permission, follow test 1, 2 & 3.

**To test version incompatible issue while doing data import**
- Install the higher version of apk from the above link and setup sync with the desktop server
- Login to the app and goto the settings page and click `Export data`
- Close the app and uninstall it
- Now install the lower version of the apk from the above link and setup sync with the desktop server
- Login to the app and goto the settings page and click `Import data`, and follow test 4, 5 & 6.

#### Tests
- [ ] 1. A file explorer will appear to choose `*.realm` file to import.
- [ ] 2. After choosing `*.realm` file mSupply app will import that file and quits the app automatically after 2 seconds.
- [ ] 3. Restart the app, login and check the data - now the datafile has previous changes in it - which means importing successfully.
- [ ] 4. A file explorer will appear to choose `*.realm` file to import.
- [ ] 5. After choosing `*.realm` file mSupply app will import that file and quits the app automatically after 2 seconds.
- [ ] 6. Restart the app, you'll see an alert as shown in the screenshot below. Clicking the `OK` will quit the app.
Screen Shot 2021-06-14 at 2 46 51 PM


raviSussol:
For resource strings translation may require for them but I've just used a plain English text for now.

joshxg (changes requested):
It's working well and the code is great! Just the translation string and maybe a restart instead of exit and I would be happy!

Some other suggestions which we could make new issues for if we think they are worthwhile:
- A button on the initialisation screen. Could maybe be hidden by `__DEV__` or not, but being able to import a file without having to initialise would be nice!
- Someway to clear the sync url and password in an exported data file or a flag "this is an exported data file, sync is paused" so we don't accidentally sync with a datafile we aren't meant to. This ones a bit trickier. Would also mean that sync is 'paused (in a way)' until the sync details are re-entered. Maybe just a `SETTINGS_KEY` for `isSyncPaused` with a button to trigger it.. 🤔  otherwise, we might need to always remove the sync url/password from exported data files, and that is not ideal!

I also wondered about the redux-persistance and if that would have some side effects, but I can't think of any..!

src/utilities/fileSystem.js
```diff
@@ -32,3 +33,38 @@ export async function backupValidation(path) {
...
+    });
+    // Check if file(s) exist of specific file type then only return their file path
+    if (fileType !== '*') {
+      const fileExtensionRegex = new RegExp(`\\.(${fileType})$`, 'g');
+      if (fileExtensionRegex.test(name)) {
+        filePath = fileCopyUri;
+      }
+    } else {
+      filePath = fileCopyUri;
+    }
```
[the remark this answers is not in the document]
raviSussol: Done b5b49b38.

src/pages/SettingsPage.js
```diff
@@ -47,7 +47,27 @@ const exportData = async () => {
...
+  const syncSiteName = UIDatabase.getSetting(SETTINGS_KEYS.SYNC_SITE_NAME);
+  const success = await UIDatabase.importData(syncSiteName);
+  const toastMessage = success ? generalStrings.imported_data : generalStrings.couldnt_import_data;
+
+  ToastAndroid.show(toastMessage, ToastAndroid.SHORT);
+
+  if (success) {
+    setTimeout(() => {
+      // Delay application exit to 2 seconds so the above toast message can be readable to the user
+      BackHandler.exitApp();
```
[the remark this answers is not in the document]
raviSussol: Done e6118570.

src/utilities/fileSystem.js
```diff
@@ -32,3 +33,38 @@ export async function backupValidation(path) {
...
+  } catch (error) {
+    // User hasn't cancelled the picker, exit any dialogs or menus and return the empty file path
+    if (!DocumentPicker.isCancel(error)) {
+      filePath = '';
+    }
```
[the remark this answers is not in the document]
raviSussol: Just wanted to be sure for `filePath` to return `''` if any error occurs but the initial value for `filePath` would be fair enough to handle for this case as well. Resolve by b5b49b38.

src/mSupplyMobileApp.js
```diff
@@ -213,6 +217,18 @@ class MSupplyMobileAppContainer extends React.Component {
     );
   };
 
+  renderVersionIncompatibleAlertDialog = () => {
+    Alert.alert(
+      'Alert...',
+      `This is version of ${appVersion} of mSupply mobile.` +
+        // eslint-disable-next-line max-len
+        `It appears that your imported data file has a newer version of ${this.databaseVersion}.` +
```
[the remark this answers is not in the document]
raviSussol: Yeah I'd thought about it initially but got stuck with the dynamic resource strings. Done e86c2a46.

raviSussol:
@joshxg,
> - A button on the initialisation screen. Could maybe be hidden by __DEV__ or not, but being able to import a file without having to initialise would be nice!

Yeah I liked the idea too. Regarding this button visibility we now have made the restore feature only available to us in the desktop app (thinking not to cause data integrity issue caused by sync due to an older version of datafile). Keeping that think on mind we should follow the same principle here in the mobile app too (i.e show under dev mode only). But it is an open source app so not sure doing so would even make really a sense?

> - Someway to clear the sync url and password in an exported data file or a flag "this is an exported data file, sync is paused" so we don't accidentally sync with a datafile we aren't meant to. This ones a bit trickier. Would also mean that sync is 'paused (in a way)' until the sync details are re-entered. Maybe just a SETTINGS_KEY for isSyncPaused with a button to trigger it.. 🤔 otherwise, we might need to always remove the sync url/password from exported data files, and that is not ideal!

Make sense! I think we can also achieve this by storing sync details in an `AsyncStorage` before exporting data, clear sync details from the datafile then export it and retain back sync details from an `AsyncStorage` to the running datafile?

joshxg (approved):
Nice, thanks a lot @raviSussol it's awesome.

@mark-prins - Do you think we should wait to release 8.1 before merging or are you OK for us to merge this now?

joshxg:
> Yeah I liked the idea too. Regarding this button visibility we now have made the restore feature only available to us in the desktop app

I don't think it's weird in an open source repo!

> Make sense! I think we can also achieve this by storing sync details in an AsyncStorage before exporting data, clear sync details from the datafile then export it and retain back sync details from an AsyncStorage to the running datafile?

That's a great idea! Does seem a little more difficult than adding a `Setting` to pause sync?

src/database/UIDatabase.js
```diff
@@ -133,6 +133,29 @@ class UIDatabase {
...
+    const { realm } = this.database;
+    const { path: realmPath } = realm;
+    const realmFilePath = await selectDocument({ fileType: 'realm' });
+
+    if (realmFilePath === '') return false;
+    if (realm.isInTransaction) return false;
+
+    // Proceed importation of the realm file to the app database location
+    try {
+      await RNFS.copyFile(realmFilePath, realmPath);
```
mark-prins, line 151: after this copy is finished, is it possible to read the database version? ( sorry, don't have time to check right now )
It would be awesome if the import could check, alert if the version is invalid, and then rollback the change.

Then you won't need the version check and alert - I find that really tiresome in mSupply desktop. Once you load up a newer datafile, you can't easily get back. It would be good if the import checked and didn't allow screwing up the app.

mark-prins:
> @mark-prins - Do you think we should wait to release 8.1 before merging or are you OK for us to merge this now?

I think OK - doesn't touch general code, so in itself seems fairly low risk. and would great to have in :)

raviSussol:
> That's a great idea! Does seem a little more difficult than adding a Setting to pause sync?

Maybe I would say no cause changes only require in the export data button and the rest of would work normally. But having a new flag or a config in the database would also require changes in other relevant places as well (say in the sychroniser, postsyncprocessor) to check this flag?

src/database/UIDatabase.js
raviSussol: Agree. I had also an initial thought about as a preliminary actions for the import functionality but I found a bit difficult to read database version from the external realm file from the app (not sure if there's an easy way to do it) so that version incompatibility can be checked before importation. Also the current mobile app doesn't keep track of app's version in anywhere. So those things made me to decide checking on the startup of the app.

raviSussol: On the second thought, here are changes 8b7f50ed @mark-prins. Now the version incompatibility is checked before importing datafile not on the startup of the app.

mark-prins (approved):
Looks great! 😄

mark-prins:
@raviSussol I've added a commit, to add the import button to the FirstUse page as well - just an idea, which helped me to test the version check. Feel free to revert the commit.. it was easier to commit in order to suggest the change, since there were a few files to update

src/database/UIDatabase.js
mark-prins: That's great! thanks :)

src/localization/modalStrings.json
```diff
@@ -70,7 +70,8 @@
     "bluetooth_permission": "Bluetooth",
     "bluetooth_permission_comment": "Required to connect to bluetooth sensors",
     "permissions_intro_1": "The following permissions and services are required for cold chain functionality.",
-    "permissions_intro_2": "If disabled, you can press an item to enable it."
+    "permissions_intro_2": "If disabled, you can press an item to enable it.",
+    "version_incompatible": "This is version of {0} of mSupply mobile. It appears that your imported data file has a newer version of {1}. It may be dangerous to continue. mSupply app will now quit. Contact info@msupply.org.nz."
```
mark-prins, line 74:
```suggestion
    "version_incompatible": "Unable to import! This is version of {0} of mSupply mobile and the selected import file has a newer version of {1}"
```

mark-prins (changes requested):
Sorry - one more thing. can you update the `version_incompatible` text please? mSupply app no longer quits :)

mark-prins (approved):
👍 
( I've already used this functionality a few times! thanks )

raviSussol:
> @raviSussol I've added a commit, to add the import button to the FirstUse page as well - just an idea, which helped me to test the version check. Feel free to revert the commit.. it was easier to commit in order to suggest the change, since there were a few files to update

That's cool! Thanks for that @mark-prins.

PR #4137, opened 2021-06-17
#4072 remote patient lookup via qr code

### LINKED ISSUE #4072: Remote patient searching via QR Code

katherine-sussol:
Fixes #4072 

## Change summary

- Adds remote patient searching by QR code (needs https://github.com/sussol/msupply/pull/8914 to actually work)
- Adds crude validation to reject QR codes that are longer than `50` chars long (sample Vanuatu National ID we have seen is 21 characters long)
- Small side quest to remove the extra `getAuthorizationHeader` method in `SearchForm`

## Testing

- [ ] Have a patient that is not available locally but is available remotely, scan a QR code corresponding to that patient -> should be taken to step 2 of the patient vaccination page with remote patient details pre-populated

Might need to look into different parsers for different countries at some point (instead of storing whole thing, depending on what their National ID QR codes look like...). Either coded in mobile or some kind of store pref in Desktop?

### DIFF
--- a/src/widgets/Tabs/PatientSelect.js
+++ b/src/widgets/Tabs/PatientSelect.js
@@ -156,7 +156,7 @@ const PatientSelectComponent = ({
     // Immediately reject QR code if it is too long (length > 50 chosen somewhat arbitrarily until
     // we get a better idea of the kinds of codes we need to handle in future)
     if (data.length > 50) {
-      ToastAndroid.show('Invalid QR code', ToastAndroid.LONG);
+      ToastAndroid.show(generalStrings.invalid_qr_code, ToastAndroid.LONG);
       return;
     }
 
@@ -175,7 +175,7 @@ const PatientSelectComponent = ({
           if (remotePatient.length) {
             selectPatient(remotePatient[0]);
           } else {
-            throw new Error('No patient found via QR code');
+            throw new Error(generalStrings.could_not_find_patient_qr);
           }
         } catch (error) {
           ToastAndroid.show(error.message, ToastAndroid.LONG);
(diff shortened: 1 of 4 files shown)


src/widgets/Tabs/PatientSelect.js
```diff
@@ -155,7 +168,19 @@ const PatientSelectComponent = ({
...
     } else {
-      ToastAndroid.show('No patient found via QR code', ToastAndroid.LONG);
+      // Do a remote search
+      withLoadingIndicator(async () => {
+        try {
+          const remotePatient = await lookupRemotePatient({ barcode: data });
+          if (remotePatient.length) {
+            selectPatient(remotePatient[0]);
+          } else {
+            throw new Error('No patient found via QR code');
```
joshxg, line 178: Think this needs to be localised since we're using it for user feedback?

joshxg (approved):
Badass

Just the localised strings!

src/widgets/Tabs/PatientSelect.js
katherine-sussol: Thanks, yeah probs a good idea! 228d105

joshxg (approved):
🎉

PR #4140, opened 2021-06-17
save addresses correctly

### LINKED ISSUE #4138: Patient lookup results in wrong address
## Describe the bug

When you have existing addresses in the database, after looking up a patient - the address shown on the edit screen is incorrect

### To reproduce

1. Create a patient on the server, specify lines 1 & 2 of the main ( billing ) address
2. Look this patient up on mobile
3. Edit the patient

### Expected behaviour

Address should be the same as on the server

mark-prins:
Fixes #4138 

## Change summary

Two issues with the addresses on looking up patients ( and prescribers )
- If all the fields are undefined then none of the address fields are strings, so the first address in the database is assigned to the incoming patient
- The properties on the incoming patient were incorrect

## Testing

- [ ] Create a patient on the server, with an address. Look them up on mobile: the address saved should be correct

### DIFF
--- a/src/actions/PatientActions.js
+++ b/src/actions/PatientActions.js
@@ -76,8 +76,11 @@ const patientUpdate = patientDetails => async (dispatch, getState) => {
     dateOfBirth: patientDateOfBirth,
     emailAddress: patientEmailAddress,
     phoneNumber: patientPhoneNumber,
-    addressOne: patientLine1,
-    addressTwo: patientLine2,
+    billAddress1: patientLine1,
+    billAddress2: patientLine2,
+    billAddress3: patientLine3,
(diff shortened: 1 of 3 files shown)


katherine-sussol (approved):
[LGTM](https://storage.googleapis.com/lgtm-cat-cats/6622e84f-9e34-4a9c-939f-e95a03ba2edd.gif)

🤣  I hope people haven't randomly been getting other people's prescriptions/invoices mailed to them

mark-prins:
Oh. Just noticed that Desktop records the comment `Age estimated` when the date is calculated based on the age
That isn't exactly accurate though, as the date is estimated, not the age! so I'm not suggesting that we do this. Merely an observation.

PR #4152, opened 2021-06-21
Add sensor to TemperatureLog incomingSync

### LINKED ISSUE #4151: TemperatureLog.sensor is not synced
## Describe the bug

in `incomingSyncUtils`, incoming `TemperatureLog` records are not related to their `Sensor`.

This means you don't get the 'current' temperature of a `Sensor` till later

### To reproduce

1. Go to sync a mobile store with an existing sensor+logs
2. Click on vaccines page

### Expected behaviour

N/A

### Proposed Solution

Add to `incomingSyncUtils`

### Version and device info

- App version:
- Tablet model:
- OS version:

joshxg:
Fixes #4151 

## Change summary

- Adds sensor to temperature log for incoming sync, ensuring temperature logs are related to a sensor when reinitialized.
- Offroad: Added hydrating of the entity stores after a reinit - currently on a full init the entity stores and refreshed, so you for example, get an empty sensor page- work around of some navigating can get you the state back but I don't want to have to explain more "Oh just navigate to this page and then that page and then presto boomo" or "have you tried restarting the app?"

## Testing

- [ ] Setup a mobile with a sensor and some temperature logs. Reinitialize. Check the records/current temperature of the sensor
- [ ] On the first navigation to 

N/A

### DIFF
--- a/src/reducers/Entities/LocationReducer.js
+++ b/src/reducers/Entities/LocationReducer.js
@@ -21,6 +21,7 @@ export const LocationReducer = (state = initialState(), action) => {
   const { type } = action;
 
   switch (type) {
+    case INITIALISE_FINISHED:
     case SYNC_TRANSACTION_COMPLETE: {
       const { byId } = state;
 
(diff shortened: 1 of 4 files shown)


joshxg:
Did this into master to give to Vanuatu.

They've had issues with the sensors 'not syncing' after reinit, which I can't recreate. However I did have this issue where I did not get a 'current temperature', so am wondering if that is what they're experiencing and assuming logs are not actually downloaded.

katherine-sussol (approved):
Strange, I don't think I've seen sensors not syncing after re-init either 🤔  but changes look good to me 😸

PR #4173, opened 2021-06-24
change code generation method

### LINKED ISSUE #4166: Patient codes are not unique
## Describe the bug

When creating patients on a tablet, the codes generated using the `NumberSequence` table are not unique if the tablet has been re-initialised. Tonga, for example, has 6,861 duplicate codes for patients.

### To reproduce

1. Create a patient
2. Re-initialise the tablet
3. Create another patient
4. Check the code of the patients

### Expected behaviour

We would like the generated code to be unique

### Proposed Solution

Create a patient code using a combination of sync id + enough characters from a uuid to bring the code to a total of 8 characters. Check for uniqueness and generate a different uuid if there is already a patient with that code.

Why? 
- Tonga would like to use the patient code in vaccination certificates and would like to have a unique code of constrained length
- I don't want to the patient code to be deterministic based on other patient information
- the codes need to be reliably unique

Other solutions:
- Sync the `NumberSequence` table to the server to that the sequence starts correctly when initialising. Note this doesn't fix existing Tonga tablets which are still creating duplicates
- Run a migration data update to determine the next number in the sequence when initialising - Would have to be run for existing tablets somehow

mark-prins:
Fixes #4166 

## Change summary

Generate a patient code independently of the `NumberSequence` as this doesn't guarantee uniqueness

## Testing

- [ ] Create patient, reset tablet, create another patient. Check that the code is unique and 8 characters long for both

### DIFF
--- a/src/database/utilities/constants.js
+++ b/src/database/utilities/constants.js
@@ -14,6 +14,7 @@ export const MILLISECONDS_PER_HOUR = MILLISECONDS_PER_MINUTE * MINUTES_PER_HOUR;
 export const MILLISECONDS_PER_DAY = MILLISECONDS_PER_HOUR * HOURS_PER_DAY;
 
 export const NUMBER_OF_DAYS_IN_A_MONTH = 365 / 12;
+export const PATIENT_CODE_LENGTH = 8;
 
 export const NUMBER_SEQUENCE_KEYS = {
   CUSTOMER_INVOICE_NUMBER: 'customer_invoice_serial_number',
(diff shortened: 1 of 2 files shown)


joshxg (approved):
More unique than the current unique codes!

katherine-sussol:
@mark-prins @joshxg I assume this should also go into release/8.1.1?

joshxg:
Oh uh.. yup!

mark-prins:
oh, yes please. Tonga have a lot of duplicate patient codes and want to use the codes as identifiers

PR #4184, opened 2021-06-28
#4182 Update getBadgeData.js

### LINKED ISSUE #4182: Breach notification badge should only display a badge when active sensors are in breach
## Describe the bug

Currently, the 'sensors in breach' badge count is incremented even if a sensor has been inactivated (but was inactivated while still in breach).

### To reproduce

1. Set up a sensor and have it enter a breach
2. Select 'Remove Sensor'
3. Cold Chain Badge count should go down (but it doesn't)

### Expected behaviour

Cold Chain Badge count should only increment if an _active_ sensor is in breach

### Proposed Solution

Add check for isActive flag on sensors when populating badge counts

### Version and device info

- App version: 8.1.0
- Tablet model:
- OS version:

### Additional context

N/A

katherine-sussol:
Fixes #4182

## Change summary
Added `sensor.isActive` check when calculating cold chain badge count
Targetting `release/8.1.1` because I think we would like it patched before the demo tomorrow.
Now I am unsure about `isPaused` as well but I think it's a little more obvious from the UI at least if a sensor is paused.

## Testing

- [ ] Reproduce #4182 

N/A

### DIFF
--- a/src/utilities/getBadgeData.js
+++ b/src/utilities/getBadgeData.js
@@ -18,7 +18,7 @@ const getBadgeData = routeName => {
       return [
         {
           count: UIDatabase.objects(dataType).filter(
-            sensor => sensor.isInHotBreach || sensor.isInColdBreach
+            sensor => sensor.isActive && (sensor.isInHotBreach || sensor.isInColdBreach)
           ).length,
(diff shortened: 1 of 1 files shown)


joshxg (approved):
👍

PR #4196, opened 2021-06-30
#4174 alert if patient already vaccinated in same day

### LINKED ISSUE #4174: Vaccine Module: Add an alert if patient has already received a vaccination today

A problem we are seeing is that users will often accidentally administer a vaccine to the same patient twice in one day. Normally this occurs minutes apart on the same device, so presumably a data entry error when a bunch of vaccinations are being entered in the system at one time.

It would be great if Mobile checked whether a patient had already been vaccinated in that store today, and gave an alert to the user e.g. "This patient already has a vaccination recorded for today, do you want to proceed?"

## Implementation

## Additional context

katherine-sussol:
Fixes #4174

## Change summary

- Pop up a modal asking the user to confirm if they want to proceed if a patient has already received a vaccine in the same day. Originally had a check if vaccine with the same item id as the vaccination in the history was selected but thought it might be too restrictive (sometimes it's bad to have more than 1 shot even if they are different shots? 🤷‍♀️ )
- Also I think this is a bit hacky so I would be keen to hear if there is a better way to do this without creating two modals 😭 , had some struggles trying to make the confirm button function conditional on which button (Confirm vs OK & Next was pressed)
- Also some weird stuff with the Confirm and OK & Next buttons because the `withOnePress` meant if the user cancelled out of the modal they wouldn't be able to press it again...also made the confirm button on the modal optionally one pressable only...I feel like there must be a better way to do this too haha.
 
## Testing

- [ ] Vaccinate a patient, try to vaccinate them again, should get a popup:
image

N/A

### DIFF
--- a/src/selectors/Entities/name.js
+++ b/src/selectors/Entities/name.js
@@ -4,6 +4,7 @@ import { UIDatabase } from '../../database';
 import { selectSpecificEntityState } from './index';
 import { DATE_FORMAT } from '../../utilities/constants';
 import { PREFERENCE_KEYS } from '../../database/utilities/preferenceConstants';
+import { MILLISECONDS_PER_DAY } from '../../database/utilities/constants';
 
 export const selectEditingNameId = state => {
   const NameState = selectSpecificEntityState(state, 'name');
@@ -88,10 +89,9 @@ export const selectVaccinePatientHistory = state => {
   return inQuery ? UIDatabase.objects('TransactionBatch').filtered(fullQuery).slice() : [];
 };
 
-export const selectWasPatientVaccinatedToday = state => {
+export const selectWasPatientVaccinatedWithinOneDay = state => {
   const history = selectVaccinePatientHistory(state);
+  const oneDayAgo = new Date().getTime() - MILLISECONDS_PER_DAY;
 
-  const currentDate = new Date();
-  return !!history.filter(historyRecord => historyRecord.confirmDate >= currentDate.getDate() - 1)
-    .length;
+  return !!history.filter(historyRecord => historyRecord.confirmDate.getTime() >= oneDayAgo).length;
 };
--- a/src/selectors/Entities/name.js
+++ b/src/selectors/Entities/name.js
@@ -87,3 +87,11 @@ export const selectVaccinePatientHistory = state => {
   const fullQuery = `(${inQuery}) AND ${baseQueryString} AND itemBatch.item.isVaccine == true`;
   return inQuery ? UIDatabase.objects('TransactionBatch').filtered(fullQuery).slice() : [];
 };
+
+export const selectWasPatientVaccinatedToday = state => {
+  const history = selectVaccinePatientHistory(state);
+
+  const currentDate = new Date();
+  return !!history.filter(historyRecord => historyRecord.confirmDate >= currentDate.getDate() - 1)
+    .length;
+};
--- a/src/selectors/Entities/name.js
+++ b/src/selectors/Entities/name.js
@@ -93,5 +93,5 @@ export const selectWasPatientVaccinatedWithinOneDay = state => {
   const history = selectVaccinePatientHistory(state);
   const oneDayAgo = new Date().getTime() - MILLISECONDS_PER_DAY;
 
-  return !!history.filter(historyRecord => historyRecord.confirmDate.getTime() >= oneDayAgo).length;
+  return !!history.filter(historyRecord => historyRecord.confirmDate.getTime() > oneDayAgo).length;
 };
--- a/src/widgets/PaperModal/PaperConfirmModal.js
+++ b/src/widgets/PaperModal/PaperConfirmModal.js
@@ -36,7 +36,7 @@ export const PaperConfirmModal = ({
 );
 
 const localStyles = StyleSheet.create({
-  questionText: { fontSize: 16, color: DARKER_GREY, fontWeight: '700' },
+  questionText: { fontSize: 16, color: DARKER_GREY, fontWeight: '700', textAlign: 'center' },
   cancelText: { textTransform: 'uppercase' },
   confirmText: { textTransform: 'uppercase', color: WHITE },
   confirmButton: { backgroundColor: SUSSOL_ORANGE },
--- a/src/widgets/PaperModal/PaperConfirmModal.js
+++ b/src/widgets/PaperModal/PaperConfirmModal.js
@@ -8,7 +8,6 @@ import { Spacer } from '../Spacer';
 import { DARKER_GREY, SUSSOL_ORANGE, WHITE } from '../../globalStyles/index';
 import { FlexRow } from '../FlexRow';
 import { PageButton } from '../PageButton';
-import { PageButtonWithOnePress } from '../PageButtonWithOnePress';
 
 export const PaperConfirmModal = ({
   Icon,
@@ -17,7 +16,6 @@ export const PaperConfirmModal = ({
   cancelText,
   onConfirm,
   onCancel,
-  withOnePress,
 }) => (
   <FlexColumn alignItems="center" justifyContent="center" flex={1} style={{ padding: 20 }}>
     <FlexView flex={1} alignItems="center" justifyContent="center">
@@ -27,21 +25,12 @@ export const PaperConfirmModal = ({
     </FlexView>
     <FlexRow justifyContent="space-evenly" alignItems="center" style={{ width: '100%' }} flex={1}>
       <PageButton onPress={onCancel} text={cancelText} textStyle={localStyles.cancelText} />
-      {withOnePress ? (
-        <PageButtonWithOnePress
-          onPress={onConfirm}
-          text={confirmText}
-          textStyle={localStyles.confirmText}
-          style={localStyles.confirmButton}
-        />
-      ) : (
-        <PageButton
-          onPress={onConfirm}
-          text={confirmText}
-          textStyle={localStyles.confirmText}
-          style={localStyles.confirmButton}
-        />
-      )}
+      <PageButton
+        onPress={onConfirm}
+        text={confirmText}
+        textStyle={localStyles.confirmText}
+        style={localStyles.confirmButton}
+      />
     </FlexRow>
   </FlexColumn>
 );
@@ -55,7 +44,6 @@ const localStyles = StyleSheet.create({
 
 PaperConfirmModal.defaultProps = {
   Icon: <HazardIcon size={50} />,
-  withOnePress: false,
 };
 
 PaperConfirmModal.propTypes = {
@@ -65,5 +53,4 @@ PaperConfirmModal.propTypes = {
   cancelText: PropTypes.string.isRequired,
   onConfirm: PropTypes.func.isRequired,
   onCancel: PropTypes.func.isRequired,
-  withOnePress: PropTypes.bool,
 };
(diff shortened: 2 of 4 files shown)


src/widgets/Tabs/VaccineSelect.js
```diff
@@ -94,6 +103,15 @@ const VaccineSelectComponent = ({
...
   );
   const runWithLoadingIndicator = useLoadingIndicator();
+
+  const wasVaccinatedToday = () => {
+    const currentDate = new Date();
+
+    return !!vaccinePatientHistory.filter(
+      history => history.confirmDate >= currentDate.getDate() - 1
+    ).length;
+  };
```
joshxg, line 113: I'm not sure of the requirement but when my daughter goes for her shots she gets like 3 vaccines in a day, so from my anecdotal evidence (the strongest form), getting multiple vaccines in a day is common?

src/widgets/Tabs/VaccineSelect.js
```diff
@@ -147,19 +165,65 @@ const VaccineSelectComponent = ({
...
+            onPress={toggleConfirmDoubleDoseModal}
+          />
+        ) : (
+          <PageButtonWithOnePress
+            text={buttonStrings.confirm}
+            style={{ marginLeft: 'auto' }}
+            isDisabled={!selectedBatches && !hasRefused}
+            onPress={confirmPrescription}
+          />
+        )}
```
joshxg, line 183: Maybe just use `useDebounce` on the `confirmPrescription` callback, and use `PageButton` for both cases? 

```
const debouncedConfirmPrescription = useDebounce(confirmPrescription);
const confirmOnPress = wasVaccinatedToday() ? debouncedConfirmPrescription : toggleConfirmDoubleDoseModal`
...
onPress={confirmOnPress}
/>
```

src/widgets/Tabs/VaccineSelect.js
```diff
@@ -147,19 +165,65 @@ const VaccineSelectComponent = ({
...
+            onPress={toggleConfirmAndRepeatDoubleDoseModal}
+          />
+        ) : (
+          <PageButtonWithOnePress
+            text={generalStrings.ok_and_next}
+            style={{ marginLeft: 5 }}
+            isDisabled={!selectedBatches && !hasRefused}
+            onPress={confirmAndRepeatPrescription}
+          />
+        )}
```
joshxg, line 199: same suggestion as above.. as long as we prevent the accidental double click on the buttons we can just use `PageButton`

src/widgets/Tabs/VaccineSelect.js
```diff
@@ -147,19 +165,65 @@ const VaccineSelectComponent = ({
...
+        isVisible={confirmAndRepeatDoubleDoseModalOpen}
+        onClose={toggleConfirmAndRepeatDoubleDoseModal}
+      >
+        <PaperConfirmModal
+          questionText={modalStrings.confirm_double_dose}
+          confirmText={modalStrings.confirm}
+          cancelText={modalStrings.cancel}
+          onConfirm={confirmAndRepeatPrescription}
+          onCancel={toggleConfirmAndRepeatDoubleDoseModal}
+          withOnePress={true}
```
joshxg, line 224: Could we do something like:

```
const modalOnConfirm = confirmDoubleDoseModalOpen ? confirmPrescription : confirmRepeatPrescription
const modalOnCancel = confirmDoubleDoseModalOpen ? toggleConfirmDoubleDoseModal : toggleConfirmAndRepeatDoubleDoseModal
...
```

and use the same modal?

Also just noticed the `PageButton` has a `debounceTimer` prop

src/widgets/PaperModal/PaperConfirmModal.js
```diff
@@ -25,25 +27,35 @@ export const PaperConfirmModal = ({
...
+          style={localStyles.confirmButton}
+        />
+      ) : (
+        <PageButton
+          onPress={onConfirm}
+          text={confirmText}
+          textStyle={localStyles.confirmText}
+          style={localStyles.confirmButton}
+        />
+      )}
```
joshxg, line 44: I think the `PageButtonWithOnePress` is generally just a bit problematic? Could just use `PageButton` and solve the spam clicking problem through the callback

Using `debounce` or..

```
const singleInvoke = (fn)  => {
    let hasRun = false;
    return (...args) => {
        if (!hasRun) {
            hasRun = true;
            fn(...args);
        }
    };
};

```

joshxg (changes requested):
Hmm few suggestions which are a bit all over the place, sorry!

General suggestion is that if the `OnePress` functionality is causing problems we might be able to solve it easier by debouncing?

src/widgets/PaperModal/PaperConfirmModal.js
katherine-sussol: Yeah the OnePress thing was messing everything up 😆 . Thanks, I missed the callback on the `PageButton`, will give that a go

src/widgets/Tabs/VaccineSelect.js
katherine-sussol: Hmm, yeah I think the requirement came in the context of COVID (where it's just one shot) but for vaccines in general that makes sense...maybe it would be better to check for duplicated vaccines then, rather than just that they have had a vaccine?

joshxg: Yeah you're right it can just be dismissed and whatever- either works for me and I'm sure if its a problem we can get feedback and change it 🤷

src/widgets/PaperModal/PaperConfirmModal.js
katherine-sussol: Resolved the `PageButtonWithOnePress` stuff with c58ff4a and 0b02810

src/widgets/Tabs/VaccineSelect.js
katherine-sussol: > ```
> const modalOnConfirm = confirmDoubleDoseModalOpen ? confirmPrescription : confirmRepeatPrescription
> const modalOnCancel = confirmDoubleDoseModalOpen ? toggleConfirmDoubleDoseModal : toggleConfirmAndRepeatDoubleDoseModal
> ...
> ```
> 
> and use the same modal?

I'm not quite following sorry 🤣  - if the modal is open it shouldn't necessarily impact which function to execute I think? It just means they have either clicked `Confirm` or `OK & Next` but we don't know which - I was thinking I could add something to the PageButton onPress to store which one was last clicked, but not sure how 🤔

joshxg: It's possibly a terrible idea, happy for you to say "no ew". It's more using which modal is open rather than if the modal is open to determine the callback.

```
const modalOnClose = confirmDoubleDoseModalOpen
  ? toggleConfirmDoubleDoseModal
  : toggleConfirmAndRepeatDoubleDoseModal;

const modalOnConfirm = confirmDoubleDoseModalOpen
  ? confirmPrescription
  : confirmAndRepeatPrescription;

const modalOpen = confirmDoubleDoseModalOpen || confirmAndRepeatDoubleDoseModalOpen;

const Modal = () => (
  <PaperModalContainer isVisible={modalOpen} onClose={modalOnClose}>
    <PaperConfirmModal
      questionText={modalStrings.confirm_double_dose}
      confirmText={modalStrings.confirm}
      cancelText={modalStrings.cancel}
      onConfirm={modalOnConfirm}
      onCancel={modalOnClose}
    />
  </PaperModalContainer>
);
```

I think this works..?

katherine-sussol: Ohhhh thanks I see, so still maintaining two toggle states for the different modals but using only one modal component.
Haha okay. Maybe a bit ew but I think a lot less ew than my version 😹

katherine-sussol: Works nicely :) - yoink 57b1f036 (and 3a847908 for consistency)

src/widgets/Tabs/VaccineSelect.js
```diff
@@ -94,6 +103,15 @@ const VaccineSelectComponent = ({
     [vaccines]
   );
   const runWithLoadingIndicator = useLoadingIndicator();
+
+  const wasVaccinatedToday = () => {
```
joshxg, line 107: Nit: Could make this a selector and have it come in as a prop

joshxg (approved):
niiice 👍

mark-prins:
I've just tested, and I'm being prompted for someone who has been vaccinated on the 20th.

katherine-sussol:
> I've just tested, and I'm being prompted for someone who has been vaccinated on the 20th.

Oh no 🙀 . Hmm will test some more

src/selectors/Entities/name.js
```diff
@@ -87,3 +87,11 @@ export const selectVaccinePatientHistory = state => {
   const fullQuery = `(${inQuery}) AND ${baseQueryString} AND itemBatch.item.isVaccine == true`;
   return inQuery ? UIDatabase.objects('TransactionBatch').filtered(fullQuery).slice() : [];
 };
+
+export const selectWasPatientVaccinatedToday = state => {
+  const history = selectVaccinePatientHistory(state);
+
+  const currentDate = new Date();
+  return !!history.filter(historyRecord => historyRecord.confirmDate >= currentDate.getDate() - 1)
```
mark-prins, line 95: `getDate()` returns day of the month!
I'm getting `30` for getDate, and the patient I'm using was vaccinated on 30th May, so it won't let me vaccinate today 😜

mark-prins: Am also concerned that I won't be able to vaccinate today if someone was jabbed yesterday

katherine-sussol: Haha.... oops. Better to just check current date is the same? (ignoring timestamp) I guess no one should be doing vaccinations at midnight or anything like that?

katherine-sussol: Oh it's all wrong, I think I somehow left out the bit that checked the actual transaction too... as the history record itself doesn't even have a confirmDate 🤦

src/selectors/Entities/name.js
```diff
@@ -87,3 +87,11 @@ export const selectVaccinePatientHistory = state => {
...
+    .length;
```
mark-prins, line 96:
```suggestion
  const history = selectVaccinePatientHistory(state);
  const currentDate = new Date().getTime() - MILLISECONDS_PER_DAY;

  return !!history.filter(historyRecord => historyRecord.confirmDate.getTime() > currentDate)
    .length;
```

src/selectors/Entities/name.js
mark-prins: The below is my suggestion :) 
Comparing the javascript version of unix time.. so we limit to vaccinations within the last 24hours.

along with this, of course:
`import { MILLISECONDS_PER_DAY } from '../../database/utilities/constants';`

mark-prins: > as the history record itself doesn't even have a confirmDate 

I thought the same! and couldn't see why it was working, and getting a confirmDate! but you were right, both times :)

src/selectors/Entities/name.js
katherine-sussol: Thanks, I took your suggestion and got a bit carried away with the modalString and variable names 😅  2217ced

mark-prins (approved):
[](https://media3.giphy.com/media/3osxYdek8wYWCOLgT6/giphy.gif)

so good, it needed a gif :)

Issue #4198, opened 2021-06-30
When saving vaccination site data - store in [transact]custom_data as an object


katherine-sussol:
Child of #4176 
 
## Is your feature request related to a problem? Please describe.
Tonga (and Kiribati??) want #4176 and this is a step required to achieve it.

## Describe the solution you'd like
See #4176

## Implementation
- When '`Confirm`' (or '`OK & Next`') are selected at the end of the vaccine dispensing flow (currently the screen where the vaccine is selected), ensure that the data entered in the form is stored in `[transact]custom_data as an object`

## Describe alternatives you've considered
Making someone unhappy? Maybe this didn't even need to be a sub-issue 🤔  can close two issues with one PR if needed though.

## Additional context
N/A

katherine-sussol:
Closed by #4209

ArpitaSussol:
Tested in mSupply V8.2.3 with Tonga-vax datafile

## Testing

Steps to reproduce or otherwise test the changes of this PR:

- [x] Create a prescription with siteData (needs FormSchema with type `VaccineSite`), check Realm Explorer (and/or export and check the contents that way) and verify that the siteData is saved to the transaction
- [x] Sync to Desktop - check via record browser that [transact].custom_data contains the site data for the vaccine prescription just created

PR #4205, opened 2021-07-02
#4176 insert step for site of vaccination

### LINKED ISSUE #4176: Vaccine dispensing: insert step for site of vaccination

Tonga would like to record the following additional data as part of the vaccination process:

- Island ( drop down field )
- Village ( drop down field, dependent on the selected Island )
- Free text field for ‘Administering Location’

- Insert a step in the wizard process which asks for
- Island, Village and administering location
- Save this data against the transaction

## Implementation

- Populate mSupply region table using the new import feature
- Create region properties to save population and geolocation data
- After this we can remove the district and population tables used by the dashboard
- Create a new JSON schema and save in mSupply ( generate island and village from the region table )
- #4197
- #4210
- #4198

katherine-sussol:
Fixes #4176

## Change summary

- Adds a step after patient editing to load a JSON form and save the submitted data into vaccinePrescription state
- Fiddled with tab heading of vaccine select page so step number is dynamic for final tab
- Hide added tab if there isn't a formSchema with type of `VaccineSite` present

## Testing

- [ ] Use realm studio to manually add a FormSchema to the realm data file and upload to device (could probably also be done on the mSupply side with 4D code and synced 😹 ), then navigate through vaccine dispensing flow

Sample schemas used (just to check stuff could be submitted):
- JSONSchema
```
{
  "title": "Step 3: Specify the vaccination site",
  "type": "object",
  "required": [],
  "properties": {
    "administeringLocation": {
      "type": "string",
      "title": "Administering Location"
    }
  }
}
```

- UISchema
```
{
  "administeringLocation": {
    "ui:emptyValue": "",
    "ui:autocomplete": ""
  }
}
```

- Nothing is done with this data yet (when the vaccine dispensing is completed it should do some stuff #4198)
- Not sure about which release version to target 🙀 
- The header style on the JSON form doesn't match the rest of the wizard 🧙 .


mark-prins:
This was my mSupply method to create the form 😉 

```
C_OBJECT($obJsonSchema; $obUiSchema; $obFormSchema)

$obJsonSchema:=New object()
$obJsonSchema.type:="object"
$obJsonSchema.title:="Step 3: Specify the vaccination site"
$obJsonSchema.required:=New collection()
$obJsonSchema.properties:=New object()
$obJsonSchema.properties.administeringLocation:=New object()
$obJsonSchema.properties.administeringLocation.type:="string"
$obJsonSchema.properties.administeringLocation.title:="Administering Location"

$obUiSchema:=New object()
$obUiSchema.administeringLocation:=New object()
$obUiSchema.administeringLocation["ui:emptyValue"]:=""
$obUiSchema.administeringLocation["ui:autocomplete"]:=""

$obFormSchema:=ds.form_schema.new()
$obFormSchema.ID:=UUID_generate
$obFormSchema.type:="VaccineSite"
$obFormSchema.version:=1

$obFormSchema.json_schema:=$obJsonSchema
$obFormSchema.ui_schema:=$obUiSchema

$obFormSchema.save()
```

mark-prins:
How about targeting v8.1.1 - there's nothing else going in to the release afaik

katherine-sussol:
> How about targeting v8.1.1 - there's nothing else going in to the release afaik

Sounds good 😸  - will leave the base as `release/8.1.1 `👍

src/widgets/Tabs/VaccineSiteSelect.js
```diff
@@ -0,0 +1,85 @@
...
+import { VaccinePrescriptionActions } from '../../actions/Entities/index';
+import { WizardActions } from '../../actions/WizardActions';
+
+const { pageTopViewContainer } = globalStyles;
+
+const VaccineSiteSelectComponent = ({ onCancel, onComplete, siteSchema }) => {
+  const [{ formData, isValid }, setForm] = useState({ formData: null, isValid: false });
+  return (
+    <FlexView style={pageTopViewContainer}>
+      <JSONForm
```
mark-prins, line 33: I wrapped this in `Paper` to get the same header style.. though that will require the JSON form to lose it's paper-ness in order to look right. Either that or pass in a header style to the JSON form.. either way, it would be nice to update the header style.

mark-prins (approved):
that's very nice! 😄 
works well, with my only real concern being the style of the header, which you know about already

src/widgets/Tabs/VaccineSiteSelect.js
katherine-sussol: Tried the `Paper` method...Removed title from the JSON Schema and made the JSON Form Paperisation dependant on the presence of a title 
83f21d0
76bc1a3

```
{
  "type": "object",
  "required": [],
  "properties": {
    "administeringLocation": {
      "type": "string",
      "title": "Administering Location"
    }
  }
}
```

image

src/localization/vaccineStrings.json
```diff
@@ -2,9 +2,11 @@
   "gb": {
     "no_vaccine_stock": "You have no vaccine stock to dispense!",
     "vaccines": "Vaccines",
+    "vaccine_step": "Step:",
     "vaccine_dispense_step_one_title": "Step: 1 Select a patient",
     "vaccine_dispense_step_two_title": "Step: 2 Edit the patients details",
-    "vaccine_dispense_step_three_title": "Step: 3 Select the vaccine",
+    "vaccine_dispense_vaccine_select_title": "Select the vaccine",
+    "vaccine_dispense_site_select_title": "Step: 3 Select the site",
```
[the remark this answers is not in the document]
katherine-sussol: I wasn't too sure what to do with this 😹  - site selection won't always be step 3 (sometimes step 3 will be the vaccine selection) so thought that might be even more confusing. Alternatively thought about renaming all of the steps to not have the number in them (and let the wizard determine the step)

src/widgets/JSONForm/templates/Field.js
```diff
@@ -59,7 +59,7 @@ export const Field = ({
   );
 
   if (schemaIsRoot(id)) {
-    return (
+    return label ? (
```
joshxg, line 62: Should this be `!label`?

joshxg (commented):
I'm late to the party and playing catchup. Going to be annoying. Sorry.

I'm a little concerned that this is not hugely extensible? It's seems to be tightly coupled to the 'site' of vaccination- is the json schema basically redundant if we have such specific components?

Also reading the issues it seems in this 

To be annoying I am going to suggest an alternative implementation even though I feel terrible for not seeing this and suggesting things earlier. Feel free to ignore it.

- Make step 2 only patient details
- Put all the JSONForms in step 3
- Make the schema with `site` something like `vaccination_transaction_data` (..?)

Not much different, but feel getting away from the `site` naming and putting all the schemas in the same component might make it clearer?

katherine-sussol:
Thanks @joshxg for the input, that is a good point r.e. extensibility. Maybe as an intermediary (just so I can be lazy and not 
 redo too much 🤣 ), we just de-couple the component/schema from the site, (i.e. point 3 in your comment)...and just call it `VaccineSupplementalData` or something (and rename the schema accordingly) 🤔.

I don't think it's _too_ terrible to have this form in a separate tab as it's more 'vaccination' specific whereas the PCD form is patient-specific, so kind of works from a flow perspective, I think?

joshxg:
sounds good to me!

src/widgets/JSONForm/templates/Field.js
katherine-sussol: The idea (and maybe it is hacky) was to not display the paper wrapper with header if there isn't a title present in the JSONSchema 😅  (for when the `JSONFrom` is loaded inside a component with its own title)

joshxg (commented):
Sorry for being annoying 🤣

katherine-sussol:
> Sorry for being annoying 🤣

🤣  I'd be the sorry one if Tonga suddenly asked for some new non-site related field ('needle type'?  🤷‍♀️ ) so thanks hah

PR #4209, opened 2021-07-05
#4198 save site data on vaccine dispense

### LINKED ISSUE #4198: When saving vaccination site data - store in [transact]custom_data as an object

katherine-sussol:
Fixes #4198

## Change summary

- Adds `Transaction.customData` to mobile schema/sync utils
- Save siteData into `Transaction.customData` when creating a vaccinePrescription if it exists

## Testing

- [ ] Create a prescription with siteData (needs FormSchema with type `VaccineSite`), check Realm Explorer (and/or export and check the contents that way) and verify that the siteData is saved to the transaction
- [ ] Sync to Desktop - check via record browser that [transact].custom_data contains the site data for the vaccine prescription just created 

Dispensing without siteData (and regular customer invoices) should not create any `customData` for the transactions

### DIFF
--- a/src/database/utilities/createRecord.js
+++ b/src/database/utilities/createRecord.js
@@ -566,7 +566,7 @@ const createCustomerRefundLine = (database, customerCredit, transactionBatch) =>
  * @param   {Name}         customer  Customer associated with invoice.
  * @return  {Transaction}
  */
-const createCustomerInvoice = (database, customer, user, mode = 'store') => {
+const createCustomerInvoice = (database, customer, user, mode = 'store', customData = null) => {
   const { CUSTOMER_INVOICE_NUMBER } = NUMBER_SEQUENCE_KEYS;
   const currentDate = new Date();
   const invoice = database.create('Transaction', {
@@ -581,6 +581,7 @@ const createCustomerInvoice = (database, customer, user, mode = 'store') => {
     otherParty: customer,
     enteredBy: user,
     mode,
+    customData: JSON.stringify(customData),
   });
 
   database.save('Transaction', invoice);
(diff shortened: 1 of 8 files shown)


katherine-sussol:
Oops something weird happened with my diff 🙀

src/database/utilities/createRecord.js
```diff
@@ -566,7 +566,7 @@ const createCustomerRefundLine = (database, customerCredit, transactionBatch) =>
  * @param   {Name}         customer  Customer associated with invoice.
  * @return  {Transaction}
  */
-const createCustomerInvoice = (database, customer, user, mode = 'store') => {
+const createCustomerInvoice = (database, customer, user, mode = 'store', customData = null) => {
```
joshxg, line 569: `JSON.stringify(null) === 'null'` - maybe not what we want? think just `''` when empty?

joshxg (approved):
sweet

src/database/utilities/createRecord.js
katherine-sussol: Eep, thanks

katherine-sussol: Did a bit more testing, think it's better to set the record value to `null` if not provided (for consistency with `parseJSONData` in `incomingSyncUtils`) so transactions without `customData` in mobile are consistent with those synced from Desktop 🤣  
(prev impl had records from Desktop with `null` and those from Mobile with `''` - oops)

Soz bit late response as well, too much fun testing different JSON schemas out 😹 
25641e3

joshxg: Ah, good points!

PR #4215, opened 2021-07-06
#4204 RequisitionLines should save ItemName

### LINKED ISSUE #4204: RequisitionLines should save `itemName`
## Describe the bug

On desktop, there is a field `itemName` on requisition lines. This preserves what the `requisitionLine.item.name` was at the time the `requisitionLine` was finalised. Mobile does not do this, so the central server doesn't have a record of this preserved item name.

For more context, this is similar to prescription transaction lines (`TransactionBatches` in mobile) saving the item name at for prescriptions to keep track of what the intent was at the time of dispensing.

Another example that's even easier to reason about, we record the cost/sell price on transaction lines. Otherwise, editing cost/sell price on the items would mess up the accounts of all invoices!

OK, point being yes please record `itemName`!

### To reproduce

1. Um... Look at the schema, we don't do it!

### Expected behaviour

When a requisitionLines are created, save the field `requisitionLine.itemName = requisitionLine.item.name`. Include with outgoing/incoming sync. Show this field value in the table for supplier requisitions.

### Proposed Solution

WIP: 

Should really display `itemName` field in the item name column rather than `requisitionLine.item.name` which I assume it currently does, but that might be a demon (I only loosely know how pages are configured and if it'd support this). 

Should we migrate? Would have to if we change the table column source! Muddies the history, but I think best bet is to update `requisitionLine.itemName = requisitionLine.item.name` with whatever the current name of the item is. 

Grey: Customer requisitions? The user will be supplying whatever they really have in stock, right? So should display the `requisitionLine.item.name`. Could (should?) keep updating `itemName` on opening the requisition until finalisation. Or make it part of sync post processing?

### Version and device info

- App version: <= v8.1.0

### Additional context

Remote authorisation web app currently shows this value. For requisitions from mobile, there are no item names!

wlthomson:
Fixes #4204.

## Change summary

Has been a while since I've done one of these, feedback appreciated!

Still awaiting confirmation that this approach will do the job, but as far as I can tell it covers the requirements. Just adds a setter for updating `item.name` which has a side-effect (*gasp*) of propagating through to all linked unfinalised `requisitionItems`.

## Testing

- [ ] Migrate existing data file to `8.2.2` (`package.json.version`).
  - [ ] No errors occur during migration.
  - [ ] App is still usable (no prolonged freezing of the app).
  - [ ] Existing requisition items are not missing names (check latest supplier/customer requisitions).
- [ ] Create a customer requistion and add an arbitrary item. Change the item name on desktop. Sync the changes and reopen the requisition. The requisition line item name is updated.
- [ ] Repeat the above, but finalise the requisition before syncing the item name change. The requisition line item name is not updated.
- [ ] Create a supplier requisition and add an arbitrary item. Change the item name on desktop. Sync the changes and reopen the requisition. The requisition line item name is updated.
- [ ] Repeat the above, but finalise the requisition before syncing the item name change. The requisition line item name is not updated.

I was contracted as a code monkey 🐵 for this task. Ask @Chris-Petty.


src/database/DataTypes/Item.js
```diff
@@ -272,6 +272,16 @@ export class Item extends Realm.Object {
...
   }
+
+  /**
+   * Updates name and linked requisition `itemName` data. Finalised requisitions are not updated.
+   */
+  updateName(name) {
+    this.name = name;
+    this.requisitionItems.filtered("requisition.status != 'finalised'").forEach(requisitionLine => {
+      requisitionLine.itemName = name;
+    });
```
wlthomson, line 283: I can't remember if it's good practice to use curlies for `forEach` one-liners... probably not, but eslint didn't complain.

src/database/DataTypes/RequisitionItem.js
```diff
@@ -294,6 +285,7 @@ RequisitionItem.schema = {
     id: 'string',
     requisition: { type: 'Requisition', optional: true },
     item: { type: 'Item', optional: true },
+    itemName: 'string',
```
joshxg, line 288: The `item` is `optional` so maybe in some edge case there's no `item` to get the `itemName`?  Does it need to be `optional` also? Also the existing lines?

src/database/DataTypes/Item.js
joshxg: Will need to `UIDatabase.save()` the `RequisitionItem` records so that they are re-synced.

joshxg (changes requested):
Great solution! Couple minors.

I also think we need to consider migration. Essentially just need to iterate over all requisition items (which aren't finalised, I guess) and assign the related items name to `itemName`.

A little concerned about the migration as there are sites with 10's of 1000's of requisitions..

src/database/DataTypes/RequisitionItem.js
wlthomson: > 
> 
> Great solution! Couple minors.
> 
> I also think we need to consider migration. Essentially just need to iterate over all requisition items (which aren't finalised, I guess) and assign the related items name to `itemName`.
> 
> A little concerned about the migration as there are sites with 10's of 1000's of requisitions..

That's a very good point. I'm assuming the majority of these are finalised (?), if so, this might be mitigated by filtering on finalised (since Realm is lazy, this should avoid pulling in any finalised records). Does Realm for RN provide any decorators for indexing fields etc?

joshxg: You can add `indexed` into the schema declaration object

wlthomson: Follow up: under what circumstances are the items undefined? Would this only be due to a partial sync? In that case, would it be better to do something like below?

```
itemName: { type: 'string', default: 'placeholder' },
```

wlthomson:
> 
> 
> Great solution! Couple minors.
> 
> I also think we need to consider migration. Essentially just need to iterate over all requisition items (which aren't finalised, I guess) and assign the related items name to `itemName`.
> 
> A little concerned about the migration as there are sites with 10's of 1000's of requisitions..

Have thought about this a bit, and I think there is a bit of a problem here.

> On desktop, there is a field itemName on requisition lines. This preserves what the requisitionLine.item.name was at the time the requisitionLine was finalised. Mobile does not do this, so the central server doesn't have a record of this preserved item name.

As I understand it, this means we need to migrate both unfinalised and finalised `requisitionItem` objects. Unfinalised is OK, as we can just set `itemName` to `item.name`, but what is the expectation for existing finalised `requisitionLines`? If we set it to the current `item.name`, we are potentially rewriting history, but if we don't, those `requisitionLine` objects will have blank names? As @joshxg says, updating all `requisitionItem` objects is also going to be expensive!

src/database/DataTypes/Item.js
joshxg: Missing a paramater on the change: https://github.com/sussol/react-native-database/blob/81e18edcfe935d796412fa8ffcae6ba69c2a4e94/src/Database.js#L163

src/database/DataTypes/RequisitionItem.js
joshxg: I guess some sync thing for the item being undefined. I'm unsure, it's just theoretically possible (so 100% going to happen!).

joshxg:
You're understanding of the expectations would be the same as mine - little to none!

Up to you how to proceed, but my suggestion:

- Check a large deployment like PNG and determine the average number of requisition lines on mobile sites.
- Test migration against at least the average number of requisition lines. There is some X time which the migration takes that makes it not feasible to migrate in this way. Maybe we're at X?

If we knew this, we would know if we migrate on mobile or need to optimise by migrating on desktop and violating some rules.

Couple notes:
- We mainly need to migrate the finalised requisitions, as the unfinalised requisitions will still be synced at some point when they're finalised.
- We might need to test against a greater number of lines. If we add the migration code to 8.X, and PNG are on 5.X and don't upgrade for a long time- maybe they have double the number of requisition lines, and eventually hit the migration code.

wlthomson:
> * We mainly need to migrate the finalised requisitions, as the unfinalised requisitions will still be synced at some point when they're finalised.

Unless we add some default logic, this means unfinalised requisitions will all be for items called `"placeholder"`!

> * We might need to test against a greater number of lines. If we add the migration code to 8.X, and PNG are on 5.X and don't upgrade for a long time- maybe they have double the number of requisition lines, and eventually hit the migration code.

I agree with this; I think if the migration doesn't scale, then it's probably something we should avoid. My current thinking now is to add some backup logic, e.g. allowing `itemName` to be optional and adding a check to iterate through `RequisitionItem` objects on opening a requisition and populate `itemName` from `item.name`. My only concern is that this might be slow for big requisitions?

Chris-Petty (commented):
Ayyy that's way better than my harebrained ideas! Nice

Chris-Petty:
> > * We mainly need to migrate the finalised requisitions, as the unfinalised requisitions will still be synced at some point when they're finalised.
> 
> Unless we add some default logic, this means unfinalised requisitions will all be for items called `"placeholder"`!
> 

This is kind of why I was bashing around the idea of flicking the column data source based on finalised/unfinalised. BUT I really like that your current solution doesn't really muddy the page logic 😕 .

> > * We might need to test against a greater number of lines. If we add the migration code to 8.X, and PNG are on 5.X and don't upgrade for a long time- maybe they have double the number of requisition lines, and eventually hit the migration code.
> 
> I agree with this; I think if the migration doesn't scale, then it's probably something we should avoid. My current thinking now is to add some backup logic, e.g. allowing `itemName` to be optional and adding a check to iterate through `RequisitionItem` objects on opening a requisition and populate `itemName` from `item.name`. My only concern is that this might be slow for big requisitions?

I think give a shot migrating it. If it's not 10s of millions of lines, then should be alright I think.

katherine-sussol:
I'm going to be annoying and poke my head in :P do you think it's worth targetting release/8.1.1? I've got some changes for #4176 and thought it might be good to roll them both into one patch (so we don't need to do 2 releases?)

wlthomson:
Have tested with a test datafile, but sounds like it would be a good idea to also test with a big datafile (e.g. PNG)?

joshxg:
> Have tested with a test datafile, but sounds like it would be a good idea to also test with a big datafile (e.g. PNG)?

Yeah. Would be great if you could upload that somewhere accessible once you've finished your testing

Let me know how the testing goes and i'll revisit the review then!

wlthomson:
> 
> 
> > Have tested with a test datafile, but sounds like it would be a good idea to also test with a big datafile (e.g. PNG)?
> 
> Yeah. Would be great if you could upload that somewhere accessible once you've finished your testing
> 
> Let me know how the testing goes and i'll revisit the review then!

The PNG data file is > 32gb, and wouldn't really want to link it anywhere publically accessible! @Chris-Petty if no offline copy available, do you have any thoughts on who is the best person to test this?

joshxg (commented):
My testing (only on develop on a Lenovo, should also test on a built APK), resulted in (average of 5 runs):

- 1000 records: .2s
- 10,000 records: 1.8s
- 50,000 records: 8.5s
- 100,000 records: 16.9s
- 300,000 records: 44.9s
- 500,000 records: 81.2s
- 1m records: `Error: mmap() failed: Out of memory size: 17158136 offset: 143654912` - failed every time

The results make me a little concerned about putting this migration in. Some things that I think we would need to work on for this:
- Chunk this to bring less into memory in case we try to migrate users with 1m+ records
- Set up a mechanism to give feedback to the user for migrating as its likely a lot of users will hit a 20-30second white screen on the app when upgrading which would be confusing.

Or.. just do the migration on desktop, somehow 🤷

wlthomson:
> 
> 
> My testing (only on develop on a Lenovo, should also test on a built APK), resulted in (average of 5 runs):
> 
>     * 1000 records: .2s
> 
>     * 10,000 records: 1.8s
> 
>     * 50,000 records: 8.5s
> 
>     * 100,000 records: 16.9s
> 
>     * 300,000 records: 44.9s
> 
>     * 500,000 records: 81.2s
> 
>     * 1m records: `Error: mmap() failed: Out of memory size: 17158136 offset: 143654912` - failed every time
> 
> 
> The results make me a little concerned about putting this migration in. Some things that I think we would need to work on for this:
> 
>     * Chunk this to bring less into memory in case we try to migrate users with 1m+ records
> 
>     * Set up a mechanism to give feedback to the user for migrating as its likely a lot of users will hit a 20-30second white screen on the app when upgrading which would be confusing.
> 
> 
> Or.. just do the migration on desktop, somehow 🤷

Can confirm that I got similar results (also on develop on emulator). I think it's clear that this solution isn't going to scale. Even if we optimise memory access patterns or provide user feedback, I'm not sure it's a good approach to be doing heavy lifting like this on mobile devices.

As I understand it, most requisition lines (using requisition line as synonymous with `RequisitionItem` and `requisition_line`) on a mobile site will have already synced to desktop. As items can't be updated on mobile, desktop is guaranteed to have the latest `requisition_line.item` records for all requisition lines. 

If migration is done on desktop, all that remains are requisition lines which have not yet been synced. Any of these which were made prior to this fix will need migrating. One way to handle these records would be to retain some mobile migration code to update only those `RequisitionItem` records which are present in the `SyncOut` queue. This should work as, as far as I understand, a requisition line can only be present on mobile and not desktop if it has been made but not yet synced. 

I can see a couple of issues with this approach, namely:

1. It requires a desktop upgrade.
2. It will generate a lot of `sync_out` records which still have to be synced to mobile.

The first issue is beyond my paygrade, but I think the second can be ameleorated by avoided pushing updated lines to the sync queue and adding some logic to mobile to fall back on `RequisitionItem.item.name` if `Requisition.itemName` is `undefined` (or `""`). This could cause some inconsistencies when comparing old requisitions between mobile and desktop, but if `itemName` is only used by desktop, I think this might be an OK trade-off? Even if we do decide all records need to be synced back, I think this is still better than migrating on mobile as it allows us to avoid re-inventing the wheel both in terms of using the existing UI for showing sync progress and in terms of past experience with large syncs (e.g. initialisation) not freezing or causing memory errors. 

@joshxg @Chris-Petty Thoughts?

Chris-Petty:
I'd bet money that no mobile site has more than 100k requisition lines. I wouldn't be surprised for some lower end tablets to run out of storage before they hit a million 😅 .

In the realm of scale hypotheticals, I'd rather 100 sites with 1m in the queue, than the server having to update 100m lines and generate corresponding sync_out records in one go (that's a looong central migration, would definitely have to be a background task), then hold that in the queue (some indefinitely, if the site isn't actually in use).

I agree there should be user feedback on migrations being run! I think that should be an issue regardless. A splash screen and some activity indicator.

What happens if you close the app half way through migrating?

> If migration is done on desktop, all that remains are requisition lines which have not yet been synced. Any of these which were made prior to this fix will need migrating. One way to handle these records would be to retain some mobile migration code to update only those `RequisitionItem` records which are present in the `SyncOut` queue. This should work as, as far as I understand, a requisition line can only be present on mobile and not desktop if it has been made but not yet synced.
> 
> I can see a couple of issues with this approach, namely:
> 
> 1. It requires a desktop upgrade.
> 2. It will generate a lot of `sync_out` records which still have to be synced to mobile.
> 
> The first issue is beyond my paygrade, but I think the second can be ameliorated by avoided pushing updated lines to the sync queue and adding some logic to mobile to fall back on `RequisitionItem.item.name` if `Requisition.itemName` is `undefined` (or `""`). This could cause some inconsistencies when comparing old requisitions between mobile and desktop, but if `itemName` is only used by desktop, I think this might be an OK trade-off? Even if we do decide all records need to be synced back, I think this is still better than migrating on mobile as it allows us to avoid re-inventing the wheel both in terms of using the existing UI for showing sync progress and in terms of past experience with large syncs (e.g. initialisation) not freezing or causing memory errors.
> 
> @joshxg @Chris-Petty Thoughts?

Mobile _should also_ use `itemName`, not just desktop. Your tests describe this behaviour, and another scan of the PR doesn't appear to change the data source for the columns in the requisition pages in the app? (Sidenote: Needs to be detective work done to check what desktop actually does, I don't think it's showing `itemName` but again _it should be_.

I wouldn't worry if the mobile site isn't guaranteed to have the most current item name. It's wrong either way, the true history has been lost. (Side note: central server isn't guaranteed either because of mirror sync! Primary is the only one truly guaranteed.)

The mobile site should be the source of truth for its site's data as best we can keep it that way. 

> Even if we do decide all records need to be synced back, I think this is still better than migrating on mobile as it allows us to avoid re-inventing the wheel both in terms of using the existing UI for showing sync progress

I don't understand, this doesn't require any changes here on either end?

> and in terms of past experience with large syncs (e.g. initialisation) not freezing or causing memory errors.

If these are reproducible you should report them! Sync issues are always high priority. Nothing worse than initialisation crashing on a deployment.

joshxg:
> @joshxg @Chris-Petty Thoughts?

Seems like more work than adding some feedback about the migration progress, but should work

wlthomson:
> 
> 
> > @joshxg @Chris-Petty Thoughts?
> 
> Seems like more work than adding some feedback about the migration progress, but should work

@joshxg 

I guess my point is that I'm not sure if adding feedback is a solution. The idea wasn't to move the heavy lifting to desktop because its easier, but because I can't see how to do it safely/responsibly on mobile. If adding feedback solves the problem, then that's fine... as long as we're not pushing out solutions which have the potential to bork the app.

@Chris-Petty 

> I'd bet money that no mobile site has more than 100k requisition lines. I wouldn't be surprised for some lower end tablets to run out of storage before they hit a million 😅 .

Fair point, although the number of requisition lines is the same either way, so wouldn't it be more reliable to do it on desktop?

> What happens if you close the app half way through migrating?

If its done via sync, I guess nothing? But what happens if it's local and the app crashes (e.g. runs out of memory)? We probably want to minimise any chance of data corruption?

> Mobile should also use itemName, not just desktop. 

I got the impression that the main requirement was for the central server to have a record of this preserved item name?

> I don't understand, this doesn't require any changes here on either end?

If we are doing it on migration, we will need to add logic for displaying a migration process bar, which as I understand it we don't already have (we do already have a sync progress bar). This isn't much work though, the main issue for me is ensuring that mobile can handle the upgrade without borking or being at risk of borking.

> If these are reproducible you should report them! Sync issues are always high priority. Nothing worse than initialisation crashing on a deployment.

Think you misunderstood. I was saying that sync is reliably tested and doesn't freeze stuff up. The logic was that if the migration is completed on desktop, we can assume that mobile sync will behave itself. Local migration, on the other hand, seems like it will be a bit more of a "test it with 100k records and cross our fingers it doesn't crash" type of situation.

@joshxg @Chris-Petty 

Personally, I'm very happy to go with the quickest solution and do it mobile side if you guys feel comfortable with it. I just wanted to ensure that there wasn't a safer way of doing it on desktop side.

wlthomson:
Next steps:

I've put down my concerns about exploding the app, so I'm happy to go with the concensus is that mobile-side migration is the way to go. I'll have a go at adding some feedback and testing on an actual tablet with between 100-500k records. From what I understand, if the wait is reasonable within these bounds, then we are OK to go ahead 👍 .

joshxg:
Feel there's some mis communication here. .

I haven't intended to advocate for or against any particular solution and only intended to give feedback/suggestions for the solution as it was in the PR (I.e. there was migration code and I suggested feedback to the user). I also gave thoughts on a proposed solution.

Chris-Petty:
> I guess my point is that I'm not sure if adding feedback is a solution. The idea wasn't to move the heavy lifting to desktop because its easier, but because I can't see how to do it safely/responsibly on mobile. If adding feedback solves the problem, then that's fine... as long as we're not pushing out solutions which have the potential to bork the app.
> 

@joshxg  suggested chunking the writes, that seems likely to solve the safety concern? Easy tweak that gets this over the line? I'd suggest making a separate issue for providing user feedback that migration is occurring via a splash screen (@katherine-sussol might be the one with most momentum on mobile to implement this?). (I'd try make sure that both are released together though, ideally)

Can stop reading here if that's enough to go off heh.

> Fair point, although the number of requisition lines is the same either way, so wouldn't it be more reliable to do it on desktop?

Kind of trying to avoid doing it on desktop, just from a record ownership point of view. If we can avoid the central editing records belonging to remotes it's almost always more straight forward. Especially giving that item names can be be out of sync (that's a pretty minimal chance, they don't change that often, but 🤷 ).

> 
> > What happens if you close the app half way through migrating?
> 
> If its done via sync, I guess nothing? But what happens if it's local and the app crashes (e.g. runs out of memory)? We probably want to minimise any chance of data corruption?

Depends if it's likely? Obviously yes we want to minimise the risk but taking this so extremely would suggest having an online only app, where nothing happens locally because it might crash at any moment?

The bigger risks IMO would simply be battery running out, or memdm trigger crashes, user turning the tablet off, user closing the app...

To that end (desktop + omsupply need to consider this as much as mobile), to improve safety (perhaps not in this issue) we need to make sure that partial migrations can be finished off. If it gets half way through the lines and crashes, when you start back up it should find where it's up to and finish.

Similar idea is in the `PostSyncProcessor.js` class, if it knows it didn't shut down gracefully it'll run functions across all data to make sure nothing was missed, because it only tracks what records to run functions on, in memory at run time. (I definitely over cooked this thing and regret not doing a dumb solution haha)

> 
> > Mobile should also use itemName, not just desktop.
> 
> I got the impression that the main requirement was for the central server to have a record of this preserved item name?

The immediate need is because some UI on remote authorisation uses it, because desktop fills the field. Generally, all historic transactions and requisitions should be showing the name that was saved at the time those records were finalised. If not addressing for requisitions in this issue, then please make a new issue(s) to make mobile compliant with this.

> > I don't understand, this doesn't require any changes here on either end?
> 
> If we are doing it on migration, we will need to add logic for displaying a migration process bar, which as I understand it we don't already have (we do already have a sync progress bar). This isn't much work though, the main issue for me is ensuring that mobile can handle the upgrade without borking or being at risk of borking.
> 
> > If these are reproducible you should report them! Sync issues are always high priority. Nothing worse than initialisation crashing on a deployment.
> 
> Think you misunderstood. I was saying that sync is reliably tested and doesn't freeze stuff up. The logic was that if the migration is completed on desktop, we can assume that mobile sync will behave itself. Local migration, on the other hand, seems like it will be a bit more of a "test it with 100k records and cross our fingers it doesn't crash" type of situation.
> 
> @joshxg @Chris-Petty
> 
> Personally, I'm very happy to go with the quickest solution and do it mobile side if you guys feel comfortable with it. I just wanted to ensure that there wasn't a safer way of doing it on desktop side.

As said at top, quick easy solution seems to be chunking the writes?

wlthomson:
@Chris-Petty Sounds good. 

I don't want to make @katherine-sussol tidy up my mess, so in the interests of getting it done, I'll see what I can do in this PR!

wlthomson:
Just updating that still working on this. Have got the basic idea working, just need to spend a bit of time stress testing it!

wlthomson:
Have pushed an idea for splitting migrations into subtasks, with process displayed to the user via a progress bar. I think in general this is a good idea, as it allows for multiple tasks per migration, rather than everything having to be chucked into one big method.

I have a few concerns about how this will work in practice, mainly whether this is going to slow down migration even more. I'm also not sure about the optimal batch size, and how much this will impact the performance.

In process of collecting some performance metrics; will update with details.

wlthomson:
Phew... looks like we are getting there. 

I've made a [test branch](https://github.com/openmsupply/mobile/tree/%234204-migration-testing-wip) which makes it easier to mock migration by bumping the versions (first bump inserts records, second migrates them). 

I managed to test with a million records without crashing (on emulator). The insertion took a whole day, but the migration was fairly speedy (a matter of minutes, exact metrics are going to vary tablet to tablet). 

Next step is to test on an actual tablet. Don't want to go through making a million records again, but will target a reasonable sized dataset (maybe ~250k).

wlthomson:
Tested migrating 250k records on a Lenovo TB-X304L, took roughly ~10mins. The migration complexity is linear, so we can expect it to grow roughly proportionally (e.g. 500k records should take about ~20mins, 1M about ~40mins). 

I think this is an acceptable speed, but more testing is always appreciated!

Chris-Petty:
Did a bit of a test run in debug:

1. Installed master branch debug
2. made some supplier requisitions
3. Confirmed in realm explorer that the column doesn't exist yet
4. Installed this branch debug
5. Checked realm explorer. The new field is there, but all values are `N/A`

image

I'm guessing the actual migrating didn't occur? Did I do this wrong

wlthomson:
@Chris-Petty 

Because of how long this has taken to get in, it's now behind `master`. I'll rebase.

In the meantime, you'll need to checkout a version of `master` < `8.1.1` and then checkout this branch and bump to `8.1.1`. That should trigger the migration.

Chris-Petty (dismissed):
Tested in debug 👍

Chris-Petty:
@wlthomson I approved this, you only changed the version bump and dismissed. No prob if I approve again?

Chris-Petty (approved):

package.json
```diff
@@ -5,7 +5,7 @@
   },
   "name": "mSupplyMobile",
   "//": "version must be in the format ${majorNumber}.${minorNumber}.${patchNumber}-rc${releaseCandidateNumber}",
-  "version": "8.2.1",
+  "version": "8.2.2",
```
Chris-Petty, line 8: Wait, shouldn't this also be `8.2.3`?

wlthomson: Plan was for the version to be bumped in `master` after merging.

wlthomson:
> 
> 
> @wlthomson I approved this, you only changed the version bump and dismissed. No prob if I approve again?

Soz, github automatically dismisses reviews if you push a new commit 🤷‍♂️.

Chris-Petty:
> > @wlthomson I approved this, you only changed the version bump and dismissed. No prob if I approve again?
> 
> Soz, github automatically dismisses reviews if you push a new commit 🤷‍♂️.

Ah that's a repo setting. Probably generally a good one tbh!

Chris-Petty (approved):

package.json
Chris-Petty: Cool beans!

PR #4217, opened 2021-07-07
#4210 Auto populate form with previously entered values

### LINKED ISSUE #4210: Vaccine Dispensing: Make site fields auto-populate to the last chosen for the previous patient
Child of #4176 

See parent issue

- Make the fields auto-populate to the last chosen for the previous patient. 
- Not sure how long to retain the last chosen values (until someone exits the vaccine dispensing tab? until someone uninstalls the app?)
- Maybe auto-populate everything? (how often do people change where they aim their 💉 ? 🤔 )

## Implementation
Prepopulate `FormData` for the `VaccineSite` Form

Not auto-populating

## Additional context

katherine-sussol:
Fixes #4210

## Change summary

- Looks in realm to try to find the last customData entry against a vaccine transaction and prepopulate supplemental data form with it
- Technically (at the moment anyway) these values _shouldn't_ change very often between jabs (might not be true later if some obscure fields are requested...)

## Testing

- [ ] Submit vaccine additional data once, do it again and see if form is prepopulated

...

### DIFF
--- a/src/selectors/Entities/vaccinePrescription.js
+++ b/src/selectors/Entities/vaccinePrescription.js
@@ -99,3 +99,19 @@ export const selectSelectedSupplementalData = state => {
 
   return supplementalData;
 };
+
+export const selectLastSupplementalData = () => {
+  const inQuery = UIDatabase.objects('Transaction')
+    .filtered('customData != $0 AND type != "cash_in" AND type != "cash_out"', null)
+    .map(({ id }) => `transaction.id == "${id}"`)
+    .join(' OR ');
+  const fullQuery = `(${inQuery}) AND itemBatch.item.isVaccine == true`;
+  const result = inQuery
+    ? UIDatabase.objects('TransactionBatch')
+        .filtered(fullQuery)
+        .sorted('transaction.entryDate', true)
+        .slice()
+    : [];
+
+  return result.length ? result[0].transaction.customData : null;
+};
--- a/src/selectors/Entities/vaccinePrescription.js
+++ b/src/selectors/Entities/vaccinePrescription.js
@@ -102,7 +102,7 @@ export const selectSelectedSupplementalData = state => {
 
 export const selectLastSupplementalData = () => {
   const inQuery = UIDatabase.objects('Transaction')
-    .filtered('customData != $0 AND type != "cash_in" AND type != "cash_out"', null)
+    .filtered("type == 'customer_invoice' AND customData != null")
     .map(({ id }) => `transaction.id == "${id}"`)
     .join(' OR ');
   const fullQuery = `(${inQuery}) AND itemBatch.item.isVaccine == true`;
--- a/src/selectors/Entities/vaccinePrescription.js
+++ b/src/selectors/Entities/vaccinePrescription.js
@@ -100,6 +100,13 @@ export const selectSelectedSupplementalData = state => {
   return supplementalData;
 };
 
+export const selectSupplementalDataIsValid = state => {
+  const VaccinePrescriptionState = selectSpecificEntityState(state, 'vaccinePrescription');
+  const { isSupplementalDataValid } = VaccinePrescriptionState;
+
+  return isSupplementalDataValid;
+};
+
 export const selectLastSupplementalData = () => {
   const inQuery = UIDatabase.objects('Transaction')
     .filtered("type == 'customer_invoice' AND customData != null")
@@ -113,5 +120,5 @@ export const selectLastSupplementalData = () => {
         .slice()
     : [];
 
-  return result.length ? result[0].transaction.customData : null;
+  return result.length ? JSON.parse(result[0].transaction.customData) : null;
 };
--- a/src/selectors/Entities/vaccinePrescription.js
+++ b/src/selectors/Entities/vaccinePrescription.js
@@ -113,12 +113,14 @@ export const selectLastSupplementalData = () => {
     .map(({ id }) => `transaction.id == "${id}"`)
     .join(' OR ');
   const fullQuery = `(${inQuery}) AND itemBatch.item.isVaccine == true`;
-  const result = inQuery
-    ? UIDatabase.objects('TransactionBatch')
-        .filtered(fullQuery)
-        .sorted('transaction.entryDate', true)
-        .slice()
-    : [];
 
-  return result.length ? JSON.parse(result[0].transaction.customData) : null;
+  if (inQuery) {
+    const result = UIDatabase.objects('TransactionBatch')
+      .filtered(fullQuery)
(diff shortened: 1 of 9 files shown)


src/widgets/Tabs/VaccineSupplementalData.js
```diff
@@ -24,11 +24,17 @@ import { PageButtonWithOnePress } from '../PageButtonWithOnePress';
...
+import { selectLastSupplementalData } from '../../selectors/Entities/vaccinePrescription';
 
 const { pageTopViewContainer } = globalStyles;
 
 const VaccineSupplementalDataComponent = ({ onCancel, onComplete, siteSchema }) => {
-  const [{ formData, isValid }, setForm] = useState({ formData: null, isValid: false });
+  const lastSupplementalData = selectLastSupplementalData();
+  const [{ formData, isValid }, setForm] = useState({
+    formData: JSON.parse(lastSupplementalData),
+    isValid: lastSupplementalData ?? false,
```
katherine-sussol, line 35: This is probably going to do weird stuff if the form schema is changed in mSupply in between transactions 🤔 . Maybe need to find a way to make the `JSONForm` validate the initial formData

src/widgets/Tabs/VaccineSupplementalData.js
```diff
@@ -37,6 +43,7 @@ const VaccineSupplementalDataComponent = ({ onCancel, onComplete, siteSchema })
         style={{ flex: 1 }}
       >
         <JSONForm
+          formData={formData}
```
joshxg, line 46: If I edit the form then step back a step or two, then when I reenter this step, the state is the same, which I'm unsure is the desired behaviour.

The original `PatientEdit` is maybe good to follow?

The `JSONForm` is a bit of a nightmare.. it has internal state and doesn't cause re-renders of its parent, so by updated `formData` in the `Suppliermentary` parent component, we end up maintaining that state if we switched patient.

If I remember correctly, `PatientEdit` seeds the data by putting the most recent filled out form into the reducer when a patient is selected, then seeding `JSONForm` with that data. When editing the form, the form is updated in the reducer state. When the patient changes, the seed changes in the reducer, which updates the form correctly... at the moment the `useState` won't be re-seeded, as `VaccineSuppl...` is never re-mounted only re-rendered.

Hope that makes sense (and is accurate?!?!)

joshxg (changes requested):
Couple of.. maybe? changes?

src/selectors/Entities/vaccinePrescription.js
```diff
@@ -99,3 +99,19 @@ export const selectSelectedSupplementalData = state => {
 
   return supplementalData;
 };
+
+export const selectLastSupplementalData = () => {
+  const inQuery = UIDatabase.objects('Transaction')
+    .filtered('customData != $0 AND type != "cash_in" AND type != "cash_out"', null)
```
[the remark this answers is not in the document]
katherine-sussol: Uhh yup 😅 . I mangled the patient history query to arrive here, but yeah it should only be `customer_invoices` haha.

src/selectors/Entities/vaccinePrescription.js
```diff
@@ -99,3 +99,19 @@ export const selectSelectedSupplementalData = state => {
...
+    .map(({ id }) => `transaction.id == "${id}"`)
+    .join(' OR ');
+  const fullQuery = `(${inQuery}) AND itemBatch.item.isVaccine == true`;
+  const result = inQuery
+    ? UIDatabase.objects('TransactionBatch')
+        .filtered(fullQuery)
+        .sorted('transaction.entryDate', true)
+        .slice()
+    : [];
```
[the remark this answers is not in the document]
katherine-sussol: Ooh subqueries! That is neat! I couldn't get it to work (transaction object doesn't have batches, so tried to reverse and query on batches first then subquery on transactions, but realm didn't like it cos it was a link and not a list property?! 😿  <- me). 
Ugly fix instead: 09ca6013

src/widgets/Tabs/VaccineSupplementalData.js
katherine-sussol: Yeah, it probably makes sense to reset the form when the patient is changed (at least that's what I think I'd expect as a user). 

I didn't realise the PCD was supposed to be seeded based on previous data (but I can see the code that does it) - I think my PCD form (even when filled out) is resulting in 'undefined' formData 🙀  I'm not sure what's going on but I'll take a look 🤔

src/widgets/Tabs/VaccineSupplementalData.js
katherine-sussol: Cool - have adjusted the schema fetching to sort by version and pull latest (affects the PCD form as well) 1133a64

The validation of seeded data was hidden in 7dc0ad8 (sorry I should have broken it into something smaller but I got a bit hack happy)

src/widgets/Tabs/VaccineSupplementalData.js
katherine-sussol: Fixed the prepopulation to not persist between patients: 7dc0ad8 9f6f033

PCD problem was just me not setting up the 'PCD' PatientEvent in mSupply correctly (if it's not set up they don't get saved 🤣 )

src/selectors/Entities/vaccinePrescription.js
```diff
@@ -99,3 +99,26 @@ export const selectSelectedSupplementalData = state => {
...
+  const inQuery = UIDatabase.objects('Transaction')
+    .filtered("type == 'customer_invoice' AND customData != null")
+    .map(({ id }) => `transaction.id == "${id}"`)
+    .join(' OR ');
+  const fullQuery = `(${inQuery}) AND itemBatch.item.isVaccine == true`;
+  const result = inQuery
+    ? UIDatabase.objects('TransactionBatch')
+        .filtered(fullQuery)
+        .sorted('transaction.entryDate', true)
+        .slice()
```
joshxg, line 120: nit: getting this linter error 🤷 

image

joshxg (approved):
amazing ❤️

src/selectors/Entities/vaccinePrescription.js
katherine-sussol: Ohh yeah, I tried removing those and got _another_ warning, other than disabling for the whole file, any ideas? 🤣 :
image

joshxg: nah .. sorry .. I usually just steal a linter config from somewhere and run with it .. never really looked to deeply into them... uhh.. I think in the `eslintrc` folder maybe just making `"indent": ["warning", 2, { "SwitchCase": 1 }],`  🤣

PR #4231, opened 2021-07-13
support for checkoxes

### LINKED ISSUE #4229: Support for list of checkboxes in JSON schema form

Would like to be able to use a list of checkboxes in a JSON form. For example, adding this to the schema:

```
    "causes": {
      "title": "Possible causes",
      "description": "Select which vaccinations you think may have caused this reaction",
      "type": "array",
      "items": {
        "type": "string",
        "enum": ["Option One", "Option Two", "Other"]
      },
      "uniqueItems": true
    },
```

and this to the UI

```
  "causes": { "ui:widget": "checkboxes" },
```

should display a list of checkboxes.

mark-prins:
Fixes #4229 

## Change summary

Adds the checkboxes component, so it is now possible to add this into the JSON form schema:

```
  "causes": {
      "title": "Possible causes",
      "description": "Select which vaccinations you think may have caused this reaction",
      "type": "array",
      "items": {
        "type": "string",
        "enum": ["Option One", "Option Two", "Other"]
      },
      "uniqueItems": true
    },
```

with the UI schema:

```
  "causes": { "ui:widget": "checkboxes" },
```

and get this in the form:

adr

There's one glitch - a warning is thrown because the form is passing in a text value for the `value` prop, when the component should receive an array. I've found that ignoring this works for me 🤷‍♂️ 

## Testing

- [ ] Add the above to your schema and check the results

### DIFF
--- a/src/widgets/JSONForm/widgets/Checkboxes.js
+++ b/src/widgets/JSONForm/widgets/Checkboxes.js
@@ -65,7 +65,11 @@ const styles = StyleSheet.create({
 
 Checkboxes.propTypes = {
   disabled: PropTypes.bool,
-  value: PropTypes.oneOf([PropTypes.string, PropTypes.boolean]),
+  value: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.bool,
+    PropTypes.arrayOf(PropTypes.string),
+  ]),
   onChange: PropTypes.func.isRequired,
   options: PropTypes.shape({
     enumOptions: PropTypes.arrayOf(PropTypes.any),
--- a/src/widgets/JSONForm/widgets/Checkboxes.js
+++ b/src/widgets/JSONForm/widgets/Checkboxes.js
@@ -1,10 +1,83 @@
-/* eslint-disable no-console */
+/* eslint-disable react/forbid-prop-types */
 import React from 'react';
-import { Text } from 'react-native';
+import { StyleSheet, Text } from 'react-native';
+import PropTypes from 'prop-types';
+import CheckBox from '@react-native-community/checkbox';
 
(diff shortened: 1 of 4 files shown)


joshxg (approved):
Works great 🎉 

I think this fixes the prop type warning?

```
value: PropTypes.oneOfType([
    PropTypes.arrayOf(PropTypes.string),
    PropTypes.arrayOf(PropTypes.bool),
  ]),
```

Then I got another one for `radio` 🙄  for which I needed to change to use `value: PropTypes.oneOfType([PropTypes.string, PropTypes.bool]),` 

then I got more about formats and the rabbit hole kept going and going and I backed out before I got to stuck...

katherine-sussol (dismissed):
Oh nice - LGTM

(soz I also released 8.1.1 just as soon as I saw the PR so might need another branch off master? 😆 )

mark-prins:
> I think this fixes the prop type warning?

Oh 🤦  of course. thanks

src/widgets/JSONForm/widgets/Checkboxes.js
```diff
@@ -1,10 +1,87 @@
...
+
+const styles = StyleSheet.create({
+  root: { paddingLeft: 10, paddingTop: 10 },
+  row: { flex: 0, flexDirection: 'row' },
+  label: { fontFamily: APP_FONT_FAMILY, marginLeft: 10, marginTop: 7 },
+});
+
+Checkboxes.propTypes = {
+  disabled: PropTypes.bool,
+  value: PropTypes.oneOfType([
```
mark-prins, line 68: oops. Needed to be `oneOfType` rather than `oneOf`
f377d25a7

PR #4244, opened 2021-07-14
#4237 Update isSupplementalDataValid state

### LINKED ISSUE #4237: Seeded supplementary data is invalid as initial state
## Describe the bug

When prescribing to a patient the second+ time such that their supplementary data is pre-filled, the form is invalid when it should be valid.

### To reproduce

1. Go to fill out supplementary data for a patient and finalise the prescription
2. Click on the patient again
3. Scroll down to the third step and try to proceed without touching the form (which is pre-filled and technically valid)

### Expected behaviour

Should be able to just 'next' if the form is pre-filled

### Proposed Solution

Getting pretty messy! I remember having to spend a bit of time on this for the patient survey. At a basic level, we need to run validation on the pre-filled form after we have found it and set the `isValid` state to the result

### Version and device info

- App version:
- Tablet model:
- OS version:

katherine-sussol:
Fixes #4237

## Change summary

- Add missed state update

## Testing

- [ ] Try to reproduce steps in #4237

N/A

### DIFF
--- a/src/reducers/Entities/VaccinePrescriptionReducer.js
+++ b/src/reducers/Entities/VaccinePrescriptionReducer.js
@@ -53,8 +53,6 @@ export const VaccinePrescriptionReducer = (state = initialState(), action) => {
       const { payload } = action;
       const { supplementalData, isSupplementalDataValid } = payload;
 
-      console.log(isSupplementalDataValid);
-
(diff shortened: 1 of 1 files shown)


joshxg (approved):
Ah!

PR #4250, opened 2021-07-15
#4238 show vaccination history when creating an ADR

### LINKED ISSUE #4238: AEFI : select from vaccination history

As per [patient hub issue #108](https://github.com/openmsupply/patient-hub/issues/108); we have a client who would like to show all patient vaccinations inside a JSON schema AEFI form, and allow the AEFI entry person to select any number of them.

If there is a JSON schema element in the correct format then update the `enum` within it with vaccinations for the selected patient. Show these as a multi-select list to the user.

## Implementation

- After patient selection, perform an online patient history check. 
- Filter these by vaccination
- populate the JSON form element

mark-prins:
Fixes #4238 

## Change summary

Modifies an ADR JSON schema in order to inject vaccinations for the current patient.

I'm very sorry.

Not an ideal approach, as it relies on knowledge of the JSON schema by the codebase, however it gives us:
- somewhere to store one-to-many vaccinations against an ADR record
- control to the client about how to present the question (wording, placement, form component)
- as well as whether to display the question

## Testing

To test, you will need this in the ADR JSON schema:

```
    "causes": {
      "title": "Possible causes",
      "description": "Select which vaccinations you think may have caused this reaction",
      "type": "array",
      "items": { "type": "string", "enum": ["No vaccination history"] },
      "uniqueItems": true
    },
```

and I have this in the UI schema for the ADR:
```
  "causes": { "ui:widget": "checkboxes" },
```

Results:

**Searching**
vax_search

**no history**
vax_none

**some vaccinations in history**
vax_list

- [ ] Add an ADR without the schema change: behaves as normal, with no search
- [ ] Add an ADR for a patient with no history: will show the checkbox 'No vaccination history'
- [ ] Add an ADR for a patient with history: searches, shows options
- [ ] Only shows vaccinations, not other dispensed items in the history
- [ ] Add when the server is offline: still lets you create an ADR after a short lookup time
- [ ] When the server is offline you can see local patient history

Another thought was to allow all history, by specifying a different element in the schema maybe. Decided not to do that

### DIFF
--- a/src/widgets/modalChildren/ADRInput.js
+++ b/src/widgets/modalChildren/ADRInput.js
@@ -54,10 +54,11 @@ export const ADRInputComponent = ({ onCancel, onSave, patient, patientHistory })
   const [ADRSchema, setADRSchema] = useState(UIDatabase.objects('ADRForm')[0]);
   const { jsonSchema, uiSchema, type, version } = UIDatabase.objects('ADRForm')[0];
   const items = getSchemaItems(jsonSchema);
+  const vaccineHistory = patientHistory.filter(h => h.itemBatch?.item?.isVaccine);
   const [{ data, loading, searched }, fetchOnline] = useLocalAndRemotePatientHistory({
     isVaccineDispensingModal: true,
     patientId,
-    initialValue: patientHistory,
+    initialValue: vaccineHistory,
     sortKey: 'itemName',
   });
   const showLoading = items && (loading || !searched);
--- a/src/widgets/modalChildren/ADRInput.js
+++ b/src/widgets/modalChildren/ADRInput.js
@@ -1,30 +1,71 @@
 /* eslint-disable react/forbid-prop-types */
-import React, { useState } from 'react';
-import { StyleSheet } from 'react-native';
+import React, { useEffect, useState } from 'react';
+import { ActivityIndicator, StyleSheet, Text } from 'react-native';
 import { connect } from 'react-redux';
 import PropTypes from 'prop-types';
 import { JSONForm } from '../JSONForm/JSONForm';
 import { UIDatabase } from '../../database/index';
 import { FlexRow } from '../FlexRow';
+import { FlexView } from '../FlexView';
 import { PageButton } from '../PageButton';
-import { SUSSOL_ORANGE } from '../../globalStyles/colors';
+import { SUSSOL_ORANGE, WHITE } from '../../globalStyles/colors';
 import globalStyles from '../../globalStyles/index';
 import { PatientActions } from '../../actions/PatientActions';
 import { selectCurrentPatient } from '../../selectors/patient';
+import { useLocalAndRemotePatientHistory } from '../../hooks/useLocalAndRemoteHistory';
+import { useLoadingIndicator } from '../../hooks/useLoadingIndicator';
+import { dispensingStrings } from '../../localization';
 
-export const ADRInputComponent = ({ onCancel, onSave, patient }) => {
+const LoadingIndicator = ({ loading }) =>
+  loading && (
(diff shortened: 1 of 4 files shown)


src/widgets/modalChildren/ADRInput.js
```diff
@@ -1,30 +1,90 @@
...
+  const { causes = {} } = properties;
+  const { items } = causes;
+  return items;
+};
+const mapHistory = history =>
+  history.map((h, index) => {
+    const { prescriberOrVaccinator, itemName, confirmDate } = h;
+    const vaccinator = prescriberOrVaccinator ? `  Vaccinator: ${prescriberOrVaccinator}` : '';
+
+    return `${index + 1}. ${itemName}  Date: ${confirmDate}${vaccinator}`;
```
katherine-sussol: It seems like (unless I'm missing something) the timestamp isn't ever anything other than `T12:00:00.000Z` so it might look a bit neater with the time portion trimmed off before display?

src/widgets/modalChildren/ADRInput.js
```diff
@@ -1,30 +1,90 @@
...
+};
+
+export const ADRInputComponent = ({ onCancel, onSave, patient, patientHistory }) => {
   const [{ formData, isValid }, setForm] = useState({ formData: null, isValid: false });
+  const patientId = patient?.id;
+  const [ADRSchema, setADRSchema] = useState(UIDatabase.objects('ADRForm')[0]);
+  const { jsonSchema, uiSchema, type, version } = UIDatabase.objects('ADRForm')[0];
+  const items = getSchemaItems(jsonSchema);
+  const [{ data, loading, searched }, fetchOnline] = useLocalAndRemotePatientHistory({
+    isVaccine: true,
```
katherine-sussol, line 56: Oh no this is my bad, I renamed `isVaccine` to `isVaccineDispensingModal` without updating the reference in the `useLocalAndRemoteHistory` hook so now non-vaccines are showing up in vaccine modals.... 😿 : https://github.com/openmsupply/mobile/blob/6a58b9377b7d50a7c3444033ebc8ae53de484797/src/sync/lookupApiUtils.js#L206

...soz

mark-prins: thanks - I missed that!
6d16feed4

src/widgets/modalChildren/ADRInput.js
mark-prins: good point, thanks!
d308e80ba

src/widgets/modalChildren/ADRInput.js
```diff
@@ -1,30 +1,94 @@
...
+      </>
+    )}
+  </FlexView>
+);
+
+LoadingIndicator.propTypes = {
+  loading: PropTypes.bool.isRequired,
+};
+
+export const ADRInputComponent = ({ onCancel, onSave, patient, patientHistory }) => {
```
katherine-sussol, line 51: I thiiink the regular dispensing local history with all items is also being passed in here - might also need an update (`selectVaccinePatientHistory` vs `selectPatientHistory`)?

katherine-sussol (commented):
One additional possible note/question - but otherwise it works/looks awesome (great vaccination injecting! 💉)

src/widgets/modalChildren/ADRInput.js
mark-prins: another good catch! thank you :)
eafe967c7

I've gone with filtering, as I have the vague though that this might want to be expanded to include all patient history for some clients and thought that it was handy to have things in one location.

katherine-sussol (approved):
[LGTM](https://storage.googleapis.com/lgtm-cat-cats/6fad12f1-c921-478d-be79-6c4b4de53e70.gif)

PR #4252, opened 2021-07-19
Add correct filtering

### LINKED ISSUE #4251: Search name is not working properly
## Describe the bug

While searching for full name shows nothing in the list.

### To reproduce

1. Go to Customer requisition/Supplier requisition.
2. Type any name from the list.
3. See an error

### Expected behaviour

Should work like item search.

### Proposed Solution

### Version and device info

- App version:808
- Tablet model: tried in the emulator(8.1)

### Additional context

photo_2021-07-16 11 54 27
photo_2021-07-16 11 54 23

joshxg:
Fixes #4251 

## Change summary

- I could only replicate on the `ByProgramModal` customer/supplier search.

## Testing

- [ ] Can search for a customer/supplier by name in the `ByProgramModal`

N/A

### DIFF
--- a/src/widgets/modalChildren/AutocompleteSelector.js
+++ b/src/widgets/modalChildren/AutocompleteSelector.js
@@ -57,8 +57,8 @@ export const AutocompleteSelector = ({
   const filterArrayData = () =>
     options.filter(
       optionItem =>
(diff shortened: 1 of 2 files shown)


mark-prins (approved):
could only repro when searching for __part__ of a name rather than the start, which I normally do.
That fix works for me!

PR #4254, opened 2021-07-19
#4239 Text not in text!

joshxg:
Fixes nothing.. really..

I was just looking through to see anywhere there might have been a mistake of the form `"" && <Something/>` by global searching `&&` and looking at all the components, then I added a bunch of `!!` and now here we are.

I fully understand just closing this PR if it's too scarey to make all these changes

## Change summary

- Adds a transform from anything that is used as a conditional in a `&&` into a boolean

## Testing

N/A?

The ones which are actually causing crashes are:
-  `<Paper headerText=""/>`
- `<IconButton label=""/>`

I can revert all the other ones if it's there's too much changes, I got carried away

### DIFF
--- a/src/pages/CustomerInvoicePage.js
+++ b/src/pages/CustomerInvoicePage.js
@@ -160,6 +160,7 @@ export const CustomerInvoice = ({
           <SearchBar
             onChangeText={onFilterData}
             value={searchTerm}
+            // eslint-disable-next-line max-len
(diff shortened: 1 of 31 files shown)


katherine-sussol (approved):
🙀  that's a lot of changes! But they don't look particularly scary and it's extra safety? (and the app still seems to be fine as far as I can tell?)
So I ticked approve :shipit:

mark-prins:
cool, that looks ok to me. seems safe enough to try

PR #4267, opened 2021-07-20
revert changes to address field names

### LINKED ISSUE #4266: Patient address is not retained
## Describe the bug

When editing the address fields of a patient the values are not retained

### To reproduce

1. Go to Dispensing and edit a patient
2. Enter values for address one and address two
3. Save and edit the patient again

### Expected behaviour

The values should save..
Also, when looking up a remote patient, the values should come through correctly.
Should work for prescribers too.

mark-prins:
Fixes #4266 

## Change summary

Essentially just reverting the incorrect changes in #4140 
The address lookup section of that PR is correct.. but the field mappings are causing problems.

## Testing

- [ ] Edit a patient address and save and check again
- [ ] Lookup a patient (which has an address 1&2 entered) and check the address arrives
- [ ] Edit and save the looked up patient and check the server
- [ ] Edit a patient record through the vaccination process
- [ ] Prescribers too?


joshxg (approved):
it is saving and syncing well 👍

Issue #4228, opened 2021-07-12
Additional COVID-19 vaccination dispensing fields for Kiribati


adamdewey:
## Is your feature request related to a problem? Please describe.

Kiribati / WHO have requested additional fields to be captured during the COVID-19 vaccination dispensing process. 

I know @katherine-sussol has been working on adding an additional intermediary page to the workflow. 

These are the fields that Kiribati / WHO require to be added to that page: 

| Field label | Field Type| Prepopulate value |
| --- | --- | --- |
| Country of Vaccination | Dropdown (Kiribati) | Kiribati|
| Island of Vaccination | Dropdown (TBD) | -
| District of Vaccination | Dropdown (Central, Linnix, Northern, South Eastern, South Western, Tarawa & Banaba) | -
| Administering Health Facility Name | Alphanumeric | -
| Anatomical Site | Dropdown (Left arm, Right arm) | -
| Dose Number | Dropdown (First, Second) | -
| Informed Consent Received | Toggle (Yes, No) | No
| Reason for Non-consent | Alphanumeric | -
| Exempted from Vaccination? | Toggle (Yes, No) | No
| Reason for Exemption | Alphanumeric | -
| Additional notes | Alphanumeric | -

## Describe the solution you'd like

The fields above added to the new page :)

## Implementation

**A description of how the solution would best be implemented in the code base**

## Describe alternatives you've considered

**A clear and concise description of any alternative solutions or features you've considered.**

## Additional context

**Add any other context or screenshots about the feature request here.**

katherine-sussol:
🥳   Aside from Additional notes, should we mark the rest as mandatory fields?
I think maybe if Informed Consent = No or Exempted = Yes, then Reason for Exemption / Non-consent should be mandatory?

adamdewey:
Hey Kat! 

Sorry, I wasn't clear here in the requirements. I've actually just separated those and added a reason field for each to make it easier to implement. 

I think we should make the exemption reason mandatory but the non-consent one not mandatory as the patient might just not give an answer?

katherine-sussol:
https://user-images.githubusercontent.com/65875762/125391574-d6cbb100-e3f8-11eb-9542-396085072bdc.mp4

Sample JSON Schema to create the above (renamed to .txt so it could be attached 🤣 ):
[testKiribatiJsonSchema.txt](https://github.com/openmsupply/mobile/files/6805941/testKiribatiJsonSchema.txt)

Sample UI Schema:
[...]

PR #4271, opened 2021-07-21
#4269 Remove supplemental data pre-population

### LINKED ISSUE #4269: On 'extra step' in Vaccine Dispensing flow (additional data page) don't prepopulate

Apparently, the pre-population isn't wanted 🤣  

The form should reset completely (i.e. not default to whatever was last entered) when a patient is selected

## Implementation

Dunno, if (true) (to be replaced with a config at some point??), don't do 
https://github.com/openmsupply/mobile/blob/9d8956dcb76570fcd6f5fd3c2a52e4298fe062fb/src/actions/Entities/VaccinePrescriptionActions.js#L183-L197 ?

Leaving it - but this leaves more room for careless mistakes from the sounds of things (auto-populate -> user scrolls without checking properly and then just clicks next)

## Additional context

N/A

katherine-sussol:
Fixes #4269

## Change summary

- Don't dispatch an action to prepopulate supplemental data form 
- Prepopulation code is still there (just not called) in case it is needed later (happy to also remove it if that's tidier) 

## Testing

- [ ] Have a vaccine supplemental form schema in mSupply (can use https://github.com/openmsupply/mobile/issues/4228). Complete a vaccination, start a new one and navigate to step 3 in Vaccine dispensing. Should be empty form (only form configured defaults should be prepopulated) 

N/A

### DIFF
--- a/src/widgets/Tabs/PatientSelect.js
+++ b/src/widgets/Tabs/PatientSelect.js
@@ -288,7 +288,6 @@ const mapDispatchToProps = dispatch => {
 
       if (selectedPatient) {
         dispatch(NameNoteActions.createSurveyNameNote(selectedPatient));
-        dispatch(VaccinePrescriptionActions.createSupplementaryData());
         dispatch(WizardActions.nextTab());
       }
     });


mark-prins (approved):
👍  that does the job!
hopefully we remember this line, for the next client.. who will want this reinstated

PR #4290, opened 2021-07-22
use trans_line.ID as the cause record

### LINKED ISSUE #4288: Store transaction ID for ADR cause

Would be good if the selection of 'possible cause' stored the transaction id rather than text.

Using a combination of `enum` and `enumNames` could do this. See [this PR](https://github.com/openmsupply/patient-hub/commit/fb522ef8027fedab33d5e3c342b0901cd2a9cdce)

mark-prins:
Fixes #4288 

## Change summary

Stores the `[trans_line]ID` instead of the display value, when selecting cause of ADR

## Testing

- [ ] Create an ADR and check the resulting JSON on the server


katherine-sussol (approved):
👍  LGTM

PR #4291, opened 2021-07-23
#4280 Turn on live validation for the supplemental data form

### LINKED ISSUE #4280: Supplemental data page in Vaccine Dispensing flow should show validation errors

When an error message is configured in the JSON Schema, e.g:
```
    "administeringHealthFacility": {
      "type": "string",
      "title": "Administering Health Facility Name"
      "errorMessage": "Please enter the administration health facility",
    },
```

They should show up on the page if validation has failed

As per description

## Implementation

Enable live validation I guess

Leaving it and not showing the messages

## Additional context
@adamdewey might have more details but I suspect someone from Kiribati mentioned it? 🤣

katherine-sussol:
Fixes #4280

## Change summary

Switch on live validation for this form

## Testing

- [ ] Have some error messages defined in the JSON Schema for some fields as per the issue, don't fill those fields in on the form -> Should see error messages

- With this enabled the initial form can't have "required" defined for fields that don't exist until parent fields are selected (e.g. "island", "village" at the top level) ... or it will crash 💥 . I think that is another issue to investigate - why so much crashing when the schema is not 100% 😿

### DIFF
--- a/src/widgets/Tabs/VaccineSupplementalData.js
+++ b/src/widgets/Tabs/VaccineSupplementalData.js
@@ -55,7 +55,7 @@ const VaccineSupplementalDataComponent = ({
             onChange={(changed, validator) => {
               onFormUpdate(changed.formData, validator);
             }}
-            liveValidate={false}
+            liveValidate={true}
           >
             <View />
           </JSONForm>


mark-prins:
oh :)
we turned it off because it made the form worse than having validation clumsy! I suppose if people try it and like it better.. then 🤷‍♂️

katherine-sussol:
> oh :)
> we turned it off because it made the form worse than having validation clumsy! I suppose if people try it and like it better.. then 🤷‍♂️

Oh, haha 🤣 . I was a bit confused because it's disabled on the `PatientEdit` page, but not the `PatientEditModal` or `ADRInput`. I'm unsure which I prefer, only that there is a request for validation errors to show up

mark-prins (approved):
let's try it :)

PR #4294, opened 2021-07-26
#4239 catch invalid json schemas

### LINKED ISSUE #4239: Ensure invalid schemas don't crash the app
## Error in mSupply Mobile

**Error** in **node_modules\react-native\Libraries\Core\InitializeCore.js:54**
Text strings must be rendered within a <Text> component.

[View on Bugsnag](https://app.bugsnag.com/sustainable-solutions-nz-ltd/msupply-mobile/errors/60ee654cb95ce400070f08f1?event_id=60ee654c007f8852bdb00000&i=gh&m=ci)

## Stacktrace

    node_modules\react-native\Libraries\Core\InitializeCore.js:54 - Al

[View full stacktrace](https://app.bugsnag.com/sustainable-solutions-nz-ltd/msupply-mobile/errors/60ee654cb95ce400070f08f1?event_id=60ee654c007f8852bdb00000&i=gh&m=ci)

joshxg:
Fixes #4239 

## Change summary

- The original error was from `json-pointer` using an invalid reference. So, Catch that and set an error state.
- Wanted to use an error boundary with `componentDidCatch` but 1) event handlers does get caught (can be worked around with `setState`, though) and the nail in the coffin for that idea 2) it's working super funky on react-native. I believe it's not catching in development builds from what I can google.

## Testing

Use this schema:

```
const invalid = {
  type: 'object',
  title: 'Extra information',
  required: ['patientDemographicData', 'island', 'x'],
};
```

- [ ] When editing a patient, the app doesn't crash with an invalid json schema

- Looks like this:
image
- Looks like this with the keyboard up.. I thought it was kinda cool looking... but 🤷 
image
- Had no idea what to put as the error message, so feel free to change/suggest changes..
- At the moment you'll be able to edit a patient that has had a name note filled out previously (Because it hasn't changed, it's fine), but you won't be able to create one. Not sure if this is the right behaviour. Think it'll require a bit more hacking around to make it possible to still create a patient with if there's an invalid schema, but i'm not sure what the desired behaviour should be here, anyway. These invalid schemas shouldn't really be making it into the wild, anyway?

### DIFF
--- a/src/localization/generalStrings.json
+++ b/src/localization/generalStrings.json
@@ -2,7 +2,7 @@
   "gb": {
     "oh_no": "Oh no!",
     "please_contact_your_administrator": "Please contact your Administrator.",
-    "theres_been_an_error_with_this_forms_configuration": "Theres been an error with this forms configuration.",
+    "theres_been_an_error_with_this_forms_configuration": "There's been an error with this form's configuration.",
     "age": "Age",
     "remove_item": "REMOVE ITEM",
     "add_item": "ADD ITEM",
--- a/src/localization/generalStrings.json
+++ b/src/localization/generalStrings.json
@@ -1,5 +1,8 @@
 {
   "gb": {
+    "oh_no": "Oh no!",
+    "please_contact_your_administrator": "Please contact your Administrator.",
+    "theres_been_an_error_with_this_forms_configuration": "Theres been an error with this forms configuration.",
     "age": "Age",
     "remove_item": "REMOVE ITEM",
     "add_item": "ADD ITEM",
--- a/src/widgets/JSONForm/JSONForm.js
+++ b/src/widgets/JSONForm/JSONForm.js
@@ -177,7 +177,14 @@ export const JSONFormComponent = React.forwardRef(
     const [hasSchemaError, setSchemaError] = useState(false);
 
     const { uiSchema, jsonSchema } = surveySchema;
-    const validator = useMemo(() => ajv.compile(jsonSchema), [jsonSchema]);
+    const validator = useMemo(() => {
+      try {
+        return ajv.compile(jsonSchema);
+      } catch (e) {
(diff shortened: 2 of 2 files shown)


src/localization/generalStrings.json
```diff
@@ -1,5 +1,8 @@
 {
   "gb": {
+    "oh_no": "Oh no!",
+    "please_contact_your_administrator": "Please contact your Administrator.",
+    "theres_been_an_error_with_this_forms_configuration": "Theres been an error with this forms configuration.",
```
katherine-sussol, line 5:
```suggestion
    "theres_been_an_error_with_this_forms_configuration": "There's been an error with this form's configuration.",
```

🙀  ... I'm being one of those annoying grammar people

src/widgets/JSONForm/JSONForm.js
```diff
@@ -147,6 +174,8 @@ export const JSONFormComponent = React.forwardRef(
   ) => {
     if (!surveySchema) return null;
 
+    const [hasSchemaError, setSchemaError] = useState(false);
+
     const { uiSchema, jsonSchema } = surveySchema;
     const validator = useMemo(() => ajv.compile(jsonSchema), [jsonSchema]);
```
katherine-sussol, line 180: I've also found that this line can throw an error if the schema is super invalid (say if "required" is not defined as an array)

E.g.
`{"type": "object", "title": "ADR", "required": "not-exist"}`

katherine-sussol (approved):
Super cool! I like how in the ADRForm the breach man really looks like he's gone to space:
image

I think if the schema is invalid it's good to be as annoying as possible (short of crashing the app 🤣 ) so they are more likely to report the issue to us (otherwise may end up in a weird situation missing lots of vaccination/patient data?)

src/widgets/JSONForm/JSONForm.js
joshxg: Ohhh nice, thanks!

0a4e851

Also thanks for the grammer lessing, I appreciate !

PR #4313, opened 2021-07-28
#4311 better json live validation

### LINKED ISSUE #4311: better validation for JSONForm
## Describe the bug

Live validation is on on the patient edit modal from the dispensary page which isn't great

### To reproduce

### Proposed Solution

### Version and device info

- App version:
- Tablet model:
- OS version:

joshxg:
Fixes #4311 

## Change summary

- Dunno if this is good- i just removed the live validation 🤷 

## Testing

- [ ] No crazy error message spam on `PatientEditModal`

🤷

### DIFF
--- a/src/widgets/JSONForm/JSONForm.js
+++ b/src/widgets/JSONForm/JSONForm.js
@@ -14,7 +14,7 @@ import { JSONFormWidget } from './widgets/index';
 import { JSONFormErrorList } from './JSONFormErrorList';
 import { PageButton } from '../PageButton';
 import { JSONFormContext } from './JSONFormContext';
-import { useDebounce } from '../../hooks';
+import { useDebounce, useIsMounted } from '../../hooks';
 import LoggerService from '../../utilities/logging';
 import { FlexColumn } from '../FlexColumn';
 import { BreachManUnhappy } from '../BreachManUnhappy';
@@ -207,6 +207,16 @@ export const JSONFormComponent = React.forwardRef(
       []
     );
 
+    const mounted = useIsMounted();
+    const [noValidate, setNoValidate] = React.useState(true);
+    const touchedFields = {};
+
+    React.useEffect(() => {
+      if (mounted()) {
+        setNoValidate(false);
+      }
+    }, [mounted]);
+
     const debouncedOnChange = useDebounce(onChange, 500);
 
     return hasSchemaError ? (
@@ -221,6 +231,10 @@ export const JSONFormComponent = React.forwardRef(
             debouncedOnChange(change, validator);
           }}
           formData={formData}
+          onBlur={id => {
+            touchedFields[id] = true;
+          }}
+          noValidate={noValidate}
           validate={(newFormData, errorHandlers) => {
             // Validate the form data, if there are any errors, an `errors` object is set on
             // on the validator object.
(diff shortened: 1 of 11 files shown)


mark-prins (commented):
Not sure about the target branch - just hoping that you merge them all in the right order

I'm still playing around - the validation is still sub-optimal; with the extra information, the validation error wasn't showing until I tried NEXT a few times, and then they don't clear so it is hard to see what is still needing to be filled in

joshxg:
Ah you're right. Hm. Maybe we leave this out for now

mark-prins:
for me, the dispensing > patient edit is still live validating even with this set to `false`
confused.
I actually like it like that - and would prefer changing the vaccination > patient edit version too. That one is line 105 in `PatientEdit` which is currently set to `false`

src/widgets/JSONForm/JSONForm.js
```diff
@@ -207,6 +207,16 @@ export const JSONFormComponent = React.forwardRef(
       []
     );
 
+    const mounted = useIsMounted();
+    const [noValidate, setNoValidate] = React.useState(true);
+    const touchedFields = {};
+
+    React.useEffect(() => {
```
mark-prins, line 214: not strictly necessary now that touched status is being used - I thought good to reduce the number of times that validation is called so left it in

src/widgets/JSONForm/JSONForm.js
```diff
@@ -207,6 +207,16 @@ export const JSONFormComponent = React.forwardRef(
       []
     );
 
+    const mounted = useIsMounted();
+    const [noValidate, setNoValidate] = React.useState(true);
+    const touchedFields = {};
```
mark-prins, line 212: using `State` caused the form to reload, losing entered values.. so went with a mutable object. Not ideal, but working.

src/widgets/JSONForm/JSONForm.js
```diff
@@ -221,6 +231,10 @@ export const JSONFormComponent = React.forwardRef(
             debouncedOnChange(change, validator);
           }}
           formData={formData}
+          onBlur={id => {
```
mark-prins, line 234: second parameter is the value, not used in this case.

src/widgets/JSONForm/widgets/Checkboxes.js
```diff
@@ -19,16 +19,23 @@ const selectValue = (value, selected, all) => {
...
-      const all = enumOptions.map(({ value }) => value);
-      onChange(selectValue(option.value, checkboxesValue, all));
-    } else {
-      onChange(deselectValue(option.value, checkboxesValue));
-    }
+    const all = enumOptions.map(({ value }) => value);
+    const newValue = checked
+      ? onChange(selectValue(option.value, checkboxesValue, all))
+      : onChange(deselectValue(option.value, checkboxesValue));
+    onBlur(id, newValue);
```
mark-prins, line 38: in all of these widgets, not necessary to add the value to the onBlur - the method on the form is expecting it though, so I thought best to be complete just in case this is used later.

joshxg (commented):
Works great!

src/widgets/JSONForm/widgets/Textarea.js
```diff
@@ -28,6 +37,7 @@ export const Textarea = ({ autofocus, disabled, placeholder, value, onChange, nu
       autoFocus={autofocus}
       multiline={true}
       numberOfLines={numberOfLines}
+      onEndEditing={handleBlur}
```
mark-prins, line 40: this method returns the value of the text input, the `onBlur` doesn't with multiline text inputs

src/widgets/JSONForm/JSONForm.js
joshxg: Maybe better to `useRef` so that there isn't future hunchy funchy with `memo`ing - but I don't think it's needed right now

mark-prins: good point!

PR #4315, opened 2021-07-29
#4314 filter out refusals

### LINKED ISSUE #4314: filter out refused vaccination history
## Describe the bug

Currently vaccination history is showing refused events, we should filter them out as there is no item name or item code and it looks funny

### To reproduce

### Proposed Solution

### Version and device info

- App version:
- Tablet model:
- OS version:

joshxg:
Fixes #4314 

## Change summary

- Filters out refusals because they look ugly.
- Maybe it would be better to have them, but I dunno how to deal with the columns in that case.. 🤷 

## Testing

- [ ] Refusals vaccination prescriptions do not show in vaccination history

N/A

### DIFF
--- a/src/widgets/modals/VaccinationHistory.js
+++ b/src/widgets/modals/VaccinationHistory.js
@@ -20,11 +20,9 @@ const EmptyComponent = () => (
 const jsonSchema = {
   type: 'object',
   properties: {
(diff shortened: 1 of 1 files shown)


mark-prins (approved):
👍

PR #4322, opened 2021-07-29
#4283 Add items to deps of useEffect

### LINKED ISSUE #4283: ADR on Mobile does not show remote vaccinations if both remote and local vaccinations exist for patient
## Describe the bug

_Note: Not an urgent issue as user can still see local vaccinations and this is the most likely scenario when creating an ADR from a tablet._ 

If a user has both remote and local vaccinations then only the local vaccinations show when creating an ADR in Mobile. 

If a user has only local vaccinations then the local vaccinations show up ok
If a user has only remote vaccinations then the local vaccinations show up ok. 

On the webform ADR both the local and remote vaccinations show up ok. 

image

image

image

### To reproduce

1. Dispense a vaccination to Patient A on Tablet A
2. Dispense a vaccination to Patient A on Tablet B
3. Open the ADR form for Patient A on Tablet A
4. Only the vaccination from Tablet A will appear. 

### Expected behaviour

Both the vaccination from Tablet A and Tablet B to appear. 

### Proposed Solution

### Version and device info

- App version: 8.2.0-rc2
- Tablet model:TB-X505F
- OS version: Android 10

joshxg:
Fixes #4283 

## Change summary

- OK I'm not gonna lie I haven't tested this at all. I was looking at the code and then when I went to test it I realised I didn't have the correct schema with causes in it - this is a complete guess and I probably shouldn't PR it......

## Testing

- [ ] Do something

### DIFF
--- a/src/widgets/modalChildren/ADRInput.js
+++ b/src/widgets/modalChildren/ADRInput.js
@@ -78,7 +78,7 @@ export const ADRInputComponent = ({ onCancel, onSave, patient, patientHistory })
 
       setADRSchema({ uiSchema, jsonSchema, properties: { type, version: version + 1 } });
     }
-  }, [data, patientId, searched]);
+  }, [data, patientId, searched, items]);
 
   return (
     <>


katherine-sussol (commented):
I tried adding items as a dep too, thinking it might work but it ends up doing this:
image.

I didn't quite manage to figure out what was going on though (so this isn't very helpful). It does seem like `useLayoutEffect()` is somehow not updating the UI (can see that the function correctly has more `items` (from spam console logging) after doing the remote lookup, but for some reason, `setADRSchema` is not refreshing the UI with those new items... 😕 )

joshxg:
Ha, sorry for wasting your time! Thought I'd take a swing..!

PR #4336, opened 2021-08-04
#3966 Adjust threshold temperatures to match the step amount

### LINKED ISSUE #3966: Breach temperature thresholds should realign on .5
## Describe the bug

When changing the breach temperature tresholds, if one of the thresholds gets close to the other, it is reduced by .1. Then, the threshold can never again align on .5

### To reproduce

1. Go to set up a new sensor
2. Click on the hot threshold and make it 8 degrees
3. Go to the cold threshold and bring it up close to 8 degrees - it will land on 7.9.
4. Then, reduce the cold threshold. It was go to 7.4, 6.9 6.4 etc, never realigning on .5

### Expected behaviour

Should realign on .5

### Proposed Solution

TBD
### Version and device info

- App version:
- Tablet model:
- OS version:

katherine-sussol:
Fixes #3966

## Change summary

- When a temperature approaches a threshold (i.e. cold temperature threshold = hot temperature threshold), make the offset 0.5 degrees celsius, rather than 0.1. This matches the step amount configured in the `TemperatureEditor` component.

## Testing

- [ ] Follow steps in #3966

Pretty lazy change (could have tried some fancy rounding maths), but it doesn't really make sense to have or allow the hot and cold temperature thresholds to be within 0.1 degrees of each other, I think (that would be a really fragile vaccine?!? 💉 )

### DIFF
--- a/src/selectors/Entities/temperatureBreachConfig.js
+++ b/src/selectors/Entities/temperatureBreachConfig.js
@@ -33,11 +33,11 @@ export const selectEditingConfigThresholds = state => {
     COLD_CUMULATIVE: coldCumulativeConfig = {},
   } = editingConfigs;
 
-  const coldConsecutiveThreshold = (hotConsecutiveConfig.minimumTemperature ?? 0) - 0.1;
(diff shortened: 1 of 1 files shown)


joshxg (approved):
That'll do it

PR #4341, opened 2021-08-05
Typescript

clemens-msupply:
Adds build and lint support for TypeScript. No functional changes intended.

The PR has some files migrated to TS. These files have been selected because I was looking at them... I only added some type defs in these files.


joshxg (commented):
I'm pretty excited for this but I want to get some other input from other mobile devs.

@mark-prins @katherine-sussol 

The plan would be to slowly migrate files? New files to use TS?

clemens-msupply:
Yes files can be slowly migrated when there is time/need for it.

katherine-sussol:
> I'm pretty excited for this 

Me too! Initial thoughts were maybe a guideline for new files to use TS? 
Could also create an epic for migration and anyone who wants to spend some time doing some migration can create an issue for themselves outlining the intended area(s)/files? 🤷  
As we work on issues we could also migrate impacted files, but I don't think we should try to set a hard and fast rule around this, as I could see it getting frustrating, especially if working on time-critical issues.

clemens-msupply:
When migrating a file its also important to make sure nobody is working on it currently because otherwise its get very difficult to merge...

wlthomson:
I'm a big fan of typescript, so my initial reaction is to approve this. Some pros/cons from my perspective:

Pros:

- Fewer runtime errors (probably the biggest pro?).
- OOP design patterns are a lot easier and nicer.
- Soft-ish introduction to static typing (Softer than Rust, anyway!)
- Easier integration with some libraries.

Cons:

- Time investment in non-functional changes.
- Maintaining a codebase in two languages (is it realistic to migrate the entire codebase?).
- Learning curve (ts error messages have a rep for being pretty opaque).
- Some libraries may not have `.ts` definitions (IME not so much a problem nowadays).
- Bad past experiences with React + ts make me more prone to adopting ts for backend rather than front-end projects.

Questions:

- Do our existing dev tools play nicely with typescript? e.g. does hot-reloading still work?

TLDR; some minor reservations, but a 👍 from me.

clemens-msupply:
Regarding your questions:

- Yes TS plays relatively nicely with the tools (there are currently some eslint issues that might need to be tweaked). Hot reload is still working.
- The most fundamental data types (e.g. DB models) and redux states would probably benefit most from being converted to TypeScript first. 
- Yes realm has TS type declarations.

Some more pros: 
- Code becomes IHMO easier to read, e.g. you don't have guess what the parameters are or what option you have.
- It is much more expressive, e.g. you can just write `test(param: string | number)` instead of putting a comment about the parameter into the docs or using JSDoc.
- Very good for catching 'undefined reference'/ null errors at compile time

andreievg:
I like the idea, I think definitely yes for any new files. My concern is with existing code, it's one of those 'pull the string on a sweater' scenarios. I think it would require very careful planning and well defined process, and confirmation from stakeholders (Craig and Tony) that we have the budget for this work (I would assume that there is a few months of work change to typescript)

clemens-msupply:
Yes I wouldn't set the goal to migrating existing files to TypeScript. Initially you don't need to migrate any files and there will probably be files that will never be migrated because they are working and nobody touches them. I would only migrate files where it makes sense. Some examples where it can make sense:
- if you can't easily verify the correctness of some code because there are too many types involved, e.g. complex types form multiple files
- as documentation for complex existing code so that the code becomes easier to understand (similar to above)
- hunting type related bugs
- to improve the tooling in the IDE, e.g. to get better code completion for common data such as app state

Migrating a file to TS can always be a bit dangerous because you can introduce some subtle bugs when fixing TS error. (On the other hand you might also catch some bugs.)

joshxg (commented):
Sorry bout the hold up.

I'm getting a lot (600+) ts errors like the below. Looks like we need to supply an `outDir` or `noEmit: true` to stop this. I think `noEmit: true`?

```
 {
	"resource": "/Users/joshuagriffin/repos/mobile/tsconfig.json",
	"owner": "typescript",
	"severity": 8,
	"message": "Cannot write file '/Users/joshuagriffin/repos/mobile/babel.config.js' because it would overwrite input file.",
	"source": "ts",
	"startLineNumber": 1,
	"startColumn": 1,
	"endLineNumber": 1,
	"endColumn": 2
}
```

---

I think the general consensus is "Would be great to use TS but we don't have the resources to do a full conversion, doing it slowly is probably fine"

If we're to do it slowly, I have some questions so that we don't have PRs getting blocked moving forward (Like this one!).

Using an examples from this PR:

The localisations have ts errors (but work fine). Seems we 'should' be using `getString` rather than accessing the string directly. I.e:

```
generalStrings.couldnt_import_data
```

Should be:

```
generalStrings.getString("couldnt_import_data")
```

Sure - loving the key check! But.. you're not touching these. So should we change them in this PR which is touching the file? Or do we make a new issue (Where they will likely pile up and not be done? Maybe this is a bad assumption).

Additionally, the types added to some of the members i.e. `objects(type: string)` I think offer little practical value. I have been caught before using `UIDatabase.objects("Items")` rather than `UIDatabase.objects("Item")` (Or is it the other way round?!), so I might suggest that we create an enum or type for each of the valid object names. But.. that seems out of scope for this PR 🤔  

So.. do we just end up with poorly typed, typescript with a bunch of `any`s, `unknown`s and little value primitives mixed in with a bunch of JS?

mark-prins:
Agree with @joshxg ( including your pessimism about things being changed if we just create an issue and wait... ) and others: great to have ts in place, we'll need to tread with caution, and go with useful type checking first.

Would be great to have the strings migrated to use `getString` in this PR - however this doesn't provide key checking for me. I updated `tsconfig.json` with the [recommended settings](https://www.npmjs.com/package/react-native-localization) and that didn't help 🤷‍♂️ 

The `PrescriptionReducer` could do with an interface for the state in order to provide a return type for the reducer.


Would also suggest something like this in `UIDatabase`:
```
export type CoreDatabaseType =
  | 'ADRForm'
  | 'ActiveLocation'
  | 'CashTransaction'
  | 'CashTransactionName'
  | 'CashTransactionReason'
  | 'Customer'
  | 'CustomerCredit'
  | 'CustomerInvoice'
  | 'CustomerTransaction'
  | 'ExternalSupplier'
  | 'InternalSupplier'
  | 'MedicineAdministrator'
  | 'NegativeAdjustmentReason'
  | 'OpenVialWastageReason'
  | 'PCDEvents'
  | 'Patient'
  | 'PatientSurveyForm'
  | 'Payment'
  | 'Policy'
  | 'PositiveAdjustmentReason'
  | 'Prescription'
  | 'PrescriptionCategory'
  | 'Provider'
  | 'Receipt'
  | 'RequestRequisition'
  | 'RequisitionReason'
  | 'ResponseRequisition'
  | 'Supplier'
  | 'SupplierCreditCategory'
  | 'SupplierInvoice'
  | 'SupplierTransaction'
  | 'Vaccine';

const translateToCoreDatabaseType = (type: CoreDatabaseType) => {
```

clemens-msupply:
Currently looking into the strings problem. The problem is that some languages have missing translations which confuses TS. Currently trying to cast to the 'gb' type...

clemens-msupply:
@mark-prins thanks for the type defs. Should the type better be named RealmDatabaseType and the return type is CoreDatabaseType? Or what would be good names?

joshxg:
@clemens-msupply yup your name is better. Other options I can think of.. `ProxyDatabaseType` `UIDatabaseObjectType` `UIObjectType` , `ViewObjectType`, `AlternateRealmType` `RealmObjectsType`...

Uh. None are better than yours..

mark-prins:
@clemens-msupply yep, your name is better :) thanks

src/database/UIDatabase.ts
```diff
@@ -17,7 +17,56 @@ import { version as appVersion } from '../../package.json';
...
+  | 'Options'
+  | 'Requisition'
+  | 'InsurancePolicy'
+  | 'InsuranceProvider'
+  | 'TransactionCategory'
+  | 'PatientEvent'
+  | 'Item'
+  | 'FormSchema'
+  | 'MedicineAdministrator'
+  | 'Location';
```
joshxg, line 67: Technically all of these within the schema are a core database type https://github.com/openmsupply/mobile/blob/4d3a1ddbb0836003f1cbc59ae4c563f1ba52c321/src/database/schema.js#L196-L259

src/database/UIDatabase.ts
```diff
@@ -17,7 +17,56 @@ import { version as appVersion } from '../../package.json';
 
 const { THIS_STORE_NAME_ID, APP_VERSION } = SETTINGS_KEYS;
 
-const translateToCoreDatabaseType = type => {
+export type RealmDatabaseType =
```
joshxg, line 20: this should be `| CoreDatabaseType` also actually. It's a superset of all the types. 

So, all the core database schema types are of type `CoreDatabaseType` and the additional types i.e. `Vaccine`, which is a filtered subset of `Item` , are the additions for `RealmDatabaseType`.

Currently, if I use `UIDatabase.objects("Transaction")`, which is valid and common, the key is not valid.

joshxg (changes requested):
Thanks, with the small change mentioned i'd be happy to merge this.

I think it would be great if we could do a small write up, maybe in the README for guidelines on when to use TS and when to convert an old file that we can refer to. Also maybe guidelines on where to place common types would be helpful?

clemens-msupply:
I fixed some more linter problems and solved the LocalizedStrings issue...

Apparently React native doesn't take TS into account when compiling the bundle but just strips the type defs from the files (!). Maybe we should put `tsc` build into a github action to catch compiler errors...

src/database/UIDatabase.ts
joshxg: Still this comment to address. Theres a set of object type strings between the two ts types which give errors but are valid i.e `UIDatabase.objects('Stocktake');`

clemens-msupply: I just had a look but I think I lack the knowledge about which table belongs where... Should I create an issue or should we go through it together once you are back in the office?

joshxg:
Sorry @clemens-msupply - I wanted to get this one merged and stop blocking you. 

I'm likely not in for at least a few days next week either, so I thought I'd commit : 1a266b5 which adds all of the schema data types from https://github.com/openmsupply/mobile/blob/4d3a1ddbb0836003f1cbc59ae4c563f1ba52c321/src/database/schema.js#L196-L259 into `CoreDatabaseType` - hopefully it makes sense (and there's no mistake!).

joshxg (approved):

Issue #4319, opened 2021-07-29
v8.3.0 Release Discussion


joshxg:
## RELEASE VERSION

v8.3.0 

### SUMMARY

This release is primarily aimed at misc bug fixing

### Chores
- #4337 

### Refactors/Enhancements
- #1546
- #4361
- https://github.com/openmsupply/mobile/pull/4321
- #4341

### Features
- #3981
- #3984
- #4080
- #4367

### Bugs
- #3771
- #3966 
- #4047
- #4334
- #4347
- #4348
- #4367
- #4380
- #4416
- #4424

## Documentation
- TBD

joshxg:
@katherine-sussol 

I think we should put an rc out and ask @ArpitaSussol if she has time to get the needs build testing issues done. Would be good to get a real build out especially to test the jsc bump and 🤞 it fixes some seg faults.. 🤣 


https://media0.giphy.com/media/1WHAlOdc9ZNUZq85JI/giphy.gif?cid=5a38a5a2r02f5ae2atyh39q0n5bx943pl27lnclxlf8xv52o&rid=giphy.gif&ct=g

ArpitaSussol:
> @katherine-sussol
> 
> I think we should put an rc out and ask @ArpitaSussol if she has time to get the needs build testing issues done. Would be good to get a real build out especially to test the jsc bump and 🤞 it fixes some seg faults.. 🤣
> 
> https://media0.giphy.com/media/1WHAlOdc9ZNUZq85JI/giphy.gif?cid=5a38a5a2fx5195ncxywrkyel3wqgcp8qpcbch6md1px7fpi4&rid=giphy.gif&ct=g
> 
> I cant drag and drop gifs from gify extension anymore this sucks
> 
> anyway...
> 
> Would you have time?

nice, i will start testing with above mentioned features and fixes once apk is available.

katherine-sussol:
> @katherine-sussol
> 
> I think we should put an rc out and ask @ArpitaSussol if she has time to get the needs build testing issues done. Would be good to get a real build out especially to test the jsc bump and 🤞 it fixes some seg faults.. 🤣
> 
> https://media0.giphy.com/media/1WHAlOdc9ZNUZq85JI/giphy.gif?cid=5a38a5a2fx5195ncxywrkyel3wqgcp8qpcbch6md1px7fpi4&rid=giphy.gif&ct=g
> 
> I cant drag and drop gifs from gify extension anymore this sucks
> 
> anyway...
> 
[...]

PR #4371, opened 2021-08-19
Bump 8.3.0-rc0

katherine-sussol:
For #4319

## Change summary

If there is another way to do this without PR-ing 😿... I feel like I have managed to bump the version before without needing a PR but I can't remember.

## Testing

- [ ] N/A

N/A


katherine-sussol:
Oh I know, I think I don't need this if I make a separate branch release/8.3.0

PR #4381, opened 2021-08-25
#4047 Remove dispensary from usingModules check

### LINKED ISSUE #4047: Module section shows on navigator when no modules enabled
## Describe the bug

Set up a new store, don't enable any modules ( experienced with the Tonga server, so check that if you can't repro )
The module section is shown on the navigator, but there are no modules in it - no menu buttons.

Screenshot:
screen

### To reproduce

1. Add a store with no items visible
2. Select the preference to use vaccines
4. Have fun!

### Expected behaviour

The module section should not be shown

katherine-sussol:
Fixes #4047

## Change summary

- Removes the presence of dispensary menu item from triggering module section to appear (as the Dispensary was moved into the same menu section as Customer Invoices / Customer Requisitions).

image

## Testing

- [ ] Have no modules enabled and no visibility of any vaccine items in a mobile store, but have the dispensary enabled. Module section should not show.

Is the dispensary actually a module? 🤔

### DIFF
--- a/src/reducers/ModulesReducer.js
+++ b/src/reducers/ModulesReducer.js
@@ -36,7 +36,7 @@ const initialState = () => {
     UIDatabase.getPreference(PREFERENCE_KEYS.CAN_EDIT_PATIENTS_FROM_ANY_STORE)
   );
 
-  const usingModules = usingDashboard || usingDispensary || usingVaccines || usingCashRegister;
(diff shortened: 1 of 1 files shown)


joshxg (approved):
Nice, thanks!

Dispensary was originally a preference (like the dashboard) and was changed to use the store mode. 

I'm not sure about having the vaccinations button show up if you have vaccine items rather than it being a module/preference..  🤷  something for another day 😁

PR #4392, opened 2021-09-01
#4387 Set requistion.isRemoteOrder to false by default

### LINKED ISSUE #4387: Set requistion.isRemoteOrder to false by default on the mobile
Set requistion.isRemoteOrder to false by default on the mobile

It seems that the field `requistion.isRemoteOrder` is set to true on the mobile where as in the desktop remote site it is set to false.  

So `isRemoteOrder` does not really literally mean remote order as both of these desktop and mobile requisitions are remote. 

As a result converting a desktop system to a mobile system is not working smoothly :  
https://support.msupply.org.nz/cerb6/profiles/ticket/ZN-4182-LB

Further for PNG we are suddenly seeing extra columns for Requisition Response forms. 

https://support.msupply.org.nz/cerb6/profiles/ticket/12969

Now @adrianwb  pointed out that `requistion.isRemoteOrder` is used by Ivory Coast to create orders in the cloud when the remote active tablet is stolen. mSupply desktop app, supports this theory. 

## The suggestion

So the conclusion : 

- The name of the field `requistion.isRemoteOrder` is confusing and it really should be `requistion.isRemoteOrderCreatedonCloud`  [ due to compatibility options we can't chnage this now] 
- We simply get the mobile tablet to ignore this field `requistion.isRemoteOrder` and set it to false.
- Data control on the Mobile to be handle in a different way and `requistion.isRemoteOrder` should not be used.

**Why?**

This will make the Mobile and desktop data structure the same. 

No can't think of any thing better.

## Additional context

If theses user column are asked by the user then we can add a store preference in the desktop.

sworup:
## Description

- Fixes #4387
- Set requisition.isRemoteOrder to false by default
- Also remove isRemoteOrder check disabling deletion of requisition if isRemoteOrder is false. Setting isRemoteOrder to false by default would have - otherwise disabled deletion of these newly created requisitions

## Change summary

- **Code change:** We are not handling the `isRemoteOrder` field of the `requisition` table, in mobile, at all. 
  - The idea is `isRemoteOrder` is used on the desktop for a purpose that is completely different from our mobile legacy code's interpretation of that field.
  - _FYI: In desktop `isRemoteOrder` is set to true to allow these requstion to be edited on the server side in case of remote devices getting lost._
  - The setting of isRemoteOrder to true is causing desktop requisition view to show extra columns that are not supposed to show for mobile created requisitions.

## Testing

- [ ] Open requisition list, none of the requisitions in the list should have deletion disabled now that we are disabling `isRemoteOrder` check.
- [ ] Create a requisition from mobile
  - [ ] Go to desktop to check, it should not have extra columns.

- Old requisitions may still show extra columns on the desktop until the foot-runner is run to set those requisitions (from mobile) with isRemoteOrder false.

### DIFF
--- a/src/database/DataTypes/Requisition.js
+++ b/src/database/DataTypes/Requisition.js
@@ -500,7 +500,7 @@ Requisition.schema = {
     otherStoreName: { type: 'Name', optional: true },
     customData: { type: 'string', optional: true },
     createdDate: { type: 'date', default: new Date() },
-    isRemoteOrder: { type: 'bool', default: true },
+    isRemoteOrder: { type: 'bool', default: false },
   },
 };
 
--- a/src/database/DataTypes/Requisition.js
+++ b/src/database/DataTypes/Requisition.js
@@ -451,6 +451,15 @@ export class Requisition extends Realm.Object {
     return finaliseStatus;
   }
 
+  /**
+   * Get if this requisition is linked to a requisition.
+   *
+   * @return  {boolean}
+   */
+  get isLinkedToRequisition() {
+    return !!this.linkedRequisition;
+  }
+
   get canFinalise() {
     return this.isRequest ? this.canFinaliseRequest : this.canFinaliseResponse;
   }
@@ -497,6 +506,7 @@ Requisition.schema = {
     linkedTransaction: { type: 'Transaction', optional: true },
     program: { type: 'MasterList', optional: true },
     period: { type: 'Period', optional: true },
+    linkedRequisition: { type: 'string', optional: true },
     otherStoreName: { type: 'Name', optional: true },
     customData: { type: 'string', optional: true },
     createdDate: { type: 'date', default: new Date() },
--- a/src/sync/incomingSyncUtils.js
+++ b/src/sync/incomingSyncUtils.js
@@ -719,6 +719,8 @@ export const createOrUpdateRecord = (database, settings, recordType, record) => 
         period = database.getOrCreate('Period', record.periodID);
       }
 
+      const linkedRequisition = record.linked_requisition_id || null;
+
       internalRecord = {
         id: record.ID,
         status: REQUISITION_STATUSES.translate(record.status, EXTERNAL_TO_INTERNAL),
@@ -733,6 +735,7 @@ export const createOrUpdateRecord = (database, settings, recordType, record) => 
         thresholdMOS: parseNumber(record.thresholdMOS),
         program: database.getOrCreate('MasterList', record.programID),
         period,
+        linkedRequisition,
         orderType: record.orderType,
         customData: parseJsonString(record.custom_data),
         isRemoteOrder: parseBoolean(record.isRemoteOrder),
--- a/src/widgets/DataTable/DataTableRow.js
+++ b/src/widgets/DataTable/DataTableRow.js
@@ -107,15 +107,25 @@ const DataTableRow = React.memo(
           // Indicator if the right hand border should be removed from styles for this cell.
           const isLastCell = index === columns.length - 1;
 
-          const { isLinkedToTransaction, isFinalised: rowIsFinalised } = rowData;
+          const {
+            isLinkedToTransaction,
+            isFinalised: rowIsFinalised,
+            isLinkedToRequisition,
+            isResponse,
+          } = rowData;
           // This cell is disabled if:
           // - the page is finalised.
           // - the row has been explicitly set as disabled.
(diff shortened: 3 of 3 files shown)


joshxg (commented):
There is existing logic that we can't delete customer requisitions which have been created through sync.

I'm not sure of why this is, exactly, I want to make sure we can remove that logic or if we need to add it back, though (We could check if a customer requisition has a linkedTransaction, and not allow deleting it, if true).

@Chris-Petty - do you know why since forever ago, we have not been able to delete Customer Requisitions from the server?

ujwalSussol (commented):
@sworup This looks good to me. 

I think we need to make a APK for testing.

ujwalSussol:
@sworup @joshxg  @Chris-Petty  :

I spent 45 minutes with @ArpitaSussol  looking at the current `Requisition - Response`  work flow and it is true that : 

-  `Requisition - Response` that is created via external order is not deletable. 
- `Requisition - Response` that is created in the mobile it self is allowed to be deleted. 

However we have a way to maintain the above workflow in the following manner : 

- we use the field [requisition]linked_requisition_id  to control the delete.
- If  [requisition]linked_requisition_id is empty or blank then the `Requisition - Response` is local and allow delete.
- if  [requisition]linked_requisition_id has a 'UUID' data then this is a external data and we don't allow delete. 
- Problem solved : We no longer need to rely on field `isRemoteOrder`

src/database/DataTypes/Requisition.js
```diff
@@ -497,6 +506,7 @@ Requisition.schema = {
     linkedTransaction: { type: 'Transaction', optional: true },
     program: { type: 'MasterList', optional: true },
     period: { type: 'Period', optional: true },
+    linkedRequisition: { type: 'string', optional: true },
```
joshxg, line 509: This won't work for historic requisitions, you would need to create migration code to send a message to the server saying the mobile site has upgraded, then have the server send all customers requisitions back to all mobile sites so we can save the id of the linked requisition.

An easier solution is to use `linkedTransaction`. All customer requisitions which have come from the server will have a `linkedTransaction` on mobile, while manually created ones on the mobile site itself will not.

This means we won't need to create migration code, nor will we need to add this field

joshxg (changes requested):

ujwalSussol:
@joshxg   if `linkedTransaction` is set on the cloud for  requistion.respose  form even when CI is not created then using this field will be OK. 

@sworup  :  you will need to do a requisition work flow to work out when  `linkedTransaction` is set.

katherine-sussol:
Hi @sworup! Would you want to look at addressing the changes requested for this one or would you prefer to have someone else take a look? (just wary this PR/issue has been open for a while and would be good to close it out)

sworup:
@katherine-sussol I am on paternity leave till 24th December so I might not be able to look into this right away.

Chris-Petty:
@sworup omg congrats!

sworup:
# About trans lines not showing in desktop after syncing when new customer requisition is created

- Here is a screenshot of the single requisition thus created, its transaction record and trans_line records in realm database:
Screen Shot 2022-02-18 at 4 57 07 PM
  - Notice here the `transactionItem inside createTransactionItem`'s id and `transaction inside createTransactionItem`'s id. The `transaction` id is same as `transactionItem` id and also the 4 `transactionItem` lines have same id. There is a `generateUUID()` method inside `createTransactionItem` that should have generated unique IDs but it is not. I think that what causes the records to no sync back to desktop hence causing the missing lines.

sworup:
@Chris-Petty 
I talked with @ujwalSussol sir and we have come to this conclusion:

- In case of conversion, we should probably do a check during store site conversion. If it is mobile store, we do not allow conversion until these suggested customer requisitions are handled. They should either be deleted or finalised. Ask the users to handle it and give them a choice to allow mSupply to delete these requisitions. This what I think, @mark-prins did to handle this issue months ago, he asked the client to do it, maybe we should put this in code during desktop to mobile site conversion.
- We do have an SG response requisition creating linked CI issue in mobile though. https://github.com/openmsupply/mobile/issues/4592

katherine-sussol:
Closing as per comments:
- https://github.com/openmsupply/mobile/pull/4392#issuecomment-1047473397
- https://github.com/openmsupply/mobile/issues/4387#issuecomment-1056245005 
 
Desktop issue:
https://github.com/sussol/msupply/issues/10490

(thanks @sworup for all the investigation!)

PR #4415, opened 2021-09-08
#4414 Update ADRInput to look in correct place for vaccinations

### LINKED ISSUE #4414: ADRInput should look at name_notes for vaccine history
## Describe the bug

Missed from #4347 - failed my own test case.
Local vaccinations are being skipped from the ADR vaccination history.
Happens only in the current develop branch, not in any released app versions.

### To reproduce

1. Create a vaccination transaction (don't sync)
2. Ensure the form schema for ADR has a vaccine history element
3. Create an ADR and try to add the vaccination as a possible cause
4. See missing vaccination

### Expected behaviour

Should see the local vaccination

### Version and device info

- App version: develop branch
- Tablet model: Emulator
- OS version: 8.1

### Additional context

N/A

katherine-sussol:
Fixes #4414

## Change summary

Update ADRInput to pull all local vaccinations (from name_note records) instead of filtering on transactions (as the filter doesn't work anymore).

## Testing

- [ ] Follow repro steps in #4414 - should see local vaccinations

### DIFF
--- a/src/selectors/Entities/name.js
+++ b/src/selectors/Entities/name.js
@@ -112,8 +112,9 @@ export const selectVaccinePatientHistory = patient => {
       ({ patientEventID, data }) =>
         patientEventID === vaccinationPatientEventID && validateJsonSchemaData(jsonSchema, data)
     )
-    .map(({ data: vaccinationNameNotes }) => ({
+    .map(({ id, data: vaccinationNameNotes }) => ({
       ...vaccinationNameNotes,
+      id,
       doses: 1, // Currently not possible to dispense more than 1 dose
       totalQuantity: 1,
       confirmDate: new Date(convertMobileDateToISO(vaccinationNameNotes.vaccineDate)),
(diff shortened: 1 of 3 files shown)


src/selectors/Entities/name.js
```diff
@@ -112,8 +112,9 @@ export const selectVaccinePatientHistory = patient => {
       ({ patientEventID, data }) =>
         patientEventID === vaccinationPatientEventID && validateJsonSchemaData(jsonSchema, data)
     )
-    .map(({ data: vaccinationNameNotes }) => ({
+    .map(({ id, data: vaccinationNameNotes }) => ({
```
katherine-sussol, line 115: This is to stop an error when 2 doses of the same vaccine (item and batch) are locally dispensed in the same day (although likely an error, the error displayed was the generic invalid schema one and not that useful for users): 
`LOG  text {"name":"JSONForm","hostname":"react-native","level":40,"msg":"Invalid JSON Schema Error: schema is invalid: data/properties/causes/items/enum should NOT have duplicate items (items ## 2 and 3 are identical)`

joshxg (approved):
Nice, sorry I missed this!

PR #4425, opened 2021-09-28
#4424 trim trailing '/' chars from sync url

### LINKED ISSUE #4424: Trim trailing '/' characters when entering sync url into mobile

When there are trailing '/' characters at the end of an otherwise valid URL, sync will fail because mobile con🐱enates this URL with `/sync/version/endpoint/`, doubling up the '/' characters.

See https://github.com/sussol/msupply/issues/9614 for related issue raised in Desktop.

Trim the trailing '/' characters before saving the sync url (which can currently be entered on the initialisation screen or settings screen)

## Implementation

- Throwing invalid url error message (though this specific case is for when URLs are valid but just have an extra '/' at the end)
- Leaving the mess for desktop to deal with (think there is already a workaround being put in Desktop, but still better practice to clean up client behaviour too)

## Additional context
We already have #276 for general URL validation, this is a more specific use-case within that.

katherine-sussol:
Fixes #4424

## Change summary

- Super minor change to trim extraneous '/' characters at the end of an otherwise valid url. See issue for history. 
- Minor offroad for grammar (side note: the 'Make sure you know what you are doing' prompt always makes me hesitate, feels like such a bold claim).

## Testing

- [ ] Set up (on the initialise or settings page) a URL ending with '/', e.g. `http://192.168.1.158:5000/` when saved, sync/login should still work

- FirstUsePage does whitespace trimming `onBlur()`, didn't put this there because that might be weird for users to see it being edited for them. Thought it was better to let them think it was fine and just clean it up in the background.

### DIFF
--- a/src/localization/authStrings.json
+++ b/src/localization/authStrings.json
@@ -9,7 +9,7 @@
     "site": "Site",
     "store": "Store",
     "invalid_username_or_password": "Invalid username or password",
-    "warning_sync_edit": "WARNING:\nThis will effect syncing for this store.\nMake sure you know what you doing!",
+    "warning_sync_edit": "WARNING:\nThis will affect syncing for this store.\nMake sure you know what you are doing!",
     "incompatible_server": "The mSupply Server you're trying to sync with is not compatible with your current version. Please contact your administrator."
   },
   "fr": {
--- a/src/pages/FirstUsePage.js
+++ b/src/pages/FirstUsePage.js
@@ -52,9 +52,12 @@ export class FirstUsePageComponent extends React.Component {
     const { onInitialised, synchroniser } = this.props;
     const { serverURL, syncSiteName, syncSitePassword } = this.state;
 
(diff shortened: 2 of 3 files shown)


src/localization/authStrings.json
```diff
@@ -9,7 +9,7 @@
     "site": "Site",
     "store": "Store",
     "invalid_username_or_password": "Invalid username or password",
-    "warning_sync_edit": "WARNING:\nThis will effect syncing for this store.\nMake sure you know what you doing!",
+    "warning_sync_edit": "WARNING:\nThis will affect syncing for this store.\nMake sure you know what you are doing!",
```
mark-prins, line 12: I clearly hadn't stopped to read this carefully before! good catch :)

mark-prins (approved):
👍  Great, thanks. we had the same issue in the dashboard back end plugin and for a moment I thought that this was a zombie issue before realising it was the same problem in a different project! 🧟

PR #4494, opened 2021-11-04
Ble module

chetstone:
- Works with msupply-ble-module
- Enables support for Laird sensors
 
Trimming of macAddress in screen not done yet: displays macAddress and sensor type. ( Note that the display is not crowded in mobile, so this may be desirable)


chetstone:
Addresses issue #4498

katherine-sussol (approved):
Sorry Chester for the delay! For this PR I've only looked at the changes in the mobile src as I noted there was a separate PR in the msupply-ble-service repo

Looks good and works well in mobile for me with both Laird and BlueMaestro running side by side 👍 .
Quite neat that mobile itself integrates nicely with the new module without needing too many code changes. Thanks also for updating the VaccinePage - I think it would be nice to have the macAddress decoupled from the deviceType (maybe we could store the type in the Sensor table as a separate field instead of as part of the macAddress?). 

But I think this is fairly minor, overall everything is working great and think it would still be good to get this merged and tested

chetstone:
Thanks a lot Kat. Yes, I agree that squeezing the device type in with the macAddress is a hack. Perhaps we can look at fixing that down the line. I came up with this when working on ColdChain-- I didn't feel confident in adding a field to the SQLite schema. And Mobile uses a different database, about which I know even less. I'm afraid this could lead to many changes being required throughout the apps. And then there's the fact that both apps synchronize with central databases. Would changes in schemas be required upstream as well? For that matter, it's possible that my hack could cause problems upstream.  For example, in ColdChain, the macaddress field is declared as a varchar of length 17. AFAIK, SQLite ignores the length spec and just stores whatever it gets. But if upstream DBs have a similar declaration, and they're more strict, some problems could arise. So I hope there's a plan to include upstream sync in the testing.via Newton MailOn Wed, Nov 10, 2021 at 8:55pm, Kat ***@***.***> wrote:

@katherine-sussol approved this pull request.



Sorry Chester for the delay! For this PR I've only looked at the changes in the mobile src as I noted there was a separate PR in the msupply-ble-service repo

Looks good and works well in mobile for me with both Laird and BlueMaestro running side by side github.githubassets.com/images/icons/emoji/unicode/1f44d.png">👍 .

Quite neat that mobile itself integrates nicely with the new module without needing too many code changes. Thanks also for updating the VaccinePage - I think it would be nice to have the macAddress decoupled from the deviceType (maybe we could store the type in the Sensor table as a separate field instead of as part of the macAddress?).

But I think this is fairly minor, overall everything is working great and think it would still be good to get this merged and tested



—You are receiving this because you authored the thread.Reply to this email directly, view it on GitHub, or unsubscribe.

[

{

***@***.***": "http://schema.org",

***@***.***": "EmailMessage",

"potentialAction": {

***@***.***": "ViewAction",

"target": "https://github.com/openmsupply/mobile/pull/4494#pullrequestreview-803386729",

"url": "https://github.com/openmsupply/mobile/pull/4494#pullrequestreview-803386729",

"name": "View Pull Request"

},

"description": "View this Pull Request on GitHub",

"publisher": {

***@***.***": "Organization",

"name": "GitHub",

"url": "https://github.com"

}

}

]

PR #4497, opened 2021-11-08
#4484 Add Name linked objects to merge

### LINKED ISSUE #4484: (Kiribati) Merged patient vaccinations not displaying in Patient Vaccination History
## Describe the bug

Kiribati have patients that have been entered into the system twice through human-error. 

Each of the patient records has had a vaccination event recorded against them. 

Kiribati now want to merge the patients records in mSupply Desktop to eliminate one of the patient records but retain both vaccination events. 

It appears that the vaccination event is correctly transferred from the eliminated patient record to the preserved patient record in mSupply Desktop, but the Patient Vaccination History in Mobile does not display the transferred vaccination event. 

image

image

image

(Seems that there's some timing discrepancy between the server and tablets here as date is off by 1 day) 

image

image

### To reproduce

1. Create a new patient record in Mobile
2. Administer a vaccination
3. Create another new patient record in Mobile
4. Administer another vaccination
5. Merge the patient records in Deskop
6. Observe that Desktop shows both vaccination events but Mobile only displays one. 

### Expected behaviour

Both the original vaccination event and the transferred vaccination event to be displayed in the Patient Vaccination History page. 
### Proposed Solution

### Version and device info

- App version: 8.3.0-rc1
- Desktop version: 5.03
- Tablet model: Lenovo TB-X505N
- OS version: Android 10

katherine-sussol:
Fixes #4484

## Change summary

Adds updates to tables linked to retained `Name` records when merged.
Doesn't fix Realm problem with deleting records that are currently being edited or viewed (going to log as a separate issue).

## Testing

- [ ] Merge two patients in Desktop (both of their home stores need to be the store you are logged into it seems), for Kiribati they will still show up in all mobile stores due to patients syncing everywhere. The vaccination records for both patient records should appear in mobile against the patient that was retained.

- Can't be working with a patient when it is merged or deleted via sync (although if you happen to be and see the crash, when relaunching the app after the initial crash it will be fine)

### DIFF
--- a/src/sync/incomingSyncUtils.js
+++ b/src/sync/incomingSyncUtils.js
@@ -1246,8 +1246,17 @@ const RECORD_TYPE_TO_TABLE = {
     },
   },
   Name: {
+    AdverseDrugReaction: {
+      field: 'name',
+    },
     ItemBatch: {
       field: 'supplier',
+    },
+    NameNote: {
+      field: 'name',
+    },
+    NameTagJoin: {
+      field: 'name',
     },
     Transaction: {
       field: 'otherParty',


joshxg (approved):
lgtm - cool these are so simple. I've regretted making this so dry as it's so confusing but seems it has worked ok for this case, at least.

I remember the originally implementation had a lot of time spent on exporting realm and converting to csv to do diffs, if that's at all helpfulto check, but I think best to get an RC done, the records to be merged are pretty visible in the app which is likely a lot easier

(this TODO: ` // TODO: log to bugsnag if merging not implemented for a certain |recordType|.` would have been good to have been implemented 🤣 )

katherine-sussol:
Thanks @joshxg! I will do an RC so Adam can do some testing too. (Will try out your suggestion for my own reassurance too 😄  - I visually checked Realm Explorer but I think a diff might be more conclusive and less error prone).

PR #4514, opened 2021-11-25
#4511 Patient search in Vaccination Module returning bad results

### LINKED ISSUE #4511: Patient search in Vaccination Module returning bad results
## Describe the bug

The patient search functionality in the Vaccination Module is not returning reliable results.

A search for a known patient first/last name combination is not returning the patient (despite the patient being present in the database). 

### To reproduce

1. Go to patient search in the Vaccination Module
2. Enter the last name for a known patient
3. Enter the first name for the known patient
4. See whether patient is returned in the results

Note this does not seem to happen for every patient - unsure exactly what determines whether a particular patient will be returned or not. Needs further digging!

Example video (_sensitive content_): https://drive.google.com/file/d/1fbwrKBi8GEYOUG1XIzRQ5lguDIi2tbFf/view?usp=sharing

### Expected behaviour

Search returns the patient searched for. 

### Proposed Solution

### Version and device info

- App version: 8.3.0-rc1
- Tablet model: Lenovo TB-X505N
- OS version: Android 10

### Additional context

This is on an a system with the `'sync all patients everywhere'` set-up

katherine-sussol:
Fixes #4511

Very weird bug 😖 

## Change summary

- Offroad: Some docstring updates (just came across while debugging)
- Handle `handleUpdate` (for some reason this function would often be holding onto an incorrect version of the '`completedForm`', especially when swapping between form fields). The `PatientSelect` component re-renders a few times every time it gets loaded or the form is modified... so suspect it was picking the value up before the form was updated properly? (and subsequently wiping out the old form data 🤔 ). Honestly, I didn't get to the very bottom of it but this seems to stop the bizarre search behaviour I was seeing from happening.

## Testing

- [ ] Follow steps in the original issue (my steps to reproduce in this comment): https://github.com/openmsupply/mobile/issues/4511#issuecomment-977295441) - search results should reflect what is typed into the form (and not something else behind the scenes)

- I've not actually been able to replicate the **exact** issue in Adam's original video (i.e using the first name field to search by last name): (https://drive.google.com/file/d/1fbwrKBi8GEYOUG1XIzRQ5lguDIi2tbFf/view?usp=sharing) but I'm convinced this is at least part of it
- Dispensing patient search works fine afaik

### DIFF
--- a/src/widgets/Tabs/PatientSelect.js
+++ b/src/widgets/Tabs/PatientSelect.js
@@ -128,12 +128,10 @@ EmptyComponent.propTypes = {
  *
  * @prop {Func}   createPatient         Callback for creating a patient.
  * @prop {object} formConfig            Configuration of the search form
- * @prop {Bool}   isAscending           Indicator if the list of patient is sorted ascending.
  * @prop {Func}   onCancelPrescription  Cancels the prescription and returns to the vaccine page
- * @prop {Func}   onFilterData          Callback for filtering patients.
- * @prop {Func}   patients              Current set of patient data.
  * @prop {Func}   selectPatient         Callback for selecting a patient.
- * @prop {String} sortKey               Current key the list of patients is sorted by.
+ * @prop {Func}   updateForm            Callback for updating form.
+ * @prop {object} completedForm         The completed form object.
  *
  */
 const PatientSelectComponent = ({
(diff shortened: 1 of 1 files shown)


joshxg (approved):
Ugh yeah that `FormControl` component has gotten very nasty! Very strange that memoising seems to be fixing some weird closure behaviour but.. sure 😁

PR #4517, opened 2021-11-28
#4516 Update name.name on patient edit (in vaccine module)

### LINKED ISSUE #4516: Kiribati: Updating patient first/last name in Vaccination Module does not update full name
## Describe the bug

This was reported by Teani in Kiribati. 

If a user edits a patient's first / last name in the Vaccination Module the full name is not updated in mSupply Desktop. 

i.e. an edit to `name.last` is not reflected in `name.name`

### To reproduce

1. Go to the Vaccination Module
2. Search for a patient and enter their record
3. Edit their last name and go through the vaccination process to administer a vaccine. 
4. Search for the patient under the new last name
5. Open their vaccination history
6. Note the name at the top of the page is displaying the old last name

### Proposed Solution

### Version and device info

- App version: 8.2.2
- Tablet model: Lenonvo Tab
- OS version: Android 10

katherine-sussol:
Fixes #4516

## Change summary

- When saving a patient after a vaccination - refresh `name.name` (as `firstName` and/or `lastName` may have been modified)

## Testing

- [ ] Try to replicate #4516 (hopefully you can't after this PR 🤞 )

Patients edited in the Dispensing module appear to be fine (name editing code is different depending on which module you are in...)

### DIFF
--- a/src/actions/Entities/NameActions.js
+++ b/src/actions/Entities/NameActions.js
@@ -100,7 +100,9 @@ const saveEditing = () => (dispatch, getState) => {
   const currentPatient = selectEditingName(getState());
   const createdDate = currentPatient?.createdDate ? new Date(currentPatient.createdDate) : null;
   const dateOfBirth = new Date(currentPatient.dateOfBirth);
-  const patientRecord = { ...currentPatient, dateOfBirth, createdDate };
+  const name = `${currentPatient.lastName}, ${currentPatient.firstName}`;
+
+  const patientRecord = { ...currentPatient, dateOfBirth, createdDate, name };
 
   UIDatabase.write(() => createRecord(UIDatabase, 'Patient', patientRecord));
   dispatch(reset());


joshxg (approved):
I trust that there is some other shenanigans going on which I can't see that is making this do what it needs to do but..

..does this work (this sounds really aggressive.. it's not meant to, but I can't think of another way to say it, so I thought I'd just talk about the aggressiveness afterward instead. How'd it work out?)?

`createRecord` which is called next does a default: https://github.com/openmsupply/mobile/blob/master/src/database/utilities/createRecord.js#L279

I think this was from before patients were added, where names couldn't be edited. How annoying. Would be much nicer to use use a `get name` 😢

katherine-sussol:
> I trust that there is some other shenanigans going on which I can't see that is making this do what it needs to do but..
> 
> ..does this work (this sounds really aggressive.. it's not meant to, but I can't think of another way to say it, so I thought I'd just talk about the aggressiveness afterward instead. How'd it work out?)?
> 
> `createRecord` which is called next does a default: https://github.com/openmsupply/mobile/blob/master/src/database/utilities/createRecord.js#L279
> 
> I think this was from before patients were added, where names couldn't be edited. How annoying. Would be much nicer to use use a `get name` 😢

Haha ... to be fair I didn't check the code in `createRecord` until you mentioned it (I did the update and just checked the history modal after editing a patient to confirm without really caring about the code 🤣  ..yes maybe a _bit_ lazy). 

But I think it is working in a bit of a messy way. `patientName` referenced above gets its value from `patientDetails.name`, where `patientDetails` refers to the object we're passing in as an arg: https://github.com/openmsupply/mobile/blob/a43354062a621a9d51a741bf52eba226c8e2a725/src/database/utilities/createRecord.js#L247-L272

Agree some kind of full name getter would be nicer. Tbh I think the whole existence of a `name` field on a patient that already has fields for `firstName` and `lastName` is a bit weird in the first place, but I guess it just reflects the tables in Desktop 🤔 .

joshxg:
Ahh! Yes of course!

PR #4540, opened 2021-12-20
#4539 update msupply ble service dependency to use npm

### LINKED ISSUE #4539: Update msupply-ble-service dependency to use npm

Currently the `msupply-ble-service` is internally published/consumed via [yalc](https://github.com/wclr/yalc).
Would like to consume via npm to align with other openmsupply public packages. 
Did an initial publish here: https://www.npmjs.com/package/@openmsupply/msupply-ble-service need to import/test/merge.

**A clear and concise description of what you want to happen.**

## Implementation

Keeping as is, means there is a copy of the msupply-ble-service code in `.yalc` (while other packages are in node_modules).

## Additional context

N/A

katherine-sussol:
Fixes #4539

## Change summary

- Removed yalc dependency and added npm dependency 

## Testing

- [ ] Try do Bluetooth stuff on mobile, it should still work as per before this change

- Should probably document package dependencies somewhere 🤔


joshxg (approved):
That's awesome, thanks a lot

PR #4553, opened 2022-01-10
#4549 missing consecutive breaches

### LINKED ISSUE #4549: Missing consecutive breaches when there is a current cumulative breach in progress
## Describe the bug

Consecutive breaches aren't being recorded if there is a pre-existing cumulative breach that hasn't yet been 'closed' e.g:
image
NB: Red box in screenshot is wrong -> hot breaches are to the right of 'average temperature' (accidentally highlighted the cold consecutive breach card)

The app is incorrectly recording a HOT_CUMULATIVE breach in the DB. The DB is supposed to only hold consecutive breaches.
image

### To reproduce

1. Set up a HOT_CUMULATIVE breach config that that will trigger soon-is (e.g. max 8 degrees, 2min duration, when it's 26 degrees where the sensor is)
2. Set up a breach config such that a HOT_CONSECUTIVE breach will be created, but AFTER the cumulative breach (e.g. max 8 degrees, 20min duration)
3. Wait for 20min
4. See error (we would expect a hot consecutive breach after 20min above the threshold....but there isn't 😭 )

### Expected behaviour
Cumulative and Consecutive breach calculations should be separate

### Proposed Solution
Think this line: https://github.com/openmsupply/mobile/blob/5c6e6b9cf98fb6ce1b44eaf79ab61d41e1e75c0c/src/actions/BreachActions.js#L35 should be modified to only look at consecutive breach configs

### Version and device info

- App version: 7.X.X
- Tablet model: N/A
- OS version: N/A

### Additional context
N/A

katherine-sussol:
Fixes #4549 (I think!?)

## Change summary

- Filters breach configs so that only consecutive configs are used when calculating consecutive breaches
- Have reordered operations so that the app will attempt to record consecutive breaches before trying to clear logs (I think I saw log clearing fail causing consec breach creation to be skipped, although now I can't replicate so I'm not 100% sure I saw what I thought I saw - think the change is safe enough anyway).

## Testing

- [ ] Attempt to replicate issue #4549

Have created this issue to try to clear logs more times before giving up: https://github.com/openmsupply/msupply-ble-service/issues/4

### DIFF
--- a/src/actions/Bluetooth/SensorDownloadActions.js
+++ b/src/actions/Bluetooth/SensorDownloadActions.js
@@ -137,9 +137,8 @@ const downloadLogsFromSensor = sensor => async dispatch => {
               false /* don't clear logs */
             );
 
-            await BleService().clearLogs(macAddress);
-
             await dispatch(BreachActions.createConsecutiveBreaches(sensor));
+            await BleService().clearLogs(macAddress);
           } catch (e) {
             dispatch(sensorDownloadError(sensor, DOWNLOADING_ERROR_CODES.E_CANT_CONNECT));
           }
--- a/src/actions/BreachActions.js
+++ b/src/actions/BreachActions.js
@@ -32,7 +32,8 @@ const createConsecutiveBreaches = sensor => async dispatch => {
       temperature,
       timestamp: moment(timestamp).unix(),
     }));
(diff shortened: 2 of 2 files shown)


joshxg (approved):
Ah! Nice. That makes sense

PR #4574, opened 2022-02-02
Add restarting scheduler on mount

### LINKED ISSUE #4568: Sensor downloads freezing intermittently
## Describe the bug

Support ticket: https://support.msupply.org.nz/cerb6/profiles/ticket/13814
This might not be a new issue as it's been observed previously - but has been reported again by another client that mSupply mobile is sometimes not downloading logs from sensors (once it gets into this state, the download never resumes until connecting to the sensor via the Blue Maestro app).

Blink commands do work - so unlikely to be purely a Bluetooth issue.
Might be related to similar cold chain issue here: https://github.com/openmsupply/msupply-cold-chain/issues/212

Not great as it's not consistently reproducible - needs investigation.

### To reproduce

1. Set up cold chain in mobile with sensors 
2. Leave it to download
3. Observe that (sometimes!?!) it doesn't and only restarts when the blue maestro app connects to the sensors.

### Expected behaviour

Downloads to happen

### Proposed Solution

TBD

### Version and device info

- App version: 8.4.1
- Tablet model: SM T505 (but not believed to be device specific)
- OS version: Android 11

### Additional context

N/A

joshxg:
Fixes #4568 ???

- The flow which causes *an* issue is that 'minimising' or sending to the app to the background 'regularly', doesn't trigger an unmount. But using the `back` button does (on the menu page), so the scheduler get's removed. When entering the foreground again, `onUpdate` isn't called, so the scheduler is not started again.

## Change summary

- Added a call to start the download scheduler on mount and left it also in the on update lifecycle method. This is because the passive download needs to potentially start after having synced from the server, also.

## Testing

- [ ] Pressing the back button on the menu page and returning to the app: Sensor temperatures are still being downloaded.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   },
   "name": "mSupplyMobile",
   "//": "version must be in the format ${majorNumber}.${minorNumber}.${patchNumber}-rc${releaseCandidateNumber}",
-  "version": "8.4.1",
+  "version": "8.4.1-rc1",
   "private": false,
   "license": "MIT",
   "description": "Mobile app for use with the mSupply medical inventory control software",
--- a/src/Store.js
+++ b/src/Store.js
@@ -27,10 +27,16 @@ const bugsnagMiddleware = () => next => action => {
   next(action);
 };
 
+const logMiddleware = () => next => action => {
+  console.log(action.type);
+
+  next(action);
+};
+
 const store = createStore(
   persistedReducer,
   {},
-  applyMiddleware(thunk, navigationMiddleware, bugsnagMiddleware)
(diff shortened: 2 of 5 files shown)


katherine-sussol (approved):
Hah - nice ❤️  so far so good - I guess if there was some other mysterious force (other than the back button) causing the main app component to unmount... this would potentially fix that too...? (I've been trying to find other ways to trigger it)

joshxg:
*Should* do 😁  

It's the only way I've been able to recreate also .. 🤷

katherine-sussol (approved):

PR #4579, opened 2022-02-07
#4567 allow all http connection

### LINKED ISSUE #4567: Unable to connect local webserver running in non SSL
## Describe the bug

From Android Pie version, Android have forced to used the https over the http, which failed to connect the local test web server for testing which runs in http.
See here for supporting docs:
https://android-developers.googleblog.com/2018/12/android-pie-la-mode-security-privacy.html
https://medium.com/mindorks/my-network-requests-are-not-working-in-android-pie-7c7a31e33330
https://developer.android.com/training/articles/security-config

### To reproduce

1. Setup a local webserver
2. Install v8 mobile app
3. Connect via credential
4. See error > It cannot be connected

### Expected behaviour

Most of the user may not face the problem as they runs in SSL, but for local testing, It will be hard for us if we need to test locally

### Proposed Solution

we could create a networkConfig.xml with this:
photo_2022-01-25_21-11-50

and in Android manifest we can link it up to use it

But there probably a better way to do this, leaving to Mobile developer

### Version and device info

- App version: V6 and higher
- Tablet model: SM-T515
- OS version: Android 11

sworup:
Fixes #4567

## Change summary

- This PR would add android device policy to allow not ssl connection which is by default disabled for android devices.
- User would be able to connect to http url apis after this change.

## Testing

- [ ] Run mSupply server on a computer locally with local ip exposed.
- [ ] Open mSupply mobile in a mobile device. Try connecting to the local ip address which obviously would not have secure https ssl connection. 
- [ ] The connection should work.

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -1,22 +1,21 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android" package="com.msupplymobile">
-
-  <uses-permission android:name="android.permission.INTERNET" />
-  <uses-permission android:name="android.permission.SYSTEM_ALERT_WINDOW"/>
-  <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
-  <uses-permission android:name="android.permission.BLUETOOTH" />
-  <uses-permission android:name="android.permission.BLUETOOTH_ADMIN" />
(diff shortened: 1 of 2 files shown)


katherine-sussol (approved):
Hi @sworup I've given this a test on Android 11 and it's working well. 
We've discussed this briefly in our patch release PR here: https://github.com/openmsupply/mobile/pull/4581 - we'll include this adjustment for 8.4.2 to avoid any future headaches with http from 8.4.2 onwards.

PR #4599, opened 2022-02-25
#4592 Fix Response requisition rouge CI problem

### LINKED ISSUE #4592: Response requisition in suggested state creates a rouge CI in mobile
## Describe the bug

- In normal flow of mSupply , when a Response requisition is created, it is created in suggested state and it does not create any linked CI yet.
- In mobile, when a mobile site creates a Response requisition,  it is created in suggested state and creates (forcefully) a linked customer invoice. 
- This CI does not show on the interface as mobile would hide ci with linked requisition but it is there. 
- These requisition and ci sync back to server and in main server we have a suggested response requisition with a linked ci. Which should never happen. 
- When we delete the suggested response requisition, this CI does not get deleted (as it was created against the flow of mSupply), and since the response requisition is deleted, the CI will finally show up in mobile. 
- This is a rouge CI issue.

### To reproduce

1. Open mobile app, and create a Customer requisition, you would have to give customer with program so do that to create a customer requisition. It would be in progress state, i.e. suggested state.
2. Look for a new confirmed Customer invoice in the Customer invoice tab of mobile, there won't be any.
3. In the mSupply server, when the Customer requisition has synced in notice that customer requisition is in suggested state. Go to Customer invoice tab, you will see a new CI has synced in which is the linked CI to the customer requisition that we had just created. 
    - The Ci might not have any trans_lines, which could be related to a different issue raised here https://github.com/openmsupply/mobile/issues/4562
4. Now delete the customer requisition (still in SG state) in desktop or mobile. 
5. Let the sync happen.
6. Now check for the customer requistion, which should have been deleted as expected.
7. Go to Customer invoice tab of mobile the linked customer invoice is not deleted, and now since its linked requistion is gone, it show shows up in the mobile Ci tab.

### Expected behaviour

- Since the Customer requisition is in SG state, it should not create a CI. 
- If it must create the CI, the customer requisition must be in confirmed state.
- The Customer requisition, if it has linked CI, should not be deletable until the CI is deleted.

### Proposed Solution

### Version and device info

- App version:
- Tablet model:
- OS version:

sworup:
Fixes #4592 

- The deletion sync was not syncing back to sever, resulting in sever still retaining the requisition that was deleted on mobile.

## Change summary

Suggested Customer Requisition deletion should delete its corresponding CI too.

## Testing

- [ ] Create a Customer Requisition on mobile
- [ ] After it syncs delete it
- [ ] Go to Customer invoice tab
- [ ] See a customer invoice corresponding (created by) the deleted Customer Requisition appear. This CI should have been deleted with Customer Requisition. 
- [ ] Check to see if the deleted requisition in mobile is also deleted on server.

## Regression tests

- [ ] Create, update and delete customer and supplier requisitions. They should all work as expected.

None

### DIFF
--- a/src/database/DataTypes/Requisition.js
+++ b/src/database/DataTypes/Requisition.js
@@ -249,7 +249,7 @@ export class Requisition extends Realm.Object {
    */
   createCustomerInvoice(database, user) {
     if (this.isRequest || this.isFinalised) {
-      throw new Error('Cannot create invoice from Finalised or Request Requistion ');
+      throw new Error('Cannot create invoice from Finalised or Request Requisition ');
     }
 
     if (
--- a/src/database/DataTypes/Requisition.js
+++ b/src/database/DataTypes/Requisition.js
@@ -52,6 +52,10 @@ export class Requisition extends Realm.Object {
    */
   destructor(database) {
     database.delete('RequisitionItem', this.items);
+
+    if (this.linkedTransaction) {
+      database.delete('Transaction', this.linkedTransaction);
+    }
   }
 
   /**
--- a/src/pages/CustomerRequisitionsPage.js
+++ b/src/pages/CustomerRequisitionsPage.js
@@ -187,7 +187,7 @@ export const CustomerRequisitions = ({
 };
 
 const mapDispatchToProps = dispatch => ({
-  ...getPageDispatchers(dispatch, 'Transaction', ROUTES.CUSTOMER_REQUISITIONS),
+  ...getPageDispatchers(dispatch, 'Requisition', ROUTES.CUSTOMER_REQUISITIONS),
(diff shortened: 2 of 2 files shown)


joshxg (approved):
nice!

sworup:
Just found another issue. When you delete Customer Requisition, it deletes from mobile but the delete never syncs back to the server. The requisition lines deletion syncs back but not the actual requisition. Investigating further.

joshxg (approved):

joshxg:
@katherine-sussol 

Sorry - are we good to just merge this into develop?

katherine-sussol:
> @katherine-sussol
> 
> Sorry - are we good to just merge this into develop?

Sounds good to me 👍

Issue #4486, opened 2021-11-02
Localise unlocalised strings in app


katherine-sussol:
## Describe the bug

As pointed out in https://github.com/openmsupply/mobile/pull/4481#issuecomment-956573016 some strings are not being translated. 

A bunch are due to not being localised and therefore not able to be easily translated - some examples:

**VaccinePage.js**
https://github.com/openmsupply/mobile/blob/a7af72900a656abc779c625948203d1164f9ff7b/src/pages/VaccinePage.js#L36

**BreachConfigRow.js**
https://github.com/openmsupply/mobile/blob/a7af72900a656abc779c625948203d1164f9ff7b/src/pages/NewSensor/BreachConfigRow.js#L12-L17

**Helpers.ts**
https://github.com/openmsupply/mobile/blob/a7af72900a656abc779c625948203d1164f9ff7b/src/authentication/Helpers.ts#L5-L12

Another bunch are because the listed json files are not included here:
https://github.com/openmsupply/mobile/blob/a7af72900a656abc779c625948203d1164f9ff7b/src/localization/utilities.js#L44-L56
-   demoUserModalStrings
-   reportStrings
-   vaccineStrings
-   validationStrings

If including these, the existing translations will need to be reviewed, as there are some which don't fit correctly:
![telegram-cloud-photo-size-1-4900228063632140783-y](https://user-images.githubusercontent.com/65875762/139759573-e8771c27-1a66-4c9c-9f32-e5956f31f2bf.jpg)

There are probably more in the app :)

### To reproduce

Steps to reproduce the behaviour:

1. Check the screens referenced in the issue comment above in a language other than English

### Expected behaviour

Strings should be easily able to be translated

### Proposed Solution

Update to localised strings

### Version and device info

- App version: 8.3.0
- Tablet model: N/A
- OS version: N/A

### Additional context

N/A

PR #4621, opened 2022-03-13
#4613 Translate remaining strings to Pt

### LINKED ISSUE #526: Phrase is not being translated: "Invalid username or password"
Build Number: 2.0.0

Description: If the user set a wrong username or password, while using other language, the app is not translating the phrase "Invalid username or password". There is no translation for that phrase.
image

Reproducible: Yes

Reproduction Steps: On login screen change the language to other than English (button 'Language' on the bottom left of the login screen). Than type an username with wrong password.

### LINKED ISSUE #4613: Portuguese translation
Sao Tome are going to start using mSupply mobile within about 2 months. We need the Portuguese version for them.

Make it speak Portuguese.

## Implementation

Hand it to Nicole who will work her tappity, tappity, tap magic.

Doing it myself. Cough. Not really.

## Additional context
None

nmadruga:
Fixes #4613 

## Change summary

Add remaining string in Pt for São Tomé

## Testing

- [ ] Check if strings in Pt fit the screen - consider **new functionality** since all main feature were already translated previously

Also fixes #526 with [mSupply/Desktop](https://github.com/sussol/msupply/pull/10468) PR. Although the solution is based on the language selected in mSupply Desktop! so it's not perfect, it just make the language selection most consistent if used both in mSupply desktop AND mobile.


joshxg (approved):
awesome!

nmadruga:
I wasn't been able to test until now, so have noticed a couple of things today:

Problem 1) Cutting of on **Customer Requisitions** button:
Screen Shot 2022-03-14 at 9 57 07 PM

Problem 2) Even though I haven't translated the term "Cold Chain" itself, it seems that most things aren't getting translated in this part of the App. Any clues?
Screen Shot 2022-03-14 at 10 00 42 PM
Screen Shot 2022-03-14 at 9 58 32 PM
Screen Shot 2022-03-14 at 9 58 18 PM
Screen Shot 2022-03-14 at 9 57 53 PM
Screen Shot 2022-03-14 at 9 57 37 PM 



@joshxg Do you knoe how I could change the UI? Maybe using something to wrap and fit the button to the size in case it's a big string like that? I think we also need some solution for Spanish strings. Could you help with a tip about the UI please?

joshxg:
Ah - 

Problem 1) Smaller string? 😁  Increase height to allow for two lines? 
Problem 2) Looks like vaccine strings needs to be added here: https://github.com/openmsupply/mobile/blob/737e0c9ca1baaf38dbfa74207ca356ba04c629a0/src/localization/utilities.js#L44

nmadruga:
> Problem 1) Smaller string? 😁 Increase height to allow for two lines?

Do you mean condition to the size of the string right? I'm not sure where to change the UI... If you could give a hand with that it would be very nice. Thanks

> Problem 2) Looks like vaccine strings needs to be added here

Will do that.

joshxg:
This component is the button on that screen. https://github.com/openmsupply/mobile/blob/737e0c9ca1baaf38dbfa74207ca356ba04c629a0/src/pages/MenuPage.js#L108-L111

I meant use less characters 😁

nmadruga:
> I meant use less characters 😁

😝 I wish, but Pt is this beautiful and prolific language.

> This component is the button on that screen.

Thank you. I think I could try changing this based on how many letter are present in the string. What do you think?
https://github.com/openmsupply/mobile/blob/737e0c9ca1baaf38dbfa74207ca356ba04c629a0/src/pages/MenuPage.js#L274

joshxg:
Good idea.

My first thought is because the second row of text within the button is sooo close to showing.. maybe some micro adjustments to the height/padding for *all* will be fine. 

I.e. if it is just a small increase and height and change in padding/margin which is not really noticeable for the other ones.. may as well do the easy solution.

The styling I think is here: https://github.com/openmsupply/mobile/blob/737e0c9ca1baaf38dbfa74207ca356ba04c629a0/src/globalStyles/buttonStyles.js#L65-L77 - there is quite a lot of padding. For example maybe padding: 10 and height: 65/70 might look OK and allow for the second line of text to show

katherine-sussol:
R.e The cold chain strings - some of them are still hardcoded in the app (in addition to needing to add vaccine strings to localisation). We found out when updating the French translations for a demo 😅 , but we didn't update at the time because Cold Chain wasn't a part of the demo.

I noted a few examples but there are possibly more:
https://github.com/openmsupply/mobile/issues/4486

nmadruga:
After a long run... Finally some good results on the Style!

See what do you think now slight change on fontSize in -1 when text is longer and fixed it being cut off:
Screen Shot 2022-03-15 at 1 04 13 PM
Screen Shot 2022-03-15 at 1 02 32 PM

nmadruga:
> R.e The cold chain strings - some of them are still hardcoded in the app (in addition to needing to add vaccine strings to localisation). We found out when updating the French translations for a demo 😅 , but we didn't update at the time because Cold Chain wasn't a part of the demo.
> 
> I noted a few examples but there are possibly more: #4486

Thanks for poping in Kat. I can have a look on that after lunch - if that's helpful. I think it should be good have everything on localized strings indeed. It helps with future work to be localising only. So, will do that on a different PR.

GaryWilletts (approved):
Parece maravilhoso! But I don't really feel qualified 😉

PR #4628, opened 2022-03-15
#4625 fix language effect on menu app

### LINKED ISSUE #4625: Changing language doesn't take effect
## Describe the bug

When you select another language and log in it doesn't  refresh the App to be using the new one - you would still see everything on the previous language selected. If I close and reopen the app and log in now then it shows the correct language selected. The only page that actually changes on the go is the login screen.

### To reproduce

1. Go to the Login page - click on language. Select one different to English by clicking twice
2. You will see the login page in the new language selected
3. Now login - The language on the main page will still be English
4. Close the app and open again - It will be showing in the new language
5. Log in - Now you can see the strings on the main menu as well

### Expected behaviour

Should change the strings on the main menu everytime you change the language on the login page. Shouldn't need to refresh the App.

### Proposed Solution

Not sure.

### Version and device info

- App version: v8.4.3
- Tablet model: Emulator 5 SDK .1
- OS version:

### Additional context

N/A

nmadruga:
Fixes #4625

## Change summary

new `currentLanguage` added to user state which gets set with the language store in settings. Also needed to add to the MenuPage mapPropsToState so it get triggered by a change on this state = reason why I added the flag showing on the bottom of the main Menupage. Hope you will like it : 9

## Testing

- [ ] Select a language on the login screen
- [ ] Login and see the language took effect (also new Flag on the bottom area)
- [ ] Try again by Logging out and selecting another language. Then login

The second time wasn't taking effect. Should be sorted now.

Sync modal, settings, other pages and Menu.

### DIFF
--- a/src/mSupplyMobileApp.js
+++ b/src/mSupplyMobileApp.js
@@ -153,8 +153,6 @@ class MSupplyMobileAppContainer extends React.Component {
   onAuthentication = user => {
     const { dispatch } = this.props;
     dispatch(UserActions.login(user));
-    // Required to refresh menuApp after language is updated (which is only on login page)
-    dispatch(UserActions.setLanguage(UIDatabase.getSetting(SETTINGS_KEYS.CURRENT_LANGUAGE)));
     this.postSyncProcessor.setUser(user);
   };
 
--- a/src/mSupplyMobileApp.js
+++ b/src/mSupplyMobileApp.js
@@ -153,6 +153,8 @@ class MSupplyMobileAppContainer extends React.Component {
   onAuthentication = user => {
     const { dispatch } = this.props;
     dispatch(UserActions.login(user));
+    // Required to refresh menuApp after language is updated (which is only on login page)
+    dispatch(UserActions.setLanguage(UIDatabase.getSetting(SETTINGS_KEYS.CURRENT_LANGUAGE)));
     this.postSyncProcessor.setUser(user);
   };
 
--- a/src/pages/MenuPage.js
+++ b/src/pages/MenuPage.js
@@ -14,9 +14,10 @@ import {
   ModulesImage,
   CustomerImage,
   IconButton,
+  InfoBadge,
+  Flag,
   SupplierImage,
   StockImage,
-  InfoBadge,
 } from '../widgets';
 
 import {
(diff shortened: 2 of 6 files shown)


src/mSupplyMobileApp.js
```diff
@@ -153,6 +153,8 @@ class MSupplyMobileAppContainer extends React.Component {
   onAuthentication = user => {
     const { dispatch } = this.props;
     dispatch(UserActions.login(user));
+    // Required to refresh menuApp after language is updated (which is only on login page)
+    dispatch(UserActions.setLanguage(UIDatabase.getSetting(SETTINGS_KEYS.CURRENT_LANGUAGE)));
```
joshxg, line 157: Nice, this works great.

I *think* it makes more sense for me to `connect` `LoginModal` component to get `dispatch`, and then `dispatch` the action in `onSelectLanguage`?

src/pages/MenuPage.js
```diff
@@ -224,14 +226,15 @@ const Menu = ({
...
+        <Flag countryCode={currentLanguage} />
         {!!isInAdminMode && <MenuButton text="Realm Explorer" onPress={toRealmExplorer} />}
         {!!isInAdminMode && <MenuButton text="Export Data" onPress={exportData} />}
         {!!isAdmin && (
           <IconButton Icon={<CogIcon />} label={buttonStrings.settings} onPress={toSettings} />
         )}
       </View>
     ),
-    [isInAdminMode, isAdmin]
+    [isInAdminMode, isAdmin, currentLanguage]
```
joshxg, line 237: This is awesome! super cute 😁

joshxg (approved):

src/mSupplyMobileApp.js
nmadruga: Ok - I'll have a look on doing this change now.

nmadruga: Hmmm seems like a slight big change because `LoginModal` is a class and not a react component, which I don't think works well with `connect`. I'll have to look into that with a little more time - TBD tomorrow.

nmadruga: Done! Finally with Chris' help. The class `LoginModal` was being exported at the top and as default, but the top one was the one in use. Sorted now, merging!

PR #4643, opened 2022-03-28
#4640 Showing invalid date

### LINKED ISSUE #4640: Showing invalid date
## Describe the bug

Mobile expects a single date format example "01/01/2021" if we send a date in any other format that gives the wrong output like "1/1/2021" which gives an invalid date.

image

### To reproduce

1. Go to mSupply and send any data with the date of format "1/1/2021"
2. Check-in tablet and it will show you an invalid date.

### Expected behaviour
It should show the correct date.

### Proposed Solution

export const convertMobileDateToISO = dateString => {
  if (!dateString) return undefined;

  const separators = ['/', '-'];
  const splitDate = dateString.split(new RegExp(separators.join('|'), 'g'));

  if (splitDate.length !== 3) return undefined;
  const dayOfMonth = splitDate[0].padStart(2, '0');
  const month = splitDate[1].padStart(2, '0');
  const year = splitDate[2].padStart(4, '0');

  return `${year}-${month}-${dayOfMonth}`;
};

### Version and device info

- App version:8.4.1
- Tablet model: In physical device Lenovo, and Emulator as well.
- OS version:

### Additional context

n/a

anildahalsussol:
By default take date format to 'MM/DD/YYYY'

Fixes #4640 

## Change summary

Change date format to 'MM/DD/YYYY'

## Testing

- [ ] Add any name notes in msupply and change date to 'm/d/yyyy'.
- [ ] Once it synced to mobile check vaccincation>history for that patient.
- [ ] Check Date it shouldn't show 'Invalid date'.

### DIFF
--- a/src/utilities/formatters.js
+++ b/src/utilities/formatters.js
@@ -74,6 +74,7 @@ export const formatDate = date => (date ? moment(date).fromNow() : generalString
  * @return  {string}              A string representing the date in 'yyyy-mm-dd' format
  */
 export const convertMobileDateToISO = dateString => {
+  dateString = moment(dateString).format('MM/DD/YYYY');
   const dayOfMonth = dateString.substring(0, 2);
   const month = dateString.substring(3, 5);
   const year = dateString.substring(6, 10);
--- a/src/utilities/formatters.js
+++ b/src/utilities/formatters.js
@@ -74,11 +74,13 @@ export const formatDate = date => (date ? moment(date).fromNow() : generalString
  * @return  {string}              A string representing the date in 'yyyy-mm-dd' format
  */
 export const convertMobileDateToISO = dateString => {
-  dateString = moment(dateString).format('MM/DD/YYYY');
-  const dayOfMonth = dateString.substring(0, 2);
-  const month = dateString.substring(3, 5);
-  const year = dateString.substring(6, 10);
-
+  if (!dateString) return undefined; // show undefined if no date
+  const separators = ['/', '-'];
+  const splitDate = dateString.split(new RegExp(separators.join('|'), 'g'));
+  if (splitDate.length !== 3) return undefined; // date needs to have month, date and year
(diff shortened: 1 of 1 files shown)


src/utilities/formatters.js
```diff
@@ -74,6 +74,7 @@ export const formatDate = date => (date ? moment(date).fromNow() : generalString
  * @return  {string}              A string representing the date in 'yyyy-mm-dd' format
  */
 export const convertMobileDateToISO = dateString => {
+  dateString = moment(dateString).format('MM/DD/YYYY');
```
katherine-sussol, line 77: https://github.com/openmsupply/mobile/issues/4640#issuecomment-1081250954

We could do something like this instead? Still brittle but just handles missing zeros better?
```
export const convertMobileDateToISO = dateString => {
  if (!dateString) return undefined;

  const separators = ['/', '-'];
  const splitDate = dateString.split(new RegExp(separators.join('|'), 'g'));

  if (splitDate.length !== 3) return undefined;
  const dayOfMonth = splitDate[0].padStart(2, '0');
  const month = splitDate[1].padStart(2, '0');
  const year = splitDate[2].padStart(4, '0');

  return `${year}-${month}-${dayOfMonth}`;
};
```

katherine-sussol: Experimented a bit more, another alternative we could use if we'd still prefer to use moment (but the input format in both of these examples still needs to be DD/MM/YYYY and not MM/DD/YYYY):
`dateString = moment(dateString, 'DD/MM/YYYY').format('DD/MM/YYYY');`

anildahalsussol: > 

sure I think we can do this... As we will just extract a day, month, and year whatever the format and whatever date it will be. I agree with your suggested solution.

katherine-sussol (approved):
👍

PR #4651, opened 2022-04-05
Added transact id

### LINKED ISSUE #4635: Link PCD name_note with VaccinationEvent name_note

Child of #4624 

When creating a vaccine prescription, the PCD `name_note` used at the time (on the patient edit step of vaccine dispensing) should be linked to the vaccination event `name_note`

## Implementation

Agreed solution: Save the PCD `name_note` ID into the VaccinationEvent name note when a vaccination is confirmed. This way we will have a record of which PCD was used at the time of vaccination without needing to create a new PCD note if the PCD is not updated.

Save the transact id into the PCD `name_note`.
But there is the possibility that the PCD was not updated at the time of vaccination (meaning we don't need to create a new one). Therefore we'd need to create a new PCD `name_note` for every vaccination, which is not ideal.

## Additional context

anildahalsussol:
Added transact id

Fixes #4635 

## Change summary

Added prescription id in name note of type "PCD".

## Testing

- [ ] Create a vaccinations.
- [ ] Edit the patient survey or create a new one and then complete all the steps.
- [ ] Click Ok & Next and check the name note for that name in msupply.
- [ ] Find patient type = "PCD" and check recent data and see if needs to have "transactID".

If there are any general areas of the codebase your changes might have side effects on, mention them here

### DIFF
--- a/src/actions/Entities/NameNoteActions.js
+++ b/src/actions/Entities/NameNoteActions.js
@@ -77,14 +77,11 @@ const updateForm = (data, validator) => ({
   payload: { data, isValid: validator(data) },
 });
 
-const saveEditing = prescriptionID => (dispatch, getState) => {
+const saveEditing = () => (dispatch, getState) => {
   const nameNote = selectCreatingNameNote(getState()) ?? {};
   const patient = UIDatabase.get('Name', nameNote?.nameID);
   const isDirty = JSON.stringify(patient?.mostRecentPCD?.data) !== JSON.stringify(nameNote?.data);
   if (isDirty) {
-    if (prescriptionID !== '') {
-      nameNote.data.transactID = prescriptionID;
-    }
     UIDatabase.write(() => createRecord(UIDatabase, 'NameNote', nameNote));
   }
   dispatch(reset());
--- a/src/actions/Entities/NameNoteActions.js
+++ b/src/actions/Entities/NameNoteActions.js
@@ -15,7 +15,6 @@ export const NAME_NOTE_ACTIONS = {
 
 const createDefaultNameNote = (nameID = '') => {
   const [pcd] = UIDatabase.objects('PCDEvents');
-
   return {
     id: generateUUID(),
     entryDate: new Date(),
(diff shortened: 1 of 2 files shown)


src/actions/Entities/NameNoteActions.js
```diff
@@ -80,15 +77,16 @@ const updateForm = (data, validator) => ({
   payload: { data, isValid: validator(data) },
 });
 
-const saveEditing = () => (dispatch, getState) => {
+const saveEditing = prescriptionID => (dispatch, getState) => {
   const nameNote = selectCreatingNameNote(getState()) ?? {};
   const patient = UIDatabase.get('Name', nameNote?.nameID);
   const isDirty = JSON.stringify(patient?.mostRecentPCD?.data) !== JSON.stringify(nameNote?.data);
-
   if (isDirty) {
```
katherine-sussol, line 84: Now that we are linking vaccination events with PCD records I'm wondering if we need to save every time (when the transaction id is updated) even if the actual PCD name note wasn't updated 🤔 .

Maybe the `isDirty` check should take into account the prescriptionID?

e.g.
If no prescriptionID: `nameNote.data.transactID = '';`
Otherwise: `nameNote.data.transactID = prescriptionID;`

Then `isDirty` check with the transactID populated:
`JSON.stringify(patient?.mostRecentPCD?.data) !== JSON.stringify(nameNote?.data);`

katherine-sussol (changes requested):
Thanks Anil! I had a quick test and it's saving nicely when dispensing a vaccine for the first time (and also if a vaccine is dispensed a second time to the same patient as long as the PCD Name Note is edited). 👍 

However I noticed a couple of things:
- The transactID is not being cleared when a patient's PCD name note is edited through the dispensary after a successful vaccine dispense.
- The transactID when dispensing a second vaccine without editing the PCD Name Note is not being saved anywhere.

I had an suggestion below but open to alternatives/other ones :)

katherine-sussol (approved):
LGTM!

PR #4654, opened 2022-04-07
#4644 Update to use selected language rather than previous language

### LINKED ISSUE #4644: Flag isn't corresponding to selected language
## Describe the bug

Issue introduced on v8.5.0-RC0 to be fixed, so the flag uses the same language as selected by the user on the login screen

### To reproduce

1. Go to login
2. Click on Language select some other than english (i.e.Spanish)
3. After loged in noticed the English flag is still showing
Screen Shot 2022-03-30 at 4 25 18 PM

4. If you go back to the login and change again, it will show the previous flag...

### Expected behaviour

Show the same flag as the language selected

### Proposed Solution

N/A

### Version and device info

- App version: v.8.5.0-RC0
- Tablet model:
- OS version:

katherine-sussol:
Fixes #4644

## Change summary

- Tiny adjustment to use the selected language rather than the previously saved one (thought I'd just fix while passing through for the Kiribati updates - and was also curious about the recent localisation updates)

## Testing

- [ ] Change language on login screen and then login, check the flag 
image

N/A

### DIFF
--- a/src/widgets/modals/LoginModal.js
+++ b/src/widgets/modals/LoginModal.js
@@ -126,7 +126,7 @@ class LoginModal extends React.Component {
 
   onSelectLanguage = ({ item }) => {
     const { settings, changeCurrentLanguage } = this.props;
-    changeCurrentLanguage(UIDatabase.getSetting(SETTINGS_KEYS.CURRENT_LANGUAGE));
+    changeCurrentLanguage(item.code);
     settings.set(SETTINGS_KEYS.CURRENT_LANGUAGE, item.code);
     setCurrencyLocalisation(item.code);
     setDateLocale(item.code);


src/widgets/modals/LoginModal.js
```diff
@@ -126,7 +126,7 @@ class LoginModal extends React.Component {
 
   onSelectLanguage = ({ item }) => {
     const { settings, changeCurrentLanguage } = this.props;
-    changeCurrentLanguage(UIDatabase.getSetting(SETTINGS_KEYS.CURRENT_LANGUAGE));
+    changeCurrentLanguage(item.code);
```
clemens-msupply, line 129: just interested, why is the db entry different to the selected language? why is it not stored in the db when changing the language?

clemens-msupply (approved):

src/widgets/modals/LoginModal.js
katherine-sussol: Oh, that's a good point, it's the next line that updates the DB with the selection (though abstracted through 'settings'):
`settings.set(SETTINGS_KEYS.CURRENT_LANGUAGE, item.code);`

So we could also just reverse the order so the DB gets updated first before trying to update the app state

mark-prins:
have to say, I really don't like the flag in the footer. way too prominent, and I don't like the British flag showing for 'English' 🤷

katherine-sussol:
> have to say, I really don't like the flag in the footer. way too prominent, and I don't like the British flag showing for 'English' 🤷

🙀  oh dear that flag has turned out to be quite controversial. I can see a new issue has been raised for it: https://github.com/openmsupply/mobile/issues/4680 so I will merge this first (as I think this also fixes the actual language state, not just the flag) and the flag can be dealt with over there

Issue #4638, opened 2022-03-25
Backfill transaction (including id) for vaccination name note and pcd name note 


katherine-sussol:
## Is your feature request related to a problem? Please describe.

Child of #4624

- Backfill:
  - `PCDNameNoteId` into `Vaccination` name_notes
  - Transaction data (including id) into `Vaccination` name_note where any don’t exist (vaccinations created since Mobile V8.3.0 already include this transaction data)

## Describe the solution you'd like

**A clear and concise description of what you want to happen.**

## Implementation

**A description of how the solution would best be implemented in the code base**

## Describe alternatives you've considered

**A clear and concise description of any alternative solutions or features you've considered.**

## Additional context

**Add any other context or screenshots about the feature request here.**

katherine-sussol:
Sample of the prescription data:
```
  "extra": {
    "prescription": {
      "serialNumber": "9",
      "enteredBy": "0763E2E3053D4C478E1E6B6B03FEC207",
      "id": "ecb172c0ae7d11ec9a08c1447abb413f",
      "confirmDate": 1648461640684,
      "comment": "",
      "customData": "{\"country\":\"Kiribati\",\"isExempted\":false,\"dateOfVaccination\":\"23/03/2022\",\"district\":\"South West\",\"island\":\"Onotoa\",\"administeringHealthFacility\":\"Tabuarorae\",\"anatomicalSite\":\"Right arm\",\"doseNumber\":\"Booster\",\"isConsented\":\"Yes\"}",
      "type": "customer_invoice",
      "entryDate": 1648461640684,
      "insuranceDiscountRate": null,
      "status": "finalised",
      "user1": null,
      "items": [
        {
          "id": "ecb25d20ae7d11ec9a08c1447abb413f",
          "item": {
            "id": "F326CB5105694DAD8A07AE1957CD512B",
            "code": "COV1",
            "description": "",
            "name": "COVID19 Vaaccine AstraZeneca",
            "defaultPackSize": 1,
            "defaultPrice": 0,
            "doses": 8,
            "isVaccine": true,
            "isVisible": true
          },
          "note": null
        }
      ],
      "theirRef": null,
      "mode": "dispensary",
      "subtotal": 0,
      "outstanding": null,
      "insuranceDiscountAmount": null,
      "isCancellation": false
    },
}
```

katherine-sussol:
[...]

PR #4668, opened 2022-04-22
#4623 View vaccination event details

### LINKED ISSUE #4623: Kiribati (1.7.a) View vaccination events in Mobile
Update mSupply Mobile to allow the user to view previously captured vaccination event data on a per-patient basis.

There's currently a vaccine history modal but it is quite sparse and can't be opened to view details.
Need to figure out a suitable UI to display detailed view with option to later edit/delete.

This ticket will add the ability to view data for vaccination events created from this point onwards. 

#4638  will cover the work needed to update existing older data to make it compatible with this new functionality. 

## Implementation

## Additional context

katherine-sussol:
Fixes #4623

## Change summary
- Updated history view to display doses and date from supplementary data form if available:
  -  image
- Adds the ability to view vaccination event records in a separate modal
  - Modal includes 
    - Vaccination event `name_note` record
    - Most recently associated PCD `name_note` record (linked either via id field in vaccination event, or by most recent PCD at the time of vaccinationEvent. 
    - image
    - If there was no PCD recorded before the vaccination event, an error will be displayed:
    image
    - Supplementary data `name_note` record details (pulled from the vaccinationEvent if it exists - entries prior to 8.3.0 will need to be backfilled: https://github.com/openmsupply/mobile/issues/4638)
- Renamed `pcdEventID` to `pcdNameNoteId`
- Removed hardcoded vaccinationEvent formSchema and retrieve from formSchemas synced from Desktop

## Testing

- Setup:
  - [ ] https://wiki.msupply.foundation/en:mobile:user_guide:vaccine_dispensing:setup
  - [ ] Dangerous patient mode needs to be enabled (patients sync everywhere and users can view all patient's history)
  - [ ] Need to add a form_schema to mSupply Desktop with type `VaccinationEvent` (schema to be added to https://github.com/sussol/client-JSON-schemas/tree/main/Kiribati/COVID-19-vaccination-rollout) 
- [ ] Dispense a vaccination to a patient
- [ ] View the patient history and select the vaccination record in the history modal

N/A - Probably more testing required with different scenarios


src/actions/Entities/NameNoteActions.js
```diff
@@ -23,36 +22,14 @@ const createDefaultNameNote = (nameID = '') => {
   };
 };
 
-const getMostRecentPCD = patient => {
```
katherine-sussol, line 26: Moved to a selector

src/actions/Entities/VaccinePrescriptionActions.js
```diff
@@ -219,9 +219,9 @@ const createVaccinationNameNote = (
   UIDatabase.write(() => UIDatabase.create('NameNote', newNameNote));
 };
 
-const getPcdEventID = (patientEventType, patientId) => {
+const getPcdNameNoteID = patientId => {
```
katherine-sussol, line 222: Renamed because the old name was a little misleading (there is also a PatientEvent with code 'PCD' with an ID but this is not what the function is returning)

src/selectors/Entities/name.js
```diff
@@ -79,28 +80,6 @@ export const selectCanEditPatient = state => {
   return UIDatabase.getPreference(PREFERENCE_KEYS.CAN_EDIT_PATIENTS_FROM_ANY_STORE) || isEditable;
 };
 
-const jsonSchema = {
```
katherine-sussol, line 82: Moved to retrieve from form schemas table in DB

src/widgets/JSONForm/JSONForm.js
```diff
@@ -177,6 +177,7 @@ export const JSONFormComponent = React.forwardRef(
       onSubmit,
       surveySchema,
       liveValidate,
+      disabled,
```
katherine-sussol, line 180: Since we're only adding view functionality here, had to add ability to disable the form (otherwise it's confusing as it looks editable but really isn't)

src/selectors/Entities/nameNote.js
```diff
@@ -11,3 +13,51 @@ export const selectNameNoteIsValid = state => {
...
+  const [mostRecentPCD] = sorted;
+  return mostRecentPCD;
+};
+
+export const selectMostRecentPCD = patient => {
+  const [pcdEvent] = UIDatabase.objects('PCDEvents');
+  if (!pcdEvent) return null;
+
+  const { id: pcdEventID } = pcdEvent;
+  const { nameNotes = [] } = patient ?? {};
```
clemens-msupply, line 48: haven't seen this before is this equivalent to: const { nameNotes } = patient ?? { nameNotes: [] };

katherine-sussol (commented):
Dropped in a few comments to try to shed a bit of light on some of the stranger looking changes

src/selectors/Entities/nameNote.js
```diff
@@ -11,3 +13,51 @@ export const selectNameNoteIsValid = state => {
   const { isValid = false } = nameNoteState ?? {};
   return isValid;
 };
+
+export const selectClosestPCDToVaccination = (patient, vaccinationEntryDate) => {
```
clemens-msupply, line 17: A short comment what closest, i.e. closest before vaccination date, means could be helpful.

clemens-msupply (approved):
Looks good (untested yet)

Just some understanding questions:

Unclear to me is why the associated PCD name_note record is the closest one to the vaccination? is it guaranteed that each vaccination has a name_note record that is earlier than the vaccination event? For example, is it possible a patient got a booster but the name_note is from the first shot? 

Also why display the closest and not the latest name_note?

src/selectors/Entities/nameNote.js
katherine-sussol: Thanks - agree that it was a little unclear before! d3c70660686444c16475bb41495d8eb7093a1a90

src/selectors/Entities/nameNote.js
katherine-sussol: I think that it behaves more like: 
`const nameNotes = patient?.nameNotes ?? [];` - i.e if patient is defined but doesn't have a `nameNotes` property, would still default to empty array.

Logic was originally copied from `selectMostRecentPCD`...but on re-reading I think this is probably less confusing than `const { nameNotes = [] } = patient ?? {}; ` ... not sure though, what do you think?

katherine-sussol: > Unclear to me is why the associated PCD name_note record is the closest one to the vaccination? is it guaranteed that each vaccination has a name_note record that is earlier than the vaccination event? For example, is it possible a patient got a booster but the name_note is from the first shot?

The UI flow to dispense a vaccination (if setup correctly) requires that a PCD name_note is filled before a vaccine can be dispensed. If, when dispensing a vaccine, there already exists a PCD `name_note` for the patient and it is not edited during the vaccine dispensing flow, no new name note will be created at the time - so the intention here is to just go back until we can find the latest... 

> Also why display the closest and not the latest name_note?
There are some things on the PCD that can change with time, e.g. pregnancy, currently breastfeeding. Kiribati would like to know what the patient's 'state' was at the time of the vaccination (e.g. they might have been pregnant at the time of vaccination, but then came in for a second dose later and were no longer pregnant)...

The overall design is ... a bit confusing. Adam has put together a slide deck to help explain the use case, might be helpful to get a picture of all the different name notes? https://docs.google.com/presentation/d/1rS0hbZTa7iAnBIGWtForcLuBiInQiGEHM33HNHsS9kI/edit#slide=id.g1197003ea89_0_564

clemens-msupply: > I think that it behaves more like: `const nameNotes = patient?.nameNotes ?? [];` - i.e if patient is defined but doesn't have a `nameNotes` property, would still default to empty array.
> 
> Logic was originally copied from `selectMostRecentPCD`...but on re-reading I think this is probably less confusing than `const { nameNotes = [] } = patient ?? {}; ` ... not sure though, what do you think?

thanks yes `const nameNotes = patient?.nameNotes ?? [];` seems much cleaner and straight forward to me. (but maybe just because I didn't know about the other syntax)

mark-prins:
@katherine-sussol there's an issue with the dispensing history - that also has the onclick enabled to display a vaccine.. but no clear indication that the row is clickable, so I think we should disable the click in that scenario. If you do happen to click you get:

Screen Shot 2022-04-27 at 10 12 46 AM

src/actions/Entities/VaccinePrescriptionActions.js
```diff
@@ -219,9 +219,9 @@ const createVaccinationNameNote = (
...
   const getSelectedNameNote = UIDatabase.objects('NameNote')
```
mark-prins, line 223: I'm unsure why this is called `getSelectedNameNote` when it isn't a method. Would you mind renaming to `selectedNameNotes` or something?

src/localization/dispensingStrings.json
```diff
@@ -1,6 +1,7 @@
 {
   "gb": {
     "vaccination_history": "Vaccination history for",
+    "vaccination_details": "Vaccination Event Details",
```
mark-prins, line 4: :nit the other strings are only uppercase for the initial word. I wonder if it should be `Vaccination event details` 🤔

src/selectors/Entities/nameNote.js
```diff
@@ -11,3 +13,53 @@ export const selectNameNoteIsValid = state => {
...
+  const sorted = filtered.sort(
+    ({ entryDate: entryDateA }, { entryDate: entryDateB }) =>
+      moment(entryDateB).valueOf() - moment(entryDateA).valueOf()
+  );
+
+  const [mostRecentPCD] = sorted;
+  return mostRecentPCD;
+};
+
+export const selectMostRecentPCD = patient => {
```
mark-prins, line 45: this and the method above are doing almost the same thing; what about creating a `selectMostRecentNameNote` which accepted a filter parameter? then both methods could use the new one

src/widgets/JSONForm/JSONForm.js
mark-prins: ah. I noticed that the Checkboxes component still allows editing - I think it will need to implement `disabled` soz

mark-prins (commented):
🎉  Wooohoo! looks awesome 😁  Nice work Kat
Wondering if you can disable the vaccine modal from the standard dispensing / patient history? Other than that, just a few ignorable comments

src/localization/dispensingStrings.json
katherine-sussol: I was conflicted on this 😆 - lots of page titles (and some other modals, e.g. 'Stocktake Details', 'Requisition Details') have all uppercased first letters, but you're also right that the other strings in this area aren't... I suppose better to keep consistency within the section 😅

src/actions/Entities/VaccinePrescriptionActions.js
katherine-sussol: Huh, yeah that's a bit confusing! b6323d5c0

src/selectors/Entities/nameNote.js
katherine-sussol: Ooh, yes they did end up being almost identical. Thanks :) https://github.com/openmsupply/mobile/pull/4668/commits/2b825ed120fd78bbc2e88375da5553f9a359335e

src/widgets/JSONForm/JSONForm.js
katherine-sussol: Oops - I didn't actually try tapping the text to enable/disable! Fixed: https://github.com/openmsupply/mobile/pull/4668/commits/7ad5a940af2926e03d27932f3c53cf0959b98a54

katherine-sussol:
Have disabled the rows if in dispensing modal... looks a bit grey, but still readable 🤔 https://github.com/openmsupply/mobile/pull/4668/commits/274d4eb4ba3e6e45baa2b921ed6185c1ec89b54e 
image.

PR #4681, opened 2022-04-27
Remove flag

### LINKED ISSUE #4680: Alter language flag icon

There has been some debate about the language flag icon and the general consensus is that something needs to be changed as it's a bit distracting / unsure of appropriateness of using country flags outside of the country the flag represents. 

Current implementation of flag icon: 

image

One of the 2 options mentioned here (the choice is yours!): 

Example of language selector (see Mobile log-in page for current use of language selector): 

image

Example of small flag icon: 

## Implementation

## Additional context

Note that the 2 options offered are not really equivalent as one just moves the displayed icon, and the other is asking for additional functionality (the ability to switch languages once past the log-in screen). 

So.... over to you 😁

mark-prins:
Fixes #4680

## Change summary

Removed the flag

## Testing

- [ ] No flag on menu page
- [ ] When changing language, this should be reflected on the menu page without the need to restart the app

### DIFF
--- a/src/pages/MenuPage.js
+++ b/src/pages/MenuPage.js
@@ -15,7 +15,6 @@ import {
   CustomerImage,
   IconButton,
   InfoBadge,
-  Flag,
   SupplierImage,
   StockImage,
 } from '../widgets';
@@ -90,7 +89,6 @@ const Menu = ({
   usingCashRegister,
   usingModules,
   usingVaccines,
-  currentLanguage,
   isAdmin,
   hasVaccines,
   toVaccineDispensingPage,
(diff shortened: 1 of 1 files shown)


katherine-sussol (approved):
LGTM! 

(I voted for it to be smaller and out of the way but agree this is much simpler, I think users will realise pretty quickly if they don't understand the text on the screen that they picked the wrong one ... 🤔 )

mark-prins:
yeah, I voted for smaller and out of the way too! but thought when it came to it that (a) Dhanya is always right and (b) it should be fairly obvious when you change language and (c) the screen has enough clutter already and (d) who wants to wade into these troubled waters?!

But mostly (a) 😉

PR #4686, opened 2022-04-29
#4684 Need to handle undefined UserJson

### LINKED ISSUE #4684: Need to handle undefined UserJson
## Describe the bug

Now even if UserJson is undefined it shows an "Unexpected error from the server" which seems to be unclear while testing locally I found that it never hit the server but later found server API was wrong as my IP address changed. 

### To reproduce

1. Go to user login.
2. Be able to log in.
3. And change your IP address.
4. See an error

### Expected behaviour

At least if it can show me a clear error so that we can fix the problem. Like User is Undefined. And I saw that it can be due to the wrong URL where UserJson is undefined.

### Proposed Solution

For now, I am thinking of at least showing the error "User Json is undefined and also the server URL". So that we can see if that's the reason behind getting "Unexpected error from the server".

### Version and device info

- App version: 8.4
- Tablet model: Nexus emulator
- OS version:

anildahalsussol:
Check if userJson is undefined or not.

Fixes #4684 

## Change summary

Added new error if userjson is undefined rather than "Unexpected response from server"
## Testing

- [ ] Sync tablet from msupply.
- [ ] Change your IP.
- [ ] Logout and login in tablet.
- [ ] Check new error with expected server ip.

### DIFF
--- a/src/authentication/UserAuthenticator.js
+++ b/src/authentication/UserAuthenticator.js
@@ -71,6 +71,11 @@ export class UserAuthenticator {
       if (userJson && userJson.error) {
         throw new Error(userJson.error);
       }
+
+      if (userJson === undefined) {
+        throw new Error(`userJson is undefined.Server Url ${serverURL}`);
+      }
+
       if (!userJson || !userJson.UserID) {
         throw new Error('Unexpected response from server');
       }


src/authentication/UserAuthenticator.js
```diff
@@ -71,6 +71,11 @@ export class UserAuthenticator {
       if (userJson && userJson.error) {
         throw new Error(userJson.error);
       }
+
+      if (userJson === undefined) {
```
katherine-sussol, line 75:
```suggestion
      if (!userJson) {
        throw new Error(`Error . Server Url ${serverURL}`);
      }
      
      if (!userJson.UserID) {
        throw new Error('Unexpected response from server');
      }
```

katherine-sussol, line 75: Sorry, I clicked too early while typing a response 😅 .
I was thinking rather than checking specifically for undefined maybe it would be better to just split out the !userJson and userJson.UserID check?

```suggestion
      if (!userJson) {
        throw new Error(`userJson not found. Server Url ${serverURL}`);
      }
      
      if (!userJson.UserID) {
        throw new Error('Unexpected response from server');
      }
```

Although I'm surprised it doesn't just log a connection failure (I haven't reproduced the specific changing IP scenario). Maybe logging the response status code might also help with troubleshooting somewhere around here?
https://github.com/openmsupply/mobile/blob/19dc0094a4ca231fee8a54f5031f538b82d85a91/src/authentication/Helpers.ts#L61
```
logger.info(`Response status code: ${response?.status}`);
```

anildahalsussol: @katherine-sussol : I think we need to handle both cases separately as the wrong URL gives undefined but if we have the correct URL and no name exists in mSupply then it might give us blank userJSON.

katherine-sussol (approved):
Can only help I think! From what you've described the current behaviour sounds particularly misleading if it's not hitting the server and you're seeing the message 'Unexpected response from the server' (although I have not been able to reproduce this myself...and I've given up trying - I seem always get 'Unable to connect').

PR #4713, opened 2022-05-17
#4691 edit pcd via vaccination event modal

### LINKED ISSUE #4691: Add the ability to edit a historical PCD via the Vaccination Event modal
Child of #4624

Kiribati would like the ability to edit PCD forms associated with a vaccination event.
Currently, when dispensing or vaccinating a patient, the user is only presented with the most recent PCD form to be updated.

This is needed because a patient's situation could have been different from the time of vaccination to the current time (e.g. maybe they were pregnant when they were vaccinated - and this is useful information for reporting purposes).

- Enable the RHS of this form:
image
- Add a 'Save' button somewhere (bottom of the modal?)
- When 'Save' is pressed, if the PCD form has any updates... then update the stored name_note? (or should we create a new one and also update the vaccination name_note 🤔 )

## Implementation

TBD

- Not really any alternatives other than just not doing it... Or using foot runner every time someone requests an update.

## Additional context
N/A

katherine-sussol:
Fixes #4691

https://user-images.githubusercontent.com/65875762/168718354-494d5a41-d6c9-4405-b5a8-1f583d2c0489.mp4

## Change summary

- Adds the ability to edit the PCD linked to a vaccination event via the vaccination event modal.

## Testing

- [ ] Follow steps in this PR to view vaccination event details: #4668 
- [ ] When a detail view is opened, the RHS panel should be editable and allow the linked PCD to be edited

- This just overwrites the historical PCD, it doesn't create a new one, due to the complexity of linking multiple dependent name_notes. This has a few implications:
  - If the same PCD was used for two vaccinations (a new PCD is only created if there were updates to the form during vaccination) - when the PCD is edited, both of the vaccination records will be linked to the updated data.
  - No continuous audit trail within mobile (although speaking with some 4D experts it seems like the changes can (if needed, and with difficulty) be traced back through the journal. 

@adamdewey tagging FYI!

Been thinking a bit on this: if we wanted to keep an audit trail (maybe) we could do something like the following (although its maybe a little bit bloated/convoluted):
- Add soft deletion flag to name_notes (also needs desktop change)
- When _anything_ in the vaccination event modal is edited, soft delete the old vaccination event name note and create a new one, changes only to be applied to the copy
- If the PCD form was edited, create a copy of the PCD note, save the updates into the copy, and link the vaccination event name note to the new PCD copy 
- Don't soft delete the PCD name note (there may be other vaccination events pointing to it)

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   },
   "name": "mSupplyMobile",
   "//": "version must be in the format ${majorNumber}.${minorNumber}.${patchNumber}-rc${releaseCandidateNumber}",
-  "version": "8.5.0-rc1",
+  "version": "8.4.2",
   "private": false,
   "license": "MIT",
   "description": "Mobile app for use with the mSupply medical inventory control software",
--- a/src/actions/Entities/NameNoteActions.js
+++ b/src/actions/Entities/NameNoteActions.js
@@ -1,5 +1,6 @@
 import { generateUUID } from 'react-native-database';
 import merge from 'lodash.merge';
+import { ToastAndroid } from 'react-native';
 import { createRecord, UIDatabase } from '../../database/index';
 import {
   selectCreatingNameNote,
@@ -7,6 +8,7 @@ import {
 } from '../../selectors/Entities/nameNote';
 import { selectSurveySchemas } from '../../selectors/formSchema';
 import { validateJsonSchemaData } from '../../utilities/ajvValidator';
+import { vaccineStrings } from '../../localization';
 
 export const NAME_NOTE_ACTIONS = {
   SELECT: 'NAME_NOTE/select',
@@ -79,6 +81,7 @@ const updateLinkedSurveyNameNote = (originalNote, updatedData) => () => {
   };
 
   UIDatabase.write(() => UIDatabase.update('NameNote', updatedNote));
+  ToastAndroid.show(vaccineStrings.vaccination_updated, ToastAndroid.LONG);
 };
 
 const createNotes = (nameNotes = []) => {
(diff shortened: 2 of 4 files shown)


mark-prins:
I'm concerned about the underlying requirement - modifying a historical record at a later date and then presenting that record as an accurate representation of what the data was at that historical time. Given the requests for corrections and updates, I can see why it is needed, but I really think we need to have an audit trail of these changes.

The mSupply `[log]` table hasn't been brought across to mobile unfortunately -  could add it, and sync but that's a big change.
There is the LoggerService, though that's more operational, errors and info, than data related.

What about creating a new `PatientEvent` type such as 'PCDModified' and augmenting the data stored to include the old, new and dates? at the point of updating you have the old and new data, you could save both as a single entry in the `[data]` field:

```
data: {
  old: {
    entryDate: ...
    data: { ... }
  },
  new: {
    entryDate: ...
    data: { ... }
  },
}
```

just requires the addition of a patient event code into mSupply - the sync and storage location is all handled

mark-prins (approved):
👍  changes look good 😁 

still uneasy about not auditing the change - it's likely to not be an issue for most clients, would still be good to have just in case

katherine-sussol:
> What about creating a new `PatientEvent` type such as 'PCDModified' and augmenting the data stored to include the old, new and dates? at the point of updating you have the old and new data, you could save both as a single entry in the `[data]` field:
> 
> ```
> {
>   name,
>   entryDate: new Date(),
>   data: {
>     old: {
>       entryDate: ...
>       data: { ... }
>     },
>     new: {
>       entryDate: ...
>       data: { ... }
>     },
>   }
> }
> ```
> 
> just requires the addition of a patient event code into mSupply - the sync and storage location is all handled

Thanks - yeah I think that's a tidier solution than `name_note` juggling 😱. Kiribati has also been wanting to modify the supplemental data (and also items/vaccinator info via this modal too... 😭 ) so going down this route, we'd probably also need an event to audit 'VaccinationEventModified' (or combine it into a generic 'NameNoteModified' event...?)

PR #4726, opened 2022-05-24
#4724 Try a wider variety of formats when parsing vaccination dates

### LINKED ISSUE #4724: Vaccination history date format inconsistency (Kiribati)
## Describe the bug

- The real 'bug' is that the backing data is a bit of a mess. From @adamdewey - these are the current vaccination date formats that have been inserted (either via import script, the UI, or footrunner):
  - DD-MM-YYYY
  - DD/MM/YYYY
  - YYYY-MM-DD
  - YYYY/MM/DD
  - MM/DD/YYYY
  - M/D/YYYY
  - YYYY/DD/MM
  - YYYY/D/M
  - D/M/YYYY
  - D-M-YYYY
- Mobile will only currently parse the following correctly:
  - DD-MM-YYYY
  - DD/MM/YYYY
  - D/M/YYYY
  - D-M-YYYY

### To reproduce

1. (This will actually take you many many hours so it might be easier to just ask Adam or Kat to see if you are curious) - Point a tablet at the kiribati vax training server
2. Download 8.5.0-rc1 of Mobile and initialise a store on it 
3. There are some users that have 'Invalid Date' displayed in their vaccination history modal (ping Adam or Kat for the name if you want some examples - won't list here for privacy reasons)
image

### Expected behaviour

The vaccination date to show up :(

### Proposed Solution

- Can give a pass as DD/MM/YYYY (or DD-MM-YYYY) first, and if that fails, try the ISO format (YYYY/MM/DD or yyyy-mm-dd) <- these seem to be the most common.
- If it still fails, someone needs to edit and make the correction at the source, we can't cater for all possible permutations 🙀  (e.g. 04/03/21 - which is the day, which is the month? Do we guess that 21 is the year and it wasn't entered in 2004?)

### Version and device info

- App version: 8.5.0-rc1
- Tablet model: N/A
- OS version: N/A (but I tested on Android 10)

### Additional context
N/A

katherine-sussol:
Fixes #4724

## Change summary

- Best effort attempt to read dodgily entered vaccination dates based on what Kiribati have in their database (🙀 )
- In cases where it's impossible to determine (e.g. 04/03/2022 could be read either as 4th of March, or 3rd of April), we use these defaults:
  - D/M/YYYY (if the year is at the end - we assume they are using the standard mobile format)
  - YYYY-MM-DD (if the year is at the beginning - we assume they were using the ISO format)

Some users may get strange results - but there's realistically no real way to completely fix this properly without a manual review of a lot of the dates.

## Testing

- [ ] Have included some unit tests to cover basic cases from the Kiribati datafile (can run with source code using `yarn test`)
- [ ] Otherwise test against the Kiribati data file or manually edit vaccination dates to be like the ones reported in the issue and see if you can view them in mobile.

I'm really sorry to anyone who has to read this, and if there is a better solution, all 👂 !

### DIFF
--- a/src/selectors/Entities/name.js
+++ b/src/selectors/Entities/name.js
@@ -88,7 +88,7 @@ export const selectVaccinePatientHistory = patient => {
   const [supplementalDataSchema = {}] = selectSupplementalDataSchemas();
 
   const hasDateOfVaccinationInSchema =
-    supplementalDataSchema?.jsonSchema?.properties?.dateOfVaccination;
+    !!supplementalDataSchema?.jsonSchema?.properties?.dateOfVaccination;
 
   const nameNotes = patient?.nameNotes
     ?.filter(
--- a/src/selectors/Entities/name.js
+++ b/src/selectors/Entities/name.js
@@ -80,15 +80,25 @@ export const selectCanEditPatient = state => {
   return UIDatabase.getPreference(PREFERENCE_KEYS.CAN_EDIT_PATIENTS_FROM_ANY_STORE) || isEditable;
 };
 
+const getDateOfVaccination = (supplementalDataSchema, vaccinationData) => {
+  const hasDateOfVaccinationInSchema = !!supplementalDataSchema?.jsonSchema?.properties
+    ?.dateOfVaccination;
+
+  if (!hasDateOfVaccinationInSchema) return null;
+  if (!vaccinationData.extra?.prescription?.customData) return 'N/A';
+
+  const { dateOfVaccination } = JSON.parse(vaccinationData.extra?.prescription?.customData);
+  if (!dateOfVaccination) return 'N/A';
+
+  return new Date(convertVaccinationEntryToISOString(dateOfVaccination));
+};
+
 export const selectVaccinePatientHistory = patient => {
   const [vaccinationPatientEvent] = UIDatabase.objects('PatientEvent').filtered(
     "code == 'vaccination'"
   );
   const { id: vaccinationPatientEventID } = vaccinationPatientEvent ?? {};
   const [supplementalDataSchema = {}] = selectSupplementalDataSchemas();
-
-  const hasDateOfVaccinationInSchema =
-    !!supplementalDataSchema?.jsonSchema?.properties?.dateOfVaccination;
 
   const nameNotes = patient?.nameNotes
     ?.filter(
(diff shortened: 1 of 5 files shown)


adamdewey:
Nice Kat! 

Yeah some background: Some of Kiribati data was imported from a spreadsheet and some was entered via an earlier version of Mobile before we had settled on a consistent way of storing the dates, hence the muckiness. 

Could either be fixed via a datascript on the server or here in Mobile, both employing the same heuristics. But as we already have an 18 hour script to run on the server as part of the upgrade release process next week we decided it would be better to put it into Mobile for better or for worse...

katherine-sussol:
Two other incoming changes needed:
- Do a schema check before trying to use `dateOfVaccination` (have just realised that Tonga do not have this in their schema) - if it doesn't exist, default `dateOfVaccination` to the `name_note` entry date
- Sort the history by `dateOfVaccination`

adamdewey:
Good catch Kat!

src/selectors/Entities/name.js
```diff
@@ -85,6 +85,10 @@ export const selectVaccinePatientHistory = patient => {
     "code == 'vaccination'"
   );
   const { id: vaccinationPatientEventID } = vaccinationPatientEvent ?? {};
+  const [supplementalDataSchema = {}] = selectSupplementalDataSchemas();
+
+  const hasDateOfVaccinationInSchema =
+    supplementalDataSchema?.jsonSchema?.properties?.dateOfVaccination;
```
mark-prins, line 91: safest to convert to a boolean, since that's the expected var type
```suggestion
  const hasDateOfVaccinationInSchema =
    !!supplementalDataSchema?.jsonSchema?.properties?.dateOfVaccination;
```

src/selectors/Entities/name.js
```diff
@@ -101,21 +105,22 @@ export const selectVaccinePatientHistory = patient => {
...
-      confirmDate:
-        (vaccinationData.extra?.prescription?.customData &&
-          JSON.parse(vaccinationData.extra?.prescription?.customData).dateOfVaccination &&
-          new Date(
-            convertMobileDateToISO(
-              JSON.parse(vaccinationData.extra?.prescription?.customData).dateOfVaccination
-            )
-          )) ??
-        'N/A',
+      confirmDate: hasDateOfVaccinationInSchema
```
mark-prins, line 108: this is getting hard to follow!
I'd create

```
const getDateOfVaccination = (supplementalDataSchema, vaccinationData) => {
  const noDateOfVaccinationInSchema =
    !supplementalDataSchema?.jsonSchema?.properties?.dateOfVaccination;

  if (noDateOfVaccinationInSchema) return null;
  if (!vaccinationData.extra?.prescription?.customData) return 'N/A';

  const { dateOfVaccination } = JSON.parse(vaccinationData.extra?.prescription?.customData);
  if (!dateOfVaccination) return 'N/A';

  return new Date(convertVaccinationEntryToISOString(dateOfVaccination));
};
```

and then 

```
      confirmDate:
        getDateOfVaccination(supplementalDataSchema, vaccinationData) ?? new Date(entryDate),
```

src/utilities/parsers/vaccinationDateParser.test.js
```diff
@@ -0,0 +1,91 @@
+import { convertVaccinationEntryToISOString } from './vaccinationDateParser';
```
mark-prins, line 1: 😁  loving the test suite!!

mark-prins (approved):
👍  looks great! thanks Kat

src/selectors/Entities/name.js
katherine-sussol: Haha - soz yes I re-read it and realised how confusing that was. Will adjust!

PR #4758, opened 2022-06-15
#4714 Add ability to edit supplemental data form

### LINKED ISSUE #4714: Add the ability to edit a historical supplemental data via the Vaccination Event modal

Kiribati would like the ability to edit supplemental data forms associated with a vaccination event.

I.e. this portion of the data in the vaccination event modal:

image

- When the form is updated and the user clicks 'save' just update the vaccination event name note? (it was originally saved into [transact]custom_data but we would like to move away from that and towards just reading out of the vaccination name note

## Implementation

TBD

Say no editing to Kiribati

## Additional context

-

katherine-sussol:
Fixes #4714

## Change summary

- Butchered the UI in an attempt to allow easier data editing:
  - image
- Updated modal to grab the actual name note - it was previously working with data mapped for the history view 
- Added 'isDirty' check to not update if the contents are the same

## Testing

Prerequisites:
- Update VaccinationEvent JSON schema:
  - [VaccinationEvent.txt](https://github.com/openmsupply/mobile/files/8905265/VaccinationEvent.txt)
  - [VaccinationEvent-UI.txt](https://github.com/openmsupply/mobile/files/8905267/VaccinationEvent-UI.txt)
- [ ] View the vaccination history for a patient and edit the supplemental data form

- Too many buttons? 🤔 Bit awkward because the data is stored in 3 different places though conceptually to users it's all supposed to be part of one event (Maybe a tabbed view would be better?)
- Maybe 'editing' should be disallowed until the user tries to press an 'edit' button of some sort?

### DIFF
--- a/src/actions/Entities/NameNoteActions.js
+++ b/src/actions/Entities/NameNoteActions.js
@@ -69,17 +69,33 @@ const saveEditing = () => (dispatch, getState) => {
   dispatch(reset());
 };
 
-const updateLinkedSurveyNameNote = (originalNote, updatedData) => () => {
-  const { id, patientEvent, name, entryDate } = originalNote;
+const updateNameNote = (originalNote, updatedData) => () => {
+  const { id, patientEvent, name, entryDate, _data } = originalNote;
 
-  const updatedNote = {
-    id,
-    patientEvent,
-    name,
-    entryDate: new Date(entryDate),
-    _data: JSON.stringify(updatedData),
-  };
+  // Quick & dirty check if the object was updated, trims out some un-needed updates
+  const isDirty = _data !== JSON.stringify(updatedData);
 
+  if (isDirty) {
+    const updatedNote = {
+      id,
+      patientEvent,
+      name,
+      entryDate: new Date(entryDate),
+      _data: JSON.stringify(updatedData),
+    };
+
+    UIDatabase.write(() => {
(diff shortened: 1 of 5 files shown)


mark-prins (approved):
👍 impressive!

I do find the three panels overwhelming - there is a lot of information on the page and it is hard to know where to look! I think that tabs would be much easier to follow here, though tricky to manage the buttons and actions.

I think that this is pretty niche functionality, so you have a lot of leeway here - only determined or knowledgable people will find their way to this screen. So the layout isn't critical.

One other oddity is why the last button is `Edit` and the first two are `Save Changes`. The edit button doesn't allow you do edit - it remains disabled, so there's no harm in making the names consistent.

Does the last panel get enabled by anything? If not - does it need to have a save button?

katherine-sussol:
Thanks @mark-prins for the review! I've been struggling a bit with what to do with that 3rd panel as it's not going to be editable if the user is on a different tablet to the one the vaccine was dispensed on (as it's transactional data that has been copied to a name note - which is not available on all tablets), and if it _is_ editable, will also be needing something a little more restrictive than the text fields that are currently there since they directly map to mSupply vaccinators and items
(hence have just kept it disabled it for now to deal with in a separate issue/PR 😓).

I do agree the mis-matched naming is odd though! Have updated!

PR #4761, opened 2022-06-15
#4760 Check for sg customer requisition in mSupply and show model to delete or try again

### LINKED ISSUE #4760: Pre existing suggested customer requisitions cannot be deleted from mobile after desktop to mobile conversion
## Describe the bug

Related to https://github.com/openmsupply/mobile/issues/4387

- When the single-user desktop is converted to a mobile store, the pending suggested customer requisitions are also converted to mobile. 
- The desktop requisition has isRemoteOrder flagged to false, while mobile has it flagged to true by default, during conversion this flag is never changed so mobile code locks the requisition and disallows edit. There are some other codes in mobile that uses isRemoteOrder to determine if the requisition was created in mobile or not. 
- In desktop, where the isRemoteOrder flag was originally created, the flag is used by the supplier hub to determine which columns and fields to unlock in form. So in desktop isRemoteOrder means created by some other means and brought in through supplier hub.
- So basically mobile and desktop have two different and sometimes conflicting use for this flag. I guess this happened by mistake where the mobile system designer misunderstood the isRemoteOrder flag's use on the desktop. It is quite late now to try and fix this, and there would a lot of overhead if we attempt to as this flagged is synced.
- Also the consequences are not too dire although they are visible and annoying and we should definitely omit them they do not affect data flow too much.

### To reproduce

1. Go to a single user database, pick a desktop store with some sg customer requisitions.
2. convert it mobile store
3. Load the site and store it to our mobile app
4. See the mobile store's customer requisition page, the sg requisitions show up on mobile but are locked.
5. On the desktop database store's requisition page, the store requisitions cannot be edited or deleted, its sync locked.
6. So affectively these requisitions cannot be deleted or edited indefinitely by the user. 

### Expected behaviour

- Either these requisitions should work as normal requisitions on mobile, be editable and delectable. Or these should be deleted before desktop to mobile conversion.
- Since these are suggested requisitions, deleting and expecting users to recreatee them from mobile should be fine.

### Proposed Solution

- We tackle this during desktop to mobile store conversion. 
- We check for sg customer requisitions; if there are any, we alert the user by saying - `There are pending customer requisitions in the store that are still in the suggested state. Please confirm or finalise them or delete them before doing the conversion. If you continue with the conversion, mSupply will auto-delete the suggested customer requestions. Do you want to continue? Yes/No`
- If the user says no, do not do the conversion, if he says yes, delete the sg customer requisition and do the conversion.
- @ujwalSussol and I discussed and this seems like the most pain-free solution. Others would involve extensive testing and potential resync of legacy mobile data. @Gwalior-84 also initially suggested this approach.

### Version and device info

- App version: All
- Tablet model: All
- OS version: All

### Additional context

None

### Agreed solution

- We trigger this when someone tries to sync a site from mobile, at the very initial sync user login screen.
  - When user enter the sync username and password and submits, after some steps on sever end, `sync_create_records_for_store` is triggered when the sync records are created. Ideally we want do our work before this method does its work. 
- Before `sync_create_records_for_store` on server. 
  - On server, we add a method that checks if the sync request is coming from mobile.
  - If it is coming from mobile we look in store data requistion tables, and see if has `sg` customer requisitions pending.
  - If yes, we do not sync the data, and we send back a respnse to the mobile, that specific response will show the pop up to mobile user saying `There are pending customer requisitions in the store that are still in the suggested state. Please confirm or finalise them or delete them before doing the conversion. If you continue with the conversion, mSupply will auto-delete the suggested customer requestions. Do you want to continue? Yes/No` 
  - If user clicks no, we stop here, nothing syncs. We are still at sync user login page on mobile.
  - If user clicks yes, we send another request to desktop... upon receiving this request, desktop will delete the `sg` customer requistions and then move on to normal sync record creation flow.

sworup:
Fixes #4760

## Change summary

## Testing

- [ ] Do something


src/localization/modalStrings.json
```diff
@@ -73,7 +73,10 @@
...
     "permissions_intro_2": "If disabled, you can press an item to enable it.",
-    "version_incompatible": "Unable to import! This is version of {0} of mSupply mobile and the selected import file has a newer version of {1}"
+    "version_incompatible": "Unable to import! This is version of {0} of mSupply mobile and the selected import file has a newer version of {1}",
+    "pending_sg_customer_requisition": "The mobile store you are trying to connect to has {0} suggested customer requisitions. We cannot sync the store while these suggested customer requisitions exist.\n\nWe suggest you to log into mSupply desktop and resolve (confirm, finalise or delete) these customer requisitions. Please click the \"Cancel\" button and try again after the issue is resolved.\n\nIf you think these suggested requisitions are obsolete, please click the \"Delete\" button below and we will delete all of the pending suggested customer requisitions automatically for you. However remember that these suggested customer requisitions would not be recoverable once deleted.\n\nDo you want to continue?",
```
katherine-sussol, line 77: Minor grammar suggestion: `We suggest you to log into mSupply desktop and resolve`
to
`We recommend logging into mSupply desktop to resolve`

katherine-sussol (commented):
Hi @sworup! Can I confirm that this is the correct associated desktop PR to test with? 
https://github.com/sussol/msupply/pull/10915 

The first half of the modal looks good to me, although I also agree with this [comment](https://github.com/sussol/msupply/issues/10490#issuecomment-1159816967) against the linked Desktop issue

arjunSussol:
After discussion with @ujwalSussol we decided to close it and will reopen if required.

PR #4762, opened 2022-06-16
#4759 Allow length of > 3 for email TLD when requesting demo store

### LINKED ISSUE #4759: .foundation email addresses aren't allowed in the demo store creation dialog
## Describe the bug

As above.

### To reproduce

I refuse to try again.

### Expected behaviour

This is an outrage.

### Proposed Solution

Make .foundation addresses illegal

katherine-sussol:
Fixes #4759

## Change summary

- Allow email TLD to have more than 3 chars

## Testing

- [ ] Test emails with some of these domains out: https://data.iana.org/TLD/tlds-alpha-by-domain.txt

N/A

### DIFF
--- a/src/authentication/DemoSiteRequest.js
+++ b/src/authentication/DemoSiteRequest.js
@@ -52,7 +52,7 @@ export class DemoSiteRequest {
   // TODO: Could be extracted to a helper file to be used elsewhere.
   // eslint-disable-next-line class-methods-use-this
   validateEmail(text) {
-    const reg = /^\w+([.-]?\w+)*@\w+([.-]?\w+)*(\.\w+)+$/;
+    const reg = /^[^\s@]+@[^\s@]+\.[^\s@]{2,}$/;
     return reg.test(text);
   }
 
--- a/src/authentication/DemoSiteRequest.js
+++ b/src/authentication/DemoSiteRequest.js
@@ -52,7 +52,7 @@ export class DemoSiteRequest {
   // TODO: Could be extracted to a helper file to be used elsewhere.
   // eslint-disable-next-line class-methods-use-this
   validateEmail(text) {
-    const reg = /^\w+([.-]?\w+)*@\w+([.-]?\w+)*(\.\w{2,3})+$/;
+    const reg = /^\w+([.-]?\w+)*@\w+([.-]?\w+)*(\.\w+)+$/;
     return reg.test(text);
   }
 


src/authentication/DemoSiteRequest.js
```diff
@@ -52,7 +52,7 @@ export class DemoSiteRequest {
   // TODO: Could be extracted to a helper file to be used elsewhere.
   // eslint-disable-next-line class-methods-use-this
   validateEmail(text) {
-    const reg = /^\w+([.-]?\w+)*@\w+([.-]?\w+)*(\.\w{2,3})+$/;
+    const reg = /^\w+([.-]?\w+)*@\w+([.-]?\w+)*(\.\w+)+$/;
```
mark-prins, line 55: If I can offer an alternative.. `/^[\w.-]+@(?:[\w-]+\.)+[\w-]{2,}$/`

I think the restriction of at least two chars for the final segment is good to retain. I've also tweaked the original because I don't understand the reasoning for it, and I don't see why we have a capturing match for a simple test.

Interestingly, the existing regex gives a 'catastrophic backtracking' error when testing on regex101.com 😁

mark-prins (approved):
👍  thanks! that should do the trick. I've added a suggestion too

src/authentication/DemoSiteRequest.js
katherine-sussol: I agree the 2 char minimum is actually quite a sensible check 😄 - I was too rash in removing that.
Been playing around with your suggested one and that seems to work just as well as the old one and is much easier to digest - thanks!

I guess while analysing the old one I had also wondered about it being overly restrictive (but didn't really want to stir the pot too much) i.e. this address is valid:
`user+mailbox@example.com` as per https://datatracker.ietf.org/doc/html/rfc3696#page-5 yet fails all versions of our validation 😆

Would I be being too rash/dramatic in suggesting we be even less restrictive and just check there is some stuff before an `@` character, and some `.` separated stuff after? Them not receiving an email should be a pretty good hint the email was invalid... 🤔? 
E.g. maybe:
`^\S+@\S+\.\S+$ `

mark-prins: I wondered the same, after looking at some of the regex suggestions out there which usually include all sorts of characters! and I agree - it is better to be lax than it is to be too restrictive - it isn't life or death having a strictly valid email here..

Even so.. I wonder about `^[^\s@]+@[^\s@]+\.[^\s@]{2,}$` to prevent having too many `@` in the email? The `\S` will allow `@`

katherine-sussol: Thanks! I totally forgot that '@' was a non-whitespace character 🤦‍♀️

PR #4794, opened 2022-07-12
#4792 Kiribati: Prevent giving vaccine to deceased patient

### LINKED ISSUE #4783: Wrong gender in remote search patients while vaccination process
## Describe the bug

If the patient's home store is different from the current mobile store then during the vaccination process fields "Is deceased" and "Gender" will show wrong values

### To reproduce

1. Go to 'Vaccinations"
2. Click on 'Search online'
3. Select any patient, see that "Is deceased" and "Gender" is wrong4. 

### Expected behaviour

Must show correct values

### Proposed Solution

### Version and device info

- App version:
- Tablet model:
- OS version:

### LINKED ISSUE #4792: Kiribati: Prevent giving vaccine to deceased patient

Do not proceed if a patient is deceased, alert.

image

**A clear and concise description of what you want to happen.**

## Implementation

## Additional context

[trello issue](https://trello.com/c/Tkz880xw/69-13b-prevent-giving-vaccine-to-deceased-patient)

chastichno:
Fixes #4792 (and small bug fixed #4783)
My first PR here YAY

## Change summary

Added an alert to the patient select and patient edit tabs to prevent vaccine dispensing if a patient is deceased.
Also, fixed a bug in the toggle styling so now "Gender" and "Is deceased" are shown correctly. Did it here as currently it is very broken and would be quite confusing for the clients.

## Testing

1.
- [ ] Go "Vaccination" -> Search for patients
- [ ] Select a patient who you know is deceased -> see an alert.
- [ ] Clicking "OK" closes an alert, you stay on the same page.

2. 
- [ ] Go "Vaccination" -> Search for patients
- [ ] Select a patient who you know is editable on the current site. Edit patient window opens.
- [ ] Change "Is deceased" to "Yes", click "Next" -> see an alert.
- [ ] Clicking "OK" closes an alert, you stay on the same page.

3. 
- [ ] Go "Vaccination" -> Search for patients
- [ ] Click "New patient". Edit patient window opens.
- [ ] Change "Is deceased" to "Yes", click "Next" -> see an alert.
- [ ] Clicking "OK" closes an alert, you stay on the same page.

4.
- [ ] Go to "Vaccination" or "Dispensing". Select any patient to edit.
- [ ] Toggles for "Gender" and "Is Deceased" must be displayed correctly in both of the cases:
  - [ ] If patient is editable on the current site
  - [ ] If patient is not editable on the current site

- I'm not sure why we are able to change patient status to "Deceased" during the vaccination -> probably we should disable it in the future

### DIFF
--- a/src/widgets/CircleButton.js
+++ b/src/widgets/CircleButton.js
@@ -18,9 +18,9 @@ const SIZE_VALUES = {
 };
 
 const ICON_SIZE_VALUES = {
-  small: 10,
-  medium: 15,
-  large: 20,
+  small: 8,
+  medium: 13,
+  large: 17,
 };
 
 /**
--- a/src/widgets/FormInputs/FormToggle.js
+++ b/src/widgets/FormInputs/FormToggle.js
@@ -54,6 +54,7 @@ export const FormToggle = ({
         toggleOnStyle={localStyles.toggleOnStyle}
         toggleOffStyle={localStyles.toggleOffStyle}
         toggleOnDisabledStyle={localStyles.toggleOnDisabledStyle}
+        toggleOffDisabledStyle={localStyles.toggleOffDisabledStyle}
       />
     </FlexView>
   );
@@ -83,6 +84,14 @@ const localStyles = StyleSheet.create({
     alignItems: 'center',
     justifyContent: 'center',
     backgroundColor: WARMER_GREY,
+    borderRadius: 20,
+    margin: 5,
+  },
+  toggleOffDisabledStyle: {
+    flex: 1,
+    alignItems: 'center',
+    justifyContent: 'center',
+    borderColor: WARMER_GREY,
     borderRadius: 20,
     margin: 5,
   },
--- a/src/widgets/Tabs/PatientEdit.js
+++ b/src/widgets/Tabs/PatientEdit.js
@@ -69,16 +69,16 @@ const PatientEditComponent = ({
   const savePatient = useCallback(
     e => {
       updatePatientDetails(completedForm);
-      if (!completedForm.isDeceased) {
-        formRef?.current?.submit(e);
-
-        if (canSaveForm) {
-          onCompleted();
-        } else {
-          ToastAndroid.show(dispensingStrings.validation_failed, ToastAndroid.LONG);
-        }
-      } else {
+      if (completedForm.isDeceased) {
         toggleIsDeceasedAlert();
+        return;
+      }
+
+      formRef?.current?.submit(e);
+      if (canSaveForm) {
+        onCompleted();
+      } else {
+        ToastAndroid.show(dispensingStrings.validation_failed, ToastAndroid.LONG);
       }
     },
     [completedForm, canSaveForm]
--- a/src/widgets/Tabs/PatientEdit.js
+++ b/src/widgets/Tabs/PatientEdit.js
@@ -14,6 +14,8 @@ import { PageButton } from '../PageButton';
 import { FlexRow } from '../FlexRow';
 import { FlexView } from '../FlexView';
 import { PageButtonWithOnePress } from '../PageButtonWithOnePress';
+import { PaperModalContainer } from '../PaperModal/PaperModalContainer';
+import { PaperConfirmModal } from '../PaperModal/PaperConfirmModal';
 
 import { selectCanEditPatient, selectEditingName } from '../../selectors/Entities/name';
 import { selectSurveySchemas } from '../../selectors/formSchema';
@@ -22,8 +24,15 @@ import { WizardActions } from '../../actions/WizardActions';
 import { VaccinePrescriptionActions } from '../../actions/Entities/VaccinePrescriptionActions';
 import { selectCanSaveForm, selectCompletedForm } from '../../selectors/form';
 import { getFormInputConfig } from '../../utilities/formInputConfigs';
+import { useToggle } from '../../hooks/useToggle';
 
-import { buttonStrings, vaccineStrings, dispensingStrings } from '../../localization';
+import {
+  buttonStrings,
+  vaccineStrings,
+  dispensingStrings,
+  modalStrings,
+  generalStrings,
+} from '../../localization';
 import globalStyles from '../../globalStyles';
 import { JSONForm } from '../JSONForm/JSONForm';
 import { NameNoteActions } from '../../actions/Entities/NameNoteActions';
@@ -55,17 +64,21 @@ const PatientEditComponent = ({
   canEditPatient,
 }) => {
   const { pageTopViewContainer } = globalStyles;
+  const [isDeceasedModalOpen, toggleIsDeceasedAlert] = useToggle(false);
   const formRef = useRef(null);
-
   const savePatient = useCallback(
     e => {
       updatePatientDetails(completedForm);
-      formRef?.current?.submit(e);
+      if (!completedForm.isDeceased) {
+        formRef?.current?.submit(e);
 
-      if (canSaveForm) {
-        onCompleted();
+        if (canSaveForm) {
+          onCompleted();
+        } else {
+          ToastAndroid.show(dispensingStrings.validation_failed, ToastAndroid.LONG);
+        }
       } else {
-        ToastAndroid.show(dispensingStrings.validation_failed, ToastAndroid.LONG);
+        toggleIsDeceasedAlert();
       }
     },
     [completedForm, canSaveForm]
@@ -118,6 +131,13 @@ const PatientEditComponent = ({
           style={{ marginLeft: 'auto' }}
         />
       </FlexRow>
+      <PaperModalContainer isVisible={isDeceasedModalOpen} onClose={toggleIsDeceasedAlert}>
+        <PaperConfirmModal
+          questionText={modalStrings.deceased_patient_vaccination}
+          confirmText={generalStrings.ok}
+          onConfirm={toggleIsDeceasedAlert}
+        />
+      </PaperModalContainer>
     </FlexView>
   );
 };
--- a/src/widgets/Tabs/PatientSelect.js
+++ b/src/widgets/Tabs/PatientSelect.js
@@ -371,11 +371,9 @@ const mapDispatchToProps = dispatch => {
       Keyboard.dismiss();
       const selectedPatient = await dispatch(NameActions.select(patient));
 
-      if (selectedPatient) {
-        if (!selectedPatient.isDeceased) {
-          dispatch(NameNoteActions.createSurveyNameNote(selectedPatient));
-          dispatch(WizardActions.nextTab());
-        }
+      if (selectedPatient && !selectedPatient.isDeceased) {
+        dispatch(NameNoteActions.createSurveyNameNote(selectedPatient));
+        dispatch(WizardActions.nextTab());
       }
     });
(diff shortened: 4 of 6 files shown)


src/widgets/Tabs/PatientEdit.js
```diff
@@ -55,17 +64,21 @@ const PatientEditComponent = ({
...
-
   const savePatient = useCallback(
     e => {
       updatePatientDetails(completedForm);
-      formRef?.current?.submit(e);
-
-      if (canSaveForm) {
-        onCompleted();
+      if (!completedForm.isDeceased) {
+        formRef?.current?.submit(e);
```
sworup, line 73: Right now if I set a patient to deceased and press next, the isDeseased checks runs before canSaveForm validation, so I think editing user would not be abled to set someone to deceased and save it from mobile.

We have to validate, allow save and check isDeseased afterwards I think.

sworup (changes requested):
Code looks pretty good, but during testing the code in ui I have a slight concern about how it work now.

src/widgets/Tabs/PatientSelect.js
```diff
@@ -253,12 +256,16 @@ const PatientSelectComponent = ({
           rowKey={keyExtractor(item)}
           columns={columns}
           onPress={name => {
-            // Only show a spinner when the name doesn't exist in the database, as we need to
-            // send a request to the server to add a name store join.
-            if (UIDatabase.get('Name', name?.id)) {
-              selectPatient(name);
+            if (!name?.isDeceased) {
+              // Only show a spinner when the name doesn't exist in the database, as we need to
```
sworup, line 260: @chastichno Similar concern to previous thing I raised, but I just noticed that when we select a patient we are showing the popup and preventing further page navigation here. I think we should allow view of the patient detail, we should only stop navigation when the app user tries to dispense to deceased user.

sworup:
@chastichno Also are we stopping edit at vaccination event details page? The one where you can edit details of previous vax events. I am working on edit but you can do the blocking fix in this pr so that we won't forget.

src/widgets/Tabs/PatientEdit.js
chastichno: The "Vaccination" process is used specifically for dispensing the vaccines. If user wants to edit any details of the patients they can do that in "Dispensing" -> clicking "Edit" button for the selected patient. I think it is a reasonable limitation not to be bale to set "Deceased" status during the vaccination process.

The behaviour in this PR is consistent with what was before -> if you don't finish the vaccination process and exit it by clicking "Cancel" then all the changes made to the patient details will not be saved.
To reproduce:
- Go to "Vaccination" -> Select an existing patient editable on the current site (let's say "Patient 1")
- Edit some details and click "Next"
- Exit vaccination process by clicking "Exit"
- Go to "Dispensing" -> select "Edit" Patient 1.
- See that details you were editing in vaccination window are not saved because you exited the process with the "Exit" button. 

In my opinion, it is the correct behaviour as clicking "Cancel" always discards any changes made.

chastichno:
>@chastichno Also are we stopping edit at vaccination event details page? The one where you can edit details of previous vax events. I am working on edit bug fixes but you can do the blocking fix in this pr so that we won't forget.

Honestly, I am not sure if we should restrict admins from editing vaccinations for deceased patients. I believe if they are doing it for improving the data quality, they will still need to edit vaccinations even for the deceased patients. @adamdewey What do you think?

src/widgets/Tabs/PatientEdit.js
sworup: Also when we try to create new Patient, deceased that is.

sworup:
> Honestly, I am not sure if we should restrict admins from editing vaccinations for deceased patients. I believe if they are doing it for improving the data quality, they will still need to edit vaccinations even for the deceased patients. @adamdewey What do you think?

@chastichno that is a good valid point. But with that argument a new dispensement should also be allowed as admin might be backdating the vaccination date there too.

Maybe we should show an alert and allow dispensing in both new and edit cases. Or if we want to keep it in edit at least we should show an alert saying you are about to edit vax event records of a deceased patient, allow admin to choose to go ahead and make the edit.

src/widgets/Tabs/PatientEdit.js
sworup: @chastichno you are right about the form not saving the changes in step 2 if admin presses cancel at point. That is a very strange flow as there is no user feedback so I kinna want an alert here too during cancel to tell user that the data they have entered so far would be discarded, but then I do not want too much alerts either.

Since we cannot edit deceased status in this form anyway, I kinda wanna disabled this option and make it viewable only.

But considering the flow you just described the code works fine as it is. Maybe we can merge it for now and make updates later if there is user feedback and demands.

src/widgets/CircleButton.js
```diff
@@ -18,9 +18,9 @@ const SIZE_VALUES = {
 };
 
 const ICON_SIZE_VALUES = {
-  small: 10,
-  medium: 15,
-  large: 20,
+  small: 8,
```
mark-prins, line 21: thanks - they look much better. the date picker icon in particular was too big

src/widgets/FormInputs/FormToggle.js
```diff
@@ -54,6 +54,7 @@ export const FormToggle = ({
         toggleOnStyle={localStyles.toggleOnStyle}
         toggleOffStyle={localStyles.toggleOffStyle}
         toggleOnDisabledStyle={localStyles.toggleOnDisabledStyle}
+        toggleOffDisabledStyle={localStyles.toggleOffDisabledStyle}
```
mark-prins, line 57: that looks better too!

src/widgets/Tabs/PatientEdit.js
mark-prins: all sounds reasonable to me. The easiest solution would be to make another issue which disables editing of a patient deceased status during the vaccination dispensing process. Like you say Nastia, that's a strange workflow anyway

src/widgets/Tabs/PatientEdit.js
```diff
@@ -55,17 +64,21 @@ const PatientEditComponent = ({
...
   const formRef = useRef(null);
-
   const savePatient = useCallback(
     e => {
       updatePatientDetails(completedForm);
-      formRef?.current?.submit(e);
-
-      if (canSaveForm) {
-        onCompleted();
+      if (!completedForm.isDeceased) {
```
mark-prins, line 72: :nit there are certainly exceptions to this.. but generally I think it reads better having positive assertions in conditionals.. in this case `if(completedForm.isDeceased) {`

:nit I prefer early returns over nested statements. in this case it could be
```
if (completedForm.isDeceased) {
  toggleIsDeceasedAlert();
  return;
}

formRef?.current?.submit(e);
if (canSaveForm) {
  onCompleted();
} else {
  ToastAndroid.show(dispensingStrings.validation_failed, ToastAndroid.LONG);
}
```
( could return early instead of the else.. but 🤷  )

:nit the `e` is unused, so we can remove it 😉

src/widgets/Tabs/PatientSelect.js
mark-prins: I think that it's ok to stop at this point, since the purpose of this workflow was to dispense a vaccine. It has been hacked around a fair bit ( at customer's request ) to enable editing and re-editing.. but the purpose here is to find a patient and dispense to them. 

One reason to show the details and allow editing would be to allow the user to bring the patient back from the dead.. which may be valid if they were changed accidentally. Even so.. I'd recommend that we push people to use the normal patient edit.

It would help though, to have an extra column so that you don't click on a dead person. Even better.. why don't we just not show dead people in the search results?

src/widgets/Tabs/PatientSelect.js
```diff
@@ -253,12 +256,16 @@ const PatientSelectComponent = ({
...
+              // send a request to the server to add a name store join.
+              if (UIDatabase.get('Name', name?.id)) {
```
mark-prins, line 262: :nit same concern as before about nesting and early returns

src/widgets/Tabs/PatientSelect.js
```diff
@@ -357,8 +371,10 @@ const mapDispatchToProps = dispatch => {
       const selectedPatient = await dispatch(NameActions.select(patient));
 
       if (selectedPatient) {
```
mark-prins, line 373:
```suggestion
      if (selectedPatient) {
```
```suggestion
      if (selectedPatient && !selectedPatient.isDeceased) {
```

mark-prins (commented):
Very nice! 😁 

Liking the extra design touches too - they look good. Am now thinking that the deceased modal needs some work.. but that's for another day.

Am wondering though.. perhaps just not return deceased people in the vaccine dispensing search? that would be easier as a user: would limit the number of patients returned in a search and prevent annoying alerts blocking you

adamdewey:
A lot of discussion in here - apologies for not sticking my neck in sooner!

My main points of clarification (hopefully): 

1. Using the Dispensing module as the method to mark a patient as Deceased would be fine. 
2. Once the patient is marked as deceased then they should still appear in the Vaccination search results so that their vaccination history can be viewed/edited (pressing on the clock icon). 
3. Pressing on the patient name (i.e. anywhere on the line except the clock icon) should just display a pop-up saying `Can not administer vaccination - this patient is deceased.` Dismissing the pop-up takes the user back to the list of search results.

src/widgets/Tabs/PatientEdit.js
chastichno: Fixed in 7cc1233311bd64ff03002b002899677c85834c60
 >:nit the e is unused, so we can remove it 
🤔 It is used in `formRef?.current?.submit(e);`

src/widgets/Tabs/PatientSelect.js
chastichno: 45f396cc5946091857b64a3a527ad97f4456a372

src/widgets/Tabs/PatientSelect.js
chastichno: af95de77ee6509e1a9d03aa3323e6b6674fcc06b

chastichno:
@mark-prins 
>It would help though, to have an extra column so that you don't click on a dead person. Even better.. why don't we just not show dead people in the search results?

I agree that an extra column would be helpful, however, I want to keep the look of the table simple. We can leave it for another PR and discuss it separately. Anyway, I don't think it's a common thing to pick a deceased patient for vaccination - I think in most cases it is done by accident. Also, as per what [Adam said](https://github.com/openmsupply/mobile/pull/4794#issuecomment-1182608164) we have to keep it in the search results.

chastichno:
@adamdewey 
Thank you for your input!
I changed the localization string with your suggestion of an alert. As per the rest - current behaviour is matching your comments 👍.

@mark-prins @sworup 
I agree that the best would be to make "Is deceased?" field in the form not editable while dispensing. I suggest we leave it for another PR as well if we have time.

mark-prins (approved):
👍  fine by me! 
I still think it would be annoying to have a patient in the list which you click and then it pops the modal - and patient history is available through the dispensing section.. but not a showstopper.

sworup (approved):
Good to merge and create another issue to handle fringe cases

PR #4812, opened 2022-07-20
#4802 Kiribati: Delete patient medical record

### LINKED ISSUE #4802: Kiribati: Delete patient medical record

Add soft deletion of the patients to the "Patient Details" page.

--------

### More context: 

**User Stories:** 

### Proposed solution

- The system will be updated to allow deletion of patient medical records via mSupply Desktop.
- Vaccination events associated with the deleted patient medical record will also be deleted. 
- The system will be updated to make automatic stock inventory adjustments when deleting a vaccination event (e.g. the issuing store will be refunded the vaccine stock that was erroneously recorded as being administered to the patient). 
- Deleted patient medical records and vaccination events will be kept in the back-end for auditing purposes. 

Some other proposed solution details:
- **If there are vaccines dispensed to the patient on a different tablet - it will not be deleted!**
- If there are any dispensary transactions made with this patient on the current tablet, the patient will not be deleted.
- All the vaccination events associated with this patient will be deleted, and stock will be returned.

## Implementation

Not sure yet

## Additional context

Trello card: [(1.7.g + 1.7.h + 1.7.i) Delete patient medical record](https://trello.com/c/DpudXeHa/76-17g-17h-17i-delete-patient-medical-record)

- I'm not sure if there's a way of checking if this patient had transactions on another tablet. As we soft delete it shouldn't cause any problems even if there are any transactions on the other tablet. However, Adam reassured me that they do not use dispensing module at all.

sworup:
Fixes #4802

## Change summary

- Allow patients without any active vaccine event history to be deleted by admin

## Testing

- [ ] Go to patients list 
- [ ] Open the patient detail of one of the patients, who has vaccine history
- [ ] See the delete button but is disabled.
- [ ] Now go back to patients list
- [ ] Select and open patient details of one of the patients whom you know- does not have active vaccine history
  - [ ] Vaccine history that is deleted are not counted 
- [ ] See the delete button and see it enabled.
- [ ] Click the delete button, a popup will show asking you if you are sure
- [ ] Select no you go back to the patients list
- [ ] Select yes and the patient will be soft deleted
- [ ] Check patient list, patient list search, vaccine patient selection windows, the soft deleted patient should not show up in the list. 
- [ ] This effects at least in UI in New patient creation button and form too so please check it works too.

The code has language file changes.

### DIFF
--- a/src/widgets/modalChildren/PatientEditModal.js
+++ b/src/widgets/modalChildren/PatientEditModal.js
@@ -9,7 +9,7 @@ import { FlexRow } from '../FlexRow';
 import { JSONForm } from '../JSONForm/JSONForm';
 import { NameNoteActions } from '../../actions/Entities/NameNoteActions';
 import { selectNameNoteIsValid, selectCreatingNameNote } from '../../selectors/Entities/nameNote';
-import { selectSortedPatientHistory } from '../../selectors/patient';
+import { selectSortedPatientHistory, selectIsCreatePatient } from '../../selectors/patient';
 import { selectCompletedForm, selectCanSaveForm } from '../../selectors/form';
 import { PatientActions } from '../../actions/PatientActions';
 import globalStyles, { SUSSOL_ORANGE } from '../../globalStyles';
@@ -30,6 +30,7 @@ export const PatientEditModalComponent = ({
   nameNoteIsValid,
   canSaveForm,
   hasVaccineEventsForm,
+  isCreatePatient,
 }) => {
   let canSave = canSaveForm;
   const hasVaccineEvents = hasVaccineEventsForm;
@@ -38,6 +39,7 @@ export const PatientEditModalComponent = ({
   }
 
   const canDelete = !isDisabled;
+  const showDelete = !isCreatePatient;
 
   const [removeModalOpen, toggleRemoveModal] = useToggle();
   const [cannotDeleteModalOpen, toggleCannotDeleteModal] = useToggle();
@@ -77,13 +79,14 @@ export const PatientEditModalComponent = ({
             textStyle={styles.saveButtonTextStyle}
             text={generalStrings.save}
           />
-          <PageButton
-            onPress={hasVaccineEvents ? toggleCannotDeleteModal : toggleRemoveModal}
-            isDisabled={!canDelete}
-            style={styles.cancelButton}
-            textStyle={styles.saveButtonTextStyle}
-            text={generalStrings.delete}
-          />
+          {showDelete && (
+            <PageButton
+              onPress={hasVaccineEvents ? toggleCannotDeleteModal : toggleRemoveModal}
+              style={styles.cancelButton}
+              textStyle={styles.saveButtonTextStyle}
+              text={generalStrings.delete}
+            />
+          )}
           <PageButton
             onPress={onCancel}
             style={styles.cancelButton}
@@ -105,6 +108,7 @@ export const PatientEditModalComponent = ({
           confirmText={generalStrings.remove}
           cancelText={buttonStrings.cancel}
           onConfirm={onDeleteForm}
+          isDisabled={!canDelete}
           onCancel={toggleRemoveModal}
         />
       </PaperModalContainer>
@@ -146,6 +150,7 @@ PatientEditModalComponent.defaultProps = {
   surveyForm: null,
   surveySchema: null,
   nameNoteIsValid: true,
+  isCreatePatient: false,
 };
 
 PatientEditModalComponent.propTypes = {
@@ -160,6 +165,7 @@ PatientEditModalComponent.propTypes = {
   onUpdateForm: PropTypes.func.isRequired,
   canSaveForm: PropTypes.bool.isRequired,
   hasVaccineEventsForm: PropTypes.bool.isRequired,
+  isCreatePatient: PropTypes.bool,
 };
 
 const mergeProps = (stateProps, dispatchProps, ownProps) => {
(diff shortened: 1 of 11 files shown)


src/widgets/modalChildren/PatientEditModal.js
```diff
@@ -73,6 +89,15 @@ export const PatientEditModalComponent = ({
           />
         </View>
       </FlexRow>
+      <PaperModalContainer isVisible={removeModalOpen} onClose={toggleRemoveModal}>
+        <PaperConfirmModal
+          questionText={modalStrings.are_you_sure_delete_patient}
+          confirmText={generalStrings.remove}
+          cancelText={navStrings.go_back}
```
chastichno, line 96: I think "Cancel" would be better here

src/widgets/modalChildren/PatientEditModal.js
```diff
@@ -9,27 +9,36 @@ import { FlexRow } from '../FlexRow';
...
   inputConfig,
   surveySchema,
   surveyForm,
   onUpdateForm,
   nameNoteIsValid,
   canSaveForm,
+  hasVaccineEventsForm,
 }) => {
   let canSave = canSaveForm;
+  const hasVaccineEvents = hasVaccineEventsForm;
```
chastichno, line 35: I think that instead of restricting to delete patients if there are vaccination events we should allow to delete of patient AND delete all the vaccination events if they were entered on the current tablet:
Screen Shot 2022-07-22 at 2 15 02 PM

However, if there are vaccination events entered on a different tablet, the user should be notified about it when clicking "Delete". I think making "Delete" enterable is actually better (and just show an alert saying that this patient cannot be deleted because there are vaccination events entered on a different tablet). Currently, it can be sometimes confusing for the user why they cannot delete the patient as there are no alerts.

sworup: @chastichno I was also thinking of showing an alert instead of disabling delete button.
Onto the next point, since we already have individual vaccine event deletion, and it would be less likely for the user to group delete the whole set if we just prevent deletion until all vaccine events associated to the patient has been deleted. Also, since it's the same functionality as we have in mSupply desktop transaction deletion. Also it is easier to do it this way.

Let's keep it this way, I will add the popup notification saying you have to delete vaccine events first. We can update if client complains.

chastichno: Yeah, I agree it is actually easier not to delete the vaccination events for the users. But a pop-up window specifying the reason why they cannot delete a patient would be nice.

src/widgets/modalChildren/PatientEditModal.js
sworup: Resolved with b7055748

src/widgets/modalChildren/PatientEditModal.js
sworup: Resolved with [983e600](https://github.com/openmsupply/mobile/pull/4812/commits/983e6006e7359a693d190312ab2f8a4e7777bf3e)

src/widgets/modalChildren/PatientEditModal.js
```diff
@@ -89,11 +89,18 @@ export const PatientEditModalComponent = ({
           />
         </View>
       </FlexRow>
+      <PaperModalContainer isVisible={cantDeleteModalOpen} onClose={toggleCantDeleteModal}>
```
chastichno, line 92: nit: maybe it's better to use something like `cannotDeleteModalOpen` and `toggleCannotDeleteModal` instead of `cantDeleteModalOpen` and `toggleCantDeleteModal`?

chastichno (changes requested):
I can remove a patient that is not editable on a current site:

https://user-images.githubusercontent.com/41636693/180670633-fb584c8d-bd82-4ce8-9288-a86180cc89b1.mov

src/widgets/modalChildren/PatientEditModal.js
sworup: Resolved with bc81d43f

sworup:
> I can remove a patient that is not editable on a current site:

Fixed with 6bd1c7ab

chastichno (changes requested):
New changes look good tome, but I think I missed one error previously while testing:

https://user-images.githubusercontent.com/41636693/180891553-e8dce89e-15a7-41b8-a53a-284e86f8cae4.mov

sworup:
Wow. @chastichno I am not sure if the error is related to delete. I will look into it and will get back.

sworup:
@chastichno upon checking I realised that the error occurs here:
https://github.com/openmsupply/mobile/blob/80824641c7e16273c75d5750d70f2e13d5e02de9/src/selectors/patient.js#L15-L18

However, at least in my branch, this selector does not and should not trigger patientHistory selector, on New Patient button click.

I can see why it would error out if it was somehow triggered.

sworup:
> New changes look good tome, but I think I missed one error previously while testing:
> 
@chastichno 
Resolved with 0b6b4bbc

chastichno (approved):
Everything works as expected!👍
The only small thing left is that the "Delete" button should be either removed or disabled on the "New patient" window as it is doing the same as the "Cancel" button.

sworup:
> Everything works as expected!👍
> The only small thing left is that the "Delete" button should be either removed or disabled on the "New patient" window as it is doing the same as the "Cancel" button.

Resolved with [8e32b74](https://github.com/openmsupply/mobile/pull/4812/commits/8e32b74cf4e58314cdfba6b2ef942ebfc06e48df)

PR #4850, opened 2022-08-03
#4849 Fix save and delete button was still showing for remote patients

### LINKED ISSUE #4849: Save and delete button are still showing active for remote patients in patient edit
## Describe the bug

- CAN_EDIT_PATIENTS_FROM_ANY_STORE is supposed to be false for Kiribati Vax code as if we allow any store can have edit rights for all other store's patients, there are all sorts of data override over sync errors.
- CAN_EDIT_PATIENTS_FROM_ANY_STORE is set from mSupply store pref where the boolean is set to true by default
- This issue does not set CAN_EDIT_PATIENTS_FROM_ANY_STORE to false but presumes if it is.

- In such case save and delete button was still showing for remote patients

### To reproduce

1. Set CAN_EDIT_PATIENTS_FROM_ANY_STORE pref to false
2. Go to Dispensing page.
3. In edit column you would see the book icon (not the pen icon). This means the record is read only.
4. Click edit to go inside.
5. See that save and delete button are both active when this record should be read only. Even the left side patient detail form is disabled.

### Expected behaviour

- Treat patient records created in other stores as read only data, the can be viewable but should not be editable or delatable.

### Proposed Solution

Update the logic to treat patient records created in other stores as read only data, the can be viewable but should not be editable or delatable.

### Version and device info

- App version:
- Tablet model:
- OS version:

sworup:
Fixes #4849

## Change summary

- Update canSave conditional logic logic.
- Add button disable option for delete button

## Testing

- [ ] Set CAN_EDIT_PATIENTS_FROM_ANY_STORE pref to false. mSupply desktop has default pref that sets CAN_EDIT_PATIENTS_FROM_ANY_STORE to true. Have to disabled.
- [ ] Go to Dispensing page
- [ ] In edit column you would see the book icon (not the pen icon). This means the record is read only. These patient records are created in other store and should not be editable in this store.
- [ ] Click icon (book) to go inside.
- [ ] See that save and delete button these should be inactive. 

### DIFF
--- a/src/widgets/modalChildren/PatientEditModal.js
+++ b/src/widgets/modalChildren/PatientEditModal.js
@@ -35,7 +35,7 @@ export const PatientEditModalComponent = ({
   let canSave = canSaveForm;
   const hasVaccineEvents = hasVaccineEventsForm;
   if (canSave) {
-    canSave = surveySchema && surveyForm ? nameNoteIsValid : !isDisabled;
+    canSave = !isDisabled && surveySchema && surveyForm && nameNoteIsValid;
   }
 
   const canDelete = !isDisabled;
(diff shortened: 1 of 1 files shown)


chastichno (approved):
Looks good!

PR #4851, opened 2022-08-04
#4844 kiribati Vax: Editing vaccination event, clicking save buttons multiple times quickly is crashing apps and which is corrupting the record

### LINKED ISSUE #4844: kiribati Vax:  Editing vaccination event, clicking save buttons multiple times quickly  is crashing apps and which is corrupting the record
## Describe the bug

While editing vaccination events and saving application crashes and while reopening the apps and seeing the history for the same patients it shows blank.

### To reproduce

1. Goto vaccination.
2. Select any patient and click on history.
3. Select the event that you are allowed to edit otherwise it will show the alert which is nice 👍
4. Now after editing try to click save multiple times which will crash the application.
5. Now reopen the application and select same patient and check same history.
6. See error

### Agreed Solution

- This happens when you click save multiple times. So we need to disable save button after 1st click and not enable it until we get a response back from the server or database.

### Version and device info

- App version: 8.4.2.2022727
- Tablet model:Samsung
- OS version:

chastichno:
Fixes #4844

## Change summary

Disabled button after clicking

## Testing

- [ ] Go to any vaccination event that you can edit from the current tablet
- [ ] Click "Save changes" with no changes - alert is shown, the button is back to be enbaled
- [ ] Change vaccination or vaccine, try to click "Save" multiple times - shouldn't be allowed.

### DIFF
--- a/src/widgets/modals/VaccinationEvent.js
+++ b/src/widgets/modals/VaccinationEvent.js
@@ -129,6 +129,9 @@ export const VaccinationEventComponent = ({
   };
 
   const trySave = useCallback(() => {
+    setIsDeletedVaccinationEvent({
+      isDeletedVaccinationEvent: true,
+    });
     const vaccineChanged = vaccine.code !== transactionBatch?.itemBatch?.item?.code;
     const vaccinatorChanged =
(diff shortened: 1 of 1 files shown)


sworup (approved):
Looks good

PR #4859, opened 2022-08-08
#4820 Disabled next button during submission enable it afterwards

### LINKED ISSUE #4820: Vaccination event can be made without filling required fields
## Describe the bug
On quickly tapping the `Next` button, we can skip the required fields and thus create new vaccination event without required details like Dose, consent, Anatomical site, etc.
### To reproduce

1. Go to Vaccinations>>New Patient
2. Fill up the patient details on Step2 page.
3. Now quickly tap Next button twice and you'll see skipping the Step 3 page
4. You've now landed on Step4 page where you can choose vaccines and save it
5. Lookup for the same patient and see their history.
6. You'll see the details with unsaved required field values. Should that even be allowed?
See the recording [here](https://user-images.githubusercontent.com/59193490/180745667-7b207c13-fc1c-46ef-b361-f250a2cf23bc.mp4)

### Expected behaviour
Do NOT allow to skip any required fields at all!

### Proposed Solution

- Immediately after the next button is clicked, disable it. All other clicks would have to wait until either data save succeeds or fails. Fail is trackable. Success we track by the fact that the flow moved to next step.

### Agreed solution

- Create a state that tracks next button enabled status. 
- Enable next button by default and disable it when save trigger happens. 
- After validation fail enable it. 
- If validation passes and data is submitted, since it s a db write through dispatch we cannot have feedback. Immediately afterwards the window is changed to next step window. When this happens the current window goes out of focus, a good time to enable the next button just before focus is lost. 

### Version and device info
- App version: Latest apk built by @sworup

sworup:
Fixes #4820

## Change summary

## Testing

- [ ] Go to Vaccinations > New Patient or Select a patient that already exists
- [ ] Fill up the patient details on Step2 page if new or just leave it as it is for existing.
- [ ] Now quickly tap Next button twice, this should not skip step 3 as it has required fields 
- [ ] In the error case when step 3 was skipped, the system would have allowed you to save this form without there required fields and trying to load this form would have crashed the app. Make sure the step 3's required fields are enforced before it allows us to move to step 4.
- [ ] Same for step 3 try filling the form properly and then multi-clicking it. It should not trigger form submission.
- [ ] Since the ok and confirm button of step 4 is just below the next button of step 3 multi-clicking was triggering it. Since step 4's code auto select the top vaccine the form would submit by accident. And since it happens so fast react-native has not time to update the render properly and it errors out. To fix it, I have added a 2 sec deactivation of the ok next button by default for step 4. 
  - [ ] Check this 2 sec delay is taking effect
  - [ ] Try mult clicking step 3 next button trying to trigger step 4 submission, it should not be possible. 

### DIFF
--- a/src/hooks/useButtonEnabled.js
+++ b/src/hooks/useButtonEnabled.js
@@ -1,0 +1,15 @@
+import { useIsFocused } from '@react-navigation/native';
+import React, { useState } from 'react';
+
+export default function useButtonEnabled(defaultValue = true) {
+  const [enabled, setEnabled] = useState(defaultValue);
+  const isFocused = useIsFocused();
+
+  React.useEffect(() => {
+    if (!isFocused) {
+      setEnabled(true);
+    }
+  }, [isFocused]);
+
+  return { enabled, setEnabled };
+}
--- a/src/widgets/Tabs/PatientEdit.js
+++ b/src/widgets/Tabs/PatientEdit.js
@@ -4,9 +4,10 @@
  * Sustainable Solutions (NZ) Ltd. 2021
  */
 
-import React, { useCallback, useRef } from 'react';
+import React, { useCallback, useRef, useState } from 'react';
 import PropTypes from 'prop-types';
 import { ToastAndroid, View } from 'react-native';
+import { useIsFocused } from '@react-navigation/native';
 import { connect } from 'react-redux';
 import * as Animatable from 'react-native-animatable';
 import { FormControl } from '../FormControl';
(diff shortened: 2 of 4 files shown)


raviSussol (approved):
Changes look good to me - though haven't tested functionally.

sworup:
@chastichno @raviSussol I have extended the logic for patient step to all steps.

chastichno (approved):
Looks good to me. The only concern I have is that this 2 seconds delay might disrupt work for someone in case they are working very fast (I don't think it's the case really though).

PR #4864, opened 2022-08-09
#4647 current stock add toggle to hide zero stock

### LINKED ISSUE #4647: Hide items with zero stock, or add toggle to hide them on the 'Current Stock' page
Currently mobile displays all items that the store has visibility of on the 'Current Stock' page, including potentially many lines with 0 stock. 

This can make it a bit hard for some users to identify what is actually in stock - so either hiding these or having an option to toggle them on/off them would be ideal (maybe better to have a toggle in case someone wants to see items that are potentially missing/uncounted?)

TBD

## Implementation

Ignoring

## Additional context
Confusion/app behaviour questioned in support ticket: https://support.msupply.org.nz/cerb6/profiles/ticket/IG-3278-HL

arjunSussol:
Fixes #4647 

## Change summary

Added a toggle on `Current stock` having label `Hide zero stocks`

## Testing

- [ ] Go to `Current stock` menu
- [ ] By defaults it shows all items having some stock and zero stock, also the `Hide zero stocks` toggle button should not be highlighted which means it showing all items with and without(zero) stocks.
- [ ] If we click the `Hide zero stocks` toggle button, it should be highlighted and hide items having zero stock.
- [ ] In another click it should not be highlighted (remains as initial state) and displays all items.
- [ ] Try sorting, searching, filtering, which must comply the state of `Hide zero stocks` toggle button.

### DIFF
--- a/src/localization/buttonStrings.json
+++ b/src/localization/buttonStrings.json
@@ -78,8 +78,7 @@
     "customer_data": "Données Clients",
     "factory_reset": "Réinitialiser",
     "check_connection": "Vérifiez la connexion",
-    "connection_status": "Connexion réussie",
-    "hide_zero_stocks": "Masquer les stocks zéro"
+    "connection_status": "Connexion réussie"
   },
   "gil": {
     "add_master_list_items": "Rinea am list",
@@ -98,8 +97,7 @@
     "new_supplier_invoice": "Bebwan am oota\nman pharmacy",
     "past": "Are I mwaina",
     "use_requested_quantities": "Kabongana maitina\nae kainnanoaki",
-    "use_suggested_quantities": "Kabongana te\nmwaiti are e katauaki",
-    "hide_zero_stocks": "Hide zero stocks"
+    "use_suggested_quantities": "Kabongana te\nmwaiti are e katauaki"
   },
   "tl": {
     "add_master_list_items": "Utiliza Lista Master",
@@ -118,8 +116,7 @@
     "new_supplier_invoice": "Distribuidor nia Invoice Foun",
     "past": "Pasadu",
     "use_suggested_quantities": "Kuantidade Sujere",
-    "use_requested_quantities": "Kuantidade Ezije",
-    "hide_zero_stocks": "Hide zero stocks"
+    "use_requested_quantities": "Kuantidade Ezije"
   },
   "la": {
     "add_batch": "​ໃສ​່​ເລກ​ຊຸດ​ຜະ​ລິດ",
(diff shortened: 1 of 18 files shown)


anildahalsussol:
@arjunSussol : LGTM 👍
@DhanyaHerath: Is that label ok or need to change to something else?

anildahalsussol (approved):

arjunSussol:
[Mobile: #4647 Hide items with zero stock, or add toggle to hide them on the 'Current Stock' page](https://trello.com/c/YBCG7y8M/436-mobile-4647-hide-items-with-zero-stock-or-add-toggle-to-hide-them-on-the-current-stock-page)

src/localization/buttonStrings.json
```diff
@@ -77,7 +78,8 @@
     "customer_data": "Données Clients",
     "factory_reset": "Réinitialiser",
     "check_connection": "Vérifiez la connexion",
-    "connection_status": "Connexion réussie"
+    "connection_status": "Connexion réussie",
```
sworup, line 81: @arjunSussol if the keys does not exist then translation would fallback to default language keys. Which means we do not need to add keys (english keys) in french or other language... it can be translated it time but for now please remove them.

src/pages/dataTableUtilities/getPageInitialiser.js
```diff
@@ -185,6 +185,7 @@ const stockInitialiser = () => {
     filterDataKeys: ['name', 'code'],
     sortKey: 'name',
     isAscending: true,
+    showAll: true,
```
sworup, line 188: @arjunSussol since seeing zero stock is annoying enough to ask for the feature addition, I think the default could be showAll false. 

However a bigger concern is the user of state variable `showAll`. I think it should be `isShowZeroStocks`. Show is a state where all filters are removed, which is the default state. There is no show all button. What we are doing is adding a filter to default state so the state should represent that filter not show all state. 

Right now there is only one filter so this works for now but if someone add one more filter my argument will start to become more relevant. So we should do this now when the PR is young.

sworup (changes requested):
@arjunSussol Code looks very nice given that this is one of your first PRs to mobile branch. Welcome to mobile branch development btw. 

I have added some comments below in the review.

DhanyaHerath:
@anildahalsussol Thanks for asking! I think the label would be more clear as "Hide items with zero stock" if that fits

arjunSussol:
> @anildahalsussol Thanks for asking! I think the label would be more clear as "Hide items with zero stock" if that fits

Hi @DhanyaHerath your suggestion is acceptable. I tried with that and notice that label is slightly unfit in one line word as attached in snapshot. 

Screen Shot 2022-08-15 at 14 19 09

To make it fit with better visual we need to improve css style in different places, cannot be fix by a single correction.
So, it looks that style could be broken from different aspects.

We decided to keep label as before `Hide zero stocks` if you agree.

src/pages/dataTableUtilities/getPageInitialiser.js
arjunSussol: @sworup I have tried to implement your suggestion, but it could break the existing workflow.
For toggle, I am using existing `toggleStockOut` reducer which is using by another functionality too.
So, If I rename the state variable `showAll` to something else then the recent toggle as well as existing toggle functionality is not working properly. Either we have to change the code all over or stick with existing.

I preferred existing code if we are using existing function/method.
Later, if we are going to create new reducer with no dependency over existing one then we can create new state variable, but for now it should be fine.

src/localization/buttonStrings.json
arjunSussol: Done!

sworup (approved):
Show all is tried to everything in our code which is a flaw but beyond the scope of this PR so approved.

Issue #4842, opened 2022-08-02
Kiribati vax: Deleted patients again shows up


anildahalsussol:
## Describe the bug

Deleted patients again show up on tablets.

Related to https://github.com/sussol/msupply/issues/11303

### To reproduce

Steps to reproduce the behaviour:

1. Create a name and sync.
2. Once it's synced ask another tablet say B to sync.
3. Now that name must be visible to another tablet as well.
4. Start creating vaccination event from tablet B and delete the name from tablet A.
5. After event is done and sync from tablet B to server.
6. Now sync tablet A and check patient and search online.
7. See error.

### Proposed Solution

Leave if you don't know how to fix/implement. Edit this issue description and explain here if you know the best path of implementing the fix within the codebase.

### Version and device info

- App version:8.4.2.20220727
- Tablet model:Samsung
- OS version:

sworup:
Our vax system is not designed to allow patient records to be editable by all stores that. It should (for now until we build proper sync logic) only allow the store owning the patient to edit that patient. 
**But in mSupply store pref there is a setting to allow patients to be editable from all store.**
I think we can start with disabling this pref by default for kiribati vax customisation. 
Now this would probably mean all data should be resynced. 

@anildahalsussol @chastichno @adamdewey @ujwalSussol opinions please.

adamdewey:
@sworup - hmm, 2 things: 

1. We need to be able to somehow tell them which tablet they need to use to edit the patient, they can't try all 40 tablets until they find the right one. (Possibly a dashboard report would be sufficient)
2. We need to be careful about re-syncing data, it would be a big effort for that much data in Kiribati, especially for 40 tablets. Would need extra coordination with the Kiribati team during the upgrade process

sworup:
> 1. We need to be able to tell them which tablet they need to use to edit the patient, they can't try all 40 tablets until they find the right one. (Possibly a dashboard report would be sufficient)

- Will create an issue for this.

> 2. We need to be careful about re-syncing data, it would be a big effort for that much data in Kiribati, especially for 40 tablets. Would need extra coordination with the Kiribati team during the upgrade process.

- @adamdewey It is just 1 pref boolean that gets updated not all data so if we set this boolean during initial tablet update we should be fine.
- We run this footrunner code https://github.com/openmsupply/mobile/issues/4876#issuecomment-1212762457 in mSupply and have all the tablet sync the all stores can edit patient boolean pref.

sworup:
[...]

PR #4882, opened 2022-08-15
#4841 Fix lookup patient overriding current changes

### LINKED ISSUE #4841: Kiribati Vax: Lookup patient is overwriting the changes done in patient detail
## Describe the bug

A clear and concise description of what the bug is. Use screenshots!

### To reproduce

1. Go to dispensing and select patient name
2. Click on edit and then edit ID (identification info like citizen) and save
3. Now go back and select the name from the look-up patient.
4. Open the same name and see the ID.

### Proposed Solution

- Since lookup is overriding the local data and the we need a viable simple solution we can do this.
  - Lookup will query all data on server like it does right now.
  - When the user clicks on the item, we check internally to see if that record exist in our database already.
  - If the record exists we do not pull that data from the server (no db update) we show the data from database even if it is old.
    - We are expecting sync will take its normal course and adjust that data especially since now we are restricting patient edit to one store.
  - If the record does not exist on our database then we pull the data.

### Version and device info

- App version: 8.4.2.20220727
- Tablet model:Samsung
- OS version:

sworup:
Fixes #4841

## Change summary

## Testing

- [ ] Do something


sworup:
@chastichno this is effectively the same changes as https://github.com/openmsupply/mobile/issues/4842. With additional change to not pull Patient data that already has a record locally. So I am merging it.

raviSussol (approved):
Lgtm!

PR #4906, opened 2022-08-26
#4905 Lock vaccination event details to home store

### LINKED ISSUE #4905: Kiribat Vax: Vaccination event details extra information pages edit gets overridden if multiple tablets are involved
# Description

- Vaccination events detail page has 3 sub-pages 1. Extra Information, 2. Addition details and 3. Vaccination event
- Sub-page 3 is only editable in the store where the vaccination event originated at. The other 2 sub-pages can be edited anywhere.
- If we edit the sub-page 1 and 2 concurrently in 2 or more tablets and save the changes, the change that is synced last will override all other changes. 
- This might be very annoying for the user.

## Agreed solution

- We cannot merge changes and do latest change check for vaccination event details edit right now given the tight deadline right now so we do a quick and dirty solution for now.
- Just like sub-page 3, sub-page 1-2 would also only be editable in the store tablet the vaccination event originated in. That would avoid the sync race condition issue for now.

sworup:
Fixes #4905

## Change summary

## Testing

- [ ] Do something

### DIFF
--- a/src/widgets/modals/VaccinationEvent.js
+++ b/src/widgets/modals/VaccinationEvent.js
@@ -254,7 +254,7 @@ export const VaccinationEventComponent = ({
                         isPCDValid: validator(changed.formData),
                       });
                     }}
-                    disabled={isDeletedVaccinationEvent}
+                    disabled={isDeletedVaccinationEvent || !transaction}
                   >
                     <></>
                   </JSONForm>
@@ -291,7 +291,7 @@ export const VaccinationEventComponent = ({
                       isSupplementalDataValid: validator(changed.formData),
                     });
                   }}
-                  disabled={isDeletedVaccinationEvent}
+                  disabled={isDeletedVaccinationEvent || !transaction}
                 >
                   <></>
                 </JSONForm>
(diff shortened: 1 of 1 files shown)


chastichno (approved):
@sworup Changes look good to me. Also, I honestly don't consider this solution to be "dirty" and I would say that this way it is actually more consistent => if users made a mistake, they can make any changes only on the tablet they used. We don't want to encourage people to make mistakes, so I think it's good for it to be a little bit harder.

One comment about the code - I think that `!transaction` is not quite readable. I understand that it indicates that the transaction for this vaccination event is located on the current tablet, but maybe we can add a variable that will just make this code more readable for the future, for example, `isTransactionCreatedOnThisTablet` (no, I'm not good with variables' names). Apart from that, everything looks great!

sworup:
@chastichno I have replaced the transaction check with a boolean. [3d10ce7](https://github.com/openmsupply/mobile/pull/4906/commits/3d10ce7d8460b5cd5e483ca3b399a3447ad15302)

Gonna merge the changes.

PR #4912, opened 2022-09-01
#4911 when new vaccine event is created all other vaccine events of the patient are updated along with it causing sync issues 

### LINKED ISSUE #4911: Kiribati Vax: When new vaccine event is created all other vaccine events of the Patient are updated, causing potential sync race conditions
## Describe the bug

When new vaccine event is created all other vaccine events of the Patient are updated along with it causing sync issues.

### To reproduce

1. Go to "Vaccinations"
2. Search of a Patient with current store as home. One with previous vax events.
3. Click on the Patient to create new Vaccine event.
4. You may edit patient details but that is irrelevant.
5. Complete the steps to create a vaccination event.
6. Visually you may not see it, but all of the vaccination events of that Patient has been updated, probably with the same data. This creates update sync out records for these vax events (Name Notes). 
7. You may use the Realm explorer to see this happen. Before syncing, go to it and look for SyncOut object string. It will show pending sync out records, and there would be multiple NameNotes updates, there should only be 2 in case of Creating new vaccine event.

### Expected behaviour

- New Vaccine event creation should not update existing vaccine events of the Patient.

### Proposed Solution

- In CreateRecords.ts in createPatient which is also used for Update patient which new vaccine event tries to do during new vaccine event creation, there is a line of code that loops through all nameNotes and updates them. 
- This code is there because during sync in of new patient these name notes had to be created. Remember create and update are same method createPatient.  We do not want to create/update all name notes for this particular case, and there is a code in NameNote create code block that creates 1 single Vax event name note. 
- So we need to add a check that ignores this code during vaccine event creation and edit too.

https://github.com/openmsupply/mobile/blob/80824641c7e16273c75d5750d70f2e13d5e02de9/src/database/utilities/createRecord.js#L337
- We do want it for sync in and other things so a boolean check would be required, we cannot remove the code block that is.

### Version and device info

- App version:
- Tablet model:
- OS version:

sworup:
Fixes #4911

## Change summary

## Testing

- [ ] Do something

### DIFF
--- a/src/actions/Entities/NameActions.js
+++ b/src/actions/Entities/NameActions.js
@@ -104,7 +104,7 @@ const saveEditing = () => (dispatch, getState) => {
 
   const patientRecord = { ...currentPatient, dateOfBirth, createdDate, name };
 
-  UIDatabase.write(() => createRecord(UIDatabase, 'Patient', patientRecord));
+  UIDatabase.write(() => createRecord(UIDatabase, 'Patient', patientRecord, true));
   dispatch(reset());
 };
 
--- a/src/actions/Entities/NameActions.js
+++ b/src/actions/Entities/NameActions.js
@@ -73,7 +73,7 @@ const select = name => async dispatch => {
     const result = await createPatientVisibility(name);
     if (result) {
       UIDatabase.write(() => {
-        selectedName = createRecord(UIDatabase, 'Patient', name);
+        selectedName = createRecord(UIDatabase, 'Patient', name, true);
       });
     } else {
       ToastAndroid.show(generalStrings.problem_connecting_please_try_again, ToastAndroid.LONG);
--- a/src/actions/Entities/NameNoteActions.js
+++ b/src/actions/Entities/NameNoteActions.js
@@ -63,7 +63,12 @@ const updateForm = (data, validator) => ({
 const saveEditing = () => (dispatch, getState) => {
   const nameNote = selectCreatingNameNote(getState()) ?? {};
   const patient = UIDatabase.get('Name', nameNote?.nameID);
-  const isDirty = JSON.stringify(patient?.mostRecentPCD?.data) !== JSON.stringify(nameNote?.data);
+  const isDirty =
+    !(
+      nameNote?.data &&
+      Object.keys(nameNote?.data).length === 0 &&
+      Object.getPrototypeOf(nameNote?.data) === Object.prototype
+    ) && JSON.stringify(patient?.mostRecentPCD?.data) !== JSON.stringify(nameNote?.data);
   if (isDirty) {
     UIDatabase.write(() => createRecord(UIDatabase, 'NameNote', nameNote));
   }
--- a/src/actions/Entities/NameNoteActions.js
+++ b/src/actions/Entities/NameNoteActions.js
@@ -9,6 +9,7 @@ import {
 import { selectSurveySchemas } from '../../selectors/formSchema';
 import { validateJsonSchemaData } from '../../utilities/ajvValidator';
 import { vaccineStrings } from '../../localization';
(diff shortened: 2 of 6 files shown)


sworup:
@chastichno would you please review this. This is a major bug. I will write test cases tomorrow morning for this.

src/actions/Entities/NameNoteActions.js
```diff
@@ -63,7 +63,12 @@ const updateForm = (data, validator) => ({
 const saveEditing = () => (dispatch, getState) => {
   const nameNote = selectCreatingNameNote(getState()) ?? {};
   const patient = UIDatabase.get('Name', nameNote?.nameID);
-  const isDirty = JSON.stringify(patient?.mostRecentPCD?.data) !== JSON.stringify(nameNote?.data);
+  const isDirty =
+    !(
+      nameNote?.data &&
+      Object.keys(nameNote?.data).length === 0 &&
+      Object.getPrototypeOf(nameNote?.data) === Object.prototype
+    ) && JSON.stringify(patient?.mostRecentPCD?.data) !== JSON.stringify(nameNote?.data);
```
chastichno, line 71: I struggle to read this piece of code, could you maybe explain a little bit what's going on?
From what I can see, the `isDirty` variable basically means that the nameNote was updated.
The condition that you added here doesn't seem clear to me...

src/actions/Entities/NameActions.js
```diff
@@ -73,7 +73,7 @@ const select = name => async dispatch => {
     const result = await createPatientVisibility(name);
     if (result) {
       UIDatabase.write(() => {
-        selectedName = createRecord(UIDatabase, 'Patient', name);
+        selectedName = createRecord(UIDatabase, 'Patient', name, true);
```
chastichno, line 76: In the same method, but on line 107 there's code:
```
UIDatabase.write(() => createRecord(UIDatabase, 'Patient', patientRecord));
```
Shouldn't we update it with the extra parameter as well? Or maybe we can add a default value for that parameter in `CreateRecord` method?

src/actions/Entities/NameNoteActions.js
sworup: The condition just check if the object is not empty is not empty, not null and is actually an object. 

The dirty variable was getting set to true when both the comaprion variable were empty ({}). 

Maybe we need to create utility function for this check, because this is for the case we are not even sure an object is sent in the variable or that object is empty or null and I see cases in our code where it can be useful. This can avoid app crashing.

src/actions/Entities/NameActions.js
sworup: Yes we should update it. And in reselector for some reason default value does not work. If I add a default value, it seem to get set permanently no matter what I sent from outside. I will try some more today to set the default value.

src/actions/Entities/NameNoteActions.js
chastichno: Thank you for the explanation, however, I can't understand why is there a `!` in front of it then:
```
!(
      nameNote?.data &&
      Object.keys(nameNote?.data).length === 0 &&
      Object.getPrototypeOf(nameNote?.data) === Object.prototype
    )
```

sworup: If the object is empty, which this condition without the bang return as true, then the data is blank and is not dirty. So it has to not empty to go to the right hand side part of the conditional. It is an not empty check. a6ce058 this might make it more clear.

src/actions/Entities/NameActions.js
sworup: Resolved with 564a263c

chastichno (approved):
Thank you Sworup! Looks very good to me 🙌✨

PR #4919, opened 2022-09-07
#4909 Subpage 2 Additional details' changes are overridden by Subpage 3 Vax Event changes

### LINKED ISSUE #4909: Kiribati Vax: Editing Vaccination events' sub-pages without syncing overrides each other
## Describe the bug
Everytime you change the vaccine event details together (Extra info, Additional details and vaccination event form), all the changes do not get saved together, i.e, only the Vaccination event changes are reflected after the change while the other changes are discarded 

### To reproduce
1. Go to a patient's vaccination history
2. Edit additional details like dose number, arms. etc ann Save
3. Now edit Vaccination Event; like vaccinator or vaccine and save.
4. Sync
5. Look back to that same patient vaccine history
6. See error: only last changes in vaccination event form is seen, the previous changes in other forms is discarded

### Expected behaviour
definitely not as above?

### Proposed Solution

### Version and device info
- App version: latest [apk](http://t.ly/0uZv)

sworup:
Fixes #4909

## Change summary

- Previously a variable customDataObject was being set that never changes with form changes.
- In subpage 3 vax event save, subpage 2 additional details are also saved.
- Now subpage 2's save uses the state object updatedSupplementalDataForm which does change with form change.
- We needed to pass in updatedSupplementalDataForm to the subpage 3 save method instead of customDataObject to have this issue fixed.

## Testing

- [ ] Select a vax event that belongs to current store so that it would be editable
- [ ] Edit Page 2: Additonal Details data, save it
- [ ] Edit Page 3: Vaccination event data and save it.
- [ ] Now sync the changes
- [ ] Load the vax event again and check, the Page 2 should not have been overridden it should have the latest changes.
- [ ] Page 3's save would auto save Page 2's change so you actually do not need to click Page 2's save button, page 3's click should save them both.
  - [ ] That is not the case for Extra Informaiton, extra information is independent from the next 2 pages (I don't know why the previous dev chose to do it this way, but it is ok, it does not break anything).

### DIFF
--- a/src/widgets/modals/VaccinationEvent.js
+++ b/src/widgets/modals/VaccinationEvent.js
@@ -159,10 +159,10 @@ export const VaccinationEventComponent = ({
 
   useEffect(() => {
     if (customDataObject) {
-      setSupplementalData({
+      setSupplementalData(prevState => ({
+        ...prevState,
         updatedSupplementalDataForm: customDataObject,
-        isSupplementalDataValid: true,
-      });
+      }));
       setLoading(false);
     }
   }, [customDataObject]);
@@ -209,7 +209,7 @@ export const VaccinationEventComponent = ({
         transactionBatch,
         vaccine,
         vaccinator,
-        customDataObject,
+        updatedSupplementalDataForm,
         vaccinationEventNameNote
       );
       toggleEditTransaction();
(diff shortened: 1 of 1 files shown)


anildahalsussol (approved):
@sworup : LGTM

PR #4921, opened 2022-09-07
#4920 Patient homes store vax event error

### LINKED ISSUE #4920: Kiribati Vax: Creating new Vax Event would still update other vax events of the patient if it is the patient's home store
## Describe the bug
- When we create new Vax event, if the store is the home store of the patient, the update all other vax event trigger still fires. 
- This creates a case where if store is not the patient's home store, the overriders of https://github.com/openmsupply/mobile/issues/4911 does not happen. But if the store is patient's home store it does happen.

### To reproduce

1. Use 2 tablets
2. In tablet 1. Create a new patient and sync
3. In tablet 2, create a vax event for that patient and sync.
4. Now, back in tablet1, sync and Go to Vaccination > Search for that patient and Create a new vaccination event for this patient in its homestore. Don't sync yet.
5. Go to other tablet2  and select the same patient, and delete their previous/existing vaccination event. And sync.
6. Now go back to tablet 1 and sync. It is important that tablet 1 would sync later as new sync would always override old ones.
7.  See error: the deleted record is still showing up.
8. Now follow same steps, just this time try editing the vax event instead of deleting it. In this case, you'll see multiple lines created for the same vax event which was edited earlier.

### Expected behaviour
Editing or deleting the vax event should reflect the changes properly 

### Proposed Solution

### Version and device info
- App version: latest apk

sworup:
Fixes #4920

## Change summary

## Testing

- [ ] Do something

### DIFF
--- a/src/actions/Entities/NameActions.js
+++ b/src/actions/Entities/NameActions.js
@@ -96,7 +96,7 @@ const filter = (key, value) => ({ type: NAME_ACTIONS.FILTER, payload: { key, val
 
 const sort = sortKey => ({ type: NAME_ACTIONS.SORT, payload: { sortKey } });
 
-const saveEditing = () => (dispatch, getState) => {
+const saveEditing = (shouldCreateNameNames = true) => (dispatch, getState) => {
   const currentPatient = selectEditingName(getState());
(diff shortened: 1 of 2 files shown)


anildahalsussol (approved):
@sworup : LGTM

PR #4948, opened 2022-09-28
#4943 Create migration to set isDeleted and isDeceased of Name and NameNotes to false by default

### LINKED ISSUE #4943: Upgrading from 8.5 to 8.6 is causing some data loss
## Describe the bug
While upgrading the existing syncing mobile store from 8.5 to 8.6, the patient lists are all gone from the dispensary.

### To reproduce
1. Get a mobile store sync setup in 8.5.
2. You can create some patients there as well for testing purpose.
3. Now, upgrade it to latest 8.6 from [here](http://t.ly/CpdV)
4. Open it and click Dispensary module
5. See error: the patient list is literally empty; not even the patient created before is there:
Screen Shot 2022-09-23 at 11 07 13
6. Search for patient in the lookup. Click on it.
7. Read that alert message toast:
Screen Shot 2022-09-23 at 11 08 00

### Expected behaviour
Data should not disappear like that whatsoever

### Proposed Solution

### Version and device info
- App version: 8.6

sworup:
Fixes #4943

## Change summary

- Currently when we upgrade from v8.5.0 to v8.6.0 the isDeleted and isDeceased fields of Name and NameNote tables are set to null. Which cause the Patients and Patients' vaccination events lists to show up blank.

## Testing

- [ ] Install mobile v8.5.0. Go to Dispensing. See the list of Patients available.
- [ ] Install mobile v8.6.1 (with this PR's fix). Wait for few seconds to few minutes while migration code runs.
- [ ] Go to Dispensing. See the list of Patients available. Previously the list would be empty.


src/database/DataTypes/Name.js
```diff
@@ -197,7 +197,7 @@ Name.schema = {
     middleName: { type: 'string', optional: true },
     lastName: { type: 'string', optional: true },
     isActive: { type: 'bool', optional: true },
-    isDeceased: { type: 'bool', optional: true },
+    isDeceased: { type: 'bool', default: false, optional: true },
```
bijaySussol, line 200: Sweet, I wondered of that

src/dataMigration.js
```diff
@@ -279,6 +279,36 @@ const dataMigrations = [
       }
     },
   },
+  {
+    version: '8.6.0',
```
bijaySussol, line 283: Super @sworup

bijaySussol (approved):
@sworup : 

Looks great to me.
Wondering of something like that required
Super Super !!!
Lets merge it and create a new installer with the same version we can test unless @nisha-dangol already tested it

Thanks

chastichno (approved):

PR #4950, opened 2022-09-28
#4908 Alert improved while vaccination event edited

### LINKED ISSUE #4908: Improve Alert: While editing vaccination event need to show atleast store id or need to query for store name from name note
## Describe the bug

Now the alert while editing the vaccination event is quite vague if the store name doesn't exist in the name note. That may lead users to search for all the tablets to find the Homestore of name note event.

### To reproduce

1. Go to Vaccination.
2. Search online the patients.
3. Choose any patients with vaccination events.
4. Check the history and select any vaccination event of that patient.
5. Now try to edit the see alert.

### Expected behaviour

It will be difficult for users to search the home store of vaccination events to edt. So good to show at least store id. Later we can make one dashboard report for this.

### Version and device info

- App version: 8.5.0 develop
- Tablet model:
- OS version:

### Additional context

Screenshot 2022-08-31 at 14 12 23

arjunSussol:
While editing another store vaccination event, the name note records neither show the created store name nor store id.

So, if store name is available(mostly in new records), then show the store name as it is. But for older records which doesn't have store_id or storeName related then show at least store_id of name_note which should be similar to prescription's store_id

Fixes #4908 

## Change summary

Updated `NameNote` schema with `storeID` field which hold the real store_id to display on alert.
Added above `storeID` field all over where NameNote is created/update and during incoming/outgoing sync records etc.

Data migration is not required as most of the existing data of mobile has wrong store_id. The store_id of prescription(transaction) is different to name_note.store_id.

So, as schema is updated we need to re-sync whole data to get records for store_id

## Testing

- [ ] First, don't use existing mobile data. Just delete that and recreate new one with fresh setup
- [ ] Look at the `Vaccinations > Select a patient > Click History button`
- [ ] There should be at least one vaccination history, click on either one.
- [ ] Better to select vaccination history created from another store which should be non-editable in home store.
- [ ] In vaccination event details window, click on `Save changes` and `Edit` button
- [ ] If this is new data and storeName is available then alert should have store name.
- [ ] If storeName is not available then it will be populated with store_id having same alert message.
- [ ] Similar to attached snapshot.
Screen Shot 2022-09-28 at 14 48 11

### DIFF
--- a/src/database/DataTypes/NameNote.js
+++ b/src/database/DataTypes/NameNote.js
@@ -28,6 +28,7 @@ export class NameNote extends Realm.Object {
       nameID: this.name?.id,
       note: this.note,
       patientEventID: this.patientEvent?.id,
+      storeID: this.storeID,
     };
   }
 }
@@ -43,6 +44,7 @@ NameNote.schema = {
     name: 'Name',
     note: { type: 'string', optional: true },
     patientEvent: 'PatientEvent',
+    storeID: { type: 'string', optional: true },
   },
 };
 
--- a/src/database/DataTypes/NameNote.js
+++ b/src/database/DataTypes/NameNote.js
@@ -44,7 +44,7 @@ NameNote.schema = {
     name: 'Name',
     note: { type: 'string', optional: true },
     patientEvent: 'PatientEvent',
-    storeID: { type: 'string', optional: true },
+    storeID: { type: 'string', default: '' },
   },
 };
 
--- a/src/actions/Entities/NameNoteActions.js
+++ b/src/actions/Entities/NameNoteActions.js
@@ -74,7 +74,7 @@ const saveEditing = () => (dispatch, getState) => {
 };
 
 const updateNameNote = (originalNote, updatedData) => () => {
-  const { id, patientEvent, name, entryDate, _data, isDeleted } = originalNote;
+  const { id, patientEvent, name, entryDate, _data, isDeleted, storeID } = originalNote;
 
   // Quick & dirty check if the object was updated, trims out some un-needed updates
   const isDirty = _data !== JSON.stringify(updatedData);
@@ -87,6 +87,7 @@ const updateNameNote = (originalNote, updatedData) => () => {
       entryDate: new Date(entryDate),
       _data: JSON.stringify(updatedData),
       isDeleted,
+      storeID,
     };
 
     UIDatabase.write(() => {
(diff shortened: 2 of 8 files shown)


src/database/DataTypes/NameNote.js
```diff
@@ -43,6 +44,7 @@ NameNote.schema = {
     name: 'Name',
     note: { type: 'string', optional: true },
     patientEvent: 'PatientEvent',
+    storeID: { type: 'string', optional: true },
```
sworup, line 47: @arjunSussol This variable is not optional, as it will be added by our code automatically. Set the default value to "" and remove optional please.

sworup (changes requested):
Code looks good. Needs to be properly tested on tablet though as it updates migration, which I have not done yet. 
Please make the migration code update suggested by the review.

src/database/DataTypes/NameNote.js
sworup: @arjunSussol have you made these changes?

arjunSussol: @sworup done by https://github.com/openmsupply/mobile/pull/4950/commits/1c6a397c698a13c8dae25748a4519b6893621965

arjunSussol: @sworup I think each name_note has store_id, so it should be mandatory and have valid record.
So, we need it as `optional: false`. However, lets discuss before reaching to any point.

sworup (approved):
LGTM.
But this might have to wait a while to go into master as it changes schema.

ujwalSussol:
@nisha-dangol   :  I have created a apk for testing.

https://drive.google.com/drive/u/0/folders/1TVGqMwjYQC7j9lOUihn0RgUEvp3cZY0V

@sworup  tells me that we need a existing tablet version that is lower than  v8.6.1 with the Kiribati data setup to test this...

sworup:
@nisha-dangol we would need to do a fresh install in v8.5.X, of Kiribati vax data file. Then let the apk created with this PR's changes to update version. 

We want to see:

1. If the apk would refuse to update from v8.5 to v8.6
2. If it updates would it trigger a resync of name notes as its schema has changed? Yes, a resync happens, then we have consider if the changes in this PR are important enough to merged into master. No, then we have check if patient events work as expected.

ujwalSussol:
Talking with @sworup and @arjunSussol  : it seems best to close this issue... This PR requires the re-sync of legacy data and the client may not be keen to do this. 

So closing this issue for now.

PR #4973, opened 2022-10-19
#4949 allow setting editable from login page

### LINKED ISSUE #4949: Ability to change Sync URL  and site password from the mobile login window

While supporting Mobile, it was evident that `Site setting` needs to be modifiable from the mSupply Mobile Login window. 

The problem : 

* A client has a Mobile issue.
* With great difficulty we get the Mobile data file. 
* We import to the local mobile tablet.
* **Problem :** It attempts to sync to the live server. We have no easy way to re-point to another clone system. 

- We have the mSupply login window as shown. 
- Lets have a setting icon. When clicked we will display a `Authorisation` window with Hardcoded password that we know.
-Once password is correctly entered, we need the following window to show the following details : 
- Site name :  [data not changeable]
- Site ID : [data not changeable]
- Site password : [Password to be update-able]
- Site URL : [URL to  update-able]
- We need a [save] and [back]
- If we have the above feature then testing and debugging mobile issue will be easier. 

Mobile_site_and_url_chnage

## Implementation

n/a

n/a

## Additional context

I think there is a setting page that you can get to once you login. So perhaps a review and we just expose this page at the beginning.

arjunSussol:
Fixes #4949 

## Change summary

Added `Sync site ID` and `Sync site name` to `SettingPage` , both fields are non editable.
Allow to visit `SettingPage` from `LoginPage` with label `SETTINGS`.
Added environment variable (.env) which is ignored in git, but there is .env.example file with same key without value.
.env file should have admin password which is not store in mobile database.

## Testing

- [ ] First, sync database then go to Login page.
- [ ] At `Login` page there is `SETTINGS` label in bottom.
- [ ] Click Settings alert with input field which required valid password
- [ ] Entered admin password and click `Done` button
- [ ] If password is valid then Setting page should be visible
- [ ] In Setting page, change the editable fields then click `Save` button
- [ ] Alert with input field which required valid password.
- [ ] Enter valid admin password as same as above entered password.
- [ ] If valid password entered then it will save settings and related pop up visible, otherwise not.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,3 @@
+# This env file should be renamed to .env and mSupply_desktop_admin_password should be replaced with real password.
+# Also, please remove this comment
+REACT_APP_ADMIN_PASSWORD_HASHED=mSupply_desktop_admin_password
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # OSX
 #
 .DS_Store
+.env
 
 # Xcode
 #
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -21,7 +21,15 @@ buildscript {
     }
 }
 
+def REACT_NATIVE_VERSION = new File(['node', '--print',"JSON.parse(require('fs').readFileSync(require.resolve('react-native/package.json'), 'utf-8')).version"].execute(null, rootDir).text.trim())
+
 allprojects {
+    configurations.all {
+        resolutionStrategy {
+            // Remove this override in 0.65+, as a proper fix is included in react-native itself.
+            force "com.facebook.react:react-native:" + REACT_NATIVE_VERSION
+        }
+    }
     repositories {
         mavenLocal()
         google()
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,3 +1,13 @@
 module.exports = {
   presets: ['module:metro-react-native-babel-preset'],
+  plugins: [
+    [
+      'module:react-native-dotenv',
+      {
+        envName: 'APP_ENV',
+        moduleName: '@env',
+        path: '.env',
+      },
+    ],
+  ],
 };
--- a/package.json
+++ b/package.json
@@ -140,6 +140,7 @@
     "lint-staged": "^10.5.4",
     "metro-react-native-babel-preset": "^0.66.0",
     "prettier": "^2.2.1",
+    "react-native-dotenv": "^3.3.1",
     "react-native-typescript-transformer": "^1.2.13",
     "typescript": "^4.3.5"
   }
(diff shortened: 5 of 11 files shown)


sworup (changes requested):
@arjunSussol I tried to run your code locally on my computer and the login part works fine.

However, have you tried to click the Save changes and Realm Explorer buttons. During my test, saven changes asks for the password and when I give it REACT_APP_ADMIN_PASSWORD_HASHED password, it fails to login. The language actually says I need to give user password, but after your changes user password would not have been entered yet so if it uses user password we would need to change that to this admin password or sync password.

Yes, we store user password and after first login it might work but that defeats the purpose of moving this setting UI outside and putting it in login page.

Same for factory reset and other buttons.

Another thing I noticed is we would still need this setting section inside after user had logged in. So if you change this password to admin password or sync password, maybe we have to set it back to user password when the UI opens inside because users would not know sync or admin password.

Another interesting thing, we added a package and used the env variable but it just dawned on me that since we had already used sync password to get to this point and sync password is saved in our mobile database. Maybe we should use that. People who know the sync password are admins anyway so they should be able to see the data. If we are using User password inside, sync password is actually more secure. 

So I think we can use sync password for this admin interface and avoid using env file. 

https://github.com/openmsupply/mobile/blob/9ecf8683e74d9cad797191f44c500ca3cd6f5fe6/src/pages/SettingsPage.js#L122-L125

Can you please experiment and try and see if you can use sync password instead of admin password.

arjunSussol:
@sworup @ujwalSussol 
After discussion I have fix issues raised by @sworup above.
Please, review this again.

ujwalSussol (approved):
@arjunSussol  : thanks for the code overview. The user interface looks great.
@sworup and @arjunSussol : Thanks for helping to get this running... 👍  I am learning new things 👍

ujwalSussol:
@AnushaUp and @santoshkhanal12   :   could you test this APK.

https://drive.google.com/drive/u/0/folders/1EhyC6ANWxV_UqiPpIV0rHAV2EjzEwpbR

AnushaUp:
Tested With latest APK: `https://github.com/openmsupply/mobile/tree/%234949-Allow-setting-editable-from-login-page`

@arjunSussol 
In this screen below which is the the just after login there is no setting button.
photo_2022-11-11 17 55 56

In this login window the screen is not responsive and setting section is showing up in the wrong section.
photo_2022-11-11 17 56 00

sworup (approved):
Looks good and tested well.

arjunSussol:
@AnushaUp @sworup the setting visibility on Login window should be fixed by this PR: https://github.com/openmsupply/mobile/pull/5001

The another issue can be recreated. I still notice the Settings functionality inside Navigator window.

PR #5002, opened 2022-11-15
#4996 Label updated as per password preference

### LINKED ISSUE #4996: Change messages in admin setting page
## Describe the bug

The more I look into this the more it does not look right.
When user is logged out and is clicking the setting page, he/she need to give a master password. The same master password is needed to save authorised changes. However the messages says user password.

I know the same interface/components are shared between the setting page here and once you get after you have logged in, but it just does not feel right as it required master password but asks for user password.

### While login in with master password

SCR-20221114-fpd

### While doing factory reset

SCR-20221114-fnv

### While saving updates to the fields, by clicking the save button

SCR-20221114-fpd

Make it "Please confirm with your master password to save the changes"

### While updating sync URL and sync password

SCR-20221114-fv6

Add this one line above the two lines shown here just below the Warning label, "Please confirm with your master password to save the changes"

### To reproduce

1. Sync a data file
2. Click the setting button and type the master password
3. From this moment on all instances ask for your user password, but you actually are supposed to give your master password.
4. Also at least 1 message can be more clear about what you are supposed to do

### Expected behaviour

- The UI should ask for master password. And in some instances should be more clear with instruction

### Proposed Solution

- Create/update message component variables.

### Version and device info

- App version:
- Tablet model:
- OS version:

arjunSussol:
Alert label will be shown differently based on logged user or admin/master password.

Fixes #4996 

## Change summary

Just added new alert label for master/admin access.
It should show `Please confirm your master password` during login and saving credentials.

## Testing

- [ ] On Login form, click the circle icon which is setting icon.
- [ ] It should ask to enter password with label `Please confirm your master password`.
- [ ] Enter accurate password and click on `Factory reset` on setting page, it gives alert with `master` keyword instead of `user`
- [ ] Click `Save Changes` button, it displays the form as step 2 with same label.
- [ ] Now, go back to login page and enter with a valid user and go to setting.
- [ ] From step 2 to 4 the above label should be same with a single word change i.e. `master` is replaced by `user`. For e.g. `Please confirm your user password`.


sworup (commented):
@arjunSussol 
The code mostly looks good.

One thing:

When you click the factory reset button and it asks for your master password or user password, as you type it, the password shows up as characters, i.e. visible to everyone. Master passwords especially are very sensitive. We should set the input type to password and hide it.

What do you think?

arjunSussol:
> @arjunSussol The code mostly looks good.
> 
> One thing:
> 
> When you click the factory reset button and it asks for your master password or user password, as you type it, the password shows up as characters, i.e. visible to everyone. Master passwords especially are very sensitive. We should set the input type to password and hide it.
> 
> What do you think?

@sworup I retested it in both new and existing one and I found that was intentional.
However, it would be good to hide password. So, going to apply your suggestion.

sworup (approved):
LGTM

PR #5018, opened 2022-12-05
Remove JSON.stringify so that createCustomerInvoice method's JSON.str…

### LINKED ISSUE #5017: JSON Malformed upon vaccination POST request Payload
## Describe the bug

JSON Malformed upon vaccination POST request Payload.

### To reproduce

1. Add a vaccination transaction
2. Get the POST request Payload from Mobile
4. Validate the JSON
5. See error

### Expected behaviour

JSON should be validated correctly.

### Proposed Solution

### Version and device info

- App version: 8.6.1
- Tablet model: Emulator
- OS version: Android 9

sworup:
Fixes #5017 

## Change summary

- You get a JSON-malformed warning log when a new vaccine event is created.
- This happens because the vaccine custom data, which is actually JSON data, is being double encoded being stored as string in customData field of Transaction.

## Testing

- [ ] Create a new vaccine event from mobile and sync it
- [ ] Technically you are fine if you do not see JSON malformed log but that won't show up on normal installation according to @bijaySussol . Bijay has been intercpting the data coming in during sync and validating the transaction json, which would malform because double encoded data would try to show up as string when data would be like this: `customData: "{"name":"value"}"`, we know this will break the json format. 
- [ ] After this fix, the above case would not happen.

### DIFF
--- a/src/widgets/modals/VaccinationEvent.js
+++ b/src/widgets/modals/VaccinationEvent.js
@@ -470,7 +470,7 @@ const mapDispatchToProps = dispatch => {
         ...vaccinationEvent.extra,
         prescription: {
           ...vaccinationEvent.extra.prescription,
-          customData: JSON.stringify(updatedSupplementalData),
+          customData: updatedSupplementalData,
         },
       },
     };


sworup:
@bijaySussol maybe this will fix it. Give it a try with the database that was causing problem please.

bijaySussol:
@sworup 
Have tested app from this PR, I get the expected JSON and that JSON is on good shape thanks.
But before all the keys on `CustomData` were set as a key pair as now it is on plain string which i think is fine for me as that is how it suppose to be as relam database that field is set with.

I am not a good caller on this when it is comes with side affect, can't think of any but would be good to get feedback from @mark-prins or @Chris-Petty before we merge this one.

But what i can tell is, The JSON is now on good shape and is validated properly.

Thanks

bijaySussol (approved):
@sworup : 
I am happy with my test cases and the expectation i have on POST Request JSON.
So, I approve.

PR #5029, opened 2022-12-12
#4974 stocktake reason should be mandatory

### LINKED ISSUE #4974: Stocktake: Reason should be mandatory if available
## Describe the bug

In Stocktake, if reason column is visible then it must be set. We should not process further or can't finalize if reason is not set. But it is broken in recent version of mobile.
Must fix it.

### To reproduce

1. The reason setting must be done in mSupply desktop side by following [Activate reason](https://docs.msupply.org.nz/preferences:options)
2. In mobile, create a stocktake
3. Go to `Batches` of any item, it display list of available batch for that item.
4. Set the `counted doses` lower/higher than `snapshot doses`
5. Then the arrow button in `Reason` column must be enable, by click it you can set reason
6. Set reason for some items and ignore for some.
7. Click finalize button
8. All process done without setting reason for all or none item.

### Expected behaviour

If Reason is visible then it must be set, don't allow to finalize stocktake

### Proposed Solution

If Reason is visible then it must be set, don't allow to finalize stocktake with an alert.

- If reason is activated in mSupply desktop, then reason must be set in mobile for each item.
- There should be alert when clicking Finalize button if reason doesn't set for all items.
- The alert `The reason must be set for each item before finalizing stocktake`
- If reason is not activated then, we should finalize without interruption.

### Version and device info

- App version:
- Tablet model:
- OS version:

arjunSussol:
Fixes #4974 

## Change summary

Check if reason is activated in desktop and not set to at least one stock take batch then gives alert when finalizing. 

## Testing

- [ ] Enable reason (positive and negative) from desktop
- [ ] Sync to mobile well
- [ ] Create an stocktake and set reason for some of its batch and ignore for some
- [ ] If at least one batch misses the reason then you can't finalize the stocktake
- [ ] When click the finalize button alert `The reason must be set for each item before finalizing stocktake` popped up.
- [ ] Repeat the test by disabling the reason from desktop and clicking finalize button
- [ ] If reason is disabled then one can easily finalize stocktake.

### DIFF
--- a/src/database/DataTypes/Stocktake.js
+++ b/src/database/DataTypes/Stocktake.js
@@ -205,7 +205,7 @@ export class Stocktake extends Realm.Object {
    * @return {boolean}
    */
   get hasReasonNotSet() {
-    return this.items.some(item => !item.hasReasonSet === false);
+    return this.items.some(item => item.hasReasonSet === true);
   }
 
   /**
--- a/src/database/DataTypes/Stocktake.js
+++ b/src/database/DataTypes/Stocktake.js
@@ -279,9 +279,11 @@ export class Stocktake extends Realm.Object {
       finaliseStatus.message = modalStrings.stocktake_no_counted_items;
     }
 
-    if (this.hasReasonNotSet && this.isReasonActive && finaliseStatus.success) {
-      finaliseStatus.success = false;
-      finaliseStatus.message = modalStrings.reason_must_set_before_finalizing_stocktake;
+    if (this.isReasonActive && finaliseStatus.success) {
+      if (this.hasReasonNotSet) {
+        finaliseStatus.success = false;
+        finaliseStatus.message = modalStrings.reason_must_set_before_finalizing_stocktake;
+      }
     }
 
     return finaliseStatus;
--- a/src/database/DataTypes/Stocktake.js
+++ b/src/database/DataTypes/Stocktake.js
@@ -8,6 +8,7 @@ import { complement } from 'set-manipulator';
 
 import { addBatchToParent, createRecord, getTotal } from '../utilities';
 import { generalStrings, modalStrings } from '../../localization';
+import { UIDatabase } from '..';
 
 /**
  * A stocktake.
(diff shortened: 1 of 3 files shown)


src/database/DataTypes/Stocktake.js
```diff
@@ -198,6 +199,29 @@ export class Stocktake extends Realm.Object {
...
+  }
+
+  /**
+   * Check wether Reason i.e. Option is activated in mSupply desktop
+   * Return True if all positive and negative reasons are activated
+   *
+   * @return {boolean}
+   */
+  // eslint-disable-next-line class-methods-use-this
+  get isReasonActive() {
```
anildahalsussol, line 218: @arjunSussol @sworup: Why do we need to check items without reason if the no reason option is activated? I think it can make a mobile slow to go through every item if the reason option is not activated.
I think it better if we can check whether the reason was activated or not first then if yes then only look for items in stock take. What do you suggest?

sworup: Good suggestion @anildahalsussol

arjunSussol: > @arjunSussol @sworup: Why do we need to check items without reason if the no reason option is activated? I think it can make a mobile slow to go through every item if the reason option is not activated. I think it better if we can check whether the reason was activated or not first then if yes then only look for items in stock take. What do you suggest?

Thanks @anildahalsussol and @sworup 
Fixed by https://github.com/openmsupply/mobile/pull/5029/commits/dd026a2196593c848db885519794fb5e5bd4894a

sworup:
I have created a built mSupplyMobile-8_6_2-rc03-universal-release.apk to test. @anildahalsussol is testing it, if approved I will merge this.

sworup:
@anildahalsussol were you able to test he apk? If this PR passes please approve it.

ujwalSussol:
@arjunSussol  : I will now aim to have some fun testing this.

sworup (approved):
LGTM. There are some other issues of stocktake reasoning but this one can be merged.

ujwalSussol:
@arjunSussol  :  Thanks was able to review the UI on my android emulator.

With @anildahalsussol :  we agreed on the minimum extra change. Thanks. 👍

PR #5046, opened 2022-12-27
5037 Fix reason would not show on Requisition on Remote desktop

### LINKED ISSUE #5037: Reason would not show on Requisition on Remote desktop
## Describe the bug

Raised here first: https://github.com/sussol/msupply/issues/11962

- Reason field not shown while sending Internal order from tablet store(V502) to Remote V7 desktop.
- We can create the internal order with reason but they would not show up on remote desktop.
- @anildahalsussol has tested this in the latest mobile version, and the issue exists there too.

### To reproduce

1. Create internal order from tablet with mSupply v-502
2. Check customer requisition on desktop remote system V7-beta40
3. See the Reason column

### Expected behaviour

If there is reason provided then it should appear on requisition

### Proposed Solution

### Version and device info

- App version: v5.0.2 and later
- Tablet model: Any
- OS version: Android

sworup:
Fixes #5037

## Change summary

- Option field from Requistion line (RequisitionItem) was not being synced. I added option id to the sync in and out code and it worked.

## Testing

- [ ] Create internal order (Supplier Requisition) from a tablet, add reasons to the internal order
- [ ] Check customer requisition on desktop remote system 
- [ ] See the Reason column, the reasons should show properly

### DIFF
--- a/src/sync/incomingSyncUtils.js
+++ b/src/sync/incomingSyncUtils.js
@@ -762,6 +762,7 @@ export const createOrUpdateRecord = (database, settings, recordType, record) => 
         incomingStock: parseNumber(record.Cust_stock_received),
         outgoingStock: parseNumber(record.Cust_stock_issued),
         daysOutOfStock: parseNumber(record.DOSforAMCadjustment),
+        option: database.getOrCreate('Options', record.optionID),
       };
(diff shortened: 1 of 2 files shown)


bijaySussol (approved):
@sworup : Awesome!

Thanks

PR #5047, opened 2022-12-28
#5039 Stocktake reasoning resolved

### LINKED ISSUE #5039: Stocktake reasoning: If no difference user should be allowed to finalize the stock take
## Describe the bug

Sometimes user use stock take just to change price, batch, suppliers not only quantity. So in that case, if the difference is 0 then it's not allowing to finalize the stock take.
### To reproduce

1. Go to Stock take.
2. Create new.
3. Just change batch or price.
4. Finalize see error.

### Expected behaviour

If differen is N/A for all lines it should show exisitng alert and if any other lines have difference 0 then still need to allow to finalize and if the difference is higher than 0 then reason set alert must be shown incase of empty.
### Proposed Solution

### Version and device info

- App version: 8.6 rc3
- Tablet model: lenovo
- OS version:

arjunSussol:
Stocktake reasoning issue fixed

Fixes #5039 

## Change summary

One should set reason to a valid batch if reasoning is activated

## Testing

- [ ] Enable reason (positive and negative) from desktop
- [ ] Sync to mobile well
- [ ] Before testing make sure to test for different type of test case scenario as below then follow from step 4
  - It looks at only counted items, then check all batches of that items.
  - If the batch has counted and there is quantity difference then look at its valid reason.
  - If reason is not set then don't allow to finalize otherwise finalize
- [ ] Create an stocktake and set reason for some of its batch and ignore for some
- [ ] If at least one batch misses the reason then you can't finalize the stocktake
- [ ] When click the finalize button alert `The reason must be set for each item before finalizing stocktake` popped up.
- [ ] Repeat the test by disabling the reason from desktop and clicking finalize button
- [ ] If reason is disabled then one can easily finalize stocktake.

### DIFF
--- a/src/database/DataTypes/Stocktake.js
+++ b/src/database/DataTypes/Stocktake.js
@@ -201,11 +201,28 @@ export class Stocktake extends Realm.Object {
 
   /**
    * Get if stocktake has any item without reason set
+   * First, get only counted items then check all of its batches
+   * then get only counted batch with some difference
+   * then check reason of those batches
+   *
+   * Return True if at least one of the valid batch doesn't have reason set.
    *
    * @return {boolean}
    */
   get hasReasonNotSet() {
-    return this.items.some(item => item.hasReasonSet === true);
+    const countedBatches = [];
+    const countedItems = this.items.filter(item => item.hasBeenCounted);
+    countedItems.forEach(item => {
+      item.batches.forEach(batch => {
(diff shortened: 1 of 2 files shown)


anildahalsussol (approved):
Looks good

ujwalSussol:
@arjunSussol  working great. 

Only thing is, it would have been good to list a item with problems to make things easier for the user to fix...  A window that can scroll on item list.

PR #5084, opened 2023-01-30
#5080 Fix crash when Entering The Details of Insurance Policy in Mobile App

### LINKED ISSUE #5080: App Crashes When Entering The Details of Insurance Policy in Mobile App
## Describe the bug

App Crashes When Entering The Details of Insurance Policy in Mobile App

### To reproduce

1. Any Data File with mobile stores
2. Sync Setup on mobile
3. Open `Dispensing` tab
4. Dispense a patient
5. Select the Prescriber
6. Select the items and quantity to the items that has been dispense
7. Click on `Next`
8. Go to payment section and click on the `+` (Plus Button)
9. Insurance policy Window will appear following things happen:
     A. Enter any text/number on Family policy Number. App gets crashed.(Bug)
     B. Enter any text/number on Personal Policy number. App gets crashed.(Bug)
     C. Try to change Policy Provider, the selection will not gets changed. (Bug)
     D. Try to change Is Active Yes or No boolean, No changes will be observed. (Bug)
      E. Try to change Policy type boolean, No changes will be observed. (Bug)

Video Reference: https://recordit.co/7NQ3UCXXQi

          

### Expected behaviour

No Bugs should be seen as mentioned.

### Proposed Solution

### Version and device info

- App version: 8.6.2-rc04
- Tablet model:Samsung
- OS version: Android

### Additional context

Found While testing the issue: https://github.com/openmsupply/mobile/issues/5043

sworup:
Fixes #5080

## Change summary

In FormControl, which is shared by all, there is a useIsFocused check that sets isFocused to true when the form control comes in focus. But the problem is in case case of Dispensing > Prescription Page the page are loaded as Wizard pages. The pages take focus but the models are intitialised way below in the hirirarchiy, in Dispensing page iteself and when the models are activated from a button click in any of the Wizard tab page the thus loaded model do not take focus.

This stops initialisation of the FormControl and cause the UI to be unresponsive or submit blank values if saved (since whatever you typed is not stored or tracked).

I did some extensive testing and realised that we do not really need the page be on foucs before initialising the FormFields. In fact we can start loading the fields as soon as the component is loaded. Which is esentially what useEffect with empty devependency fields does. So replacing isFoucsed check with simple useEffect with empty depenedencies. It will run only once when the component mounts which is exactly what we want.

## Testing > 

- [ ] Click Dispensing and dispense to any of the patients. Create one if there are no patients.
- [ ] Select a prescriber. Create a new prescriber if there are none.
- [ ] Select one or more items with quantity, and give quantity to this prescription. 
- [ ] Click Next
- [ ] Now in the Payment section click the red circle with + in it to add new insurance policy.
- [ ] Fill the form and update the toggle button dropdown etc. Previously none of these changes were registering and the app was crashing.
- [ ] Make sure the insurance policy would be selected on the Prescription Finalise page.

### Regression checks

- [ ] This change effects all of the FormControl pages. So anything from New patient creation/editing, lookup patient, prescriber create and edit, vaccine survay forms etc. anything that has form loaded in a model needs to retested as this change will affect all of them.

### DIFF
--- a/src/pages/DispensingPage.js
+++ b/src/pages/DispensingPage.js
@@ -12,7 +12,7 @@ import { ToggleBar, DataTablePageView, SearchBar, PageButton } from '../widgets'
 import { DataTable, DataTableRow, DataTableHeaderRow } from '../widgets/DataTable';
 import { SearchForm } from '../widgets/modals/SearchForm';
 import { PatientHistoryModal } from '../widgets/modals/PatientHistory';
-import { FormControl } from '../widgets/FormControl';
+
 import { ModalContainer } from '../widgets/modals/ModalContainer';
 
 import { recordKeyExtractor, getItemLayout } from './dataTableUtilities';
@@ -20,11 +20,9 @@ import { createPrescription } from '../navigation/actions';
 import { useNavigationFocus, useSyncListener, useDebounce } from '../hooks';
 
 import { UIDatabase, generateUUID } from '../database';
-import { getFormInputConfig } from '../utilities/formInputConfigs';
 
 import { PatientActions } from '../actions/PatientActions';
 import { PrescriberActions } from '../actions/PrescriberActions';
-import { InsuranceActions } from '../actions/InsuranceActions';
 import { DispensaryActions } from '../actions/DispensaryActions';
 
 import {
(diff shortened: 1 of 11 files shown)


sworup:
After testing with @anildahalsussol we found that removing isFoucs will break  PatientSelectComponent close. So need isFocus true check so have to find another way to fix this.

anildahalsussol:
@sworup dai: As we checked for some windows it's creating a memory leak. I think we can add another prop in FormControl like inFocus and set true for the child container and check isFocused  from hook API which works for the parent container and inFocused which we pass as a prop and can be useful for the child container.

arjunSussol (approved):
Got understanding with @sworup about issue and tested it.
It works without crashing and codes look great.

PR #5087, opened 2023-02-02
#5086 Reason column should be visible to all items

### LINKED ISSUE #5086: Reason column should be visible to all items if reason is enabled
## Describe the bug

Make reason column visible to all items if enabled

### To reproduce

Take datafile where reasoning is enabled.
1. In mSupply desktop, enable positive and negative reason, and disable `Open vial wastage`
2. Create a stock take for the vaccine item
3. The reason column is invisible to this vaccine item
4. Enter quantity and see if the difference is not 0.
5. Try to finalize and it will not because reason must be set.

### Expected behaviour

Make reason visible to all items if positive/negative reasons are enabled, include `Open vial wastage` reason if enabled to vaccine items only

### Proposed Solution

First check if both positive and negative reasons are enabled then make reason column visible to all items.
If `Open vial wastage`  reason is enabled then include it along with negative reason as existing for vaccine item, otherwise just show positive/negative reason to vaccine items too as similar to normal item.
After that check stock take reason validation on finalization.

### Version and device info

- App version:Develop
- Tablet model:Emulator
- OS version:

arjunSussol:
If reason is active (if both negative and positive reasons are enabled) then reason column must be visible to all items including vaccines.

Fixes #5086 

## Change summary

Reason column is visible to vaccine items too even `Open vial wastage` reason is disabled

## Testing

- [ ] Take a data having both vaccine and non vaccine items for mobile site, if not then create both
- [ ] Go to mSupply desktop > Preferences > Options > Create some reasons for `Positive`/`Negative` inventory adjustment and `Open vial wastage`.
- [ ] Make a tick to them to enable the respective reason.
- [ ] Setup a mobile and sync must be up to date.
- [ ] Mobile, create a stocktake with vaccine and non vaccine items
- [ ] Click `Batch` column, the `Reason` column must be visible to each batch
- [ ] If difference is greater than zero then positive reason should be visible
- [ ] If difference is less than zero and not vaccine then negative reason should be visible
- [ ] If difference is less than zero and vaccine item then negative and vial reasons should be visible
- [ ] Then finalize stocktake, if at least one batch missed reason then you can't finalize.
- [ ] Repeat same process from point 2, just disable the `Open vial wastage` in desktop
- [ ] Now, all items including vaccines the reason column must visible with either positive or negative reason.
- [ ] You should be able to select one and finalize.
- [ ] Stocktake finalization takes place if reason set to all counted items having quantity difference greater or less than zero

### DIFF
--- a/src/widgets/modals/StocktakeBatchModal.js
+++ b/src/widgets/modals/StocktakeBatchModal.js
@@ -57,7 +57,6 @@ export const StocktakeBatchModalComponent = ({
   usingVaccines,
   usingHideSnapshotColumn,
   dispatch: reduxDispatch,
-  usingOpenVialWastageReasons,
 }) => {
   const initialState = useMemo(() => {
     const pageObject = stocktakeItem;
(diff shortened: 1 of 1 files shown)


anildahalsussol:
@arjunSussol : Code seems good but strange that I can no longer select the reason.

arjunSussol:
> @arjunSussol : Code seems good but strange that I can no longer select the reason.

@anildahalsussol it should be fixed by https://github.com/openmsupply/mobile/pull/5087/commits/38dee226ad2c8995536a17b8a16cb7817a7d60ae

anildahalsussol (approved):
Awesome

sworup (approved):
LGTM

PR #5098, opened 2023-02-06
5097 Fix Instances of same model in different nested routes opens multiple  models in background

### LINKED ISSUE #5097: Instances of same model in different nested routes opens multiple models in background
## Describe the bug

- This is introduced by https://github.com/openmsupply/mobile/issues/5080. The changes there fix crashing bugs but introduce this case when if you load 2 instances of the same model in nested routes. For example, PatientEditModel is loaded in the dispensary and prescriber routes. You have a case where both these models open when you trigger it from inside Prescriber route. It is kinda un-noticable for the client but it does happen, and kinda be seen in a slow tablet that gives i-frames to animations (you can kinda seek another model opening in the background).
- This does not affect data and since the models are loaded at start and the system only keeps one version of the model in momory and treates the instances are references, it should not do too much memory leak issue either. But since it does happen I want to tackle it.

### To reproduce

1. Go to Patient edit/create or prescriber edit/create. Open it from the prescriber select window.
2. You would see the patient edit or prescriber edit model open. It would work properly and will let you edit data properly.
3. If you have a slow system and you give good attention you can see 2 models open, one at the background and one at the foreground. Most of the time this is unnoticable. Also has not brearing on funtionality but it is a problem.

### Expected behavior

Only the model in the foreground should open and the background one should not open.

### Proposed Solution

Use useIsFocused to check if the model is in focus. The background one would have this value returned to false and foreground one has it to true. Add this isFocused check in model visible prop of the model.

### Version and device info

- App version: All
- Tablet model: 8.6.2
- OS version: All

### Additional context

None

sworup:
Fixes #5097

## Change summary

Use useIsFocused to check if the model is in focus. The background one would have this value returned to false and foreground one has it to true. Add this isFocused check in model visible prop of the model.

## Testing

- [ ] Go to Patient edit/create or prescriber edit/create. Could you open it from the prescriber select window?
- [ ] You would see the patient edit or prescriber edit model open. It would work properly and will let you edit data properly.
- [ ] Previously If you have a slow system and you give good attention you can see 2 models open, one at the background and one at the foreground. Most of the time this is unnoticeable. Also has not bearing on functionality but it is a problem. This multiple model opening should not happen.

### DIFF
--- a/src/widgets/modalChildren/PatientEditModal.js
+++ b/src/widgets/modalChildren/PatientEditModal.js
@@ -3,6 +3,7 @@ import React from 'react';
 import { View, StyleSheet } from 'react-native';
 import { connect } from 'react-redux';
 import PropTypes from 'prop-types';
+import { useIsFocused } from '@react-navigation/core';
 import { FormControl } from '..';
 import { PageButton } from '../PageButton';
 import { FlexRow } from '../FlexRow';
@@ -47,6 +48,8 @@ const PatientEditModalComponent = ({
   isCreatePatient,
   patientEditModalOpen,
 }) => {
+  const isFocused = useIsFocused();
+
   let canSave = canSaveForm && canEditPatient;
 
   const hasVaccineEvents = hasVaccineEventsForm;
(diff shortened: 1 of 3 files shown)


anildahalsussol (approved):
looks good

PR #5139, opened 2023-03-13
#5132 check for local duplicity of patient

### LINKED ISSUE #5132: Check for local duplicates of patient name + date of birth

Trello: https://trello.com/c/EB4F1nRk/108-414a-warn-user-that-duplicate-patient-records-exist

They want the system to check that there are no local duplicate patients on the tablet before allowing the user to create a new patient. 

This verification would be done by checking for duplicates of 'patient name + date of birth'.

 1) Just show a warning when duplicate patients are detected but still allow the user to proceed then it will be fine to add this functionality for all clients

2) Ignore the lookup patient interference, client is just wanting to try and reduce the number of duplicates that are being created, I don't think they're expecting to eliminate the chance of them being created entirely.

## Implementation

## Additional context

Jobcode: KIY:MOB

arjunSussol:
Fixes #5132 

## Change summary

Check for local duplicity of patient when creating new.
If patient already exists with same lastName, firstName, and dateOfBirth then we need a confirmation modal.
Either we can cancel the duplicity or can create duplicate patient.

## Testing

- [ ] Go to Dispensary
- [ ] Create a new patient
- [ ] Enter all mandatory fields, the `Save` button will be enable to create new patient.
- [ ] Save above patient.
- [ ] Try to recreate the same patient having same firstName, lastName, and dateOfBirth as above
- [ ] If above fields are matched and mandatory fields are entered then an confirmation alert `Are you sure you want to duplicate this patient? A patient with name: {0} and date of birth: {1} already exist.` will be generated.
- [ ] If you click `Cancel` then nothing will be saved and page return to patient list.
- [ ] If you click `Save` then duplicate patient will be created having same details as above.
- [ ] Repeat above steps for vaccination module, if you try to duplicate a new patient by entering firstName, lastName, and dateOfBirth then above alert will be generated.
- [ ] If you click `Cancel` then page go to previous opening tab.
- [ ] If you click `OK` then page remains in the entry form and keep continue to enter all mandatory fields to create new patient.
- [ ] Make sure that this is applied to new patient only not editing old patient.

@bijaySussol once the above confirmation title `Are you sure you want to duplicate this patient?` approved, we could need its Chinese translation.

### DIFF
--- a/src/localization/modalStrings.json
+++ b/src/localization/modalStrings.json
@@ -3,6 +3,7 @@
     "adr_form_for": "ADR Form for:",
     "are_you_sure_delete_sensor": "Are you sure you want to delete this sensor?",
     "are_you_sure_delete_patient": "Are you sure you want to delete this patient?",
+    "are_you_sure_duplicate_patient": "Are you sure you want to duplicate this patient?",
     "patient_cant_delete_with_vaccine_events": "You can only delete records of patients who do not have any active vaccine events. Please make sure that the vaccine events has been deleted if you want to delete this patient.",
     "add_at_least_one_item_before_finalising": "You need to add at least one item before finalising",
     "and": "and",
--- a/src/localization/modalStrings.json
+++ b/src/localization/modalStrings.json
@@ -3,7 +3,7 @@
     "adr_form_for": "ADR Form for:",
     "are_you_sure_delete_sensor": "Are you sure you want to delete this sensor?",
     "are_you_sure_delete_patient": "Are you sure you want to delete this patient?",
-    "are_you_sure_duplicate_patient": "Are you sure you want to duplicate this patient?",
+    "are_you_sure_duplicate_patient": "Are you sure you want to duplicate this patient? A patient with name: {0} and date of birth: {1} already exist.",
     "patient_cant_delete_with_vaccine_events": "You can only delete records of patients who do not have any active vaccine events. Please make sure that the vaccine events has been deleted if you want to delete this patient.",
     "add_at_least_one_item_before_finalising": "You need to add at least one item before finalising",
     "and": "and",
--- a/src/selectors/patient.js
+++ b/src/selectors/patient.js
@@ -93,11 +93,12 @@ export const selectPatientByNameAndDoB = ({ lastName, firstName, dateOfBirth }) 
   if (dateOfBirth) {
     const dob = moment(dateOfBirth).format('L');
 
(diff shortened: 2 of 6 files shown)


arjunSussol:
[(4.1.4.a) Warn user that duplicate patient records exist](https://trello.com/c/EB4F1nRk/108-414a-warn-user-that-duplicate-patient-records-exist)

bijaySussol:
> @bijaySussol once the above confirmation title `Are you sure you want to duplicate this patient?` approved, we could need its Chinese translation.

I think we should be super fine here, since x client does not uses the dispensary module at all, so we should be super fine for now.
Thanks for checking, If they ask later after the switch we can add them up quite easily but do not worry for that client

bijaySussol:
@arjunSussol as on standup, You probably doing more changes on this one, so will hold off the review until you get those new changes into here
Thanks

arjunSussol:
@bijaySussol @anildahalsussol @ujwalSussol @adamdewey this PR has been improved after @adamdewey suggestions.
Now, you can have a look to review it.

anildahalsussol:
@arjunSussol : Few things: 
1. Isdeleted is not handled which is soft delete in msupply.
2. currentPatient is required field but its sending null while dispening.
Screenshot 2023-03-22 at 16 33 29

arjunSussol:
@anildahalsussol 
> @arjunSussol : Few things:
> 
>     1. Isdeleted is not handled which is soft delete in msupply.
> 
should be fixed by https://github.com/openmsupply/mobile/pull/5139/commits/3fe45fc9496a26a51254eaed7de44c12bd1f325e
>     2. currentPatient is required field but its sending null while dispening.
> 
Can't recreate this warning
> 
> Screenshot 2023-03-22 at 16 33 29

arjunSussol:
> @arjunSussol : While adding date if we type just 22 its converting to current date and if any name with current DOB it gives alert where 22 can be from this year or previous year which is unknown. While debugging it seems in method:
> 
> ```
> getFormInputConfig : line 131
> let inputDate = moment(input, DATE_FORMAT.DD_MM_YYYY, null, true);
> Here moment is putting a current date.
> ```

@anildahalsussol as we discussed, for now we are going to disable the manual DoB entry. We just focus on DoB entry bu datePicker.
For future we have an issue for this: https://github.com/openmsupply/mobile/issues/5157

anildahalsussol (approved):
Looks good.

PR #5189, opened 2023-05-11
Add notifications for AI API interactions

### LINKED ISSUE #5188: Add notification when macro eyes API runs
Curently it's hard to see if the Macro Eyes API is being called or not in the Application when the user clicks the 'Use Suggested' button on a Program Requistion.

Ideally the system should show some kind of notification to confirm if a macro eyes api call has been made.

Perhaps something like at toast message saying

"X usage predictions updated from STRIATA API"

(STRIATA is the forecasting system from MacroEyes)

`ToastAndroid.show();` might be useful?

If the API returns an error, and error toast with error information should be displayed.

## Implementation

I would expect this message to be triggered around here:
https://github.com/openmsupply/mobile/blob/3b21d2587b82a1a4ee92ebb4470fc860a8f0eb56/src/pages/SupplierRequisitionPage.js#L288

Do nothing

## Additional context

There's a pilot underway with this code, and we need better visibility of usage.
Branch: `feature/macro-eyes`

jmbrunskill:
Fixes #5188 

## Change summary

- Use positive daily usage as item_filter (should result in more reliable requests to API)
- Move API Key db lookup inside request, to avoid global state not being updated correctly? - Possible improvement

- Add notification toasts to inform user of AI API interactions

*Success*
Kapture 2023-05-12 at 10 03 05

*Error*
Kapture 2023-05-12 at 10 04 16

## Testing

- [ ] Setup api requests in Store config
- [ ] Test with API Errors

Performance issue on old devices?

### DIFF
--- a/src/utilities/prediction/macroEyes.js
+++ b/src/utilities/prediction/macroEyes.js
@@ -6,15 +6,12 @@
  * - Update suggested values in requisition item list
  *
  */
+import { ToastAndroid } from 'react-native';
 import { useEffect, useState } from 'react';
 
 import { UIDatabase } from '../../database';
 import { SETTINGS_KEYS } from '../../settings';
-
-const API_URL =
-  UIDatabase.getSetting(SETTINGS_KEYS.ME_PREDICTION_API_URL) || 'http://civapi.dev.macro-eyes.com';
-
-const API_KEY = UIDatabase.getSetting(SETTINGS_KEYS.ME_PREDICTION_API_KEY);
+import { programStrings } from '../../localization';
 
 const RETRY_COUNT = 3;
 const TIMEOUT_MS = 10 * 1000;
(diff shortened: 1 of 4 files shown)


Chris-Petty (approved):
🚀

src/utilities/prediction/macroEyes.js
```diff
@@ -6,15 +6,12 @@
...
+import { ToastAndroid } from 'react-native';
 import { useEffect, useState } from 'react';
 
 import { UIDatabase } from '../../database';
 import { SETTINGS_KEYS } from '../../settings';
-
-const API_URL =
-  UIDatabase.getSetting(SETTINGS_KEYS.ME_PREDICTION_API_URL) || 'http://civapi.dev.macro-eyes.com';
-
-const API_KEY = UIDatabase.getSetting(SETTINGS_KEYS.ME_PREDICTION_API_KEY);
```
Chris-Petty, line 17: I was looking at some react code somewhere else and thinking about this again. It's not immediately obvious because these constants are simply used inside the methods below, but they've probably worked functionally like the database singleton pattern below:

https://github.com/openmsupply/mobile/blob/1f849502e6fb6cbb7984ca17b83522643e30ac32/src/database/BaseDatabase.js#L1-L21

bijaySussol (approved):
@jmbrunskill : 

Much Much better thanks @jmbrunskill 
Approved!

PR #5219, opened 2023-06-06
#5215 fixed alert issue for name duplicity check

### LINKED ISSUE #5215: Vaccinations module: Wrong alert when using patient
## Describe the bug

On the vaccination module on mSupply mobile while trying to select patient its giving wrong alert the name and date already exist and also there is same behaviour while adding the new patient which doesn't even exist

### To reproduce

1. Turn on the preference `mobile: Uses Vaccine Module` for the particular mobile store
2. Go to vaccination module
3. Try to search and use the existing patient
5. Also go to new patient and try creating new one
6. See error

### Expected behaviour

Should give appropriate error if there is some duplications 

### Proposed Solution

### Version and device info

- App version: v706
- Tablet model: mSupply mobile v8.6.5

### Additional context
 
https://github.com/openmsupply/mobile/assets/115983430/2507241f-2418-4a09-8b72-58064d478dd6

arjunSussol:
Fixes #5215 

## Change summary

Mobile > In the vaccination module > Name duplicity check alert should be generated for new patient.
Check for local duplicity of patient when creating new.
If patient already exists with same lastName, firstName, and dateOfBirth then we need a confirmation modal.
Either we can cancel the duplicity or can create duplicate patient.

## Testing

- [ ] Go to Vaccination, if not activated then active from desktop
- [ ] Create a new patient
- [ ] Enter all mandatory fields, the `Save` button will be enable to create new patient.
- [ ] Save above patient.
- [ ] Try to recreate the same patient having same firstName, lastName, and dateOfBirth as above
- [ ] If above fields are matched then an confirmation alert `Are you sure you want to duplicate this patient? A patient with name: {0} and date of birth: {1} already exist.` will be generated.
- [ ] If you click `Cancel` then nothing will be saved and page return to first tab of vaccination page.
- [ ] If you click `OK` then you are allowed to create a duplicate patient having same details as above.
- [ ] After clicking `OK`, and entering all mandatory fields click `Next` button and process all step to create a patient and its related vaccination event.
- [ ] In vaccination module a patient will be created successfully after vaccination.
- [ ] After this test same process from `Dispensing` module too.

### DIFF
--- a/src/selectors/Entities/name.js
+++ b/src/selectors/Entities/name.js
@@ -138,6 +138,8 @@ export const selectIsNewName = state => {
   const { editing } = nameState;
   const { id } = editing ?? '';
 
-  const isNewName = UIDatabase.objects('Name').filtered('id = $0', id).length === 0;
-  return isNewName;
+  if (id) {
+    return UIDatabase.objects('Name').filtered('id = $0', id).length === 0;
+  }
+  return false;
 };
--- a/src/selectors/Entities/name.js
+++ b/src/selectors/Entities/name.js
@@ -132,3 +132,12 @@ export const selectWasPatientVaccinatedWithinOneDay = state => {
 
   return !!history.filter(historyRecord => historyRecord.confirmDate.getTime() > oneDayAgo).length;
 };
+
+export const selectIsNewName = state => {
+  const nameState = selectSpecificEntityState(state, 'name');
+  const { editing } = nameState;
+  const { id } = editing ?? '';
+
+  const isNewName = UIDatabase.objects('Name').filtered('id = $0', id).length === 0;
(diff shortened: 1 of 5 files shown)


anildahalsussol:
@arjunSussol : Still getting the same issue as raised.

arjunSussol:
> @arjunSussol : Still getting the same issue as raised.

@anildahalsussol should be fixed by https://github.com/openmsupply/mobile/pull/5219/commits/bd02957795ea0a6015a776df17842ec3243c5897

anildahalsussol (approved):
Looks good. No more issue but there is another policies issue which is already reported by @AnushaUp.
And as agreed: duplicity while editing is ignored.

PR #5228, opened 2023-06-13
#5223 fixed dob issue and dispense vaccine

### LINKED ISSUE #5223: Doesn't let you create prescription through vaccination if the patient is from another store
## Describe the bug

There's some wrong alert while creating prescription from the vaccination module if the patient is from another store. It gives the alert if the patient is from same store as well but we can edit date and only it will give access to go forward but for patient of other store edit mode is disabled. 

https://github.com/openmsupply/mobile/assets/115983430/6f6c81c7-5518-475e-a6ed-1781783aa82c

### To reproduce

1. Go to Vaccination module and create prescription 
2. Click on Patient from another store and see error
3. It either redirect you back or doesn't let you go ahead and stops you there

### Expected behaviour

Should let you create prescription

arjunSussol:
Fixes #5223 

## Change summary

Able to dispense vaccine to existing patient without playing with dob.

## Testing

- [ ] Go to Vaccination module
- [ ] View existing patient and click Next button, should go to next tab without touching date of birth field as it should be accurately set.
- [ ] Dispense a vaccine to them
- [ ] Try to do same to the new patient
- [ ] Also, test in Dispensing module too.

### DIFF
--- a/src/utilities/formInputConfigs.js
+++ b/src/utilities/formInputConfigs.js
@@ -128,7 +128,7 @@ const FORM_INPUT_CONFIGS = seedObject => ({
     isRequired: true,
     validator: input => {
       let inputDate = moment(input, DATE_FORMAT.DD_MM_YYYY, null, true);
-      if (input instanceof Date && !Number.isNaN(input)) {
+      if (typeof input === 'number') {
         inputDate = moment(input);
       }
 


anildahalsussol (approved):
Reverting to the same old. So looks good approved.

